PR #16, opened 2016-04-20
Update index.html

kcoyle:
Ignore this one; just setting up. Will not make all edits at once.


jimkont:
ignoring but the action looks good +1

PR #17, opened 2016-04-21
Update index.html

kcoyle:
wording: notion -> concept; eliminated one instance of engine


jimkont:
Hi Karen, although we decided to remove most references to the "validation engine" I think we should have a way to refer to them  in some cases where it is necessary and keeping the mention in the introduction will help.
Let's remove it for now and see how many we can eliminate till the end. If we can remove them all we are fine, otherwise we might have to re-introduce it

PR #21, opened 2016-09-07
typo

stuartyeates:


ericprud:
Changed result table to test ApprovedIssueShape rather than changing the schema ApprovedIssueShape to IssueShape.

PR #22, opened 2016-10-30
examples use tabular summary

ericprud:
This changes the examples to use a tabular summary of node/shape pairs and their validation results.


HolgerKnublauch:
Hi Eric, I am cleaning up old pull requests and notice this didn't make it into the doc. As we are near CR, the document has been significantly edited since your branch and it was basically a clarification, we would prefer to close and move on.

PR #48, opened 2017-04-12
Added netage SHACL Engine implementation report

nicky508:
Added netage SHACL Engine implementation report


HolgerKnublauch:
Thanks, it's online now :) Feel free to send updates to kill off the red bars.

PR #50, opened 2017-04-13
test of validation report

pfps:
This has two separate additions:
1/ A new test for a shape where two property constraints share a shape and that shape is validated on the same node.
2/ Two tests for pre-binding.


HolgerKnublauch:
I have merged them in, but had to add triples along the lines of

<>
  rdf:type mf:Manifest ;
  mf:entries (
      <shared>
    ) ;
.

I'd appreciate if you could put similar triples into future files.

pfps:
I'll try to do so.  I copied from the existing files, and because the test
control information was interspersed with the test information I did not
notice that this information needed to be included.

peter

PR #66, opened 2017-04-20
run shacl-shacl on shacl-shacl

pfps:
eat your own (eat your own dogfood)


HolgerKnublauch:
Invalid, the shapes and data files must be relative URLs.

pfps:
Why is this invalid?  I don't see any prohibitions against absolute URLs there.

HolgerKnublauch:
We don't want pointers to any external files (that may change etc). It should be sufficient to download the folder and run it from there. That's why the URLs must be relative too. I cannot accept this branch until that's fixed, e.g. you could place a copy of that external file somewhere.

pfps:
How are people supposed to submit test cases if the rules aren't stated?  If there is such a requirement, put it in the document.

pfps:
I pulled the current shacl-shacl file into the test suite.   This isn't as good as tracking the file, but it is better than nothing.

PR #69, opened 2017-04-23
Revert "run shacl-shacl on shacl-shacl"

HolgerKnublauch:
Reverts w3c/data-shapes#66

Contains broken test cases (again). Please fix and make sure they work before resubmitting.


pfps:
What's broken?

HolgerKnublauch:
A manifest file had the same errors that I had fixed on your other commits (swapped data/shapes graph, mismatching URIs) and I then stopped looking further.

PR #78, opened 2017-04-28
tests of double sh:class, sh:minInclusive on dateTime, allowed qualified values

pfps:


HolgerKnublauch:
I have deleted class-003 because it doesn't look right (e.g. one result has two focusNodes).

pfps:
I fixed and resubmitted.

PR #79, opened 2017-04-28
tests for subselect that doesn't bind $this

pfps:


HolgerKnublauch:
Test cases are now invalid because the use of VALUES has been made ill-formed in SHACL-SPARQL.

PR #81, opened 2017-04-28
two tests of strange paths

pfps:


HolgerKnublauch:
I don't plan to merge this in, see https://lists.w3.org/Archives/Public/public-rdf-shapes/2017Apr/0078.html and previous discussions on this particular topic.

PR #97, opened 2017-10-21
test updates for issue #96

jimkont:
This PR provides tests related to issue #96.

What it does is downgrades the problematic tests from `approved` to `suggested` as well as adds two new suggested tests that verify the behavior according to the spec.
It also suggests another test for node validators with a SELECT query that was not tested before.

instead of deleting the tests, I thought it would be better to keep all to make it more obvious for people that this is a point that needs attention until it gets resolved

### DIFF
--- a/data-shapes-test-suite/reports/rdfunit-shacl-earl.ttl
+++ b/data-shapes-test-suite/reports/rdfunit-shacl-earl.ttl
@@ -23,320 +23,10 @@
   earl:assertedBy  <http://aksw.org/DimitrisKontokostas> ;
   earl:result      [ a             earl:TestResult ;
                      earl:mode     earl:automatic ;
-                     earl:outcome  earl:passed
-                   ] ;
-  earl:subject     <http://aksw.org/projects/RDFUnit> ;
-  earl:test        <urn:x-shacl-test:/core/node/hasValue-001>
-] .
-
-[ a                earl:Assertion ;
-  earl:assertedBy  <http://aksw.org/DimitrisKontokostas> ;
-  earl:result      [ a             earl:TestResult ;
-                     earl:mode     earl:automatic ;
-                     earl:outcome  earl:passed
-                   ] ;
-  earl:subject     <http://aksw.org/projects/RDFUnit> ;
-  earl:test        <urn:x-shacl-test:/core/node/datatype-001>
-] .
-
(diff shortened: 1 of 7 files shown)


HolgerKnublauch:
If I read this correctly, you have added tests that will produce a Failure if malformed validators exist. However, section 6.2.3 states "If no suitable validator can be found, a SHACL-SPARQL processor ignores the constraint.". I believe that rule fires on them.

Furthermore, having both the "new" tests and the "old" tests in place (albeit proposed) would be a contradiction. Again, I believe the simplest would be to simply delete the two offending tests because the behavior of them (and their ill-formed shapes graphs) is "undefined".

jimkont:
>
> If I read this correctly, you have added tests that will produce a Failure
> if malformed validators exist. However, section 6.2.3 states "If no
> suitable validator can be found, a SHACL-SPARQL processor ignores the
> constraint.". I believe that rule fires on them.
>
I see, I added another 2 test for this case then.

As I am not comfortable taking the decision on what to delete and what to
leave, I provide tests for all options as input.
As I see there are 4 optione here:

 - delete tests, spec has this option undefined
 - adjust the spec to allow this behavior
 - adjust the spec to forbid this bahavior
 - the spec is well defined and should ignore these cases

jimkont:
>  - the spec is well defined and should ignore these cases

what I meant is, according to Holger: "If no suitable validator can be found, a SHACL-SPARQL processor ignores the constraint." so the tests cases should be adjusted and return no violation (ignore the validators)

HolgerKnublauch:
I have removed the two tests that were misusing ASK validators. They contained ill-formed shapes graphs (because the values of sh:nodeValidator and sh:propertyValidator must be SELECT validators according to the spec). We have in general applied the policy that ill-formed shapes graphs are not covered by tests. Therefore most of the tests from your branch are now not needed, except the additional (valid) one on sh:nodeValidator. I have added that test, however it was missing the sh:value in the results.

https://github.com/w3c/data-shapes/commit/50dd59eb780ecee148c8cdc8380ec501c783b443

With this, I believe this branch can be closed? I assume you may want to rerun your tests and submit them with a separate PR.

jimkont:
To me, any other option would be much better in terms of interoperability but if this is what the WG decides, happy to close this

PR #106, opened 2018-09-24
Submitting pySHACL EARL validation report

ashleysommer:


HolgerKnublauch:
Thanks, I have integrated it. Note there was a https: mix-up in the earl namespace, in case you want to submit an update in the future (the one proposed test might be easy to pass for your implementation).

ashleysommer:
Woops, I missed that one proposed test.
You're right, it is an easy pass.
https://github.com/w3c/data-shapes/pull/107

PR #112, opened 2019-03-29
Update SHACL compact syntax link

Hydrotoast:
Hi w3c

The current link to `SHACL Compact-Syntax` on the `README.md` references a nonblank page with another link to the actual current draft. To remove a layer of indirection, I updated the link to reference the current draft.

Thanks for making this work available!


HolgerKnublauch:
For better or worse, there were political reasons for the current situation. I think someone from W3C asked us to empty the page and make clear that the compact syntax is not officially published by the data shapes WG, and the reason for that is well-known to those who had the pleasure of being in the WG.

So at this stage I don't have authority to accept the change request. The place to evolve this document further is the SHACL Community Group (and this is where the link goes - /shacl instead of /data-shapes). If you are interested in moving this further you are more than encouraged to post to the CG's mailing list.

PR #116, opened 2019-05-19
Fix SPARQL-based Target example

### LINKED ISSUE #115: Ill-formed example for SPARQL-based Targets
The example for [SPARQL-based Targets](https://www.w3.org/TR/shacl-af/#h-sparqltarget) contains an ill-formed prefix declaration with an IRI sh:namespace.
According to [namespace-datatype](https://www.w3.org/TR/shacl/#syntax-rule-namespace-datatype) in 5.2.1, "The values of sh:namespace are literals of datatype xsd:anyURI."

langsamu:
Fixes #115.

### DIFF
--- a/shacl-af/index.html
+++ b/shacl-af/index.html
@@ -509,7 +509,7 @@
 ex:
 	sh:declare [
 		sh:prefix &quot;ex&quot; ;
-		sh:namespace &lt;http://example.com/ns#&gt; ;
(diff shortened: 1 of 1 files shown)


HolgerKnublauch:
Thanks, but the fix would need to be xsd:anyURI, not xsd:anyUri.

PR #126, opened 2020-09-30
typing namespace string

pwin:


HolgerKnublauch:
Thanks, this is correct. I would apply the patch but only an official WG would be permitted to make such changes to a recommendation spec. Until a SHACL 1.1 is published, this issue will remain recorded in the errata (https://www.w3.org/2017/shacl/errata) which is basically a link to this github repo for now. So I suggest we just keep the ticket open for now, until a critical mass of changes triggers a new publication.

pwin:
Holger, the "Process 2020" of W3C allows ad hoc changes, and the turtle
file probably isn't constrained in the same way that the recommendation is.
Perhaps a word with PLH needed for clarification.

HolgerKnublauch:
Thanks for the update to the Process. I haven't tracked this, and the changes sound positive to me.

I can see this sentence in https://www.w3.org/wiki/Process2020#REC_Route

> Changes cannot initially be directly incorporated into the document, but must be merely annotated into it, until after they have received wide review and interoperable implementations.

I don't know what this means for a TTL file and how to even get interoperable implementations - there is nothing to do here for implementers.

BTW there is another small glitch - the rdfs:label of sh:nodeValidator is currently "shape validator" while it should be "node validator". That fix could go into the same update.

Given that this change here is just to the TTL file and then only to the github repo, not yet the published version, I will merge the PR in and also fix the node validator label above. When this goes "live" is another question to address later. Having the changes online will hopefully give it more exposure.

Issue #131, opened 2021-01-13
Proposal for removing the domain assertion for 'sh:declare'


wouterbeek:
I'm proposing to remove the following triple from the SHACL vocabulary:

```ttl
sh:declare rdfs:domain owl:Ontology.
```

While there is a good use case for asserting prefix declarations for OWL ontologies, there are also good use cases for asserting prefix declarations for things that are not OWL ontologies.  Examples include (SPARQL) queries, (SPARQL) endpoints, and datasets that are not OWL ontologies.

The removal of this triple will also make the SHACL vocabulary easier to understand and apply in light of the following paragraph from [the SHACL standard document](https://www.w3.org/TR/shacl/#sparql-prefixes), which explicitly allows `sh:declare` to be asserted for resources that are not OWL ontologies:

> The recommended subject for values of `sh:declare` is the IRI of the named graph containing the shapes that use the prefixes. These IRIs are often declared as an instance of `owl:Ontology`, but this is not required.

HolgerKnublauch:
To summarize from https://github.com/w3c/data-shapes/issues/130, the main reasons for having sh:declare rdfs:domain owl:Ontology were
- sh:declare is in practice typically used in conjunction with owl:imports and owl:imports rdfs:domain owl:Ontology
- we considered it a good practice (recommendation) to put such prefix declarations into a single place per graph, not repeat them for each occurrence.
- to help forms and other tools suggest sh:declare to users

The main reason against this domain statement seems to be that it may cause an undesired inference of X rdf:type owl:Ontology if X is not (meant to be) an owl:Ontology.

I have no strong opinion yet it seems the advantages outweigh the disadvantage. What specific problems does the inference of rdf:type owl:Ontology cause to anyone?

wouterbeek:
> What specific problems does the inference of rdf:type owl:Ontology cause to anyone?

One specific problem:

```turtle
prefix sd: <http://www.w3.org/ns/sparql-service-description#>

[ a sd:SPARQL11Query;
  sh:declare [ sh:prefix "..."; sh:namespace "..." ] ].
```

^ After reading the SHACL documentation I thought that this was ok.  Later I discover, to my surprise, that I am actually asserting that my queries are ontologies.


> The main reason against this domain statement seems to be that it may cause an undesired inference of X rdf:type owl:Ontology if X is not (meant to be) an owl:Ontology.

[...]

Issue #130, opened 2021-01-09
Unclear applicability of prefix declarations


wouterbeek:
The SHACL standard introduces support for representing prefix declarations.

According to the standard, these prefix declarations can be used on any node (not only ontologies):

> The recommended subject for values of sh:declare is the IRI of the named graph containing the shapes that use the prefixes. These IRIs are often declared as an instance of owl:Ontology, but this is not required.

But the vocabulary contains the following assertion (requiring everything for which a prefix is declared to be an ontology):

```ttl
sh:declare rdfs:domain owl:Ontology.
```

What is the intended utility of SHACL prefix declarations?  Are they intended to be generic?  Users would be able to represent prefix declarations for dataset and queries, among other things.  Or are they intended to only apply to ontologies?

HolgerKnublauch:
The rdfs:domain assertion does not "require everything for which a prefix is declared to be an ontology". This is RDFS, so it would only ever be an inference. The statement was meant to be a hint for tools that rely on RDFS/OWL to suggest that owl:Ontologies are a good place to store sh:declare triples. The reason why we recommend owl:Ontologies is that those also carry the owl:imports statements, and those are automatically followed when prefixes are resolved. Basically, whenever someone publishes a vocabulary, it would be a nice gesture to also declare the preferred prefix for that namespace at the owl:Ontology (or other resource that has the URI of the graph itself). However, you can theoretically also place them at non-Ontology resources, if you have use cases for that.

wouterbeek:
There are indeed properties that are used to provide domain hints.  An example of this is [`sdo:domainIncludes`](https://schema.org/domainIncludes) which indicates that a property may be used on instances of a certain class.

But `rdfs:domain` is not such a property at all.  `p rdfs:domain C.` states that everything with property `p` is also an instance of `C`.  That's just the way it is according to the RDF standard.

Indeed, from `p rdfs:domain C` and `s p o` the entailment `s a C` follows, but this is only a syntactic derivation.  `s` is already an instance of `C` semantically.  Entailment only makes the implicit knowledge explicit.

I must be overlooking something here, since you are obviously aware of these standards as well :-/

HolgerKnublauch:
Yes I believe I understand these standards :)

I am aware of various tools that use rdfs:domain as a hint for forms. It is much more widely used than sdo:domainIncludes, even if this approach can be considered a historical hack.

In any case, has the original question been answered?

wouterbeek:
[...]

PR #132, opened 2021-01-15
Erratum clarifying sh:declare and owl:Ontology

TallTed:
addresses #130 and #131

### DIFF
--- a/shacl/index.html
+++ b/shacl/index.html
@@ -4132,7 +4132,7 @@ ex:
 					</p>
 					<p>
 						The recommended <a>subject</a> for values of <code>sh:declare</code> is the IRI of the named graph containing the shapes that use the prefixes.
-						These IRIs are often declared as an instance of <code>owl:Ontology</code>, but this is not required.
+						These IRIs are often declared as an instance of <code>owl:Ontology</code>, but this <code>rdf:type owl:Ontology</code> declaration is not required.
 					</p>
 					<p>
 						<a>Prefix declarations</a> can be used by <a>SPARQL-based constraints</a>,


wouterbeek:
> The recommended <a>subject</a> for values of <code>sh:declare</code> is the IRI of the named graph containing the shapes that use the prefixes.
> These IRIs are often declared as an instance of <code>owl:Ontology</code>, but this <code>rdf:type owl:Ontology</code> declaration is not required.

For the second sentence I propose: "Notice that by asserting <code>sh:declare</code>, the subject IRI is also asserted to denote an OWL ontology."

My motivation:
- "These IRIs are often declared [...]" is an empirical observation that may be true or false, but whose truth is difficult to verify in practice.  This empirical observation may also change over time, as practice changes.  In my proposed formulation this empirical assertion is removed for this reason.
- While "this <code>rdf:type owl:Ontology</code> declaration is not required" is correct, it is also an obvious statement.  Because *any* implicit assertion is not required to be expressed explicitly in RDF, I removed this from my proposed formulation.
- Since the second sentence acts as a notice to the reader, I use the word "Notice" explicitly.
- The thing the reader should be noticed about is that by asserting `<s> sh:declare <o>` they are *also* asserting that `<s>` denotes an OWL ontology.  The latter is not obvious from reading the rest of the standard, so can act as a warning to readers who would otherwise have used `sh:declare` incorrectly.  (Specifically, it would have warned me that `sh:declare` should not be used to relate prefix declaration to queries and other things that are not OWL ontologies.)

HolgerKnublauch:
If you're OK I suggest closing this PR in favor of #133 - it is also easier to change the TTL file than the text of the spec.

PR #151, opened 2024-01-15
Submitting shaWell EARL validation report

cem-okulmus:
Here is the validation report for the tool [shaWell](https://github.com/cem-okulmus/shawell). It uses a provided Sparql endpoint to perform validation. The provided validation report here is for the Shacl CORE subset only. Since the approach already uses a Sparql  endpoint, it should be fairly straight-forward to extend it to support Shacl Sparql as well, but this is not supported as of this moment. I might provide another PR in the future with an extended validation report if and once this functionality is available in a future version of shaWell. 

If there is an interest in recreating this validation report, please write me. It can be produced automatically via test functions, but there is some hard-coded stuff that one would need to manually adapt to run it locally on a different machine. The tool currently also only supports GraphDB (other systems would require more work, mostly around rights management as the test needs to use Sparql Update to add the data ad-hoc and many systems do not easily expose this).

### DIFF
--- a/data-shapes-test-suite/reports/shawell_report.ttl
+++ b/data-shapes-test-suite/reports/shawell_report.ttl
@@ -1,0 +1,1188 @@
+@prefix sht:   <http://www.w3.org/ns/shacl-test#> .
+@prefix doap: <http://usefulinc.com/ns/doap#> .
+@prefix earl: <http://www.w3.org/ns/earl#>  .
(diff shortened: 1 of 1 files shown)


HolgerKnublauch:
Thanks! Please note that the W3C WG has finished and we cannot update the published documents at this stage. But I take your PR as input for a future iteration.

Issue #228, opened 2025-02-05
IRI introspection


ajnelson-nist:
There is a function available in SPARQL that I do not believe is available in SHACL 1.0.  If this actually is available somewhere or already under draft for 1.2, I very much welcome a link.

I have some SHACL shapes I would like to write that, for various purposes, require some review of the spelling of the subject IRI.

## Use case 1

Use case 1 is a prescribed-name-form checker.  With an ontology I work with, there is an expectation that `owl:NamedIndividual`s using the ontology should end with a UUID if they have no other practice in place to prevent IRI collisions.  E.g., `http://example.org/kb/Thing-36b67df4-1a42-4588-808a-19dfb79efbeb`.  My understanding is the only way to enforce this in SHACL 1.0 is a SPARQL constraint, because in SPARQL we can review the IRI as a string with `STR($this)`.  For reference, the shape we use[^1] is [here](https://ontology.unifiedcyberontology.org/documentation/shape-coreucothing-identifier-regex-shape.html)[^2][^3], but for saving a click, the form SPARQL constraint reads like this:

```turtle
[]
            a sh:SPARQLConstraint ;
            rdfs:seeAlso <https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.3> ;
            sh:message "UcoThings are suggested to end with a UUID."@en ;
            sh:select """
			PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
			PREFIX core: <https://ontology.unifiedcyberontology.org/uco/core/>
			SELECT $this
			WHERE {
			        FILTER (
			                ! REGEX (
			                        STR($this),
			                        "[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[0-9a-f]{4}-[0-9a-f]{12}$",
			                        "i"
			                )
			        )
			}
		""" ] ;
```

## Use case 2

Use case 2 is infer-able class assignment based on the URL form.  For instance, suppose an RDF data model wanted to model GitHub's Issues and Pull Requests as `ex:Issue` and `ex:PullRequest`.  If a `owl:NamedIndividual` follows the pattern `https://github.com/[^/]+/[^/]/issues/\d+` (with whatever regex escaping's needed to make that work), I should be able to write a `sh:TripleRule` that assigns the type `ex:Issue`.  Again, using SPARQL's `STR($this)`, I can write a `CONSTRUCT` query to handle this, but I don't see an easy way to target based only on IRI spelling.

[...]

PR #231, opened 2025-02-05
Issue 230: Update description of `sh:pattern` behavior with IRI nodes

### LINKED ISSUE #230: sh:pattern does not allow testing IRIs
> `sh:pattern` is defined by:
> 
> > The values of `sh:pattern` in a shape are valid pattern arguments for the [SPARQL REGEX function](https://www.w3.org/TR/sparql11-query/#func-regex).
> 
> Right. And the first argument is a defined as literal. 
> 
> ```sparql
> # this is valid
> BIND(regex(str(<>), 'foo') as ?foo)
> 
> # but this is not
> BIND(regex(<>, 'foo') as ?foo)
> ``` 

 _Originally posted by @tpluscode in [#228](https://github.com/w3c/data-shapes/issues/228#issuecomment-2637039780)_

> Isn't this highlighted line a test of `sh:pattern` against an IRI ...
> 
> https://github.com/w3c/data-shapes/blob/380d4ee454cb7c2b3426754a343c1927dba4509b/data-shapes-test-suite/tests/core/node/pattern-001.ttl#L15
> 
> and this line the corresponding XFAIL result?
> 
> https://github.com/w3c/data-shapes/blob/380d4ee454cb7c2b3426754a343c1927dba4509b/data-shapes-test-suite/tests/core/node/pattern-001.ttl#L41 

 _Originally posted by @ajnelson-nist in [#228](https://github.com/w3c/data-shapes/issues/228#issuecomment-2637163242)_

ajnelson-nist:
This patch series will address update test results, and possibly other documentation, around `sh:pattern` and how it behaves against IRI nodes.  E.g., there will be clarification on what should happen with use cases like this:

```turtle
@prefix ex: <http://example.org/> .

ex:Shape-1
    a sh:NodeShape ;
    sh:pattern "^https" ;
    sh:targetNode ex:Node-1 ;
    .
```

I.e., should the `sh:sourceConstraintComponent` be `sh:PatternConstraintComponent` for violating the pattern (not starting with `https`)?  `sh:NodeKindConstraint` for being an IRI node, even though `sh:nodeKind` is not used in this shape?

The intent of this PR is to at least close #230 , and possibly generate editor's draft revisions to cover use case 1 in #228 .

I suspect the working group needs to confirm whether this will be a backwards-compatible change.

I could also see SHACL-SHACL being updated to further constrain where `sh:pattern` can be used.

Aside: This is also my first PR against this repo, so rules-of-the-road pointers would be appreciated, such as appropriate target branch.
ACKs are due to @afs, @tfrancart, and @tpluscode from conversation on #228 ; what's the appropriate `git log` tag to use?  `Cc:`?
I'll force-push the first patch for the benefit of `git log` and `git blame` after guidance on ACKs.

### DIFF
--- a/data-shapes-test-suite/tests/core/node/pattern-001.ttl
+++ b/data-shapes-test-suite/tests/core/node/pattern-001.ttl
@@ -40,7 +40,7 @@ ex:TestShape
           rdf:type sh:ValidationResult ;
           sh:focusNode ex:Test ;
           sh:resultSeverity sh:Violation ;
-          sh:sourceConstraintComponent sh:PatternConstraintComponent ;
+          sh:sourceConstraintComponent sh:NodeKindConstraintComponent ;
           sh:sourceShape ex:TestShape ;
           sh:value ex:Test ;
         ] ;


data-shapes-test-suite/tests/core/node/pattern-001.ttl
```diff
@@ -40,7 +40,7 @@ ex:TestShape
           rdf:type sh:ValidationResult ;
           sh:focusNode ex:Test ;
           sh:resultSeverity sh:Violation ;
-          sh:sourceConstraintComponent sh:PatternConstraintComponent ;
+          sh:sourceConstraintComponent sh:NodeKindConstraintComponent ;
```
VladimirAlexiev, line 43: The original was right, this is wrong

ajnelson-nist: I have no objection with rolling this back, but I'd like the reasoning to be unambiguous from updates to the SHACL-Core document.

afs: I think should be `sh:PatternConstraintComponent` (which `jena-shacl` passes).

Even if it were a blank node - so some kind of argument error - it is `sh:sourceConstraintComponent` which is the constraint failing, not some kind of an error class.

afs:
> I'll force-push the first patch for the benefit of git log and git blame after guidance on ACKs.

We-all need to agree on process. We have one repo for many documents. Force-push may loose work on other documents.

The `editor-drafts-1.2` branch was created recently (Jan 6) by @HolgerKnublauch so I though it was his PR branch.

We need a protected branch for the documents.

ajnelson-nist:
> > I'll force-push the first patch for the benefit of git log and git blame after guidance on ACKs.
> 
> We-all need to agree on process. We have one repo for many documents. Force-push may loose work on other documents.
> 
> The `editor-drafts-1.2` branch was created recently (Jan 6) by @HolgerKnublauch so I though it was his PR branch.
> 
> We need a protected branch for the documents.

Apologies, I never had intention of force-pushing the `editor-drafts-1.2` branch.  My comments here were only meant to be about revising the first commit of this patch series, [`1064859`](https://github.com/w3c/data-shapes/pull/231/commits/1064859de98fd9651f8474876aae117227524d26), so its log-message reflected contributors accurately.

That effort might be moot if the patch'll be superseded, though.

simonstey:
> > I'll force-push the first patch for the benefit of git log and git blame after guidance on ACKs.
> 
> We-all need to agree on process. We have one repo for many documents. Force-push may loose work on other documents.
> 
> The `editor-drafts-1.2` branch was created recently (Jan 6) by @HolgerKnublauch so I though it was his PR branch.
> 
> We need a protected branch for the documents.

+1, besides a protected main/dev branch, what about having dedicated branches for each of the individual documents? e.g., one for shacl-core, shacl-sparql, shacl-inference, shacl-cs, ... ?

@gkellogg How was this handled in the JSON-LD WG with its documents on framing etc.?

edit: JSON-LD had multiple repos https://github.com/search?q=topic%3Ajson-ld-wg+org%3Aw3c&type=Repositories

tpluscode:
> dedicated branches for each of the individual documents

sounds like overkill

caribouW3:
> We need a protected branch for the documents.

The WG decided during yesterday's call to have protection and approval for merging on the gh-pages branch (which is the main branch for all documents).
If the proposal in a PR is wrong and the WG decides against it, we simply drop the PR, no roll-back involved.
if a merged PR was a wrong decision and the WG revisits it, then a new decision is recorded and a new PR is made.
Reverting is reserved to editorial mistakes, basically.

VladimirAlexiev:
As discussed in https://github.com/w3c/data-shapes/issues/230, sh:pattern does allow testing IRIs.
So this PR is wrong, I'll close it.

Issue #234, opened 2025-02-07
Contents of SHACL-AF


afs:
> I am unclear why shacl-af and shacl-js were edited here too, as those documents are from the previous WG and will not be moved forward.

_Originally posted by @HolgerKnublauch in https://github.com/w3c/data-shapes/issues/167#issuecomment-2638979866_

afs:
Is this where some of the sections from SHACL-AF will go?

* SPARQL Custom Targets
  This moves to SHACL-SPARQL (if not subsumed by `targetNode` +node expressions)

* Annotation Properties
  This moves to SHACL-SPARQL?

* SHACL Functions (Node Expressions)
  Part of node expressions

HolgerKnublauch:
@afs please double-check the title of this issue, doesn't look right.

- Yes I believe custom SPARQL targets will be solved with the sh:select node expression and sh:targetNode. Custom target types do not have an equivalent unless we introduce a higher-level template language for node expressions.

- I have no strong opinion on whether the Annotation Properties need to be upgraded as I have never seen them used.

- User-defined SHACL functions are a general extension point that I believe makes sense in SHACL-SPARQL 1.2 but may be out of scope due their implementation complexity.

- The use of functions as node expressions would become a node expression in SHACL-SPARQL if we elect to limit them to SPARQL functions, otherwise no clue. For example, we also support JS-based custom functions.

afs:
> [@afs](https://github.com/afs) please double-check the title of this issue, doesn't look right.

Done! (Side effect of GH "create new issue" on the original comment - lesson learnt)

afs:
> * The use of functions as node expressions would become a node expression in SHACL-SPARQL if we elect to limit them to SPARQL functions, otherwise no clue. For example, we also support JS-based custom functions.

Being able to give a name to a node expression so it can be used in several places seems a likely need. That doesn't need the full function definition (it does not need to declare parameters, or return type necessarily). 

```
my:nodeExpr_numTables
    rdf:type sh:nodeExpression ;
    sh:datatype xsd:integer ;     ## Optional
    sh:description "Calculate the number of tables in this database" ;
    sh:values [
        sh:count [
            sh:path [
                sh:inversePath edg:tableOf ;
            ] ;
        ] ;
    .
```

afs:
The two sections "Node Expressions" and "SHACL Rules" could be two documents, depending on how big the node expressions document is likely to become. The machinery for node expressions needs to be in core.

[...]

PR #237, opened 2025-02-12
shacl12-node-expr, shacl12-inf-rules

afs:
This PR adds two documents, one for node expressions and one for inference rules. (The document are empty in this PR. The CSS is lifted from an existing SHACl document.)

See https://github.com/w3c/data-shapes/issues/234#issuecomment-2650440403 for the initial discussion.

The important point of this PR is the short names.
The short names in the initial PR are "shacl-node-expr" and "shacl-inf-rules".
Short names are the path component after `/TR/` so they show up,

We could use longer "short names"
"shacl-node-expressions", "shacl-inference-rules"


HolgerKnublauch:
Shouldn't this include 12, e.g. shacl12-node-expr?

afs:
There is the short name for the series and the short name for each version.

"Short names" are keys in various W3C systems so they get copied around. Changing them later is disruptive.
cc @caribouW3 

I think it might need a WG resolution to point to so the W3C system team can action things. I'm not sure which one is is most important, or maybe both. (This isn't something I have done very often.)

Changed to `*12` for now.

And about the names themselves? Longer or shorter?

HolgerKnublauch:
The names look good to me, as they are supposed to be "short".

HolgerKnublauch:
And if we do the 12 in the name, shouldn't then the folder also be like shacl12-inf-rules?

caribouW3:
> There is the short name for the series and the short name for each version.

Yes, we have both in some cases: when the new version does not supersede the previous one.
Otherwise I think it's fine without a versioned shortname.
 
> "Short names" are keys in various W3C systems so they get copied around. Changing them later is disruptive. cc @caribouW3

True

> I think it might need a WG resolution to point to so the W3C system team can action things. I'm not sure which one is is most important, or maybe both. (This isn't something I have done very often.)

The decision is to be made formally by the WG when we publish the first public Working Draft. We can have different naming in GH (e.g. shacl-node-expr in GH but /TR/shacl-node-expressions when finally published)

simonstey (approved):
I fixed respec errors and aligned the section structure with the one of shacl12-core

robert-david (approved):

shacl-node-expr/.prettierrc
```diff
@@ -0,0 +1,14 @@
+{
+  "useTabs": true,
```
afs: Let's not use tabs. They come out differently for different people/IDEs.
If they get mixed, then indentation can look quite irregular.

Does prettier work with reSpec? 
The usual problem I encounter is the breaking lines inside links - legal but then the link e.g. a `data-cite` is unclear.

HolgerKnublauch (approved):

Issue #224, opened 2025-02-03
What is the SHACL version this group is working to produce?


ajnelson-nist:
I noticed in [Issue 223](https://github.com/w3c/data-shapes/issues/223) that there is discussion of a "SHACL 1.1 ..." deliverable.

Yet, the [deliverables](https://www.w3.org/2024/12/data-shapes.html#deliverables) noted today all describe SHACL 1.2 artifacts.

My understanding is that the 2017 SHACL Recommendation is "1.0", though the HTML documentation does not include the string "1.0," and the Turtle encoding of SHACL also does not include the string "1.0" or an `owl:versionIRI`.

Was there ever a 1.1?  If not, where will the documentation go for why a 1.1 is being skipped?

This feels to me like a "Core"-taggable issue, but I'll leave that decision to the chairs.

HolgerKnublauch:
The intention with 1.2 was

1. to align with RDF 1.2
2. to make sure that this is not just an incremental improvement over what was a 1.1 working draft from the CG. Nobody should mix up the historic 1.1 drafts with the shiny new 1.2 in the future as they have completely new deliverables (e.g. 1.1 SHACL-AF will just disappear and be distributed across multiple 1.2 docs).

ajnelson-nist:
Thank you, @HolgerKnublauch !  That's straightforward enough to me.  Will this be noted in the 1.2 Core document?

With your reply of ACK or NACK, I'm fine with this Issue being closed unless it's helpful as a reminder to include the 1.1-vs-1.2 in whichever document is appropriate.

HolgerKnublauch:
As the 1.1 drafts have no official status, they can be updated to clearly point at their successor documents once they are in a stable state.

From the 1.2 docs, I would suggest we link to the official 1.0 specs where appropriate

```
	previousPublishDate: "2017-07-20",
	previousMaturity: "REC",
	prevRecShortname: "shacl",
	prevRecURI: "https://www.w3.org/TR/shacl/",
```

So I would link from 1.2 to 1.0 and from 1.1 to 1.2 but not from 1.2 to the 1.1 drafts.

If you're OK, this could be closed.

ajnelson-nist:
Your plan sounds great to me.  Thank you for the walkthrough.

caribouW3:
> https://www.w3.org/TR/shacl/

It should use the dated URI instead, because that one will point to the new spec, eventually.

afs:
[...]

PR #244, opened 2025-02-13
Copied most recent TTL files into the new shacl12-vocabularies folder with a comment that these are under development

### LINKED ISSUE #243: Add TTL files for the SHACL-related vocabularies (or decide to keep the existing ones)
The previous WG has produced TTL files shactl.ttl and shacl-shacl.ttl in 

https://github.com/w3c/data-shapes/tree/gh-pages/shacl

As we need to update this (frequently) from now on, we need to decide where we will maintain these files in the new WG. In particular, this is important once we start merging pull requests that add new features, as those PRs should be accompanied by the changes to the TTL files.

I suggest we create a new folder shacl12-vocabularies where the new versions can be placed, to cleanly separate from the historic documents.

HolgerKnublauch:


ajnelson-nist:
@HolgerKnublauch - would you be fine linking this to #243 with the "Development" interface on the right?  We're having enough Issues and PRs pop up that it would be helpful to have GitHub interface hints for which are tied to one another.

afs:
>  would you be fine linking this to #243 with the "Development" interface on the right? We're having enough Issues and PRs pop up that it would be helpful to have GitHub interface hints for which are tied to one another.

Putting the text:
```
Closes #243
```
in the description will close the issue on merge as well.

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

bergos (approved):

YoucTagh:
Should we use `http://www.w3.org/ns/shacl-shacl#` in the ttls or the dated versions as discussed in #224?

HolgerKnublauch:
@YoucTagh where in particular? The namespaces used as prefix should not be dated as they are long-term.

YoucTagh:
For example `http://www.w3.org/ns/shacl-shacl` in: https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L29

or using in `rdfs:seeAlso https://www.w3.org/TR/shacl/#syntax-rule-SHACL-list` in: https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl-shacl.ttl#L19C17-L19C49

HolgerKnublauch:
Ok, thanks. I wouldn't do this as the TTL files don't really have the same publishing cycle as the TR documents, so I don't even see what URLs we could link to here (and the links should be resolvable in the linked data sense).

Issue #148, opened 2023-04-03
Button "Hide SPARQL Definitions" in 1.6 Relationship between SHACL and SPARQL seems to be broken


megoth-capgemini:
The button under section 1.6 Relationship between SHACL and SPARQL does not toggle between hiding and showing SPARQL definitions.

![bilde](https://user-images.githubusercontent.com/122792971/229473293-a45ccdd3-bfbf-4fd3-9d41-4aa691fa5090.png)

HolgerKnublauch:
Thanks for pointing this out. Judging by the error, the page no longer knows about jQuery ($). Maybe that was available to earlier page templates and got lost later.

megoth-capgemini:
Looks like a good assessment. Just to suggest it, I reckon you could use `Document.querySelectorAll()` or similar standard (nowadays) functions in place of the jQuery-specific parts.

HolgerKnublauch:
Fixed, and removed jQuery. Thanks for the report.

PR #245, opened 2025-02-13
#148: Removed use of jQuery and fixed the Hide/Show SPARQL buttons

HolgerKnublauch:


bergos:
@HolgerKnublauch Is there a reason you switched from `pre` to `div` for the shacl-shacl Turtle? It looks like all the examples are using `pre`.

HolgerKnublauch:
Yes, because if it's <pre> the hyperlinks don't work. I have not investigated this in depth but <div> fixed it. Maybe a respec thing.

bergos (approved):

PR #254, opened 2025-02-16
#212: Added sh:ShapeClass

### LINKED ISSUE #212: Add sh:ShapeClass
Related to #169 SHACL should become more stand-alone and grow up in a formerly OWL-dominated world.

In the dash namespace we have added 

```
dash:ShapeClass
  a dash:ShapeClass ;
  rdfs:comment "A class that is also a node shape. This class can be used as rdf:type instead of the combination of rdfs:Class and sh:NodeShape." ;
  rdfs:label "Shape class" ;
  rdfs:subClassOf rdfs:Class ;
  rdfs:subClassOf sh:NodeShape ;
.
```

so that classes using implicit class targets can be written with just one rdf:type instead of two:

```
ex:Person
    a dash:ShapeClass ;
    sh:property ex:Person-birthDate ;
...
```

instead of

```
ex:Person
   a rdfs:Class, sh:NodeShape ;
    sh:property ex:Person-birthDate ;
...
```

This could be leveraged into the sh: namespace as sh:ShapeClass.

HolgerKnublauch:
I know this issue has not been discussed yet, but I have prepared this branch to implement it in case it gets accepted, and also to illustrate how the process for future changes could/should look like. The PR includes the changes to the HTML text, adds at least one test case and updates the TTL file.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1147,6 +1147,10 @@ ex:House a ex:Nephrologist .
 							If <code>s</code> is a <a>SHACL instance</a> of <code>sh:ShapeClass</code> in a <a>shapes graph</a> <code>SG</code>
 							then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
 						</div>
+						<p>
+							Please keep in mind that <code>sh:ShapeClass</code> may not be understood to be a subclass of <code>rdfs:Class</code> by some SHACL-unaware implementations.
+							It is therefore recommended (but not required) that graphs that use <code>sh:ShapeClass</code> include an <code>owl:imports sh:</code> statement.
+						</p>
 						<p><em>The remainder of this section is informative.</em></p>
 						<p>
 							In the following example, <code>ex:Alice</code> is a focus node, because it is a <a>SHACL instance</a> of
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1158,7 +1158,7 @@ WHERE {
 							in <code>SG</code> then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
 						</div>
 						<p id="ShapeClass">
-							The SHACL namespace includes a dedicated metaclass <code>sh:ShapeClass</code> that can serve as a syntactic shortcut for the implicit class targets pattern.
+							The SHACL namespace includes a dedicated class <code>sh:ShapeClass</code> that can serve as a syntactic shortcut for the implicit class targets pattern.
 						</p>
 						<div class="def">
 							<div class="def-header">TEXTUAL DEFINITION</div>
@@ -4501,7 +4501,7 @@ shsh:EntailmentShape
 		<section class="appendix informative" id="changes-12">
 			<h2>Changes between SHACL 1.0 Core and SHACL 1.2 Core</h2>
 			<ul>
-				<li>Added the new metaclass <a href="#ShapeClass"><code>sh:ShapeClass</code></a> as syntactic sugar for implicit class targets, see <a href="https://github.com/w3c/data-shapes/issues/212">Issue 212</a></li>
+				<li>Added the new class <a href="#ShapeClass"><code>sh:ShapeClass</code></a> as syntactic sugar for implicit class targets, see <a href="https://github.com/w3c/data-shapes/issues/212">Issue 212</a></li>
 			</ul>
 		</section>
 		
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1157,6 +1157,16 @@ WHERE {
 							in a <a>shapes graph</a> <code>SG</code> and <code>s</code> is also a <a>SHACL instance</a> of <code>rdfs:Class</code>
 							in <code>SG</code> then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
 						</div>
+						<p id="ShapeClass">
+							The SHACL namespace includes a dedicated metaclass <code>sh:ShapeClass</code> that can serve as a syntactic shortcut for the implicit class targets pattern.
+						</p>
+						<div class="def">
+							<div class="def-header">TEXTUAL DEFINITION</div>
+							The class <code>sh:ShapeClass</code> is an <code>rdfs:subClassOf</code> of both <code>sh:NodeShape</code> and <code>rdfs:Class</code>.
+							<code>sh:ShapeClass</code> has itself as its <code>rdf:type</code>.
+							If <code>s</code> is a <a>SHACL instance</a> of <code>sh:ShapeClass</code> in a <a>shapes graph</a> <code>SG</code>
+							then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
+						</div>
 						<p><em>The remainder of this section is informative.</em></p>
 						<p>
 							In the following example, <code>ex:Alice</code> is a focus node, because it is a <a>SHACL instance</a> of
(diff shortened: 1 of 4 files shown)


shacl12-core/index.html
```diff
@@ -1157,6 +1157,16 @@ <h4>Implicit Class Targets</h4>
 							in a <a>shapes graph</a> <code>SG</code> and <code>s</code> is also a <a>SHACL instance</a> of <code>rdfs:Class</code>
 							in <code>SG</code> then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
 						</div>
+						<p id="ShapeClass">
+							The SHACL namespace includes a dedicated metaclass <code>sh:ShapeClass</code> that can serve as a syntactic shortcut for the implicit class targets pattern.
```
afs: "meta" is unnecessary. A subclass is a subset of it's super.

HolgerKnublauch: Unnecessary, but isn't it helpful to spell this out? Feedback from the 1.0 version was often that the spec is a bit too brief and too formal...

afs: > The class <code>sh:ShapeClass</code> is an <code>rdfs:subClassOf</code> of both <code>sh:NodeShape</code> and <code>rdfs:Class</code>.

For SC to be a "subClassOf" both X1 and X2 means element z of SC is both in X1 and X2.

A node shape (instance in SG) is a member of sh:NodeShape (the class).
The node shape instance is a _member_ of `sh:ShapeClass` because of the subClass relationship.

[The text](https://github.com/w3c/data-shapes/pull/254/files#diff-8b57d0f37f503174858ec5f096b5e33258e9130dda585f0410645c5f8f0de9ccR1164-R1168) at 1164-1168 has  two parts one for SG (instance) and one for DG (set of instances).

I think the definition of sh:ShapeClass needs to involve "SHACL Class" which has the two level SG/DG relationship.

RDF isn't strong on the metaclass/class distinction.

tldr: I'd avoid "meta" :smile:

HolgerKnublauch: sh:ShapeClass is rdfs:subClassOf rdfs:Class, which IMHO makes it a metaclass just like owl:Class

afs: sh:ShapeClass is rdfs:subClassOf rdfs:Class and is a rdfs:subClassOf of sh:NodeShape. 

Both apply to elements of sh:ShapeClass.

"meta" isn't in SHACL 1.0. There would be other places to add it for consistency.

HolgerKnublauch: It's fine. I'll take it out.

shacl12-core/index.html
```diff
@@ -4721,12 +4743,19 @@ <h2>Acknowledgements</h2>
...
 			</p>
 			<ul>
 				<li><b>2024-02-14</b>: New work started by cloning the main SHACL spec and splitting it into SHACL Core and SHACL-SPARQL</li>
 			</ul>
 		</section>
+
+		<section class="appendix informative" id="changes-12">
+			<h2>Changes between SHACL 1.0 Core and SHACL 1.2 Core</h2>
+			<ul>
+				<li>Added the new class <a href="#ShapeClass"><code>sh:ShapeClass</code></a> as syntactic sugar for implicit class targets, see <a href="https://github.com/w3c/data-shapes/issues/212">Issue 212</a></li>
```
afs: It has behavior so "syntactic sugar" is misleading
```suggestion
				<li>Added the new class <a href="#ShapeClass"><code>sh:ShapeClass</code></a> for implicit class targets, see <a href="https://github.com/w3c/data-shapes/issues/212">Issue 212</a></li>
```

shacl12-core/index.html
```diff
@@ -1241,6 +1241,16 @@ <h4>Implicit Class Targets</h4>
 							in a <a>shapes graph</a> <code>SG</code> and <code>s</code> is also a <a>SHACL instance</a> of <code>rdfs:Class</code>
 							in <code>SG</code> then the set of <a>SHACL instances</a> of <code>s</code> in a data graph <code>DG</code> is a <a>target</a> from <code>DG</code> for <code>s</code> in <code>SG</code>.
 						</div>
+						<p id="ShapeClass">
+							The SHACL namespace includes a dedicated class <code>sh:ShapeClass</code> that can serve as a syntactic shortcut for the implicit class targets pattern.
+						</p>
+						<div class="def">
+							<div class="def-header">TEXTUAL DEFINITION</div>
+							The class <code>sh:ShapeClass</code> is an <code>rdfs:subClassOf</code> of both <code>sh:NodeShape</code> and <code>rdfs:Class</code>.
+							<code>sh:ShapeClass</code> has itself as its <code>rdf:type</code>.
```
VladimirAlexiev: This is not true.
```
sh:ShapeClass rdfs:subClassOf rdfs:Class , sh:NodeShape. # true
sh:ShapeClass a rdfs:Class. # true
sh:ShapeClass a sh:NodeShape. # FALSE!
```

VladimirAlexiev (changes requested):

shacl12-core/index.html
HolgerKnublauch: But the triple sh:ShapeClass rdf:type sh:ShapeClass is explicitly stated in the TTL file:

https://github.com/w3c/data-shapes/pull/254/files#diff-e7ad12ee6ccf9c41e835d5d12955c739ac6ead85384719f9a3cd0127f7a1c5fdR58

afs: In the textual definition, the sentence 

> `sh:ShapeClass` has itself as its `rdf:type`.

has two readings: as part of the definition and being an authoritative statement, or as stating a consequence of the first part. Replace the preceeding `.` with `;` and it does make it part of the definition.

> The class sh:ShapeClass is an rdfs:subClassOf of both sh:NodeShape and of rdfs:Class; sh:ShapeClass has itself as an rdf:type.

("an rdf:type" - it has unrelated ones as well)


The current 3rd sentence, which is definitely not adding to the definition, and is stating a consequence of "SHACL instance":

> If s is a [SHACL instance](https://github.com/w3c/data-shapes/pull/254#dfn-shacl-instance) of sh:ShapeClass in a [shapes graph](https://github.com/w3c/data-shapes/pull/254#dfn-shapes-graphs) SG then the set of [SHACL instances](https://github.com/w3c/data-shapes/pull/254#dfn-shacl-instance) of s in a data graph DG is a [target](https://github.com/w3c/data-shapes/pull/254#dfn-targets) from DG for s in SG.

could be moved to after the textual definition box so the definition box is all definition.

VladimirAlexiev: Why should `sh:ShapeClass rdf:type sh:ShapeClass` be true?

afs: As I read it - because the definition says it is.

If it is the other reading where `sh:ShapeClass rdf:type sh:ShapeClass` is an implication, it does not need to be in the definition box. I don't see the implication path that concludes that fact but maybe I've missed it.

HolgerKnublauch: Ok, what about I just remove that sentence? We do not make these assertions explicit in the spec of other things (e.g. we don't say that sh:NodeShape rdf:type rdfs:Class. This information is present in the sh: namespace TTL and that's probably good enough?

HolgerKnublauch: I have removed the offending sentence and made sh:ShapeClass a rdfs:Class in the TTL file as this was not really necessary. I hope this addresses the concern here and will ask for another review

afs: @HolgerKnublauch --

Are you saying it is a natural conclusion from the rest of SHACL? What path leads to that?

If on the other hand, it is a necessary additional requirement about `sh:ShapeClass` then it should be in the definition. (i.e the definition should be complete and self-contained without the namespace document.)

Concretely - what requires `sh:ShapeClass rdf:type sh:ShapeClass` to work?

HolgerKnublauch: >  what requires sh:ShapeClass rdf:type sh:ShapeClass to work?

Nothing really, so I have taken it out and made it rdfs:Class only. Maybe our messages crossed paths.

VladimirAlexiev (approved):

shacl12-core/index.html
VladimirAlexiev: Thanks!

(`sh:ShapeClass` is not a `sh:NodeShape` because it doesn't define any properties.)

ajnelson-nist:
Uh oh - I don't think this PR was ready to be accepted.  It's slated for a discussion Monday.

Issue #259, opened 2025-02-18
Could we drop SHACL-for-SHACL (shsh) from the Core spec?


HolgerKnublauch:
The SHACL 1.0 doc and current Core draft contains an appendix showing the whole shsh: 

https://www.w3.org/TR/shacl/#shacl-shacl

My preference is to keep this outside of the Core document and instead give it a separate publication cycle through its TTL file only.

- Removing it will shorten the Core spec
- We can keep the Appendix but just as a link to the TTL file
- The intro of the Appendix already states that the graph may be wrong and that it cannot overrule the textual definitions, so it's not really normative anyway
- I find it adds too little value to give it so much space
- It only covers a subset of SHACL Core, so will be hard to update when we introduce the other documents (thus a separate publication cycle is better)
- It is not covered by test cases
- It was written (by me) within a few hours a few days before the 1.0 publication to make one distinguished reviewer happy. A proper design would introduce better shapes that could, for example, define a suggested rendering as form layout, potentially unifying SHACL use across tools
- Requiring to maintain it takes away resources from more important tasks.

caribouW3:
In the Recommendation it does not seem to be linked from anywhere in the spec text. It's not mentioned as either normative nor non-normative clearly in the appendix title, but it does not seem to be normative but as @HolgerKnublauch commented, the spec texts takes precedence.
so keeping this as a separate document, maintained in GH, seems a good idea.

HolgerKnublauch:
Ok, PR created but I'd like to hear from anyone with concerns in a WG meeting.

ajnelson-nist:
> Ok, PR created but I'd like to hear from anyone with concerns in a WG meeting.

@HolgerKnublauch - I started the agenda for next Monday, [here](https://github.com/w3c/data-shapes/blob/agenda/admin/agenda-2025-02-24.md), and primed it with your Issue and my scribing.  Would you mind cross-linking the PR on this Issue and adding any agenda text to describe the topic?

caribouW3:
Minutes of Feb 24th discussion: https://www.w3.org/2025/02/24-data-shapes-minutes.html#716f

PR #261, opened 2025-02-18
#259: Removed full shacl-shacl TTL snippet, outsourced as a link

HolgerKnublauch:
This should be discussed and confirmed by the WG


shacl12-core/index.html
```diff
@@ -4262,424 +4255,13 @@ <h2>Summary of SHACL Syntax Rules</h2>
...
-				that indicates an error in the following shapes graph.
-				Please see the <a href="https://www.w3.org/2017/shacl/errata">Errata Page</a> for an enumeration and analysis of possible errors that have been reported.
-        		This shapes graph is available at <a href="http://www.w3.org/ns/shacl-shacl">http://www.w3.org/ns/shacl-shacl</a>.
-        		That version may be more up-to-date than this specification as errata are noted against this specification.
+        		The shapes graph available at <a href="https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl-shacl.ttl">http://www.w3.org/ns/shacl-shacl</a>
+				is intended to enforce many of the syntactic constraints related to SHACL Core in this specification. 
+				As such, it can be understood as a machine-readable version of a subset of those constraints.
+			</p>
+			<p class="todo">
+				TODO: Replace the hyperlink to the official site once that has been updated.
```
afs, line 4263: The new file can be included in this PR.

The TR is `/TR/shacl/index.html`. Associated files go in the same directory. The link is `href="./shacl-shacl.ttl"`.

HolgerKnublauch: Ok thanks. Here my preference is to decouple the release cycles of these files, so that the shacl-shacl is not just about core but also about any future doc such as node expressions. This means it would be more like a living standard and live outside of Core.

ajnelson-nist: I would think that SHACL-SHACL's review scope should cover every class and predicate with the `sh:` prefix, so it makes sense to me that shacl-shacl would live outside of core.  Would it get a standalone document accompanying the TTL file?  If so, I'd guess it would be a pretty short document.

HolgerKnublauch:
I believe this PR was approved in the meeting yesterday but I am formally unable to merge it without an approval here.

caribouW3 (approved):
Minutes of yesterday's meeting: https://www.w3.org/2025/02/24-data-shapes-minutes.html

bergos (approved):

Issue #222, opened 2025-01-29
Node Expressions


afs:
This issue is for general discussion of _Node Expressions_.

[Node Expressions in SHACL-AF (May 2017)](https://w3c.github.io/data-shapes/shacl-af/#node-expressions).

[SHACL CG draft with a longer list of possible node expressions](https://w3c.github.io/shacl/shacl-af/index.html#node-expressions)

Wiki doc: [Proposal on Node Expressions for SHACL Core 1.2](https://github.com/w3c/data-shapes/wiki/Proposal-on-Node-Expressions-for-SHACL-Core-1.2)

Use cases: 
* Issues [labelled "Inferencing and UCR"](https://github.com/w3c/data-shapes/issues?q=is%3Aissue%20state%3Aopen%20label%3Aucr%20label%3AInferencing)

Please add use cases, small or large, as [issues](https://github.com/w3c/data-shapes/issues/) and label them "Inferencing" and also "UCR". We can do further organisation later if the volume is high enough.

The full work on this area is phase2.

However, some of the ideas also apply to to SHACL core in phase1. We need to know what is the vision of "node expressions" so that phase1 documents are aligned.

tpluscode:
I have been using Node Expressions extensively for generating SPARQL queries from shapes: https://shape-to-query.hypermedia.app/docs/

afs:
@HolgerKnublauch 

On the description page, I don't understand the second `edg:Database-tableCount` example. It uses `sh:path` on the property shape but it does not evaluate to a path. What would the new SHACL-core definition of sh:path look like?

```turtle
edg:Database-tableCount
    a sh:PropertyShape ;
    sh:inferredProperty edg:tableCount ;
    sh:path [                       ## ????
        sh:count [
            sh:path [
                sh:inversePath edg:tableOf ;
            ] ;
        ] ;
    ] .
    sh:datatype xsd:integer ;
    sh:description "The number of tables in this database, automatically computed." ;
```

From the description of it, this seems to be a prototype/template for a first example `edg:Database-tableCount` - a "property shape" but it does not have a `sh:path` (= reachability for value nodes), it has `sh:inferredProperty` instead and values of the object. `sh:inferredProperty`.

```turtle
edg:Database-tableCount
    a sh:DerivedPropertyShape ;
    sh:inferredProperty edg:tableCount ;
    sh:values [
        sh:count [
            sh:path [
                sh:inversePath edg:tableOf ;
            ] ;
        ] ;
    ] .
[...]

Issue #181, opened 2023-03-07
Inverse Property Value Rules


tpluscode:
At the moment SHACL-AF only defines the shorthand syntax and behaviour of Property Value Rules when the path is an IRI

I think it should also be allowed to have property rules set on properties whose property is `sh:InversePath`. That can be trivially achieved by switching places of `sh:object` and `sh:subject` of the implicit triple rule:

```
8.5.2 Inverse Property Value Rules

Similar to [Property Value Rule](#8.5.2) but with `sh:object` and `sh:subject` reversed, 
that is for the implicit rule `T`:

- T sh:subject V
- T sh:object sh:this
```

HolgerKnublauch:
Sounds plausible.

tpluscode:
Would you accept PR, actually?

HolgerKnublauch:
Have you implemented this and encountered problems? I do remember that I thought about this but didn't go there because that inverse case caused complications. I guess it is because they may produce "cycles" where inferring one "forward" value may later also infer an "inverse" value. Or, at the minimum, I didn't see enough value to include that feature, although from a symmetry point of view it would make sense.

What is your specific use case?

tpluscode:
I don't remember exact details now but I found this missing when implementing a SPARQL Query generator

HolgerKnublauch:
In my implementation of sh:path node expressions, the system will automatically "infer" properties that have a sh:values rule. If we were to add this for the inverse direction, it would at least complicate these cases. Unless there is a strong use case and evidence that this has a reasonable cost-benefits ratio, I would rather table this for a future WG.

tpluscode:
> I would rather table this for a future WG.

Fair enough

HolgerKnublauch:
With Node Expressions directly used at sh:path, I believe this ticket can be closed, @tpluscode?

tpluscode:
Can you point to an example of what you have in mind?

HolgerKnublauch:
https://github.com/w3c/data-shapes/blob/issue-222-node-expressions/shacl12-core/index.html#L1640

What you are originally asking here assumes that there is a separate property that will be inferred (because sh:values was separate from the sh:path). Above these two are linked.

The SHACL inference rules spec may introduce a new property sh:alias or sh:inferredProperty that would materialize the derived properties. Maybe that could be inverse too and your ticket remains valid?

tpluscode:
[...]

PR #274, opened 2025-02-25
#222: New (better) definition of node expressions

HolgerKnublauch:
I have revisited my earlier branch and believe this one here is a better starting point for node expressions.

In particular, this definition no longer makes property paths a subset of node expressions but leaves the two concepts separate. However, the values of sh:path can either be property paths or node expressions.

This design means that there is less confusion between node expressions and property path syntaxes:
- IRIs would otherwise either be constant node expressions or serve as IRI property paths
- rdf:Lists would otherwise either be list node expressions or sequence paths.

The node expressions spec should introduce a node expression type [ sh:path XY ] that allows the use of property paths as node expressions, creating a much cleaner design.

This is now a PR, open for review and comments. There are a couple of TODOs that can be addressed later.

Summary of changes:

- Moved non-validating property shape properties to a separate section at the end
- Added a top-level section SHACL Expressions
- Moved definition of property paths into that new section (basically unchanged from 1.0)
- Added sub-section on Node Expressions
- Generalized sh:targetNode to allow Node Expressions instead of just constants
- Generalized sh:path to allow property paths OR node expressions (in that order, to make sure that IRIs and Lists are handled as paths, not node expressions)
- Updated the definition of values nodes for the extended semantics of sh:path


shacl12-core/index.html
```diff
@@ -2232,20 +2193,35 @@ <h4>Severity (sh:resultSeverity)</h4>
...
+					</p>
+					<p>
+						If <code>p</code> is a <a>node expression</a> the value nodes are the set of <a>output nodes</a> of <code>p</code>
+						when evaluated with the given <a>focus node</a> and the <a>data graph</a> as <a>active graph</a>.
+					</p>
+					<p>
+						Unless stated otherwise, the value of <code>sh:resultPath</code> of each validation result is a "deep copy"
+						of <code>p</code> and any <a>triples</a> that can be reached by transitively traversing the <a>blank nodes</a>
+						that appear in the <a>object</a> position of these triples,
+						see the <a href="https://www.w3.org/submissions/CBD/">Consise Bounded Description</a>.
```
afs: I don't understand what this means.

HolgerKnublauch: How can I make it clearer? The intent is that the validation report is self-contained. So whenever you have something like sh:path [ sh:inversePath ex:prop ] in a property shape, then that whole blank node needs to be present including any nested triples.

afs: Do you have an example where this text makes a difference from SHACL 1.0?

HolgerKnublauch: I cannot think of a difference, except for node expressions, which were not in 1.0 anyway. For property paths the requirement was already like this. But happy to be proven wrong.

shacl12-core/index.html
```diff
@@ -1546,7 +1538,8 @@ <h3>Property Shapes</h3>
 					<span data-syntax-rule="PropertyShape">A <dfn data-lt="property shapes">property shape</dfn> is a <a>shape</a> in the <a>shapes graph</a>
 					that is the <a>subject</a> of a <a>triple</a> that has <code>sh:path</code> as its <a>predicate</a>.</span>
 					<span data-syntax-rule="path-maxCount">A shape has at most one <a>value</a> for <code>sh:path</code>.</span>
-					<span data-syntax-rule="path-node">Each <a>value</a> of <code>sh:path</code> in a shape must be a <a>well-formed</a> <a>SHACL property path</a>.</span>
+					<span data-syntax-rule="path-node">Each <a>value</a> of <code>sh:path</code> in a shape must be a <a>well-formed</a> <a>SHACL property path</a>
+					or (if it is not a <a>well-formed</a> <a>SHACL property path</a>) a <a>well-formed</a> <a>node expression</a> but not a <a>literal</a>.</span>
```
tpluscode: As said, keeping Node Expressions and Property Paths mutually exclusive should simplify this definition

shacl12-core/index.html
tpluscode: > For property paths the requirement was already like this.

Kinda. It says that `sh:resultPath` must point at a well-formed SHACL property path. Easy to overlook and even the original shacl-js library only copied a single blank node without the rest of the path.

It's good to spell it out explicitly

tpluscode (commented):
> The node expressions spec should introduce a node expression type [ sh:path XY ] that allows the use of property paths as node expressions, creating a much cleaner design.

So just like [Path Expression](https://w3c.github.io/shacl/shacl-af/#node-expressions-path) in SHACL-AF 1.1 draft?

> Generalized sh:targetNode to allow Node Expressions instead of just constants

What do you think about also generalizing all Targets so that Node Expressions could be used with `sh:targetClass` etc, and even custom targets.

> Generalized sh:path to allow property paths OR node expressions (in that order, to make sure that IRIs and Lists are handled as paths, not node expressions)

I think the emphasis is unnecessary, unless you intend to deviate from the current definition of Node Expressions. All of them are required to be blank nodes, so IRIs must be interpreted as Predicate Path anyway. And lists? I don't see how a list can be a Node Expression either, thus it can only be a Sequence Path. Am I missing something?

Technically, an IRI would be a Constant Term Expression too, right? Should we then merge its definition with Predicate Path?

afs:
This should be part of the FPWD.

Separating path expressions and node expressions is better. 

1/ Say somewhere "path expressions" are "node expressions" introduced by `sh:path`.

2/ For path expressions (and only path expressions) keep the SPARQL equivalents.
The shalc12-sparql is going to be about whole SPARQL queries: `sh:select` and `sh:ask`.


These comments based on reading the node expressions section.

3/ "Node Expression type"

"type" is word that is used a lot in the document and here there is no `rdf:type`

"node expression types" -> "node expression functions" or "function name"
(which goes with functions have parameters)

4/ Example 15: Why is this a property shape?

This seems an artificial distinction. Keep `sh:path` for path uses.

The only case seems to be derived properties (to be in shacl12-node-expr?)
and that could be handled by saying a property shape has exactly one of "sh:path or sh:derivedPath".

NE are for "property values" (objects) and path calculations.
Call the node expression a "value shape"?
Or is it a Node Shape?

5/ "Node expression types can declares one or more node expression parameters."

Is that as a declaration written in SHACL/RDF? and that declaration is required?
Or as their definition/function signature (which could be a textual description) can talk about parameters?
If former, every NE needs to have a paired definition resource which seems a bit of a burden.

6/ Will shacl12-core or shalc12-node-expr define how to write custom node expression functions?

7/ Option "input focus node" - the terminology needs defining.

8/ No `sh:this`? This is more general than an input focus node. It can be used as a parameter value to functions like `fn:string-join`.

shacl12-core/index.html
HolgerKnublauch: Thanks, I missed that place further down. Fixed: https://github.com/w3c/data-shapes/commit/e508b68152fa988fcbcaa0850c4e279b218a470d

In order to keep track, I'll mark this discussion as resolved - please reopen if you disagree.

shacl12-core/index.html
HolgerKnublauch: Yes, which they are (mutually exclusive) so unless I hear otherwise I mark this as resolved.

HolgerKnublauch:
@afs thanks for your feedback. I almost wish you had broken the comment down into multiple snippets to make them easier to respond to, but I'll process them individually.

HolgerKnublauch:
> Say somewhere "path expressions" are "node expressions" introduced by sh:path.

Do you mean [ sh:path ex:somePath ] as node expression as defined in https://w3c.github.io/shacl/shacl-af/#node-expressions-path ? I don't want to introduce this because this would be the (first and only) real node expression type in core and thus have implications for all core implementations. For example we would then need to add test cases for the use of node expressions in sh:targetNode.

HolgerKnublauch:
> 2/ For path expressions (and only path expressions) keep the SPARQL equivalents.
> The shalc12-sparql is going to be about whole SPARQL queries: sh:select and sh:ask.

Yes the SPARQL equivalents are still around.

HolgerKnublauch:
> 3/ "Node Expression type"
> "type" is word that is used a lot in the document and here there is no rdf:type
> "node expression types" -> "node expression functions" or "function name"
> (which goes with functions have parameters)

It's difficult to find a suitable term here. The term function is used by SPARQL functions and MAY be added to SHACL-SPARQL using sh:Function, see https://w3c.github.io/shacl/shacl-af/#functions

I believe down the road some users may want to explicitly add rdf:type triples to their node expressions, eg.

    [
        a sh:SelectNodeExpression ;
        sh:select "SELECT * { ... }" ;
    ]

in which case using the term "type" is not too far off. In the SHACL TTL file we may actually introduce all of these types explicitly, together with their parameter declarations. Then those type triples would fit naturally. The choice of word here tries to keep these options open in the future, but we can always revisit once we have started work on the node expr document.

I am open to other terms except type and function.

HolgerKnublauch:
> 4/ Example 15: Why is this a property shape?
> This seems an artificial distinction. Keep sh:path for path uses.
> The only case seems to be derived properties (to be in shacl12-node-expr?)
> and that could be handled by saying a property shape has exactly one of "sh:path or sh:derivedPath".
> NE are for "property values" (objects) and path calculations.
> Call the node expression a "value shape"?
> Or is it a Node Shape?

We may need higher bandwidth to discuss this. I don't quite understand the intent here.
Example 15 is a property shape because this is exactly the main use case of node expressions to derive properties.
IMHO the Node expr  spec should introduce a dedicated property such as sh:inferredProperty or sh:inferredInverseProperty which point to a forward property IRI or an inverse IRI, not a path.

Unless there are other topics maybe we can use the Monday meeting to go through this and finalize the design?

tpluscode:
> I am open to other terms except type and function.

Node Expression Kind?

HolgerKnublauch:
> 5/ "Node expression types can declares one or more node expression parameters."
> Is that as a declaration written in SHACL/RDF? and that declaration is required?
> Or as their definition/function signature (which could be a textual description) can talk about parameters?
> If former, every NE needs to have a paired definition resource which seems a bit of a burden.

We can handle this like in the constraint components: each NE type would have a textual definition of the parameters. The TTL file can declare them too, and/or the shacl-shacl file can declare shapes to validate them. For example, there could be a shape sh:SelectNodeExpressionShape sh:targetSubjects sh:select and sh:datatype xsd:string for sh:select.

HolgerKnublauch:
> 6/ Will shacl12-core or shalc12-node-expr define how to write custom node expression functions?

This would be cool, and it sounds like it could go into shacl12-sparql for SPARQL-based node expressions. In fact, the sh:Function stuff in SHACL-AF is already one such extension point.

HolgerKnublauch:
> 7/ Option "input focus node" - the terminology needs defining.

What in particular? Currently it reads 

    <li>An optional <a>node</a> called the <dfn>input focus node</dfn>.</li>

which defines that this is an argument to the evaluation. Maybe it should become a formal evalExpr(activeGraph, inputFocusNode, { params } ) expression?

Would it make sense to accept the current PR at some stage and then work on these refinements individually?

HolgerKnublauch:
> 8/ No sh:this? This is more general than an input focus node.
> It can be used as a parameter value to functions like fn:string-join.

Ouch, this is an issue. When I say that IRIs are interpreted as the set of exactly that IRI, then we cannot add the special handling of sh:this later. But if I introduce sh:this then all Core engines would need to support that special case everywhere we allow node expressions, we would need test cases for all this etc.

Note there is another potential exception: rdf:nil could become the empty list, also so that we can introduce blank node lists with rdf:first as key parameter as a new NE type.

One option is to say that SHACL Core implementations do not allow sh:this and rdf:nil as node expressions.

Another option is to instead introduce a blank node to access the input focus node, like any other variable: [ sh:varName "focusNode" ] or a new datatype "focusNode"^^sh:var

Or we open the floodgates and define some node expressions in Core (which would be against the original goal of minimal changes).

Any other idea?

afs:
> > 3/ "Node Expression type"
> > "type" is word that is used a lot in the document and here there is no rdf:type
> > "node expression types" -> "node expression functions" or "function name"
> > (which goes with functions have parameters)
> 
> It's difficult to find a suitable term here. The term function is used by SPARQL functions and MAY be added to SHACL-SPARQL using sh:Function, see https://w3c.github.io/shacl/shacl-af/#functions
> 
> I believe down the road some users may want to explicitly add rdf:type triples to their node expressions, eg.
> 
> ```
> [
>     a sh:SelectNodeExpression ;
>     sh:select "SELECT * { ... }" ;
> ]
> ```
> 
> in which case using the term "type" is not too far off.

This branch text says : "A node expression is a function with exactly one node expression type" 

SPARQL functions are `sh:SPARQLFunction` - I can't find `sh:Function`.
Aren't "SPARQL functions" subsumed by defining node expressions?

To me that is a good example for *not* using type. `sh:select` is not being used in the way that types occur in RDF which is the object of `rdf:type`.


`sh:select` is the property giving the relationship of the `[ ]` to a string to be understood as SPARQL code.
It can be seen as being in the role of a function name.

"node expression name"

"kind" sort of works but the word appears quite a lot already, not just in `sh:nodeKind`.

afs:
> I almost wish you had broken the comment down into multiple snippets to make them easier to respond to, but I'll process them individually.

OK - next time, I'll do that.
(The notes were written offline while reading the section)

Maybe we can enable GH discussions, which have threading.

In this vein, review is hard because it is not clear what has changed across the whole document.  Code diffs are opaque because reformatting and new content looks the same.

tpluscode:
> In this vein

... I almost wish we had this document in markdown ...

HolgerKnublauch:
Ok, Andy's argument that sh:SPARQLFunction may just become a kind of node expressions has convinced me that the term function is not too bad, and the "type" is the function name (also represented by an IRI but not used as rdf:type). Unless I hear otherwise, I'll update the branch in the next few days.

afs:
> Would it make sense to accept the current PR at some stage and then work on these refinements individually?

Yes. "This should be part of the FPWD."

The items need recording. I think we're at the stage of deciding what the issues to raise actually are. The most important ones at the moment are those that are about general approach and material across documents. There has been some concern expressed in the group about too many issues. Maybe that is seeing issues as specific work items.

caribouW3:
> > Would it make sense to accept the current PR at some stage and then work on these refinements individually?
> 
> Yes. "This should be part of the FPWD."

It would not be ideal to merge during the Call for Consensus, but please respond to the CfC email stating that you wish to see this PR merged if it is what you mean (it's unclear to me whether you wish to merge it as is or refine that work before FPWD)

afs:
>  (it's unclear to me whether you wish to merge it as is or refine that work before FPWD)

My concern is around the fact that merging a PR closes it and the discussion becomes implicitly frozen.

My preference is that FPWD gives the best indication of the direction of the working group. Node expressions are a major new feature.

Ideally, list issues in the FPWD (ReSpec:: `<p class="issue" data-number="222"> ...`) but I recognize that may slow down getting FPWD published which isn't desirable either and getting a prompt FPWD is more important. I can put in some separate issues if that helps.

ajnelson-nist:
Re @HolgerKnublauch (from a ways back in the thread):

> [...]
> I believe down the road some users may want to explicitly add rdf:type triples to their node expressions, eg.
> [...]

FWIW, I am one of those folks that prefers the opportunity to add `rdf:type` whenever practical, for the sake of self-documenting code.  My initial learning curve for SHACL was, I believe, steeper, because `rdf:type sh:NodeShape` and `rdf:type sh:PropertyShape` not being required led to me not appreciating their differences for some time.  Nor did I quickly realize when I was quietly using `sh:NodeShape`s (like with `sh:not [ sh:class ex:Foo ]`).

Now, this is not me advocating for **requiring** `rdf:type` on node expressions.  I just want to openly appreciate the _opt-in_ opportunity to go a little overkill for the sake of self-explaining code.

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+				<p>
+					This section introduces the concept of <a>node expressions</a> as an extension point used by other specifications such as <a href="shacl12-sparql">SHACL-SPARQL</a>.
+					This SHACL Core specification does not define any <a>node expression types</a> but defines how they are generally handled.
+					<span class="todo">TODO: Add link to shacl12-node-expr once that is stable.</span>
+				</p>
+				<div class="def">
+					<div class="def-header">TEXTUAL DEFINITION</div>
+					A <dfn>node expression</dfn> is a function with exactly one <dfn>node expression type</dfn> as its type.
+					Each <a>node expression type</a> has an <a>IRI</a>.
+					Node expression types can declares one or more <dfn>node expression parameters</dfn>.
```
ajnelson-nist: Nit: "can declares"

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					One of these parameters can be the <dfn>key parameter</dfn> that uniquely identifies the node expression type.
+					<br />
+					<br />
+					During <dfn>evaluation</dfn> a node expression takes the following input:
+					<ul>
+						<li>A <a>graph</a>, called the <dfn>active graph</dfn>.</li>
+						<li>An optional <a>node</a> called the <dfn>input focus node</dfn>.</li>
+						<li>One <a>node</a> as value for each of the declared parameters, optional for any but the key parameter unless stated otherwise in the definition of the node expression type.</li>
+					</ul>
+					The result of the evaluation of a node expression is a list of <a>nodes</a> (possibly empty and with duplicates) called the <dfn>output nodes</dfn>.
```
ajnelson-nist: _List_ of nodes, or _set_ of nodes?  The "possibly ... with duplicates" point would be moot for a set.  If a list, would there be conditions under which an order of returned nodes could be guaranteed?

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					</p>
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:Person-fullName
+	a sh:PropertyShape ;
+	sh:name "full name" ;
+	sh:path <b>[
+		sh:select """
+			PREFIX ex: &lt;http://example.org/ns#&gt;
+			SELECT ?fullName
```
ajnelson-nist, line 1790: In 1.0 uses of SHACL-SPARQL, the `SELECT` clause binds `$this`.  How should this named variable `?fullName` be bound in SHACL?

shacl12-core/index.html
```diff
@@ -2232,20 +2194,35 @@ <h4>Severity (sh:resultSeverity)</h4>
...
-				</ul>
+					</p>
+					<p>
+						If <code>p</code> is a <a>node expression</a> the value nodes are the set of <a>output nodes</a> of <code>p</code>
+						when evaluated with the given <a>focus node</a> and the <a>data graph</a> as <a>active graph</a>.
+					</p>
+					<p>
+						Unless stated otherwise, the value of <code>sh:resultPath</code> of each validation result is a "deep copy"
+						of <code>p</code> and any <a>triples</a> that can be reached by transitively traversing the <a>blank nodes</a>
+						that appear in the <a>object</a> position of these triples,
```
ajnelson-nist: Nit on the line below: s/Consise/Concise/

ajnelson-nist:
> > 8/ No sh:this? This is more general than an input focus node.
> > It can be used as a parameter value to functions like fn:string-join.
> 
> Ouch, this is an issue. When I say that IRIs are interpreted as the set of exactly that IRI, then we cannot add the special handling of sh:this later. But if I introduce sh:this then all Core engines would need to support that special case everywhere we allow node expressions, we would need test cases for all this etc.
> 
> Note there is another potential exception: rdf:nil could become the empty list, also so that we can introduce blank node lists with rdf:first as key parameter as a new NE type.
> 
> One option is to say that SHACL Core implementations do not allow sh:this and rdf:nil as node expressions.
> 
> Another option is to instead introduce a blank node to access the input focus node, like any other variable: [ sh:varName "focusNode" ] or a new datatype "focusNode"^^sh:var
> 
> Or we open the floodgates and define some node expressions in Core (which would be against the original goal of minimal changes).
> 
> Any other idea?

I saw this in the diffs:

```patch
+               <section id="expressions">
+                       <h2>SHACL Expressions</h2>
+                       <p><em>The introduction of this section is informative.</em></p>
+                       <p>
+                               This section introduces the concepts of <a href="#property-paths">SHACL property paths</a>
+                               and <a href="#node-expressions">node expressions</a>.
+                               Informally, both are mechanisms to produce a list of nodes, typically based on a given focus node as input.
```

"Based on a given focus node as input" sets me up to expect `sh:this`.  But I think I see where @afs asked after `sh:this`:

```patch
+                                       <span data-syntax-rule="node-expression-iri">A node expression that is an <a>IRI</a> evaluates to the list consisting of exactly this IRI.</span>
```

`sh:this` here would probably be a concern for exactly the SHACL-SHACL graph.  I don't know under what other conditions someone would want to use `sh:this` meaning the node `sh:this` rather than the focus node.  But, SHACL-SHACL is a pretty significant use case of SHACL.

Seeing this line of the change log gave me an idea on how to resolve this, which I could've swore came up before somewhere among the Issues or PRs, but I can't find it:

```patch
+                               <li>Introduced <a>node expressions</a> as an extension point to dynamically compute lists of nodes. Generalized sh:targetNode and sh:path to support node expressions.</li>
```

I suspect we truly have a reason to introduce `sh:targetNodes` (plural).  `sh:this` could have a different behavior as the object of `targetNode` vs. `targetNodes`:

* `sh:targetNode sh:this` - truly the node `sh:this`, which SHACL-SHACL could use.
* `sh:targetNodes sh:this` - the subject node, i.e. taking `sh:this` and "resolving" (functionally-executing?) it.

shacl12-core/index.html
HolgerKnublauch: Yes it should be a list because then people can do sh:orderBy and sh:limit, e.g. when produced by sh:select

shacl12-core/index.html
HolgerKnublauch: The engine will just return all bindings for the first variable in the result set. But details of this would be written up in the SHACL-SPARQL spec.

shacl12-core/index.html
afs: `sh:order By` involves coordination between two lists. That is going to need some work. It is pushing the boundaries of what a node expression is.

```ttl
[
    sh:nodes [ sh:path ex:child ] ;
    sh:orderBy [ sh:path rdfs:label ] ;
] .
```
`sh:nodes` is a list of `?child`, not a list of pairs `(?child , ?label)`

`[ sh:path rdfs:label ]` is another list that might be a different length and/or may have matched so it does not arrive in the same sequence as `[ sh:path ex:child ]` so they can't be lined up to produce pairs.

```
[ sh:orderBy [ sh:path ex:child ] ]
```
may be more appropriate for NE without SPARQL.

HolgerKnublauch: Yes I am not particularly fond of the old definition of sh:orderBy.

shacl12-core/index.html
```suggestion
					<span data-syntax-rule="path-node">Each <a>value</a> of <code>sh:path</code> in a shape MUST be
					either a <a>well-formed</a> <a>SHACL property path</a>
					or a <a>well-formed</a> <a>node expression</a>;
					i.e., not a <a>literal</a>.</span>
```

shacl12-core/index.html
```diff
@@ -1555,6 +1548,11 @@ <h3>Property Shapes</h3>
 					<a>focus node</a> either by directly following a given property (specified as an <a>IRI</a>) or any other <a>SHACL property path</a>,
 					specified using <code>sh:path</code>.
 				</p>
+				<p>
+					If the <code>sh:path</code> is not a <a>well-formed</a> property path, it must be a <a>node expression</a> and the constraints will
+					be validated against the <a>output nodes</a> of this node expression.
+					Note that support for node expressions is not required by SHACL Core, but for extensions such as [[shacl12-sparql]].
```
```suggestion
					If the <code>sh:path</code> is not a <a>well-formed</a> property path, it must be a <a>node expression</a>
					and the constraints will be validated against the <a>output nodes</a> of this node expression.
					Note that support for node expressions is not required by SHACL Core,
					but is necessary for extensions such as [[shacl12-sparql]].
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
 						</div>
 					</div>
 				</aside>
+			</section>
+		</section>
+
+		<section id="expressions">
+			<h2>SHACL Expressions</h2>
+			<p><em>The introduction of this section is informative.</em></p>
```
TallTed: It would be better to always use distinct (sub)sections, where the parent (e.g., `id="expressions"`) is normative (informative sections may not have normative subsections), and each subsection (e.g., `id="expressions-intro"`, `id="property-paths"`, etc.) is labeled normative or informative, as appropriate.

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+			<p>
+				This section introduces the concepts of <a href="#property-paths">SHACL property paths</a>
+				and <a href="#node-expressions">node expressions</a>.
+				Informally, both are mechanisms to produce a list of nodes, typically based on a given focus node as input.
```
```suggestion
				Informally, each is a mechanism to produce a list of nodes, typically based on a given focus node as input.
```

shacl12-core/index.html
```diff
@@ -1111,13 +1111,13 @@ <h3>Targets</h3>
...
-							<span data-syntax-rule="targetNode-nodeKind">Each <a>value</a> of <code>sh:targetNode</code> in a shape is either an <a>IRI</a> or a <a>literal</a>.</span>
+							<span data-syntax-rule="targetNode-nodeKind">Each <a>value</a> of <code>sh:targetNode</code> in a shape is a well-formed <a>node expression</a>.</span>
 						</p>
 						<div class="def">
 							<div class="def-header">TEXTUAL DEFINITION</div>
 							If <code>s</code> is a <a>shape</a> in a <a>shapes graph</a> <code>SG</code> and <code>s</code> has
-							<a>value</a> <code>t</code> for	<code>sh:targetNode</code> in <code>SG</code>
-							then <code>{ t }</code> is a <a>target</a> from any data graph for <code>s</code> in <code>SG</code>.
+							<a>value</a> <code>expr</code> for <code>sh:targetNode</code> in <code>SG</code>
+							then the <a>output nodes</a> of <code>expr</code> are <a>targets</a> for the data graph <code>DG</code> as <a>active graph</a>.
```
```suggestion
							<a>value</a> <code>expr</code> for <code>sh:targetNode</code> in <code>SG</code>,
							then the <a>output nodes</a> of <code>expr</code> are <a>targets</a>
							for the data graph <code>DG</code> as <a>active graph</a>.
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+				<section id="property-path-predicate">
+					<h4>Predicate Paths</h4>
+					<p class="syntax">
+						A <dfn>predicate path</dfn> is an <a>IRI</a>.
+					</p>
 					<p>
-						SHACL includes RDF terms to represent the following subset of <a>SPARQL property paths</a>:
-						<code>PredicatePath</code>, <code>InversePath</code>, <code>SequencePath</code>, <code>AlternativePath</code>,
-						<code>ZeroOrMorePath</code>, <code>OneOrMorePath</code> and <code>ZeroOrOnePath</code>.
+						If <code>p</code> is a <a>predicate path</a> then <code>path(p,G)</code> is a SPARQL <code>PredicatePath</code> with <code>p</code> as <code>iri</code>.
```
```suggestion
						If <code>p</code> is a <a>predicate path</a>, then <code>path(p,G)</code> is
						a SPARQL <code>PredicatePath</code> with <code>p</code> as <code>iri</code>.
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+						A <dfn>sequence path</dfn> is a <a>blank node</a> that is a <a>SHACL list</a>
+						with at least two <a>members</a> and each member is a <a>well-formed</a> SHACL property path.
 					</p>
 					<p>
-						The following sub-sections provide syntax rules of <a>well-formed</a> <a>SHACL property paths</a>
-						together with mapping rules to <a href="https://www.w3.org/TR/sparql12-query/#pp-language">SPARQL 1.2 property paths</a>.
-						These rules define the <dfn>path mapping</dfn> <code>path(p,G)</code> in an RDF graph <code>G</code> of an RDF term <code>p</code> that is a SHACL property path in <code>G</code>.
-						Two SHACL property paths are considered <dfn data-lt="equivalent path">equivalent paths</dfn> when they map to the exact same SPARQL property paths.
+						If <code>p</code> is a <a>sequence path</a> in <code>G</code> with list <a>members</a>
+						<code>v<sub>1</sub></code>, <code>v<sub>2</sub></code>, ..., <code>v<sub>n</sub></code>
```
```suggestion
						<code>v<sub>1</sub></code>, <code>v<sub>2</sub></code>, ..., <code>v<sub>n</sub></code>,
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					<h4>Alternative Paths</h4>
+					<p class="syntax" data-syntax-rule="path-alternative">
+						An <dfn>alternative path</dfn> is a <a>blank node</a> that is the subject of exactly one triple in <code>G</code>.
+						This triple has <code>sh:alternativePath</code> as predicate, <code>L</code> as object,
+						and <code>L</code> is a <a>SHACL list</a> with at least two <a>members</a>
+						and each member of <code>L</code> is a <a>well-formed</a> SHACL property path.
 					</p>
 					<p>
-						The following example illustrates some valid SHACL property paths, together with their SPARQL 1.2 equivalents.
+						If <code>p</code> is an <a>alternative path</a> in <code>G</code> then, for the members of its SHACL list <code>L</code>:
```
```suggestion
						If <code>p</code> is an <a>alternative path</a> in <code>G</code>,
						then, for the members of its SHACL list <code>L</code>:
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					<p class="syntax" data-syntax-rule="path-inverse">
+						An <dfn>inverse path</dfn> is a <a>blank node</a> that is the <a>subject</a> of exactly one <a>triple</a> in <code>G</code>.
+						This triple has <code>sh:inversePath</code> as predicate, and the <a>object</a> <code>v</code> is a <a>well-formed</a> SHACL property path.
+					</p>
 					<p>
-						While the previous sections introduced properties that represent validation conditions,
-						this section covers properties that are ignored by SHACL processors.
-						The use of these so-called <dfn data-lt="non-validating property">non-validating properties</dfn> is entirely optional and not subject to formal interpretation contracts.
-						They MAY be used for purposes such as form building or predictable printing of RDF files.
+						If <code>p</code> is an <a>inverse path</a> in <code>G</code> then <code>path(p,G)</code> is a
```
```suggestion
						If <code>p</code> is an <a>inverse path</a> in <code>G</code>, then <code>path(p,G)</code> is a
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
-	rdfs:label "Name" .
+				</section>
+				<section id="property-path-zero-or-more">
+					<h4>Zero-Or-More Paths</h4>
+					<p class="syntax" data-syntax-rule="path-zero-or-more">
+						A <dfn>zero-or-more path</dfn> is a <a>blank node</a> that is the <a>subject</a> of exactly one <a>triple</a> in <code>G</code>.
+						This triple has <code>sh:zeroOrMorePath</code> as <a>predicate</a>, and the <a>object</a> <code>v</code> is a <a>well-formed</a> SHACL property path.
+					</p>
+					<p>
+						If <code>p</code> is a <a>zero-or-more path</a> in <code>G</code> then <code>path(p,G)</code> is a
```
```suggestion
						If <code>p</code> is a <a>zero-or-more path</a> in <code>G</code>, then <code>path(p,G)</code> is a
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					</p>
+				</section>
+				<section id="property-path-one-or-more">
+					<h4>One-Or-More Paths</h4>
+					<p class="syntax" data-syntax-rule="path-one-or-more">
+						A <dfn>one-or-more path</dfn> is a <a>blank node</a> that is the <a>subject</a> of exactly one <a>triple</a> in <code>G</code>.
+						This triple has <code>sh:oneOrMorePath</code> as <a>predicate</a>, and the <a>object</a> <code>v</code> is a <a>well-formed</a> SHACL property path.
+					</p>
+					<p>
+						If <code>p</code> is a <a>one-or-more path</a> in <code>G</code> then <code>path(p,G)</code> is a
```
```suggestion
						If <code>p</code> is a <a>one-or-more path</a> in <code>G</code>, then <code>path(p,G)</code> is a
```

shacl12-core/index.html
```diff
@@ -1587,267 +1585,230 @@ <h3>Property Shapes</h3>
...
+					</p>
+				</section>
+				<section id="property-path-zero-or-one">
+					<h4>Zero-Or-One Paths</h4>
+					<p class="syntax" data-syntax-rule="path-zero-or-one">
+						A <dfn>zero-or-one path</dfn> is a <a>blank node</a> that is the <a>subject</a> of exactly one <a>triple</a> in <code>G</code>.
+						This triple has <code>sh:zeroOrOnePath</code> as <a>predicate</a>, and the <a>object</a> <code>v</code> is a <a>well-formed</a> SHACL property path.
+					</p>
+					<p>
+						If <code>p</code> is a <a>zero-or-one path</a> in <code>G</code> then <code>path(p,G)</code> is a
```
```suggestion
						If <code>p</code> is a <a>zero-or-one path</a> in <code>G</code>, then <code>path(p,G)</code> is a
```

shacl12-core/index.html
```diff
@@ -2232,20 +2194,35 @@ <h4>Severity (sh:resultSeverity)</h4>
...
+				</section>
+				<section id="value-nodes-property-shapes">
+					<h4>Value Nodes of Property Shapes</h4>
+					<p>
 						For <a>property shapes</a> with a <a>value</a> for <code>sh:path</code> <code>p</code> the
-						<a>value nodes</a> are the set of <a>nodes</a> in the <a>data graph</a>
+						<a>value nodes</a> are defined as follows.
+					</p>
+					<p>
+						If <code>p</code> is a <a>SHACL property path</a> the value nodes are the set of <a>nodes</a> in the <a>data graph</a>
```
```suggestion
						If <code>p</code> is a <a>SHACL property path</a>, then its value nodes are the set of <a>nodes</a> in the <a>data graph</a>
```

shacl12-core/index.html
```diff
@@ -2232,20 +2194,35 @@ <h4>Severity (sh:resultSeverity)</h4>
...
 						that can be reached from the <a>focus node</a> with the <a>path mapping</a> of <code>p</code>.
-						Unless stated otherwise, the value of <code>sh:resultPath</code> of each validation result is a
-						<a>SHACL property path</a> that <a href="#property-paths">represents</a> an <a>equivalent path</a> to the one provided in the shape.
-					</li>
-				</ul>
+					</p>
+					<p>
+						If <code>p</code> is a <a>node expression</a> the value nodes are the set of <a>output nodes</a> of <code>p</code>
```
```suggestion
						If <code>p</code> is a <a>node expression</a>, then its value nodes are the set of <a>output nodes</a> of <code>p</code>
```

shacl12-core/index.html
```diff
@@ -2232,20 +2194,35 @@ <h4>Severity (sh:resultSeverity)</h4>
...
+						when evaluated with the given <a>focus node</a> and the <a>data graph</a> as <a>active graph</a>.
+					</p>
+					<p>
+						Unless stated otherwise, the value of <code>sh:resultPath</code> of each validation result is a "deep copy"
+						of <code>p</code> and any <a>triples</a> that can be reached by transitively traversing the <a>blank nodes</a>
+						that appear in the <a>object</a> position of these triples,
+						see the <a href="https://www.w3.org/submissions/CBD/">Concise Bounded Description</a>.
```
```suggestion
						that appear in the <a>object</a> position of these triples.
						See the <a href="https://www.w3.org/submissions/CBD/">Concise Bounded Description</a>.
```

shacl12-core/index.html
```diff
@@ -4298,6 +4275,141 @@ <h4>sh:in</h4>
 				</section>
 			</section>
 		</section>
+				
+		<section id="nonValidation" class="informative">
+			<h2>Non-Validating Property Shape Characteristics</h2>
+			<p>
+				While the previous sections introduced properties that represent validation conditions,
+				this section covers properties that are ignored by SHACL processors.
+				The use of these so-called <dfn data-lt="non-validating property">non-validating properties</dfn> is entirely optional and not subject to formal interpretation contracts.
```
```suggestion
				The use of these so-called <dfn data-lt="non-validating property">non-validating properties</dfn>
				is entirely optional and is not subject to formal interpretation contracts.
```

shacl12-core/index.html
```diff
@@ -4298,6 +4275,141 @@ <h4>sh:in</h4>
...
+				They MAY be used for purposes such as form building or predictable printing of RDF files.
+			</p>
+			<section id="name">
+				<h3>sh:name and sh:description</h3>
+				<p>
+					Property shapes may have one or more <a>values</a> for <code>sh:name</code> to provide human-readable labels for the property in the target where it appears.
+					If present, tools SHOULD prefer those locally specified labels over globally specified labels at the <code>rdf:Property</code> itself.
+					For example, if a form displays a node that is in the target of a given property shape with an <code>sh:name</code>, then the tool SHOULD use the provided name.
+					Similarly, property shape may have values for <code id="description">sh:description</code> to provide descriptions of the property in the given context.
+					Both <code>sh:name</code> and <code>sh:description</code> may have multiple <a>values</a>, but should only have one <a>value</a> per language tag.
```
```suggestion
					Property shapes may have one or more <a>values</a> for <code>sh:name</code>
					to provide human-readable labels for the property in the target where it appears.
					If present, tools SHOULD prefer such locally specified labels
					over globally specified labels at the <code>rdf:Property</code> itself.
					For example, if a form displays a node that is in the target of a given property shape
					with an <code>sh:name</code>, then the tool SHOULD use the provided name.
					Similarly, property shapes may have values for <code id="description">sh:description</code>
					to provide descriptions of the property in the given context.
					Both <code>sh:name</code> and <code>sh:description</code> may have
					multiple <a>values</a>, but should only have one <a>value</a> per language tag.
```

shacl12-core/index.html
```diff
@@ -4298,6 +4275,141 @@ <h4>sh:in</h4>
...
+				</p>
+			</section>
+			<section id="order">
+				<h3>sh:order</h3>
+				<p>
+					Property shapes may have one <a>value</a> for the property <code>sh:order</code> to indicate the relative order of the property shape for purposes such as form building.
+					The values of <code>sh:order</code> are decimals.
+					<code>sh:order</code> is not used for validation purposes and may be used with any type of subjects.
+					If present at property shapes, the recommended use of <code>sh:order</code> is to sort the property shapes in an ascending order, for example so that
+					properties with smaller order are placed above (or to the left) of properties with larger order.
```
```suggestion
					Property shapes may have one <a>value</a> for the property <code>sh:order</code>
					to indicate the relative order of the property shape for purposes such as form building.
					The values of <code>sh:order</code> are decimals.
					<code>sh:order</code> is not used for validation purposes
					and may be used with any type of subjects.
					If present at property shapes, the recommended use of <code>sh:order</code> is
					to sort the property shapes in an ascending order, for example so that
					properties with smaller order are placed above (or to the left)
					of properties with larger order.
```

shacl12-core/index.html
```diff
@@ -4298,6 +4275,141 @@ <h4>sh:in</h4>
...
+				</p>
+			</section>
+			<section id="group">
+				<h3>sh:group</h3>
+				<p>
+					Property shapes may link to an <a>SHACL instance</a> of the class <code>sh:PropertyGroup</code> using the property <code>sh:group</code> to indicate that
+					the shape belongs to a group of related property shapes.
+					Each group may have additional triples that serve application purposes, such as an <code>rdfs:label</code> for form building.
+					Groups may also have an <code>sh:order</code> property to indicate the relative ordering of groups within the same form.
```
TallTed: "an `rdfs:label`" because it reads "an arr-dee-eff-ess". But does `sh:group` read "an ess-aych-group" or "a shh-group"?
```suggestion
					Property shapes may link to a <a>SHACL instance</a> of the class <code>sh:PropertyGroup</code>
					using the property <code>sh:group</code> to indicate that
					the shape belongs to a group of related property shapes.
					Each group may have additional triples that serve application purposes,
					such as an <code>rdfs:label</code> for form building.
					Groups may also have an <code>sh:order</code> property to indicate
					the relative ordering of groups within the same form.
```

shacl12-core/index.html
```diff
@@ -4298,6 +4275,141 @@ <h4>sh:in</h4>
...
+				</p>
+			</section>
+			<section id="defaultValue">
+				<h3>sh:defaultValue</h3>
+				<p>
+					Property shapes may have a single value for <code>sh:defaultValue</code>.
+					The default value does not have fixed semantics in SHACL, but MAY be used by user interface tools to pre-populate input widgets.
+					The value type of the <code>sh:defaultValue</code> should align with the specified <code>sh:datatype</code> or <code>sh:class</code> of the same shape.
```
```suggestion
					The default value does not have fixed semantics in SHACL,
					but MAY be used by user interface tools to pre-populate input widgets.
					The value type of the <code>sh:defaultValue</code> SHOULD align with
					the specified <code>sh:datatype</code> or <code>sh:class</code> of the same shape.
```

shacl12-core/index.html
```diff
@@ -4380,6 +4492,7 @@ <h2>Revision History</h2>
 		<section class="appendix informative" id="changes-12">
 			<h2>Changes between SHACL 1.0 Core and SHACL 1.2 Core</h2>
 			<ul>
+				<li>Introduced <a>node expressions</a> as an extension point to dynamically compute lists of nodes. Generalized sh:targetNode and sh:path to support node expressions.</li>
```
```suggestion
				<li>Introduced <a>node expressions</a> as an extension point to dynamically compute lists of nodes. Generalized <code>sh:targetNode</code> and <code>sh:path</code> to support node expressions.</li>
```

TallTed (changes requested):
Punctuation and typo fixes, adjusted line lengths in source, and a couple of open questions.

shacl12-core/index.html
HolgerKnublauch: The revision would lose an important aspect in the ordering: a parser needs to *first* check if it's a property path and only then check for node expression. This is because IRIs as properties have precedence over IRIs as node expression, and lists as sequence paths have precedence over lists as node expressions.

Also, "i.e. not a literal" is not right here - literals are also node expressions but they are explicitly excluded here because there is no use case for a constant literals to be the sh:path, and it would be a change to Core.

shacl12-core/index.html
HolgerKnublauch: Not sure if this would really be better. I think this is a matter of preference.

HolgerKnublauch:
On the topic of sh:this (and potentially rdf:nil) I suggest we consistently switch to blank nodes and treat all IRIs as constants. Only because sh:this was suggested in SHACL-AF does not mean that we must carry it forward.

(Besides, the name sh:this is not intuitive as it's really the focus node. I picked sh:this because sh:focusNode is already used in the results vocabulary.)

So: Instead of sh:this we could add a general variable scope mechanism, allowing engines to pass nodes into the evaluation algorithm. Other examples beside the focusNode could be the URI of the shapes graph, the URI of the current shape or whatever else we may need in the future. Then, accessing a variable value could be a blank node

    [ sh:var "focusNode" ]

The other special case could have been rdf:nil, for the empty list. However, this would mean that people can no longer use rdf:nil for other purposes, e.g. sh:targetNode. So instead, we could simply use an empty blank node:

    []

for the empty list. With these assumptions, we can define these two node expressions in the Node Expr document and leave Core untouched. In particular, Core engines wouldn't need to make any changes to support node expressions in 1.2 Core.

bergos (changes requested):
I also struggle a bit with the term *node expression type*.

I'm not sure if the core specification should not define any node expression type, as I don't know where `sh:call` could be placed otherwise. But that's a topic for a separate PR. I created an [issue](https://github.com/w3c/data-shapes/issues/279) for it so it doesn't get lost between the other non-core node expression topics.

shacl12-core/index.html
tpluscode: > The revision would lose an important aspect in the ordering: a parser needs to first check if it's a property path and only then check for node expression. This is because IRIs as properties have precedence over IRIs as node expression, and lists as sequence paths have precedence over lists as node expressions.

This is what I meant in my earlier [comment](https://github.com/w3c/data-shapes/pull/274#discussion_r1972411305). You said yourself that Property Paths and Node Expressions are mutually exclusive. Thus, it should not matter in which order an implementation considers them

tpluscode:
> (Besides, the name sh:this is not intuitive as it's really the focus node. I picked sh:this because sh:focusNode is already used in the results vocabulary.)

Overall I may be biased as I've used SHACL-AF extensively but I do admit that it is sometimes a bit difficult to assume the right Focus Node when Focus Node Expression is the default in some other expressions.

> So: Instead of sh:this we could add a general variable scope mechanism, allowing engines to pass nodes into the evaluation algorithm.

Off the top of my head, this sounds like it is going to make the overall solution more complicated to use. Could you show what that would look like in a practical example?

shacl12-core/index.html
HolgerKnublauch: @tpluscode how would this then be able to distinguish IRIs and sequence paths (assuming blank node rdf:Lists also become a node expression function)?

So, I may have misunderstood you, the syntax of property paths and NEs does overlap.

HolgerKnublauch:
I have just pushed an update to these definitions, which is hopefully much cleaner. I have introduced more formal definitions of the terms, replaced node expression type with function (name), added an official eval function and then introduced the two built-in expression types for IRIs and Literals. I have generalized the notion of focusNode variable to a "scope" which IMHO is a good investment into the future and generalizes how people can access variables (such as focusNode).

tpluscode:
> On the topic of sh:this (and potentially rdf:nil) I suggest we consistently switch to blank nodes and treat all IRIs as constants

In the SHACL-AF draft, `sh:this` is the only exception which is not a blank node, right? So if we switch it to blank node, as you say. And maybe rename to `sh:focusNode` for clarity? 

For example, updating this [Triple Rule example](https://w3c.github.io/shacl/shacl-af/#x8-1-examples-of-shacl-rules)

```diff
ex:Rectangle
	sh:rule [
		a sh:TripleRule ;
-		sh:subject sh:this ;
+		sh:subject [ sh:focusNode () ] ;
		sh:predicate rdf:type ;
		sh:object ex:Square ;
		sh:condition ex:Rectangle ;
		sh:condition [
			sh:property [
				sh:path ex:width ;
				sh:equals ex:height ;
			] ;
		] ;
	] .
```

The object of `sh:focusNode` would be irrelevant, or... we switch the whole Node Expressions to lists, similar to Function Expressions. But that's not for now to discuss.

tpluscode:
And now that I look at my example above, it would indeed be possible to circle back to `sh:this` (or any other name) as per @afs suggestion that it should be possible to have "named node expressions"

```turtle
sh:this [
  a sh:NodeExpression ;
  sh:values [ sh:focusNode () ] ;
] .
```

This way allowing `sh:subject sh:this` all over again, without the special handling in the spec

HolgerKnublauch:
@tpluscode please check if you can live with the new design that introduces a general "scope" mechanism to pass in variables. We could then introduce [ sh:this () ] as you suggest, as syntactic sugar for [ sh:var "focusNode" ].

Or, we limit sh:LiteralExpression to all literals except those with datatype sh:var so that variables could be accessed using "focusNode"^^sh:var. I prefer the blank node though, for consistency.

Note that using [ sh:focusNode () ] could clash with the use of sh:focusNode as predicate in the results graph - those would then appear as potential node expressions.

In general, I believe a scope mechanism is a good thing. In the future, there will be more scenarios where people will want to invoke node expressions, not just at sh:path and sh:targetNode but to support any kind of computation on graphs.

shacl12-core/index.html
tpluscode: The only place for potential conflict is `sh:path`?

```
[
  sh:property [ 
    sh:path ex:foo # Predicate Path or Constant Node Expression
  ] ; 
  sh:property [ 
    sh:path ( ex:foo ex:bar ) # Sequence Path or List Node Expressions
  ] ; 
] .
```

_What is a **List Node Expression** anyway?_

We can revisit this definition later but my gut tells me that the syntax to Property Paths and Node Expressions better be unambiguous.

tpluscode: > `sh:path ex:foo # Predicate Path or Constant Node Expression`

Here, the difference appears irrelevant. The behaviour is the same, the value would become the property's path in both case.

> `sh:path ( ex:foo ex:bar ) # Sequence Path or List Node Expressions`

I guess the equivalent of List NE would be an Alternative Path? Here, the unambiguity could be avoided by having something like this

> `sh:path [ sh:values ex:foo, ex:bar ]`

I didn't even use a list because order does not matter

HolgerKnublauch: List Node Expression is not in the spec but I have it locally implemented. It would have rdf:first as key parameter and rdf:rest to point at the rest, so basically a normal rdf list as blank node.

sh:path ex:foo  as property path is very different from sh:path ex:foo as node expression. The latter would return ex:foo as value node, while the former would return all values of ex:foo as value nodes.

sh:path (ex:a ex:b) as a property path would be the SPARQL ex:a/e:b sequence, while if interpreted as an rdf:List it would have ex:a and ex:b as value nodes.

Constant IRIs as node expressions are very important because we can then say, for example, that the values of sh:class must be node expressions and any existing sh:class ex:Foo shapes graph would work as before even if the engine treats ex:Foo as a node expression.

shacl12-core/index.html
afs: In this case, drop the `<p><em>...</em></p>`. 

I don't see anything in the intro that would have an effect. The intro is a overview.

Could s/Informally, both/Both/ if that worries you.

(Not this PR:: It looks to me like most uses of "Informally, " in the doc are unnecessary. Descriptive text can go in normative sections. The use of definition boxes makes it clear which text is the most significant.)

afs:
Constants can be thought of as functions of zero arguments that return themselves. It is functions all the way down!

Written directly is a short-hand form for `[ :constant () ]` and we may end up with places where the short-hand form does not work but that's OK - it's "syntax" not "meaning".

afs:
About scope: on an initial reading:

Is like environment variables? A fixed map, not updated within an expression. We need this (focus node, logical time of day, locale, ...) and it is like XQuery/XPath "[dynamic context](https://www.w3.org/TR/xpath-3/#eval_context)".

Maybe only a fixed set of function-like operations can do this (sh:path would change the focus node) - the focus node value is set on function call and reset of the way out for `sh:path` and possibly certain defined URIs.

Or is it a general mutable binding scope and _any_ NE function can set (remove?) arbitrary variables, reset on the way out of a function (probably this is the same as passing named arguments between functions). Are scope settings visible after the function returns? (in-out parameters).

Or is a global, mutable map (issues with the order of evaluation in parallel)?

* +1 to the environment case. This seems like a minimum to have.
* +1 to some specific, defined modifications during execution - e.g. the focus node moves.
* Cautious +1 to named arguments "maybe, probably can be made to work"

For the general mutable binding scope, some things look to be complicated (for users composing functions, and for implementation/optimization/introspection) because now functions have a form of side-effect.
I'd want to understand the use cases more for this.

HolgerKnublauch:
@afs on scope, I have no strong opinion. I would be happy to simplify it and hard-code just for the focus node if this allows us to move forward. The branch here has been going on for very long already...

I do remember needing variables in the past but since they were not supported I have found other hacks and forgotten details now.

The scope should IMHO be like from a stack-based programming language, where each node expression can expand the previous scope without overwriting it, e.g. by defining additional variables or overriding values such as the focusNode. So basically, it would become a linked chain of ChildScope objects, with a Map each and getVar() would look for the first Map that defines it.

Use cases off the top of my head:
- Pass in a limit depending on user preferences or UI needs - sh:limit could do COALESCE(var "limit", value of sh:limit)
- Pass in a preferred locale/language of the user, for something like a label function that returns the most suitable label for the user.
- Pass in the current user and its permissions
- Pass in the URI of the shapes graph so that it can be queried from SPARQL which by default executes on the data graph

We solve some of these use cases currently with ThreadLocals and other hacks. An official approach might be better but, again, I have no strong opinion and can also see arguments to keep it as simple as possible.

HolgerKnublauch:
On the focusNode variable, it occurred to me that the name "focus node" is quite overloaded and may not be the best choice here. For example, sh:path expressions can shift the focus node to something else during the evaluation. Therefore, would it be better to call that node "active node" or "current node"? This would avoid the conflicts with either $this or sh:focusNode which are already used elsewhere. It would then open up using [ sh:activeNode () ] as a built-in function, avoiding the legacy of sh:focusNode and sh:this.

shacl12-core/index.html
```diff
@@ -1538,8 +1539,8 @@ <h3>Property Shapes</h3>
 					<span data-syntax-rule="PropertyShape">A <dfn data-lt="property shapes">property shape</dfn> is a <a>shape</a> in the <a>shapes graph</a>
 					that is the <a>subject</a> of a <a>triple</a> that has <code>sh:path</code> as its <a>predicate</a>.</span>
 					<span data-syntax-rule="path-maxCount">A shape has at most one <a>value</a> for <code>sh:path</code>.</span>
-					<span data-syntax-rule="path-node">Each <a>value</a> of <code>sh:path</code> in a shape must be a <a>well-formed</a> <a>SHACL property path</a>
-					or (if it is not a <a>well-formed</a> <a>SHACL property path</a>) a <a>well-formed</a> <a>node expression</a> but not a <a>literal</a>.</span>
+					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape MUST be a <a>well-formed</a> <a>SHACL property path</a>
+					or (if it is not a <a>well-formed</a> <a>SHACL property path</a>) a <a>well-formed</a> <a>node expression</a>, but not a <a>literal</a>.</span>
```
TallTed: Responding to https://github.com/w3c/data-shapes/pull/274#discussion_r1974902004

> The revision would lose an important aspect in the ordering: a parser needs to _first_ check if it's a property path and only then check for node expression. This is because IRIs as properties have precedence over IRIs as node expression, and lists as sequence paths have precedence over lists as node expressions.

I think it is an error to _imply_ what a parser should do and in what order, rather than _explicitly stating_ such things. The "precedence order" you provide in this comment should likewise be *stated in the spec*.

It might be worthwhile to write out your complete thinking of how a parser should do its job, which might be included as an appendix (if it is informative, as I expect) or a section late in the document (if it is normative, as your comment suggests it might be).

> Also, "i.e. not a literal" is not right here — literals are also node expressions but they are explicitly excluded here because there is no use case for a constant literals to be the `sh:path`, and it would be a change to Core.

Ah, OK. It might be good to also work some of the above paragraph into the text. Revised suggestion (albeit without that inclusion) follows...

```suggestion
					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape MUST be
					either a <a>well-formed</a> <a>SHACL property path</a>
					or a <a>well-formed</a> <a>node expression</a> that is not a <a>literal</a>.</span>
```

HolgerKnublauch: I have added extra text to clarify the execution order. I hope this addresses your concerns so I am marking this conversation resolved. Please reopen if you disagree.

HolgerKnublauch:
I have re-requested reviews from those who have marked this PR as Request Changes. Any of those is currently blocking this PR. I hope we can make progress in the meeting on Monday.

afs (approved):

afs:
I've approved the PR.
I hope it makes the FPWD.

There are still points to discuss but there is diminishing returns on fine-tuning this PR.  We won't know whether the design is "right" until we try it in the wider evolution of SHACL.

shacl12-core/index.html
```diff
@@ -1539,8 +1539,13 @@ <h3>Property Shapes</h3>
 					<span data-syntax-rule="PropertyShape">A <dfn data-lt="property shapes">property shape</dfn> is a <a>shape</a> in the <a>shapes graph</a>
 					that is the <a>subject</a> of a <a>triple</a> that has <code>sh:path</code> as its <a>predicate</a>.</span>
 					<span data-syntax-rule="path-maxCount">A shape has at most one <a>value</a> for <code>sh:path</code>.</span>
-					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape MUST be a <a>well-formed</a> <a>SHACL property path</a>
-					or (if it is not a <a>well-formed</a> <a>SHACL property path</a>) a <a>well-formed</a> <a>node expression</a>, but not a <a>literal</a>.</span>
+					<br/><br/>
+					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape is either a <a>well-formed</a>
+					<a>SHACL property path</a> or a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-expr-order">The syntax rules for <a>SHACL property paths</a> take precedence over <a>node expressions</a>,
+					for example so that <a>IRIs</a> are interpreted as <a>predicate path</a> instead of <a>IRI expression</a>.</span>
```
```suggestion
					<span data-syntax-rule="path-expr-order">The syntax rules for <a>SHACL property paths</a> take precedence over <a>node expressions</a>;
					for example, this means that <a>IRIs</a> are interpreted as <a>predicate path</a> instead of <a>IRI expression</a>.</span>
```

TallTed (commented):
I think the tweak above is worth making, now or later. Please turn it to an issue if you prefer not to merge it in.

bergos (approved):

shacl12-core/index.html
```diff
@@ -1587,266 +1592,275 @@ <h3>Property Shapes</h3>
...
+					One of these parameters can be the <dfn>key parameter</dfn> that uniquely identifies the <a>function name</a>.
+				</div>
+				<p>
+					The following example illustrates the syntax of a node expression, used as a blank node value of <code>sh:path</code>:
+				</p>
+				<aside class="example" title="A dynamically computed property using a node expression based on a SPARQL query">
+					<p>
+						Here is an example use of a node expression based on <a href="shacl12-sparql">SHACL-SPARQL</a>, computing
+						the values of a property shape for the property "full name" as the concatenation of <code>ex:firstName</code>,
+						a space and the <code>ex:lastName</code>.
```
tpluscode, line 1780: Sorry for dragging this PR further. I re-read this example when looking at #181 and I misunderstand something.

In SHACL-AF, the values inferred for a property (denoted by `sh:path`, as usual) were the result of Node Expression at `sh:values`.

Below, the value of `sh:path` is a Node Expression instead. The Expression returns the values of full name (say, a single-element list `( "Tomasz Pluskiewicz" )`). What then is the inferred triple?

The description says something like

```turtle
$this ?prop "Tomasz Pluskiewicz" . # <- what is ?prop ?
```

Did you intend to keep `sh:values` and have some `sh:path` here?

I'd expect a Node Expression at `sh:path` to compute the property. Not the values

HolgerKnublauch: @tpluscode the intent is for node expressions to behave like property path expressions, namely to compute some value nodes starting at the focus node. For example, sh:zeroOrMorePath also doesn't compute a property but produces the value nodes that can be reached by transitive traversal.

Core does NOT infer triples here. It only computes derived property values for the duration of the (validation) process. However, I expect that the Rule Inf document will define how triples can be derived from it. As this is a fundamentally different architecture than just deriving value nodes, it deserves its own document, alongside other types of rules that infer triples.

So, assuming Rule Inf declares the property sh:inferProperty then the complete example would be

```
    ex:Person-fullName
	a sh:PropertyShape ;
	sh:name "full name" ;
	sh:path <b>[
		sh:select """
			PREFIX ex: <http://example.org/ns#>
			SELECT ?fullName
			WHERE {
				$this ex:firstName ?firstName .
				$this ex:lastName ?lastName .
				BIND (CONCAT(?firstName, " ", ?lastName) AS ?fullName) .
			}
		"""
	] ;
        sh:inferProperty ex:fullName ;
	sh:datatype xsd:string .
```

Does this make sense now?

afs: This is a longer term discussion.
We should merge this PR for FPWD.

I also find this use of `sh:path` odd. It is generating the objects of the property : property-values.

For me, `sh:path` indicates navigating the data graph.

We could define a property shape as having a `sh:path` XOR `sh:inferProperty` (which could take an NE to calculate the property IRI).

tpluscode: Of course, I do not mean to block the merging of this PR.

But to me too, it is a little confusing for anyone coming  from SHCL 1.1 and very confusing for those familiar with SHACL-AF

Not necessary now, but I would like to see this reconsidered or at least clarified.

HolgerKnublauch: I also thought about introducing a separate property to avoid misleading assumptions about sh:path not really just being a path. We can do that in principle - it's just a syntactic difference.

However, I don't see how node expressions can only be used to compute property IRIs. This would basically make them useless. How would something like "full name" (or the current use cases of sh:values) be computed otherwise, what would happen with node expressions that return more than one value.

I think we need to get that out of the way before merging the PR.

tpluscode: > However, I don't see how node expressions can only be used to compute property IRIs.

Yes, I can see your point and I do agree. The small problem here may be that I (others?) understood Property Paths in SPARQL pattern sense: `$this ?path ?value`. Not the functional way of Node Expressions. I see now what your reply to #181 meant.

> Does this make sense now?

In fact it does! I was preconditioned to think about Node Expressions in terms of rules. Now I see how the new design makes new use cases possible. Without `sh:inferProperty`, as the example is current written, I only now realise it means to validate that the values of `?fullName` returned from the query are validated against the Property Shape's constraints. Very clean. Only... subtly different from SPARQL 1.1 on an uncanny level.

I have nothing against going forward with the current example. I will only ask to revisit the phrasing and/or the example itself to add the necessary clarity.

tpluscode: Side note about `sh:inferProperty`. My proposal would be `sh:materialise` or `sh:materialiseAs`

YoucTagh (approved):

shacl12-core/index.html
robert-david: I think introducing a separate property for NE is a good idea to avoid confusion. What about the mentioned sh:derivedPath? Or sh:derivedNodes, sh:inferredNodes?

afs: > Side note about `sh:inferProperty`. My proposal would be `sh:materialise` or `sh:materialiseAs`

I read "materialise" as coming with an implication of creating the triple (forward-chaining and/or actually inserting the triple into the data graph) and not "virtual" triples (backward-chaining; on demand matching). Something neutral would be better - I prefer "derived".

tpluscode: >I think introducing a separate property for NE is a good idea to avoid confusion. What about the mentioned sh:derivedPath? Or sh:derivedNodes, sh:inferredNodes?

I do like the the new logic explained by @HolgerKnublauch because it unifies the utility of Node Expressions across shapes.

If we were to introduce a new predicate, I would propose that it works as outline in the PR, and keep `sh:path` for backward-compatibility. Might treat it as alias or keep only Property Paths as allowed values.

Think, in the lines of `sh:nodeSelector`/`sh:selectNodes`.

```turtle
sh:selectNodes 
  a rdf:Property ;
  # rdfs:range sh:PropertyPathOrNodeExpression ?
.

sh:path 
  rdfs:subPropertyOf sh:nodeSelector ;
  # rdfs:range sh:PropertyPath ?
  # owl:deprecated true ?
.
```

tpluscode (approved):

ajnelson-nist (approved):
I am fine with this merging, even as part of FWPD, because I believe we need to look deeper at Node Expressions.

shacl12-core/index.html
```diff
@@ -1542,8 +1542,8 @@ <h3>Property Shapes</h3>
 					<br/><br/>
 					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape is either a <a>well-formed</a>
 					<a>SHACL property path</a> or a <a>well-formed</a> <a>node expression</a>.</span>
-					<span data-syntax-rule="path-expr-order">The syntax rules for <a>SHACL property paths</a> take precedence over <a>node expressions</a>,
-					for example so that <a>IRIs</a> are interpreted as <a>predicate path</a> instead of <a>IRI expression</a>.</span>
+					<span data-syntax-rule="path-expr-order">The syntax rules for <a>SHACL property paths</a> take precedence over <a>node expressions</a>;
+					for example, this means that <a>IRIs</a> are interpreted as <a>predicate path</a> instead of <a>IRI expression</a>.</span>
```
TallTed: Either 
```suggestion
					for example, this means that <a>IRIs</a> are interpreted as <a>predicate paths</a> instead of <a>IRI expressions</a>.</span>
```

TallTed: or
```suggestion
					for example, this means that an <a>IRI</a> will be interpreted as a <a>predicate path</a> instead of an <a>IRI expression</a>.</span>
```

shacl12-core/index.html
TallTed: If this sub-thread needs further discussion, I *strongly* advise that someone well versed in its specifics create a new issue for that discussion (where I suggest a summarization of the thread so far, with a link back here to the original), and that this PR be merged as it represents several steps forward.

FPWD doesn't require that this thread (nor any other) be satisfied.

TallTed (approved):
A couple of very small tweaks. Presuming they'll be applied, I'm approving. If any of them is not being applied, please open a new issue from each suggested change that is not being applied.

HolgerKnublauch:
I am not yet deleting this (merged) branch in case we need to revisit the discussions attached to it.

tpluscode:
You can safely delete the branch. All discussion and history will be preserved anyway

PR #280, opened 2025-02-28
Prefixes are defined

afs:
"binding" is an overloaded word.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -534,7 +534,7 @@
 			<section id="conventions">
 				<h3>Document Conventions</h3>
 				<p>
-					Within this document, the following namespace prefix bindings are used:
+					Within this document, the following namespace prefix definitions are used:
 				</p>
 				<table class="term-table">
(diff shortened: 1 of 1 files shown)


HolgerKnublauch (approved):

ajnelson-nist:
FWIW, [Turtle, 2.4](https://www.w3.org/TR/turtle/#sec-iri) uses "associates".

afs:
> FWIW, [Turtle, 2.4](https://www.w3.org/TR/turtle/#sec-iri) uses "associates".

SPARQL 1.2 uses "defines" ... because we just merged the same change as here :smile:.

ajnelson-nist:
> > FWIW, [Turtle, 2.4](https://www.w3.org/TR/turtle/#sec-iri) uses "associates".
> 
> SPARQL 1.2 uses "defines" ... because we just merged the same change as here 😄.

Touché.  =)

nicholascar (approved):

PR #290, opened 2025-03-03
#136: Fix property reference in second example of 4.7.3

### LINKED ISSUE #136: Wrong property name "ex:property" in second example of 4.7.3
https://www.w3.org/TR/shacl/#x4.7.3-sh:qualifiedvalueshape,-sh:qualifiedmincount,-sh:qualifiedmaxcount :
The second informative example on qualified value constraints mentions "at most 5 values of ex:property" where the property name should have been ex:digit.

Originally reported here: https://groups.google.com/g/topbraid-users/c/x9PNJDmkpl8

ajnelson-nist:
This PR appears to have no conflicts with #222 .

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -4055,7 +4055,7 @@ ex:Jane
 					</aside>
 					<p>
 						The following example illustrates the use of <code>sh:qualifiedValueShapesDisjoint</code>
-						to express that a hand must have at most 5 values of <code>ex:property</code> (expressed using <code>sh:maxCount</code>),
+						to express that a hand must have at most 5 values of <code>ex:digit</code> (expressed using <code>sh:maxCount</code>),
 						and exactly one of them must be an instance of <code>ex:Thumb</code> while exactly 4 of them must be an instance of <code>ex:Finger</code>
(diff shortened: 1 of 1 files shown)


ajnelson-nist:
Should this PR also include an errata note somewhere?

HolgerKnublauch (approved):

HolgerKnublauch:
I don't think this needs an extra errata - the whole SHACL Core document is one big errata in a sense, and there is always the GitHub history for historians.

ajnelson-nist:
> I don't think this needs an extra errata - the whole SHACL Core document is one big errata in a sense, and there is always the GitHub history for historians.

This is fine with me in this case.  It feels like a nit-level fix.

nicholascar:
In future, if a PR specifically addresses an Issue, can `Closes Issue #XXX` be added to the end of the Description? That will then automatically close the Issue when merged rather than it needing a manual close.

ajnelson-nist:
> In future, if a PR specifically addresses an Issue, can `Closes Issue #XXX` be added to the end of the Description? That will then automatically close the Issue when merged rather than it needing a manual close.

Fine by me.  I'll try to remember.

Issue #264, opened 2025-02-23
Drop SHACL-JS terms from shacl.tt


HolgerKnublauch:
The SHACL vocabulary TTL file contains some terms from the SHACL-JS spec, which is discontinued. These terms should be removed.

caribouW3:
Discussed during Feb 24th meeting: https://www.w3.org/2025/02/24-data-shapes-minutes.html#7233

ajnelson-nist:
Before closing this ticket, can we have a follow-n PR to #265 that logs in some 1.2 document changes since SHACL 1.0---in particular, noting the effects of 265?

HolgerKnublauch:
I just remembered that there is already a section on the top of the TTL file that could be used here:

https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L9

I should have recorded the dropping of the JS terms there as part of the PR. Could be addressed in a future PR.

HolgerKnublauch:
I think this change is now sufficiently captured in the TTL file itself, so I'll close the ticket.

Issue #177, opened 2024-02-23
Add something similar to `dash:singleLine`


HolgerKnublauch:
dash:singleLine (https://datashapes.org/constraints.html#SingleLineConstraintComponent) can be used to specify that string values must not contain line breaks. This is commonly used for things like labels or ID strings.

While this can be expressed with sh:pattern, it is better to have an explicit boolean for this. For example, user interface builders can determine whether they should display a textfield vs a textarea.

HolgerKnublauch:
In the Feb 24 meeting the WG accepted this feature into core, but marked as "At risk" which means that the WG will revisit the inclusion of this feature when it has gathered more evidence about the actual need. In a nutshell, the concern is where to draw the line and whether adding too many little features will open the floodgates.

PR #291, opened 2025-03-03
#264, #177: Added change log entry about deleting SHACL-JS terms and …

HolgerKnublauch:
…sh:singleLine


ajnelson-nist (approved):
Looks good to me.  Thank you.

Issue #271, opened 2025-02-24
Where should "Potential definition in SPARQL" go?


bergos:
We want to remove the "Potential definition in SPARQL" in the core specification, but there is interest in keeping them elsewhere. A new location needs to be found. See #235

HolgerKnublauch:
They could become an appendix in the SHACL-SPARQL document.

PR #294, opened 2025-03-03
#271: Moved SPARQL definitions from Core into an appendix of SHACL-SPARQL

HolgerKnublauch:


bergos (changes requested):
Four definitions are missing in the SPARQL Extensions spec:

- `sh:targetNode`
- `sh:minInclusive`
- `sh:maxExclusive`
- `sh:maxInclusive`

HolgerKnublauch:
@bergos the three constraints are listed under sh:minExclusive because they are all so similar.
sh:targetNode cannot be expressed in SPARQL anymore because they may be node expressions now.

bergos:
@HolgerKnublauch Sorry, I missed the min/max when comparing "POTENTIAL DEFINITION IN SPARQL" count for completeness.

Depending on where we allow node expressions, also others could no longer be directly expressed in SPARQL. e.g., `$minExclusive` must be substituted with / bound to the result of the evaluated node expression in case a node expression is given as the value.

HolgerKnublauch:
> Depending on where we allow node expressions, also others could no longer be directly expressed in SPARQL. e.g., $minExclusive must be substituted with / bound to the result of the evaluated node expression in case a node expression is given as the value.

Yes, good point. I have added a TODO note for this, as Core is not quite there yet either.

bergos (approved):

PR #297, opened 2025-03-04
#288: Added sh:SelectExpression, with test cases and TTL changes

### LINKED ISSUE #288: Add sh:SelectExpression to SHACL-SPARQL
The Core spec (assuming #222 gets merged in) contains a reference to a node expression function sh:SelectExpression, in an example, with a reference to SHACL-SPARQL where it is supposedly defined. This actually should happen now, as the first non-trivial node expression type.

I believe defining it is straight-forward, but the real work here is to cover it with test cases, because this will be the first time that node expressions will be making a real difference to features such as sh:targetNode and sh:path. This also means that engines will need to add this extra flexibility.

HolgerKnublauch:


ajnelson-nist:
@HolgerKnublauch - I added an update to SHACL-SHACL.  It seems one update should have gone in for #274 : If `sh:targetNode` can now allow blank nodes in their object position, then it seems there's currently no `sh:PropertyShape` necessary because there's no constraint to specify.  Does this need to be broken out into a separate PR, possibly also to address any updates to `targetNode-nodeKind` text?

bergos (approved):
LGTM, as mentioned in #301, we may need to revisit the `sh:path` property. But that should be done in a separate PR.

ajnelson-nist:
Oh dear.  For awareness, there is a comment thread tied to just the last commit, [here](https://github.com/w3c/data-shapes/commit/8632a3e0067c5bfde8f70de79a669b1e3d95e9f5), documenting that I'm going to undo the SHACL-SHACL changes and move them to another PR.  It's not displaying in this PR page.  Sorry @bergos , you'll need to re-review.

ajnelson-nist:
Ok, the updates to SHACL-SHACL I had here are now moved out-of-band to #302 .

shacl12-vocabularies/shacl.ttl
```diff
@@ -1174,6 +1176,23 @@ sh:xone
...
+	a rdfs:Class ;
+	rdfs:label "Node expression"@en ;
+	rdfs:comment "The base class for the available node expression functions."@en ;
+	rdfs:subClassOf sh:Parameterizable ;
+	rdfs:isDefinedBy sh: .
+
+sh:keyParameter
+    a rdf:Property ;
+	rdfs:label "key parameter"@en ;
+	rdfs:comment "Set to true for the parameter of a node expression function that unique identifies it."@en ;
```
```suggestion
	rdfs:comment "Set to true for the parameter of a node expression function that uniquely identifies it."@en ;
```

TallTed (dismissed):
typo

HolgerKnublauch:
@TallTed could you kindly recheck whether you can unblock this PR? Thanks.

afs (approved):
This looks good for a FPWD.
I can't check the test case (no code).
There details recorded elsewhere for the example using sh:path.

HolgerKnublauch:
@TallTed this PR is still blocked because you have requested changes. As I have long applied your suggestion, I believe you still need to change the status of your review away from Request Changes.

(I don't think simple change requests such as typo fixes should ever block progress)

caribouW3:
> @TallTed this PR is still blocked because you have requested changes. As I have long applied your suggestion, I believe you still need to change the status of your review away from Request Changes.
> 
> (I don't think simple change requests such as typo fixes should ever block progress)

Since that request has been already accepted, I'll dismiss the review comment to unlock merging.

HolgerKnublauch:
Thanks a lot, @caribouW3. I am going to merge this PR into the main branch now because it has sufficient approvals and has been widely agreed to be done by the group when #222 was accepted. The Core spec even contains a reference to this already, so it's best to get this consistent now. Also, there are follow-up branches based on this already piling up.

Issue #306, opened 2025-03-05
Change the order of properties in the `sh:order` example


YoucTagh:
In [Example 44](https://w3c.github.io/data-shapes/shacl12-core/#example-44) about Non-Validating Property Shape Characteristics, it is better if the groups and property shapes in the example are not in order, and then display them in order in the form to illustrate the idea.

HolgerKnublauch:
I have mixed feelings about this. I think it improves readability of the shapes graph if people place the properties in the same order as they are visually displayed. I see your point about wanting to explain that the order in TTL is irrelevant but everyone should know this already because TTL files do not change the "order" of triples.

Having said this, no strong opinion, but I'll not submit a review of the PR.

PR #307, opened 2025-03-05
#306: Change the order of properties in the sh:order example

YoucTagh:


shacl12-core/index.html
```diff
@@ -6290,6 +6289,170 @@ <h3>sh:group</h3>
...
+					"sh:order": {
+						"@type": "xsd:integer",
+						"@value": "11"
+					},
+					"sh:path": {
+						"@id": "ex:streetAddress"
+					}
+				},
+				{
+					"sh:description": "The suburb, city or town of the address",
```
TallTed: This should perhaps lose the "suburb" since that is generally an adjective describing towns around a city (the "urb").
```suggestion
					"sh:description": "The suburb, city, or town of the address",
```

shacl12-core/index.html
```diff
@@ -6127,7 +6127,7 @@ <h3>sh:group</h3>
 	sh:property [
 		sh:path ex:locality ;
 		sh:name "locality" ;
-		sh:description "The suburb, city or town of the address" ;
+		sh:description "The suburb, city, or town of the address" ;
```
TallTed: Unless someone can provide an example of a region where "suburb" is the class of some habitations, I believe it is roughly a synonym of town, in context. Therefore, the following should be applied to all instances of `"The suburb, city, or town of the address"` —
```suggestion
		sh:description "The city or town of the address" ;
```

afs: "suburbs" are parts of towns, cities. 

In the UK, they often traditional names for areas built during the growth of commutting and they are not a unit of government organisation, although they may align with one.

See wikipedia.

HolgerKnublauch: In Australia, suburb is an official geographical term for post codes, but if that's not well-known elsewhere, let's get rid of it.

YoucTagh: I will apply the proposed changes tomorrow if there are no objections.

TallTed (approved):

Issue #301, opened 2025-03-04
Alternative property for computed values


bergos:
As discussed in the [definition of node expressions PR](https://github.com/w3c/data-shapes/pull/274), the `sh:path` property could be misleading for the use case of creating intermediate values for validations. One may expect that `sh:path` keeps the functionality of traversing the path given as the object value. In the case of a node expression, the result of the node expression should be used for traversing. A new property could be defined that skips the traversing part and only sets the value to the computed value of the node expression.

Below are two property shapes and how they could be translated into SPARQL:

```turtle
ex:propertyShape1 a sh:PropertyShape;
  sh:path [
    shn:iri [
      shn:concat ("http://example.org/" "property")
    ]
  ].

ex:propertyShape2 a sh:PropertyShape;
  sh:computed [
    shn:iri [
      shn:concat ("http://example.org/" "property")
    ]
  ].
```

```sparql
# for ex:propertyShape1
SELECT * WHERE {
  $this IRI(CONCAT("http://example.org/", "property")) $value
}
```

```sparql
# for ex:propertyShape2
SELECT * WHERE {
  BIND(IRI(CONCAT("http://example.org/", "property")) AS $value)  
}
```

There were multiple proposals in the PR for the name of the property. *derived* sounds good, but I would prefer *computed*. As shown in the example, it's not required to derive the value from any existing data. Maybe also combine it with *value*, like *computedValue*. *computedNode* may give the impression that only blank node and named node results are expected from the node expressions, but literals would also be allowed.

I took the term *value* from the SPARQL definitions of the constraint components as I think we don't have a defined term for the result of the traversing step from focus node over path.

afs:
Is it really a `sh:NodeShape`?  or even a new "ValuesShape"?

> Informally, property shapes specify constraints that need to be met with respect to [nodes](https://www.w3.org/TR/rdf12-concepts/#dfn-node) that _can be reached_ from the [focus node](https://github.com/w3c/data-shapes/issues/301#dfn-focus-node)

HolgerKnublauch:
I don't think the use of node expressions where we previously already allowed complex SPARQL paths is fundamentally different. In the future, people may naturally switch between the node expression features and paths. It is probably just for the transition period that this confusion will happen.

What are the use cases to only compute the predicate, and not its values directly?

[...]

PR #310, opened 2025-03-06
#301: sh:values and sh:defaultValue

HolgerKnublauch:
This implements my comment at https://github.com/w3c/data-shapes/issues/301#issuecomment-2701697096 which is basically what we have in SHACL-AF and what has been implemented in TopBraid for many years.

Note the PR includes the changes for SHACL-SPARQL introducing sh:select for technical reasons, because the separate PR for that is currently blocked for process reasons. https://github.com/w3c/data-shapes/pull/297#issuecomment-2702495195

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1660,7 +1660,7 @@ ex:SomeClass-alternativePathExample
 					<p>
 						Here is an example use of a node expression based on <a data-cite="shacl12-sparql#SelectExpression">SHACL-SPARQL</a>, computing
 						the values of a property shape for the property "full name" as the concatenation of the <code>ex:firstName</code>,
-						a space and the <code>ex:lastName</code>.
+						a space, and the <code>ex:lastName</code>.
 					</p>
 					<div class="shapes-graph">
 						<div class="turtle">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1775,7 +1775,7 @@ ex:SomeClass-alternativePathExample
 				</p>
 				<aside class="example" title="A dynamically computed property using a node expression based on a SPARQL query">
 					<p>
-						Here is an example use of a node expression based on <a href="shacl12-sparql">SHACL-SPARQL</a>, computing
+						Here is an example use of a node expression based on <a data-cite="shacl12-sparql#SelectExpression">SHACL-SPARQL</a>, computing
 						the values of a property shape for the property "full name" as the concatenation of <code>ex:firstName</code>,
 						a space and the <code>ex:lastName</code>.
 					</p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1434,7 +1434,7 @@ ex:JohnDoe a ex:Person .
 				</p>
 				<p>
 					Informally, property shapes specify constraints that need to be met with respect to <a>nodes</a> that can be reached from the
-					<a>focus node</a> either by directly following a given property (specified as an <a>IRI</a>) or any other <a>SHACL property path</a>,
+					<a>focus node</a> by either (a) directly following a given property (specified as an <a>IRI</a>), (b) directly following any other <a>SHACL property path</a>
 					specified using <code>sh:path</code>, or evaluating the node expression specified using <code>sh:values</code>, or (if no other values exist)
 					by evaluating the node expression specified using <code>sh:defaultValue</code>.
 				</p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2160,7 +2160,7 @@ ex:Person-fullName
 						<li>
 							If the set is still empty and <code>d</code> is the <a>value</a> of <code>sh:defaultValue</code> at the <a>property shape</a>,
 							then add the <a>output nodes</a> of <code>eval(d, <a>data graph</a>, scope)</code> where <code>scope</code>
-							contains the <a>focus node</a> as value of the variable <code>focusNode</code>.
+							contains the <a>focus node</a> as the value of the variable <code>focusNode</code>.
 						</li>
 					</ol>
 				</section>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1547,7 +1547,7 @@ ex:JohnDoe a ex:Person .
 					<br/><br/>
 					<span data-syntax-rule="path-values">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:values</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
 					<span data-syntax-rule="path-defaultValue">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:defaultValue</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
-					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is an <a>IRI</a>.</span>
+					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
 				</p>
 				<p>
 					Informally, property shapes specify constraints that need to be met with respect to <a>nodes</a> that can be reached from the
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5,18 +5,6 @@
 		<meta charset="utf-8">
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer></script>
 		<script class="remove">
-
-			function hideSPARQL() {
-				document.querySelectorAll('.def-sparql').forEach(element => element.style.display = "none");
-				document.getElementById('hide-sparql').style.display = "none";
-				document.getElementById('show-sparql').style.display = "";
-			}
-
-			function showSPARQL() {
-				document.querySelectorAll('.def-sparql').forEach(element => element.style.display = "");
-				document.getElementById('show-sparql').style.display = "none";
-				document.getElementById('hide-sparql').style.display = "";
-			}
 		
 			function prepareSyntaxRules() {
 				document.querySelectorAll("[data-syntax-rule]").forEach(element => {
@@ -173,14 +161,6 @@
 				border-left-width: .5em;
 				border-color: #c0c0c0;
 				margin-bottom: 16px;
-			}
-			
-			.def-sparql {
-			}
-			
-			.def-sparql-body {
-				margin-top: 0px;
-				margin-bottom: 0px;
 			}
 			
 			.def-text {
@@ -447,16 +427,6 @@
 						<dfn data-lt="language tag">Language tags</dfn> are defined as in [[!BCP47]].
 					</div>
 				</div>
-				<div class="def" id="shacl-terminology">
-					<div class="term-def-header">SHACL-SPARQL Terminology</div>
-					<div>
-						This document uses the terms
-						<dfn data-cite="shacl12-sparql#dfn-parameter-names" data-lt="parameter names">parameter name</dfn>,
-						<dfn data-cite="shacl12-sparql#dfn-pre-binding" data-lt="pre-bind|pre-binding|pre-bound">pre-binding</dfn>, and
-						<dfn data-cite="shacl12-sparql#dfn-substituted" data-lt="substitution|substituted">substitution</dfn>
-						as defined in the SHACL 1.2 SPARQL Extensions specification [[!shacl12-sparql]].
-					</div>
-				</div>
 				<div class="def">
 					<div class="term-def-header">Property Value and Path</div>
 					<div>
@@ -598,15 +568,6 @@
 					<div class="turtle">
 # This box represents an output results graph
 					</div>
-				</div>
-
-				<p>
-					SHACL Definitions appear in blue boxes:
-				</p>
-				<div class="def def-sparql">
-					<div class="def-header">SPARQL or TEXTUAL DEFINITIONS</div>
-<pre class="def-sparql-body">
-# This box contains SPARQL or textual definitions. </pre>
 				</div>
 				
 				<p class="syntax">
@@ -832,38 +793,7 @@ ex:PersonShape
 					Otherwise, the SHACL processor MUST provide the entailments for all of the values of <code>sh:entailment</code> in the <a>shapes graph</a>,
 					and any inferred triples MUST be returned by all queries against the <a>data graph</a> during the <a>validation</a> process.
 				</p>
-			</section>
-			
-			<section id="shacl-sparql" class="informative">
-				<h3>Relationship between SHACL and SPARQL</h3>
-				<p>
-					This specification uses parts of SPARQL 1.2 in non-normative alternative definitions of the semantics of <a>constraint components</a> and <a>targets</a>.
-					While these may help some implementers, SPARQL is not required for the implementation of the SHACL Core language.
-				</p>
-				<p>
-					SPARQL variables using the <code>$</code> marker represent external <a>bindings</a> that are <a>pre-bound</a> or, in the case of <code>$PATH</code>, <a>substituted</a> in the SPARQL query before execution (as explained in <a href="shacl12-sparql#constraint-components-validation"></a>).
-				</p>
-				<p>
-					The definition of some <a>constraints</a> requires or is simplified through access to the <a>shapes graph</a> during query execution.
-					SHACL-SPARQL processors MAY <a>pre-bind</a> the variable <code>shapesGraph</code> to provide access to the <a>shapes graph</a>.
-					Access to the <a>shapes graph</a> is not a requirement for supporting the SHACL Core language.
-				</p>
-				<p>
-					Note that at the time of writing, SPARQL EXISTS has been imperfectly defined and implementations vary.
-					While a <a href="https://www.w3.org/community/sparql-exists/">W3C Community Group</a> is working on improving this situation,
-					users of SPARQL are advised that the use of EXISTS may have inconsistent results and should be approached with care.
-				</p>
-				<div class="syntaxmenu">
-					<p>The button below can be used to show or hide the SPARQL definitions.</p>
-					<form>
-						<p>
-							<input id="hide-sparql" onclick="hideSPARQL()" type="button" value="Hide SPARQL Definitions" />
-							<input id="show-sparql" onclick="showSPARQL()" style="display:none" type="button" value="Show SPARQL Definitions" />
-						</p>
-					</form>
-				</div>
-			</section>
-			
+			</section>			
 		</section>
 
 		<section id="constraints-section">
@@ -1127,7 +1057,7 @@ ex:MultiplePatternsShape
 						<aside class="example">
 							<div class="shapes-graph">
 								<div class="turtle">
-	ex:PersonShape
+ex:PersonShape
 	a sh:NodeShape ;
 	sh:targetNode ex:Alice .
 								</div>
@@ -1139,12 +1069,6 @@ ex:Bob a ex:Person .
 								</div>
 							</div>
 						</aside>
-	
-						<p class="def-sparql">
-							The following query expresses a potential definition of node targets in SPARQL.
-							The variable <code>targetNode</code> will be <a>pre-bound</a> to the given value of <code>sh:targetNode</code>.
-							All <a>bindings</a> of the variable <code>this</code> from the <a>solution</a> become focus nodes.
-						</p>
 					</section>
 					
 					<section id="targetClass">
@@ -1194,19 +1118,6 @@ ex:House a ex:Nephrologist .
 								</div>
 							</div>
 						</aside>
-						<p class="def-sparql">
-							The following query expresses a potential definition of class targets in SPARQL.
-							The variable <code>targetClass</code> will be <a>pre-bound</a> to the given value of <code>sh:targetClass</code>.
-							All <a>bindings</a> of the variable <code>this</code> from the <a>solutions</a> become focus nodes.
-						</p>
-						<div class="def def-sparql">
-							<div class="def-header">POTENTIAL DEFINITION IN SPARQL</div>
-<pre class="def-sparql-body">
-SELECT DISTINCT ?this    <span class="triple-can-be-skipped"># ?this is the focus node</span>
-WHERE {
-	?this rdf:type/rdfs:subClassOf* $targetClass .    <span class="triple-can-be-skipped"># $targetClass is <a>pre-bound</a> to ex:Person</span>
-}</pre>
-						</div>
 	
 					</section>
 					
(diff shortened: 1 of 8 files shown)


shacl12-core/index.html
```diff
@@ -1539,25 +1539,24 @@ <h3>Property Shapes</h3>
...
+					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape is a <a>well-formed</a>
+					<a>SHACL property path</a>.</span>
 					<br/><br/>
 					It is recommended, but not required, for a <a>property shape</a> to be declared as a <a>SHACL instance</a> of <code>sh:PropertyShape</code>.
 					<span data-syntax-rule="PropertyShape-path-minCount"><a>SHACL instances</a> of <code>sh:PropertyShape</code> have one <a>value</a> for the property <code>sh:path</code>.</span>
+					<br/><br/>
+					<span data-syntax-rule="path-values">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:values</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-defaultValue">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:defaultValue</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is an <a>IRI</a>.</span>
```
tpluscode:
```suggestion
					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
```

Might eat our own dogfood by using SHACL terminology?

shacl12-core/index.html
```diff
@@ -1539,25 +1539,24 @@ <h3>Property Shapes</h3>
...
 				</p>
 				<p>
 					Informally, property shapes specify constraints that need to be met with respect to <a>nodes</a> that can be reached from the
 					<a>focus node</a> either by directly following a given property (specified as an <a>IRI</a>) or any other <a>SHACL property path</a>,
-					specified using <code>sh:path</code>.
+					specified using <code>sh:path</code>, or evaluating the node expression specified using <code>sh:values</code>, or (if no other values exist)
+					by evaluating the node expression specified using <code>sh:defaultValue</code>.
```
tpluscode: Would this be more readable as a bullet list?

shacl12-core/index.html
tpluscode: And does it mean that we are also keeping the SHACL-AF way for materialising new triples when `sh:path` is used together with `sh:values/sh:defaultValue`?

shacl12-core/index.html
```diff
@@ -1539,25 +1539,24 @@ <h3>Property Shapes</h3>
...
 				</p>
 				<p>
-					If the <code>sh:path</code> is not a <a>well-formed</a> property path, it must be a <a>node expression</a>
-					and the constraints will be validated against the <a>output nodes</a> of this node expression.
-					Note that support for node expressions is not required by SHACL Core,
+					Note that support for <code>sh:values</code> and <code>sh:defaultValue</code> is not required by SHACL Core,
```
tpluscode, line 1442: For a SHACL Core engine, what should be its behaviour when it encounters a Property Shape with `sh:values` and `sh:path`?

Does it ignore `sh:values` and look for node by following the path?
Does it ignore the property altogether?
Does it fail?

tpluscode (commented):
Do we actually want to include SPARQL expressions in Core?

shacl12-core/index.html
HolgerKnublauch: Maybe. The section of value nodes does use a bullet list, so I am not too worried here.

shacl12-core/index.html
HolgerKnublauch: Do you mean sh:path used as a node expression? Yes, we should use the SHACL-AF "backward chaining" but the details of that will need to be handled by the Node Expr document, not core.

shacl12-core/index.html
HolgerKnublauch: The idea here is that we don't want to break SHACL Core implementations. So we cannot expect them to even know about sh:values and therefore we cannot even expect them to fail. In other words, they can just look for the path values and ignore sh:values/defaultValue.

tpluscode: Right, but then in practice it means that in cases of constraints which require existence of values (`sh:hasValue`, `sh:MinCount`) you will likely get false-negative validation result (no nodes at derived property). And skipping the property altogether implies a potential false-positive outcome for nodes which would otherwise fail validation if `sh:values` would have been resolved.

Maybe Core client should at least acknowledge that it saw and recognised a property it cannot process and add a `sh:Warning` result to the validation report? 

Or is it something to leave to implementations? One can think of many such cases when SHACL silently ignores problems, leading to subtle errors in downstream code

HolgerKnublauch: Good question. The situation is similar to any new constraint component that 1.2 introduces. 1.0 engines would just ignore them and then probably produce incorrect or missing results. Maybe this discussion requires its own ticket.

bergos: That should be covered by the profiles. Maybe we can use SHACL-SHACL to identify such problems. I opened an issue for [that topic](https://github.com/w3c/data-shapes/issues/312).

bergos (approved):

HolgerKnublauch:
Now that the main branch contains the sh:SelectExpression, I have updated this branch against the main branch, which will make it much easier to review.

shacl12-sparql/index.html
```diff
@@ -1246,14 +1246,15 @@ <h3>Select Expressions</h3>
 				<aside class="example" title="A dynamically computed property using a node expression based on a SPARQL query">
 					<p>
 						Here is an example use of a <a>select expression</a>, computing the values of a property shape for the property 
-						"full name" as the concatenation of <code>ex:firstName</code>, a space and the <code>ex:lastName</code>.
+						"full name" as the concatenation of the <code>ex:firstName</code>, a space and the <code>ex:lastName</code>.
```
```suggestion
						"full name" as the concatenation of the <code>ex:firstName</code>, a space, and the <code>ex:lastName</code>.
```

shacl12-core/index.html
```diff
@@ -1422,25 +1422,24 @@ <h3>Property Shapes</h3>
...
+					<br/><br/>
+					<span data-syntax-rule="path-values">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:values</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-defaultValue">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:defaultValue</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
 				</p>
 				<p>
 					Informally, property shapes specify constraints that need to be met with respect to <a>nodes</a> that can be reached from the
 					<a>focus node</a> either by directly following a given property (specified as an <a>IRI</a>) or any other <a>SHACL property path</a>,
```
TallTed: The `specified ...` clauses should either all be parenthesized (as in the "either") or all separated into comma-delineated phrases (as with the first "or"). Either of these is better than no delineation (as now in the second and third "or"). I think that the parentheses will be clearest, so I'll suggest that (in 2 suggestions, because there's a red line between other lines I want to change. This is change 1 of 2.).

While working through the above, the complexity of the options became more clear. I hope the `(a)` enumeration works. These might want to be changed to an `<ol>`.
```suggestion
					<a>focus node</a> by either (a) directly following a given property (specified as an <a>IRI</a>), (b) directly following any other <a>SHACL property path</a>
```

shacl12-core/index.html
```diff
@@ -1422,25 +1422,24 @@ <h3>Property Shapes</h3>
...
-					specified using <code>sh:path</code>.
+					specified using <code>sh:path</code>, or evaluating the node expression specified using <code>sh:values</code>, or (if no other values exist)
+					by evaluating the node expression specified using <code>sh:defaultValue</code>.
```
TallTed: (This is change 2 of 2.)
```suggestion
					(specified using <code>sh:path</code>), (c) evaluating the node expression (specified using <code>sh:values</code>),
					or, (d) if no other values exist, evaluating the node expression (specified using <code>sh:defaultValue</code>).
```

shacl12-core/index.html
```diff
@@ -1654,20 +1654,21 @@ <h3>Node Expressions</h3>
...
 				<p>
-					The following example illustrates the syntax of a node expression, used as a blank node value of <code>sh:path</code>:
+					The following example illustrates the syntax of a node expression, used as a blank node value of <code>sh:values</code>:
 				</p>
 				<aside class="example" title="A dynamically computed property using a node expression based on a SPARQL query">
 					<p>
 						Here is an example use of a node expression based on <a data-cite="shacl12-sparql#SelectExpression">SHACL-SPARQL</a>, computing
-						the values of a property shape for the property "full name" as the concatenation of <code>ex:firstName</code>,
+						the values of a property shape for the property "full name" as the concatenation of the <code>ex:firstName</code>,
 						a space and the <code>ex:lastName</code>.
```
```suggestion
						a space, and the <code>ex:lastName</code>.
```

shacl12-core/index.html
```diff
@@ -1702,7 +1703,7 @@ <h3>Node Expressions</h3>
 					The evaluation may also result in an <dfn>evaluation failure</dfn>.
 				</div>
 				<p>
-					This SHACL Core specification only defines exactly the <a>node expression functions</a> from the following two sections.
+					This SHACL Core specification only defines exactly the <a>node expression functions</a> from the following two subsections.
```
TallTed: makes the `exactly` apply more clearly to `defines` than to `the ... functions ...`
```suggestion
					This SHACL Core specification only exactly defines the <a>node expression functions</a> from the following two subsections.
```

shacl12-core/index.html
```diff
@@ -2064,10 +2065,13 @@ <h5>Focus node (sh:focusNode)</h5>
 					<section id="results-path">
 						<h4>Path (sh:resultPath)</h4>
 						<p>
-							Validation results may have a value for the property <code>sh:resultPath</code> pointing at a <a>well-formed</a> <a>SHACL property path</a>
-							or <a>node expression</a>.
+							Validation results may have a value for the property <code>sh:resultPath</code> pointing at a <a>well-formed</a> <a>SHACL property path</a>.
```
TallTed, line 2068:
```suggestion
							Validation results may have a value for the property <code>sh:resultPath</code> that points at a <a>well-formed</a> <a>SHACL property path</a>.
```

shacl12-core/index.html
```diff
@@ -2142,22 +2146,23 @@ <h4>Value Nodes of Node Shapes</h4>
...
+						set of <a>value nodes</a> is produced by the following steps:
 					</p>
+					<ol>
+						<li>
+							Add all <a>nodes</a> in the <a>data graph</a> that can be reached from the <a>focus node</a> with the <a>path mapping</a> of <code>p</code>.
+						</li>
+						<li>
+							If <code>e</code> is the <a>value</a> of <code>sh:values</code> at the <a>property shape</a>,
+							then add the <a>output nodes</a> of <code>eval(e, <a>data graph</a>, scope)</code> where <code>scope</code>
+							contains the <a>focus node</a> as value of the variable <code>focusNode</code>.
```
```suggestion
							contains the <a>focus node</a> as the value of the variable <code>focusNode</code>.
```

shacl12-core/index.html
```diff
@@ -2142,22 +2146,23 @@ <h4>Value Nodes of Node Shapes</h4>
...
+						</li>
+						<li>
+							If the set is still empty and <code>d</code> is the <a>value</a> of <code>sh:defaultValue</code> at the <a>property shape</a>,
+							then add the <a>output nodes</a> of <code>eval(d, <a>data graph</a>, scope)</code> where <code>scope</code>
+							contains the <a>focus node</a> as value of the variable <code>focusNode</code>.
```
```suggestion
							contains the <a>focus node</a> as the value of the variable <code>focusNode</code>.
```

TallTed (commented):
Not requesting changes with the GitHub button, but if any of my requests aren't acceptable as they are, please don't merge until we come to agreement on the revisions to these areas.

shacl12-core/index.html
```suggestion
					This SHACL Core specification only defines exactly the <a>node expression functions</a> in the following two subsections.
```

shacl12-core/index.html
```diff
@@ -2142,22 +2146,23 @@ <h4>Value Nodes of Node Shapes</h4>
...
-						that appear in the <a>object</a> position of these triples.
-						See the <a href="https://www.w3.org/submissions/CBD/">Concise Bounded Description</a>.
+						set of <a>value nodes</a> is produced by the following steps:
 					</p>
+					<ol>
+						<li>
+							Add all <a>nodes</a> in the <a>data graph</a> that can be reached from the <a>focus node</a> with the <a>path mapping</a> of <code>p</code>.
+						</li>
+						<li>
+							If <code>e</code> is the <a>value</a> of <code>sh:values</code> at the <a>property shape</a>,
```
afs, line 2156: This precludes only having node expressions because `sh:path` is required. Is that intended?

shacl12-core/index.html
```diff
@@ -1422,25 +1422,24 @@ <h3>Property Shapes</h3>
...
-					specified using <code>sh:path</code>.
+					<a>focus node</a> by either (a) directly following a given property (specified as an <a>IRI</a>), (b) directly following any other <a>SHACL property path</a>
+					specified using <code>sh:path</code>, or evaluating the node expression specified using <code>sh:values</code>, or (if no other values exist)
+					by evaluating the node expression specified using <code>sh:defaultValue</code>.
 				</p>
 				<p>
-					If the <code>sh:path</code> is not a <a>well-formed</a> property path, it must be a <a>node expression</a>
-					and the constraints will be validated against the <a>output nodes</a> of this node expression.
-					Note that support for node expressions is not required by SHACL Core,
+					Note that support for <code>sh:values</code> and <code>sh:defaultValue</code> is not required by SHACL Core,
```
afs, line 1442: Can we just say:

```suggestion
                    Note that `sh:values` and `sh:defaultValue` are not used by SHACL Core,
```

shacl12-core/index.html
```diff
@@ -1422,25 +1422,24 @@ <h3>Property Shapes</h3>
...
-					<span data-syntax-rule="path-expr-order">The syntax rules for <a>SHACL property paths</a> take precedence over <a>node expressions</a>;
-					for example, this means that an <a>IRI</a> will be interpreted as a <a>predicate path</a> instead of an <a>IRI expression</a>.</span>
-					<span data-syntax-rule="path-not-literal">The <a>value</a> of <code>sh:path</code> cannot be a <a>literal</a>.</span>
+					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape is a <a>well-formed</a>
+					<a>SHACL property path</a>.</span>
 					<br/><br/>
 					It is recommended, but not required, for a <a>property shape</a> to be declared as a <a>SHACL instance</a> of <code>sh:PropertyShape</code>.
 					<span data-syntax-rule="PropertyShape-path-minCount"><a>SHACL instances</a> of <code>sh:PropertyShape</code> have one <a>value</a> for the property <code>sh:path</code>.</span>
+					<br/><br/>
+					<span data-syntax-rule="path-values">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:values</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
```
afs, line 1431: Is it the case that only one of `sh:values` and `sh:defaultValue` is allowed otherwise it is a SHACL-syntax error?

Section 4.7.2 allows for having both. `sh:values` may return an empty list.

The "or" on line 1433 isn't clear.

shacl12-core/index.html
```diff
@@ -1422,25 +1422,24 @@ <h3>Property Shapes</h3>
...
+					<span data-syntax-rule="path-defaultValue">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:defaultValue</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
+					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
```
afs: A sequence path would be possible as well.

shacl12-core/index.html
```diff
@@ -1654,20 +1654,21 @@ <h3>Node Expressions</h3>
 					One of these parameters can be the <dfn>key parameter</dfn> that uniquely identifies the <a>function name</a>.
```
afs, line 1654: I read "TEXTUAL DEFINITION" as implying:

> Each node expression function has an IRI as its function name.
> Node expression functions can declare one or more node expression parameters.
> Each of these parameters has an IRI. 

reads as if the only way to invoke a function is by its function name which isn't true (IRI expression, literal expression)

Alternative:

A node expression is one of
1. A function called by its function name
2. An IRI expression
3. A literal expression

and the definitions of "IRI expression" and "literal expression" need changing (they can be used as direct terms or via their function name).

shacl12-core/index.html
```diff
@@ -1654,20 +1654,21 @@ <h3>Node Expressions</h3>
...
 				</div>
 				<p>
-					The following example illustrates the syntax of a node expression, used as a blank node value of <code>sh:path</code>:
+					The following example illustrates the syntax of a node expression, used as a blank node value of <code>sh:values</code>:
 				</p>
 				<aside class="example" title="A dynamically computed property using a node expression based on a SPARQL query">
 					<p>
 						Here is an example use of a node expression based on <a data-cite="shacl12-sparql#SelectExpression">SHACL-SPARQL</a>, computing
```
afs, line 1676: The example should go after "EVALUATION OF NODE EXPRESSIONS".

shacl12-core/index.html
```diff
@@ -1702,7 +1703,7 @@ <h3>Node Expressions</h3>
 					The evaluation may also result in an <dfn>evaluation failure</dfn>.
```
afs: `eval(expr, activeGraph, scope)`

1:: Does scope include the parameters named by IRI?
scope is SPARQL variable name to node (and used as "focusNode")

2:: What about postional arguments `[ fn:strlen ( someWayToGetInputFocuNode   2   3 ) ]`?

shacl12-core/index.html
ajnelson-nist: I prefer the pre-edit wording.  Presence in SHACL-Core notes the terms should be recognized as in the SHACL vocabulary, but knowing their functionality is explicitly "not required" is a clearer statement to me.

shacl12-core/index.html
TallTed: Resolved but not applied, and no discussion?

shacl12-core/index.html
TallTed: These changes to lines 1438-1439 were not applied. They follow on [changes which *were* applied to lines 1437-1439](https://github.com/w3c/data-shapes/pull/310#discussion_r1988075396). Lines 1438-1439 no longer feed into lines 1437-1439.

shacl12-core/index.html
TallTed: Terminal/Oxford commas never add, and often remove, confusion.

shacl12-core/index.html
TallTed: Why no `the`?

shacl12-core/index.html
TallTed: Why no `the`?

shacl12-sparql/index.html
TallTed: Terminal/Oxford comma, again...

shacl12-core/index.html
afs: Recognizing SHACL vocabulary would imply some basic "support"?

I wanted to capture that they can be syntactically correct and not processed in the "SHACL Core profile" (to come...).

ajnelson-nist: I suppose I'm fine with it.  Though, the next line would need its grammar aligned.

shacl12-core/index.html
HolgerKnublauch: Are you sure? What for ex:first/ex:second - it is not clear which triples need to be asserted as there may be multiple objects of ex:first, and then then multiple triples would match this path (or just one). The idea here is that the sh:path could be used to assert/infer actual triples, in a predictable way.

shacl12-core/index.html
HolgerKnublauch: I have no clue what went wrong. I used the Files tab to add all suggestions to batch, but maybe some of them were not captured. Sorry - not intentional.

shacl12-core/index.html
HolgerKnublauch: Yes, this was intended. Because it consistently handles the case when sh:values were materialized (inferred or asserted). After that, the results of the sh:path query include all values from the sh:values clause, so no further dynamic computation needs to happen.

Also I think it's more consistent to have the policy that *all* property shapes have a sh:path.

(Marking resolved, please reopen if you disagree).

shacl12-core/index.html
HolgerKnublauch: 1) No, the definition of the eval function (or process or algorithm) would include details on how to fetch the parameter values from the given expr node. In some cases these are further node expressions that require nested evaluation. But the scope is a separate concept that gets passed into the execution engine from the outside.

2) I hope we can move the details of the syntax into SHACL 1.2 Node Expr. Core doesn't yet need this and it will make more sense when there are real functions in the library (for example etc). Alternatively, we could cover some of it in SHACL-SPARQL, in the part about sh:Functions.

(For housekeeping I'll mark this resolved but please reopen if this is not good enough).

shacl12-core/index.html
HolgerKnublauch: This question is not directly related to this PR. This definition was there before. It wouldn't surprise anyone if the details of node expressions will be refined in the next couple of iterations. I suggest if you think this block needs changing, you could raise a fresh issue and maybe make a specific PR, because I am not exactly sure what you mean. The relationship between the function IRI and the concept of the key parameter is also unclear to me. It will become clearer once we start on the Node Expr document (which may make sense pretty soon).

For housekeeping I'll mark this question resolved - please reopen if you disagree.

shacl12-core/index.html
HolgerKnublauch: Ok, I have reformulated this sentence to hopefully clarify it. Both values can co-exist.

shacl12-core/index.html
afs: There could be a predicable policy (e.g. only create if ex:first absent, use a bnode) which would cover some usage patterns. Rules might be better to give control options - I haven't thought about whether defaulting in rule will need special handling.

shacl12-core/index.html
afs: > This question is not directly related to this PR.

The previous update was pushed through to get into FPWD.

shacl12-core/index.html
afs: SHACL Core needs to define node expression evaluation if it is to be the basis for other documents. A partial just means the NE document is actually starting point and SHACL-SPARQL is based on SHACL-NE.

In WG discussions, parameter lists have been common.

(And then the doc can have another example that is not using `sh:select` - `sh:targetNode` would seem like a good one.)

shacl12-core/index.html
afs: Issue raised: #322.

shacl12-core/index.html
```diff
@@ -1430,7 +1430,7 @@ <h3>Property Shapes</h3>
 					<br/><br/>
 					<span data-syntax-rule="path-values">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:values</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
 					<span data-syntax-rule="path-defaultValue">A <a>property shape</a> has at most one <a>value</a> for the property <code>sh:defaultValue</code> and this <a>value</a> is a <a>well-formed</a> <a>node expression</a>.</span>
-					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have a <a>value</a> for <code>sh:values</code> or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
+					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have <a>values</a> for <code>sh:values</code> and <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
```
TallTed: Are `sh:values` and `sh:defaultValue` always present together? Perhaps the `and` should be `and/or`?
```suggestion
					<span data-syntax-rule="path-values-iri">A <a>property shape</a> can only have <a>values</a> for <code>sh:values</code> and/or <code>sh:defaultValue</code> when its <a>value</a> for <code>sh:path</code> is a <a href="#property-path-predicate">Predicate Path</a>.</span>
```

Issue #110, opened 2019-03-14
Clarity of prose in sec. 1.4


skynx:
Section 1.4, "SHACL Example" articulates the following caveat when introducing the JSON-LD variant of an example.

>The example below shows the same shape definition as a possible JSON-LD [json-ld] fragment. Note that we have left out a `@context` declaration, and depending on the `@context` the rendering may look quite different. Therefore this example should be understood as an illustration only.

I think the choice to leave out the `@context` part of the JSON description (which ungrounds the small number of namespace prefixes) and to explain this in preceding ancillary text is confusing. Rather than rely on the reader to bring in background knowledge at such an early point in the text, it could be more effective to alter the example to include the assumed namespace prefixes in `@context`. If desired, the meaning of `@context` could then be illustrated by also showing the URI-expanded JSON-LD representation (i.e. what you need without providing prefixes in `@context`) of the same example shapes graph.

This is an editorial/stylistic issue.

HolgerKnublauch:
Thanks for your comment, which will remain open to be considered by a future revision. I agree it would be better to have this self-contained. The WG did not deliver an official JSON-LD context, although some work on this was actually done in the group and the result can be found at https://github.com/w3c/shacl/blob/master/shacl-jsonld-context/shacl.context.ld.json and https://github.com/w3c/data-shapes/tree/gh-pages/shacl-jsonld-context. Users can access that using

https://w3c.github.io/data-shapes/shacl-jsonld-context/shacl.context.ld.json

(Without checking details I think if we embed the context into the example then we'd also need to declare that ignoredProperties is @list, not just the namespace prefixes).

bergos:
fixed in #317

PR #317, opened 2025-03-08
#283: Added JSON-LD serialization for examples

### LINKED ISSUE #283: [style]: Use tabs to show alternate serializations
When both the Turtle and JSON-LD serialization of a shape is given (and presumably soon SHACL-C), it would be better to have the different serializations navigated through tabs rather than listed in the document.

A ReSpec example of this is [here](https://www.w3.org/TR/vc-data-model-2.0/#example-use-of-issuer-properties).

bergos:
- added tabs for JS and CSS for serializations
- JSON-LD examples based on existing `<pre class="jsonld">` element
- removed separate JSON-LD example
- updated description about examples & formats
- updated some Turtle examples to make them fully parsable & comparable
- validated JSON-LD examples with a semi-automatic script

fixes #283 #110

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -663,8 +663,8 @@
 }</pre>
 				<p>
 					Note that the URI of the graph defining the SHACL vocabulary itself is equivalent to
-					the namespace above, i.e. it includes the <code>#</code>.
-					References to the SHACL vocabulary, e.g. via <code>owl:imports</code> should include the <code>#</code>.
+					the namespace above, i.e., it includes the <code>#</code>.
+					References to the SHACL vocabulary, e.g., via <code>owl:imports</code> should include the <code>#</code>.
 				</p>
 				<p>
 					Throughout the document, color-coded boxes containing RDF graphs in Turtle and JSON-LD will appear.
@@ -677,7 +677,7 @@
 					<div class="turtle">
 # This box represents an input shapes graph
 
-# Triples that can be omitted are marked as grey e.g.
+# Triples that can be omitted are marked as grey, e.g. --
 <span class="triple-can-be-skipped">&lt;s&gt; &lt;p&gt; &lt;o&gt; .</span>
 						</div>
 				  <div class="jsonld">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -64,7 +64,46 @@
 					element.id = "validator-" + validatorId;
 				});
 			}
-		
+
+			document.addEventListener("DOMContentLoaded", () => {
+				// Replace code div blocks in shapes, data, and results graph with tabs
+				for (const graph of document.querySelectorAll(".shapes-graph, .data-graph, .results-graph")) {
+					const tabs = document.createElement("aside");
+					tabs.classList.add("ds-selector-tabs");
+					graph.firstElementChild.classList.add("selected");
+
+					for (const child of graph.children) {
+						child.classList.add("tab");
+					}
+
+					tabs.append(...graph.children);
+					graph.append(tabs)
+				}
+
+				// Generate buttons for the selection logic
+				for (const tabs of document.querySelectorAll(".ds-selector-tabs")) {
+					const selectors = document.createElement("div");
+					selectors.classList.add("selectors");
+					selectors.innerHTML = `
+						<button class="selected" data-selects="turtle">Turtle</button>
+						<button data-selects="jsonld">JSON-LD</button>
+					`
+
+					tabs.prepend(selectors);
+				}
+
+				// Add example button selection logic
+				for (const button of document.querySelectorAll(".ds-selector-tabs .selectors button")) {
+					button.onclick = () => {
+						const ex = button.closest(".ds-selector-tabs");
+						ex.querySelector("button.selected").classList.remove("selected");
+						ex.querySelector(".selected").classList.remove("selected");
+						button.classList.add('selected');
+						ex.querySelector("." + button.dataset.selects).classList.add("selected");
+					}
+				}
+			});
+
 			var respecConfig = {
 				group: "wg/data-shapes",
 				github: "w3c/data-shapes",
@@ -103,7 +142,7 @@
 						mailto:     "simon.werner@taxonic.com",
 						w3cid:      164386
 					},
-					
+
 				],
 				formerEditors: [
 					{
@@ -327,7 +366,7 @@
 				margin-top: -1.5em;
 				white-space: pre;
 			}
-			
+
 			.data-graph { 
 				background: #eeb; 
 				border: 1px solid #cc9;
@@ -361,6 +400,12 @@
 				padding-left: 0.4em;
 			}
 
+			/* no dark mode, keep colors for shapes-graph, data-graph, and results graph background */
+			code.hljs {
+				--base: transparent;
+				--mono-1: #383a42;
+			}
+
 			/* our syntax menu for switching */
 			div.syntaxmenu {
 				border: 1px dotted black;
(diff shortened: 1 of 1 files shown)


shacl12-core/index.html
afs, line 1: Change needed:

https://github.com/w3c/data-shapes/blob/1f758aa099736ecda96dd90e86040a1e5d74da3c/shacl12-core/index.html#L489

afs, line 1: 1. The JSON-LD has extra vertical space, before and after the example text. This might be because of class-inside class in `div-pre`.
2. The font size and possibly weight for JSON-LD and Turtle is different.

The Turtle shows too big and heavy. This is the case pre-this PR and is a change from SHACL 1.0.

On my display (Linux desktop, Chrome): I'm getting:

Turtle is 14.4px, line height 21.6px.
JSON-LD 12.6px, 18.9px

At SHACL 1.0 Turtle is 13px, 19px

shacl12-core/index.html
```diff
@@ -592,12 +702,30 @@ <h3>Document Conventions</h3>
 # Elements highlighted in red are focus nodes that fail validation
 <span class="focus-node-error">ex:Alice</span> a ex:Person .
 					</div>
+					<div class="jsonld">
+						<pre class="text">// This box represents an input data graph</pre>
```
afs, line 706: Two `<pre>` leading to a lot of space between the two `<pre>`

shacl12-core/index.html
```diff
@@ -562,14 +648,29 @@ <h3>Document Conventions</h3>
 						<td><code>http://example.com/ns#</code></td>
 					</tr>
 				</table>
+
+				<p>
+					Within this document, the following JSON-LD context is used:
+				</p>
+				<pre class="jsonld">{
+  "@context": {
```
afs, line 656: Will there be more context added to make the JSON-LD more plain JSON-like?

(It may become impossible to have a better `@context` because of node expressions everywhere.)

shacl12-core/index.html
```diff
@@ -701,35 +859,54 @@ <h3>SHACL Example</h3>
...
+	"sh:property": [
 		{
-			"path" : "ex:ssn",
-			"maxCount" : 1,
-			"datatype" : "xsd:string" ,
-			"pattern" : "^\\d{3}-\\d{2}-\\d{4}$"
+			"sh:datatype": {
+				"@id": "xsd:string"
+			},
+			"sh:maxCount": {
```
afs, line 882: Change from using JSON native type to inline RDF literal - see comment about `@context` above.

shacl12-core/index.html
```diff
@@ -4427,7 +6359,7 @@ <h3>sh:defaultValue</h3>
 	sh:property [
 		sh:path ex:postalCode ;
 		sh:name "postal code" ;
-		sh:name "zip code"@en-US ;
+		sh:name "zip code"@en-us ;
```
afs, line 6362: `en-US` is BCP-47 preferred form.

shacl12-core/index.html
```diff
@@ -3150,14 +4340,50 @@ <h4>sh:languageIn</h4>
 							<div class="turtle">
 ex:Mountain
 	ex:prefLabel "Mountain"@en ;
-	ex:prefLabel "Hill"@en-NZ ;
+	ex:prefLabel "Hill"@en-nz ;
```
afs, line 4343: `en-NZ` is valid. BCP-47 preferred form is `en-NZ`.
Language tags are not required to be lower case.

afs:
Can we keep the Turtle and JSON-LD aligned as RDF? Is there tooling to add a github workflow?

Will the JSON-LD also have the highlighting used in Turtle examples to show import features like focus node, invalid data and classes when it's the purpose of the example to show these?

shacl12-core/index.html
bergos: highlight.js is used for JSON-LD content. That was already the case before my changes. highlight.js makes it complicated to highlight features like focus node. I can try to disable it and use the existing Turtle formatting. It's a tradeoff between highlight.js and custom formatting, but I think highlight.js doesn't offer us many benefits.

That's also the reason for the styling differences you have reported.

:+1: if I should disable highlight.js and switch to the Turtle formatting.

bergos: The tooling I have built extracts all Turtle and JSON-LD blocks parses the content and compares the canonical output. Generating pretty JSON-LD is a bit more complicated as it would require a frame to have inline blank nodes. The pretty JSON-LD is hand-tweaked. However, I could clean up my tool to compare the content and add it to a GitHub workflow. Then, we would at least know if they are not in sync anymore. I will create an issue for that and add it with a separate PR.

bergos:
@afs highlight.js is used for JSON-LD content. That was already the case before my changes. highlight.js makes it complicated to highlight features like focus node. I can try to disable it and use the existing Turtle formatting. It's a tradeoff between highlight.js and custom formatting, but I think highlight.js doesn't offer us many benefits.

That's also the reason for the styling differences you have reported.

:+1: if I should disable highlight.js and switch to the Turtle formatting.

bergos:
@afs The tooling I have built extracts all Turtle and JSON-LD blocks parses the content and compares the canonical output. Generating pretty JSON-LD is a bit more complicated as it would require a frame to have inline blank nodes. The pretty JSON-LD is hand-tweaked. However, I could clean up my tool to compare the content and add it to a GitHub workflow. Then, we would at least know if they are not in sync anymore. I will create an issue for that and add it with a separate PR.

shacl12-core/index.html
bergos: The JSON-LD parser makes them lowercase, and rdf-canonize doesn't normalize them. As the JSON-LD from the serializer would have them lowercase, and both styles are allowed, I normalized it. It would be possible to preprocess the language tags to check if both serializations are in sync.

afs:
@bergos -- thanks for the background.

If the JSON-LD needs additional updating to highlight key points, then better done once the Turtle is stable.

Can we agree that for doc development at the moment, the Turtle is primary?

shacl12-core/index.html
```diff
@@ -562,14 +648,29 @@ <h3>Document Conventions</h3>
...
+    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "sh": "http://www.w3.org/ns/shacl#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "ex": "http://example.com/ns#"
+  }
+}</pre>
 				<p>
 					Note that the URI of the graph defining the SHACL vocabulary itself is equivalent to
 					the namespace above, i.e. it includes the <code>#</code>.
```
```suggestion
					the namespace above, i.e., it includes the <code>#</code>.
```

shacl12-core/index.html
```diff
@@ -562,14 +648,29 @@ <h3>Document Conventions</h3>
...
 					References to the SHACL vocabulary, e.g. via <code>owl:imports</code> should include the <code>#</code>.
```
```suggestion
					References to the SHACL vocabulary, e.g., via <code>owl:imports</code> should include the <code>#</code>.
```

shacl12-core/index.html
```diff
@@ -578,7 +679,16 @@ <h3>Document Conventions</h3>
 
 # Triples that can be omitted are marked as grey e.g.
```
```suggestion
# Triples that can be omitted are marked as grey, e.g. --
```

TallTed (changes requested):
Editorial, trivial.

afs (approved):

TallTed (approved):
My trivial editorial adjustments have been applied.

Without a PR-Preview or similar, I cannot judge the other aspects.

Issue #315, opened 2025-03-08
Add Node Expression sh:eval to execute SPARQL expressions


HolgerKnublauch:
Similar to sh:select node expressions which execute SPARQL SELECT queries, there should be a node expression function that executes SPARQL expressions (known from BIND and FILTER): https://www.w3.org/TR/sparql12-query/#CreatingValuesWithExpressions

Example (computes the length of the URI of the current focus node):

```
    ex:Resource-uriLength
        a sh:PropertyShape ;
        sh:path ex:uriLength ;
        sh:values [
            sh:eval "STRLEN(STR($this))" ;
        ]
```

This will return a single node, and can also be used to express ASK-like queries using sh:eval "EXISTS { ... }".

It would take parameters sh:eval and sh:prefixes and expand to [prefixes] SELECT ($eval AS ?result) WHERE {}

tpluscode:
In general, I find this too much reliant on string literals and unnecessarily limited to SPARQL

Re my [reply](https://github.com/w3c/data-shapes/issues/314#issuecomment-2708300357) to #314, I found [Function Expressions](https://w3c.github.io/shacl/shacl-af/#node-expressions-function) and `dash-sparql:` very useful and flexible in its implementation-agnostic form (minus the verbose blank node structure).

```diff
    ex:Resource-uriLength
        a sh:PropertyShape ;
        sh:path ex:uriLength ;
        sh:values [
-           sh:eval "STRLEN(STR($this))" ;
+           fn:strlen ( [ fn:str ( [ sh:activeNode () ] ) ] )
        ]
```

HolgerKnublauch:
Yes, both syntaxes should (and will) be possible. Many things can be expressed in the blank node syntax, but SPARQL strings are often the only way to express things, and can be more compact. For example the < and != operators, NOT EXISTS { ... }

If engines already have to implement sh:select anyway, then doing sh:eval is a trivial addition.

afs:
+1 to the concept `sh:eval`. (It may not be the right name - `eval` may be more useful elsewhere.)

As expressions become more complicated, writing out RDF-NE syntax gets less practical.

It does assume access to the expression parser/evaluator of a SPARQL processor.

ajnelson-nist:
[...]

PR #328, opened 2025-03-17
#315: Added support for sh:eval to SHACL-SPARQL

HolgerKnublauch:


afs:
I don't think `sh:eval` is a good name here. "eval" is very general concept and not specific to SPARQL expressions. 

Suggestion: `sh:sparqlExpr` which allows for other expression languages and makes it clear it is an expression.

We might have similar naming issues elsewhere in SHACL grows in SHACL 1.2 so adopting a naming convention (often applied, not always)

I'd like to reserve for `eval` for a general usage, maybe dynamic invocation:

```
[ sh:eval ( functionExpr arg1 arg2 ) ]
```

where `functionExpr` is a node expression to decide the function name to call.

HolgerKnublauch:
Right, I had similar concerns about "eval". But sparqlExpr is quite a mouthful too, and will lead to awkward names because the term "expression" is already used in node expressions. So we would have sh:SelectExpression but sh:SPARQLExprExpression (?)

Also note that we use sh:select and sh:ask instead of sh:sparqlSelect and sh:sparqlAsk.

sh:eval is quite short to write for people who edit node expressions. But there may be a better stand-alone name from the SPARQL universe that doesn't need to be prefixed with "sparql"?

ajnelson-nist:
Does the [SPARQL Grammar](https://www.w3.org/TR/sparql11-query/#sparqlGrammar) provide any helpful hints for what could work instead of `eval`?  E.g., is there something in the neighborhood of `BuiltInCall` (121)? (Note - 1.1 permalink cited in case of edits to 1.2.)

tpluscode:
Given the hypothetical `[ sh:eval "IRI($this)" ]`, one could interpret the object as argument to SPARQL `BIND(IRI($this) as ?value)`

So what if we called this a `sh:BindExpression`, used as `[ sh:bind "IRI($this)" ]`. And then when we adopt the `sh:var`, one could almost write SPARQL like `[ sh:bind "IRI($this)"; sh:as "varName" ]`

HolgerKnublauch:
The "beauty" of sh:eval is the similarity to the JS eval method, which also takes a string and executes it as a sub-script. At least sh:sparqlEval should be better than sh:sparqlExpr to avoid the double use of "expression".

afs:
JS eval is an "action". This use in SHACL is declaring the expression written in SPARQL.

We already have

> The evaluation of a node expression is defined as a function [`eval(expr, activeGraph, scope)`](https://w3c.github.io/data-shapes/shacl12-core/#node-expressions)

and we are talking about the `expr` part.

> Does the [SPARQL Grammar](https://www.w3.org/TR/sparql11-query/#sparqlGrammar) provide any helpful hints for what could work instead of eval?

In SPARQL it is an [expression](https://www.w3.org/TR/sparql11-query/#rExpression) so `sh:expr`.

Other:

`sh:value`  
`sh:valueOf` (singular `value`).

HolgerKnublauch:
> JS eval is an "action". This use in SHACL is declaring the expression written in SPARQL.

I am in general not sure yet whether those node expressions should sound like verbs or nouns or a mix. Right now it's a mix. "select" and "ask" can be either a pointer to a SELECT query or be the verb "to select".

Any name is a compromise and language is always a bit ambiguous. We just need to pick something that we can reasonably live with.

In any case, I can of course replace eval(expr, activeGraph, scope) with evalExpr(expr, activeGraph, scope) in the specs as part of this PR in case we elect to stay with sh:eval.

afs:
I think evaluation is going to be 

```
   evalExpr(expr, activeGraph, positionalArgs, context)
   evalExpr(expr, activeGraph, namedArgs, context)
```

Whether this part of the NE document or in Core isn't clear to me.

Separating out the environment is cleaner to implement - the name parameter map is a stack across calls (and could even have named results), the environment is unchanging and maybe made immutable.

Use in SPARQL will be simple names so the use of URIs isn't going to so easy to keep things apart.

HolgerKnublauch:
Closing this PR because it is also part of https://github.com/w3c/data-shapes/pull/340 and people have made suggestions there that actually affect sh:eval. Let's continue with the other PR in a streamlined manner.

Issue #327, opened 2025-03-17
Add support for SHACL-AF Annotation Properties (for validation results)


HolgerKnublauch:
As mentioned in #234 SHACL-AF defines a section on https://w3c.github.io/shacl/shacl-af/#sparql-constraints-annotations

These are already in the shacl.ttl file and the description looks pretty much straight-forward. It was one of the features that were moved out of SHACL Core 1.0 due to lack of time.

They should be copied into SHACL-SPARQL.

PR #330, opened 2025-03-18
#327: Added section on Annotations

HolgerKnublauch:


shacl12-sparql/index.html
```diff
@@ -1211,7 +1211,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
 				</p>
 			</section>
 		</section>
-		
+				
+		<section id="sparql-constraints-annotations">
+			<h2>Annotation Properties</h2>
+			<p>
+				This section extends the general <a href="#sparql-constraints-variables">mechanism</a>
+				to produce validation results using of <a href="sparql-constraints">SPARQL-based constraints</a> or
```
```suggestion
				to produce validation results using <a href="sparql-constraints">SPARQL-based constraints</a> or
```

shacl12-sparql/index.html
```diff
@@ -1211,7 +1211,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
...
+				<a href="sparql-constraint-components">constraint components</a>.
+			</p>
+			<p>
+				Implementations that support this feature make it possible to inject so-called <dfn>annotation properties</dfn>
+				into the validation result nodes created for each <a>solution</a> produced by the SELECT queries of a
```
```suggestion
				Implementations that support this feature make it possible to inject <dfn>annotation properties</dfn>
				into the validation result nodes created for each <a>solution</a> produced by the <code>SELECT</code> queries of a
```

shacl12-sparql/index.html
```diff
@@ -1211,7 +1211,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
...
+				</tr>
+				<tr>
+					<td><code>sh:annotationValue</code></td>
+					<td>
+						Constant <a>RDF terms</a> that shall be used as default values.
+					</td>
+				</tr>
+			</table>
+			<p>
+				For each <a>solution</a> of a SELECT result set, a SHACL processor that supports annotations
```
```suggestion
				For each <a>solution</a> of a <code>SELECT</code> result set, a SHACL processor that supports annotations
```

shacl12-sparql/index.html
```diff
@@ -1211,7 +1211,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
...
+				walks through the declared result annotations.
+				The mapping from result annotations to SPARQL variables uses the following rules:
+			</p>
+			<ol>
+				<li>Use the <a>value</a> of the property <code>sh:annotationVarName</code></li>
+				<li>If no such <a>value</a> exists, use the <a>local name</a> of the <a>value</a> of <code>sh:annotationProperty</code> 
+				as the variable name</li>
```
```suggestion
				as the variable name.</li>
```

TallTed (changes requested):
Small fixes.

TallTed (approved):

ajnelson-nist:
1. Are there forbidden values for `sh:annotationProperty` or `sh:annotationVarName`?  E.g., what should happen if I say `sh:annotationVarName "value" .`?  Or `"this"`?

2. Is it ok to route multiple annotations into the same predicate, e.g.,

```turtle
		sh:resultAnnotation [
			sh:annotationProperty rdfs:seeAlso ;
			sh:annotationVarName "seeThis" ;
		] ;
		sh:resultAnnotation [
			sh:annotationProperty rdfs:seeAlso ;
			sh:annotationVarName "seeThat" ;
		] ;
```

HolgerKnublauch:
@ajnelson-nist 

1. I cannot think of invalid cases. The properties must be IRIs, and any RDF property is allowed there. For pre-existing properties, such as sh:focusNode (coming from ?this) it will mean that the validation result may have two values, which is unusual and usually unwanted, but creating these corner cases doesn't harm anything either - it's typically a user error. The variable names must be valid SPARQL variables in order to match those in the solution bindings - otherwise they simply, silently don't find matches which is OK.

2. Yes, all triples there add up. A property may get multiple constant default values or get values from multiple annotation property declarations. Again, this may be unusual but not harmful IMHO.

shacl12-sparql/index.html
```diff
@@ -1206,7 +1206,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
...
+						The name of the SPARQL variable to take the annotation values from.
+						<span data-syntax-rule="annotationVarName">Each <a>result annotation</a> has at most 1 <a>value</a>
+						for the property <code>sh:annotationVarName</code> and this <a>value</a> is <a>literal</a> with
+						<a>datatype</a> <code>xsd:string</code>.</span>
+					</td>
+				</tr>
+				<tr>
+					<td><code>sh:annotationValue</code></td>
+					<td>
+						Constant <a>RDF terms</a> that shall be used as default values.
```
ajnelson-nist, line 1256: For the sake of cross-linking consistency in `shacl-shacl.ttl`, can a `data-syntax-rule` `span` be added into this line?

shacl12-sparql/index.html
```diff
@@ -1206,7 +1206,113 @@ <h3>Validation with SPARQL-based Constraint Components</h3>
...
+			</table>
+			<p>
+				For each <a>solution</a> of a <code>SELECT</code> result set, a SHACL processor that supports annotations
+				walks through the declared result annotations.
+				The mapping from result annotations to SPARQL variables uses the following rules:
+			</p>
+			<ol>
+				<li>Use the <a>value</a> of the property <code>sh:annotationVarName</code></li>
+				<li>If no such <a>value</a> exists, use the <a>local name</a> of the <a>value</a> of <code>sh:annotationProperty</code> 
+				as the variable name.</li>
```
ajnelson-nist, line 1268: One of those niche corner cases: `http://example.org/☃` ([snowman, code point U+2603](https://www.compart.com/en/unicode/U+2603)) is a valid IRI ([RFC 3987, Section 2.2](https://datatracker.ietf.org/doc/html/rfc3987#section-2.2), grammar rule `ucschar`).  `☃` cannot be a SPARQL variable name (SPARQL grammar rule [`VARNAME`](https://www.w3.org/TR/sparql11-query/#rVARNAME)).  Do we need to leave a reminder here of SPARQL's `VARNAME` grammar rule?  If someone does try some unicode IRI---probably not an emoji , but perhaps there would be non-Latin use cases---and it wouldn't work in SPARQL, should we specify that it's the SHACL processor's responsibility to catch this at shapes graph load-time; or, instead warn the shapes graph authors that incompatible IRIs would likely be reported by the SPARQL engine (i.e., SHACL delegating to SPARQL)?  Note the latter case leaves the shapes graph _users_ scratching their heads if the shapes graph _authors_ don't see the warning.

ajnelson-nist: I'll move this to an Issue.

shacl12-sparql/index.html
ajnelson-nist: Done [in 377](https://github.com/w3c/data-shapes/pull/377/commits/e68d239e2e35644fef71429dfcd5e460f66fc0c0).

shacl12-sparql/index.html
ajnelson-nist: Moved to #378 .

ashleysommer:
I like this feature, thanks @HolgerKnublauch and thank you @ajnelson-nist for reviewing in my absence.

PR #336, opened 2025-03-24
#246: Create SHACL OWL concept-casting resource

### LINKED ISSUE #246: A SHACL OWL "profile" to aid ontologies mixing SHACL and OWL
I'm posting this mindful of the scope of the current SHACL working group not including other modeling languages.  I am hopeful that this stays within bounds for a modeling language that has already had some concepts used by SHACL, OWL 2, specifically for `owl:Ontology`, and incidentally `owl:imports`.

Summary: This is a request to add a graph as a resource in the SHACL suite for the benefit of ontologies that want to use SHACL, AND be conformant to OWL 2 DL (or other profile), and not fall into OWL 2 FULL.

There are ontologies that use a mixture of SHACL and OWL for the two languages' non-overlapping features.  There is some roundabout compatibility support for OWL class modeling with SHACL, such as in [SHACL 1.0's Section 2.1.3.3](https://www.w3.org/TR/2017/REC-shacl-20170720/#implicit-targetClass) definition of implicit class targets and [SHACL Class](https://www.w3.org/TR/2017/REC-shacl-20170720/#dfn-shacl-class)'s positional description that lets `owl:Class` substitute straightforwardly for `rdfs:Class`.

However, usage of SHACL in an OWL ontology currently does not satisfy OWL 2 DL rules, without a little extra work left to the implementing ontologists.  Syntactically, there is a requirement that can be elided from "skipping ahead" in the OWL specs to [OWL 2 Web Ontology Language Mapping to RDF Graphs (Second Edition)](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/), that:

* Whenever `o` is the object in a triple `s rdf:type o`, `o` needs to be an instance of `owl:Class` (or `owl:Restriction` and other reserved OWL IRIs, but they're beside the point for this issue).
* Whenever `p` is the predicate in a triple `s p o`, `p` must be exactly one of `owl:ObjectProperty`, `owl:DatatypeProperty`, or `owl:AnnotationProperty`.

SHACL and SHACL-SHACL are currently defined primarily with RDFS concepts.  Every class and property are `rdfs:Class` and `rdf:Property`.  The SHACL 1.0 implementation includes the triple `sh: a owl:Ontology .`, so in a sense, a toe is already dipped into the OWL waters.

The issue is, if an ontology is using SHACL and OWL, the ontology will not be conformant to OWL 2 DL unless it takes the extra step of assigning OWL modeling concepts to every concept in SHACL used in the ontology.  E.g., if `sh:declares` is used, it needs to be defined as an `owl:ObjectProperty` or `owl:AnnotationProperty`.  I think there is no reason to leave every SHACL-and-OWL adopter the task of adding these statements, especially if there is any pondering they might do on whether every property should be an `owl:AnnotationProperty` or not.

For the benefit of ontologies mixing OWL and SHACL, can a graph be provided that:

* For each class in SHACL, asserts it is an `owl:Class`;
* For each property in SHACL, asserts it is an `owl:AnnotationProperty`.

From my experience with OWL 2 DL conformance testing, I believe this would be sufficient to enable OWL 2 DL conformance testing with SHACL concepts not being an impediment.

If anyone wants to dig into the reasoning for global assignment of `owl:AnnotationProperty` - again, being mindful of scoping of working group activities - this is to:
* Allow the freedom to let SHACL properties have ranges of literals or non-literal nodes without OWL concerns, and
* Implicitly separate SHACL modeling from OWL modeling involving `owl:Restriction`s (c/o the reminder that `owl:AnnotationProperty`s cannot be restricted - see all occurrences of `owl:onProperty` in the OWL-RDF mapping I cited above, and note "AP" for "Annotation Property" never occurs).

I suggest this graph file be standalone, and not use `owl:imports` to bring in anything (including the core SHACL spec graph).  The new graph should be able to be `owl:imports`-able by itself.

This graph can be typed up in about 10 minutes with a plain text editor, and I believe would be straightforward to maintain with the evolving SHACL 1.2 and beyond specifications with a mutual coverage checking script.  The harder question for the group is the decision to provide this.

Somewhat related issues I can recall offhand:
* #169 - A suggestion to use another type instead of `owl:Ontology`
* #196 - A note that RDFS entailment can assign the type `owl:Ontology` to SHACL-SHACL too; however, IIRC (corrections welcome), OWL entailment will terminate reporting inconsistency while properties are not additionally typed per OWL 2 DL specification.

ajnelson-nist:
This PR will close #246 .

Items remaining to move out of Draft status:

- [ ] Find spot among SHACL documents to note graph file.
- [ ] Confirm placeholder `owl:versionIRI` OK.


ajnelson-nist:
In discussion on today's Profiles task force call, we determined this PR might better target the [`shacl-resources`](https://github.com/w3c/shacl-resources/) repository.  I'll be moving it there.

ajnelson-nist:
I'm closing this PR, soft-migrating to [shacl-resources PR 4](https://github.com/w3c/shacl-resources/pull/4).

PR #340, opened 2025-03-27
Issue 322: Node expressions base definition

### LINKED ISSUE #322: Node Expressions : base definition and evaluation
The current (2025-03-12) definition of Node Expression in SHACL Core reads as:

> Each node expression function has an IRI as its function name.
> Node expression functions can declare one or more node expression parameters.
> Each of these parameters has an IRI.

as if the only way to invoke a function is by its function name which isn't true (IRI expression, literal expression).

Alternative:

A node expression is one of

* A function called by its function name
* An IRI expression
* A literal expression

and the definitions of "IRI expression" and "literal expression" need changing (they can be used as direct terms or via their function name).

Evaluation is defined as `eval(expr, activeGraph, scope)` 

How to paramaters and positional arguments e.g. `[ fn:strlen ( someWayToGetInputFocusNode 2 3 ) ]`
get passed to the evaluation step?

1. What goes in `scope`?
2. Does default properties modify the active graph during evaluation?
3. Positional arguments are not in the node expression definition.

HolgerKnublauch:
This diff looks larger than it is, mainly because I moved the Property Path sections up one level.

I have simplified the definition of Node Expressions to simply define IRI and Literal expressions and then leave all other details to other documents - this is for blank nodes.

I have changed the TTL system ontology to roughly match what we had discussed, with the separation of Named and List Parameters, so that I could use sh:NamedParameterExpression in the definitions of sh:SelectExpression and sh:EvalExpression. But these details may change later as the Node Expr work starts in earnest.

The way that it's currently defined, SHACL-SPARQL does not really depend on the NodeExpr document, but Andy has suggested changes to the exprEval function that may go in there too.

Warning: This Diff also includes the not-yet-approved branch that introduced sh:eval.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2157,46 +2157,32 @@ ex:ExamplePropertyShape
 			</section>
 		</section>
 
-		<section id="expressions">
-			<h2>SHACL Expressions</h2>
+		<section id="property-paths">
+			<h2>SHACL Property Paths</h2>
 			<p>
-				This section introduces the concepts of <a href="#property-paths">SHACL property paths</a>
-				and <a href="#node-expressions">node expressions</a>.
-				Each is a mechanism to produce a list of nodes, typically based on a given focus node as input.
+				Property paths can be used at <a href="#property-shapes"><code>sh:path</code></a> to derive the <a>value nodes</a> of a property shape.
 			</p>
 			<p>
-				SHACL Core supports such expressions in the following features:
+				SHACL includes RDF terms to represent the following subset of <a>SPARQL property paths</a>:
+				<code>PredicatePath</code>, <code>InversePath</code>, <code>SequencePath</code>, <code>AlternativePath</code>,
+				<code>ZeroOrMorePath</code>, <code>OneOrMorePath</code> and <code>ZeroOrOnePath</code>.
 			</p>
-			<ul>
-				<li>Property paths can be used at <a href="#property-shapes"><code>sh:path</code></a> to derive the value nodes of a property shape.</li>
-				<li>Node expressions can be used at <a href="#property-shapes"><code>sh:values</code> and <code>sh:defaultValue</code></a> to derive the value nodes of a property shape.</li>
-				<li>Node expressions can be used at <a href="#targetNode"><code>sh:targetNode</code></a> to dynamically compute the targets of a shape.</li>
-				<li>Node expressions can be used as parameter values of most <a href="#core-components">constraint components</a> to represent constraints that may be different depending on each focus node. <span class="todo">TODO: This change needs to be made still.</span></li>
-			</ul>
-
-			<section id="property-paths">
-				<h3>SHACL Property Paths</h3>
-				<p>
-					SHACL includes RDF terms to represent the following subset of <a>SPARQL property paths</a>:
-					<code>PredicatePath</code>, <code>InversePath</code>, <code>SequencePath</code>, <code>AlternativePath</code>,
-					<code>ZeroOrMorePath</code>, <code>OneOrMorePath</code> and <code>ZeroOrOnePath</code>.
-				</p>
-				<p>
-					The following sub-sections provide syntax rules of <a>well-formed</a> <a>SHACL property paths</a>
-					together with mapping rules to <a href="https://www.w3.org/TR/sparql12-query/#pp-language">SPARQL 1.2 property paths</a>.
-					These rules define the <dfn>path mapping</dfn> <code>path(p,G)</code> in an RDF graph <code>G</code> of an RDF term <code>p</code> that is a SHACL property path in <code>G</code>.
-					Two SHACL property paths are considered <dfn data-lt="equivalent path">equivalent paths</dfn> when they map to the exact same SPARQL property paths.
-				</p>
-				<p class="syntax">
-					<span data-syntax-rule="path-metarule">A node in an RDF graph is a <a>well-formed</a> <dfn data-lt="SHACL property paths">SHACL property path</dfn> <code>p</code> if it satisfies exactly one of the syntax rules in the following sub-sections.</span>
-					<span data-syntax-rule="path-non-recursive">A node <code>p</code> is not a <a>well-formed</a> SHACL property path if <code>p</code> is a blank node and any path mappings of <code>p</code> directly or transitively reference <code>p</code>.</span>
-				</p>
-				<p>
-					The following example illustrates some valid SHACL property paths, together with their SPARQL 1.2 equivalents.
-				</p>
-				<aside class="example" title="Some valid property path expressions and their SPARQL 1.2 equivalents">
-					<div class="shapes-graph">
-						<div class="turtle">
+			<p>
+				The following sub-sections provide syntax rules of <a>well-formed</a> <a>SHACL property paths</a>
+				together with mapping rules to <a href="https://www.w3.org/TR/sparql12-query/#pp-language">SPARQL 1.2 property paths</a>.
+				These rules define the <dfn>path mapping</dfn> <code>path(p,G)</code> in an RDF graph <code>G</code> of an RDF term <code>p</code> that is a SHACL property path in <code>G</code>.
+				Two SHACL property paths are considered <dfn data-lt="equivalent path">equivalent paths</dfn> when they map to the exact same SPARQL property paths.
+			</p>
+			<p class="syntax">
+				<span data-syntax-rule="path-metarule">A node in an RDF graph is a <a>well-formed</a> <dfn data-lt="SHACL property paths">SHACL property path</dfn> <code>p</code> if it satisfies exactly one of the syntax rules in the following sub-sections.</span>
+				<span data-syntax-rule="path-non-recursive">A node <code>p</code> is not a <a>well-formed</a> SHACL property path if <code>p</code> is a blank node and any path mappings of <code>p</code> directly or transitively reference <code>p</code>.</span>
+			</p>
+			<p>
+				The following example illustrates some valid SHACL property paths, together with their SPARQL 1.2 equivalents.
+			</p>
+			<aside class="example" title="Some valid property path expressions and their SPARQL 1.2 equivalents">
+				<div class="shapes-graph">
+					<div class="turtle">
 # SPARQL Property path: ex:parent
 ex:SomeClass-predicateExample
 	sh:path ex:parent .
@@ -2226,9 +2212,9 @@ ex:SomeClass-alternativePathExample
 	sh:path [ 
 		sh:alternativePath ( ex:father ex:mother  ) 
 	] .						
-						</div>
-						<div class="jsonld">
-							<pre class="jsonld">{
+					</div>
+					<div class="jsonld">
+						<pre class="jsonld">{
 	"@graph": [
 		{
 			"@id": "ex:SomeClass-alternativePathExample",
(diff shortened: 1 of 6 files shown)


shacl12-sparql/index.html
```diff
@@ -1315,6 +1315,78 @@ <h3>Select Expressions</h3>
...
+			<section id="EvalExpression">
+				<h3>Eval Expressions</h3>
+				<p>
+					A <a>node expression</a> that has a <a>value</a> for <code>sh:eval</code> is called an <dfn>eval expression</dfn> with the <a>function name</a>
+					<code>sh:EvalExpression</code>.
+				</p>
+				<p class="syntax">
+					<span data-syntax-rule="EvalExpression-syntax-eval">A node in an RDF graph is a <a>well-formed</a> <a>eval expression</a> if it is a <a>blank node</a>
+					that is the <a>subject</a> of exactly one <a>triple</a> with <code>sh:eval</code> as <a>predicate</a> and a <a>literal</a> as <a>object</a>
+					with <a>datatype</a> <code>xsd:string</code>.</span>
```
TallTed: I've suggested changes to make clearer what I think was the intended meaning (i.e., that the blank node could be the subject of multiple triples, but only one of these may have the specified predicate and object). It may be that I've misunderstood the intent, which will lead to different change suggestions (e.g., that the blank node may be the subject of only one triple, which has the specified predicate and object). My suggestion could be ignored, in favor of construction similar to that in lines 2312–2362.
```suggestion
					<span data-syntax-rule="EvalExpression-syntax-eval">A node in an RDF graph
					is a <a>well-formed</a> <a>eval expression</a>
					if it is a <a>blank node</a>
					that is the <a>subject</a> of exactly one <a>triple</a>
					with <code>sh:eval</code> as <a>predicate</a>
					and a <a>literal</a> with <a>datatype</a> <code>xsd:string</code> as <a>object</a>.</span>
```

shacl12-sparql/index.html
```diff
@@ -1315,6 +1315,78 @@ <h3>Select Expressions</h3>
...
+					<span data-syntax-rule="EvalExpression-syntax-prefixes">A <a>well-formed</a> <a>eval expression</a> can have at most one <a>value</a> for the property
+					<code>sh:prefixes</code> and this value is a <a>IRI</a> or a <a>blank node</a>.</span><br/><br/>
```
```suggestion
					<code>sh:prefixes</code> and this value is an <a>IRI</a> or a <a>blank node</a>.</span><br/><br/>
```

shacl12-test-suite/tests/sparql/property/property-eval-001.ttl
```diff
@@ -0,0 +1,57 @@
...
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix sht: <http://www.w3.org/ns/shacl-test#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+ex:Resource-uriLength
+	a sh:PropertyShape ;
+  sh:targetNode ex:Invalid ;  # The URI of this has 29 chars
+  sh:targetNode ex:Valid ;    # The URI of this has 27 chars
+	sh:name "URI length" ;
```
TallTed: Indents should only use spaces, else we'll have ongoing issues of differing indents because of differing tab settings and differing mixes of tabs and spaces. Here, I've changed two one-tab indents to eight-space indents, and two two-space indents to eight-space indents.
```suggestion
        a sh:PropertyShape ;
        sh:targetNode ex:Invalid ;  # The URI of this has 29 chars
        sh:targetNode ex:Valid ;    # The URI of this has 27 chars
        sh:name "URI length" ;

```

TallTed (changes requested):

shacl12-sparql/index.html
```diff
@@ -1315,6 +1315,81 @@ <h3>Select Expressions</h3>
...
+					A <a>node expression</a> that has a <a>value</a> for <code>sh:eval</code> is called an <dfn>eval expression</dfn> with the <a>function name</a>
+					<code>sh:EvalExpression</code>.
+				</p>
+				<p class="syntax">
+					<span data-syntax-rule="EvalExpression-syntax-eval">A node in an RDF graph
+					is a <a>well-formed</a> <a>eval expression</a>
+					if it is a <a>blank node</a>
+					that is the <a>subject</a> of exactly one <a>triple</a>
+					with <code>sh:eval</code> as <a>predicate</a>
+					and a <a>literal</a> with <a>datatype</a> <code>xsd:string</code> as <a>object</a>.</span>
```
TallTed: I still can't tell whether this parenthetical (or its negative) is intended to be understood from the curren text. Therefore, I think this parenthetical (or its negative) should be added.
```suggestion
					and a <a>literal</a> with <a>datatype</a> <code>xsd:string</code> as <a>object</a>.
					(It may also be the <a>subject</a> of multiple <a>triples</a> 
					with other <a>predicate</a> and/or <a>object</a> values.)</span>
```

TallTed: This would be adding the negative.
```suggestion
					and a <a>literal</a> with <a>datatype</a> <code>xsd:string</code> as <a>object</a>.
					(It may not be the <a>subject</a> of any <a>triples</a> 
					with other <a>predicate</a> and/or <a>object</a> values.)</span>
```

TallTed (changes requested):

shacl12-sparql/index.html
HolgerKnublauch: Will change this to "... a blank node that has exactly one value for the predicate sh:eval and that value is a literal with datatype xsd:string". I hope this clarifies it enough.

In general, if something is not stated then it is allowed. Nothing prevents such blank nodes to have other properties, e.g. an rdf:type or annotations, and that is by design.

shacl12-core/index.html
```diff
@@ -2289,219 +2275,179 @@ <h3>SHACL Property Paths</h3>
...
+				<div class="def-header">EVALUATION OF NODE EXPRESSIONS</div>
+				The <dfn>evaluation</dfn> of a node expression is defined as a function <code>evalExpr(expr, activeGraph, scope) -> outputNodes</code>
+				where
+				<ul>
+					<li><code>expr</code> is a <a>node expression</a>.</li>
+					<li><code>activeGraph</code> is a <a>graph</a>, called the <dfn>active graph</dfn>.</li>
+					<li><code>scope</code> is a map from <a href="https://www.w3.org/TR/sparql12-query/#defn_QueryVariable">variable names</a> to individual <a>nodes</a>.
+						The value of the variable <code>focusNode</code> (if it exists) is called the <dfn>input focus node</dfn>.
+					</li>
+				</ul>
```
bergos: It's not clearly stated what is the `activeGraph`, but I guess it means the data graph. But shouldn't an expression also have access to the shapes graph? For example, the `sh:prefixes` argument of a `sh:sparqlExpr` is a named node pointing to the prefix declaration in the shapes graph and, therefore, requires access to the shapes graph.

One could see `expr` as a reference to the node expression in the shapes graph, including the shapes graph. But that should be better emphasized.

bergos (changes requested):

shacl12-core/index.html
HolgerKnublauch: The activeGraph is just a variable and which graph that is doesn't matter to the algorithm. Usually it will be the data graph (of a validation engine) but I assume node expressions will be used from various other places in the future too.

I agree the shapes graph access needs to be clarified. Will have pushed a small update attempting to do so.

HolgerKnublauch:
Hi @TallTed did my edits address your concerns? If yes, could you remove the blocking state of your review?

bergos (approved):

shacl12-sparql/index.html
```diff
@@ -1218,20 +1218,22 @@ <h3>Select Expressions</h3>
 				</p>
 				<p class="syntax">
 					<span data-syntax-rule="SelectExpression-syntax">A node in an RDF graph is a <a>well-formed</a> <a>select expression</a> if it is a <a>blank node</a>
-					that is the <a>subject</a> of exactly one <a>triple</a> with <code>sh:select</code> as <a>predicate</a> and a <a>literal</a> as <a>object</a>
-					with <a>datatype</a> <code>xsd:string</code>.</span>
+					that has exactly one <a>value</a> for the <a>predicate</a> <code>sh:select</code>
+					and this <a>value</a> is <a>literal</a> with <a>datatype</a> <code>xsd:string</code>.</span>
```
```suggestion
					and this <a>value</a> is a <a>literal</a> with <a>datatype</a> <code>xsd:string</code>.</span>
```

shacl12-sparql/index.html
```diff
@@ -1218,20 +1218,22 @@ <h3>Select Expressions</h3>
...
+					<span data-syntax-rule="SelectExpression-syntax-prefixes">A <a>well-formed</a> <a>select expression</a> can have at most one <a>value</a> for the property
+					<code>sh:prefixes</code> and this value is an <a>IRI</a> or a <a>blank node</a>.</span><br/><br/>
```
```suggestion
					<code>sh:prefixes</code> and this value can only be an <a>IRI</a> or a <a>blank node</a>.</span><br/><br/>
```

shacl12-sparql/index.html
```diff
@@ -1315,6 +1317,79 @@ <h3>Select Expressions</h3>
 					</div>
 				</aside>
 			</section>
+
+			<section id="SPARQLExprExpression">
+				<h3>SPARQL Expr Expressions</h3>
+				<p>
+					A <a>node expression</a> that has a <a>value</a> for <code>sh:sparqlExpr</code> is called an <dfn>SPARQL expr expression</dfn> with the <a>function name</a>
```
```suggestion
					A <a>node expression</a> that has a <a>value</a> for <code>sh:sparqlExpr</code> is called a <dfn>SPARQL expr expression</dfn> with the <a>function name</a>
```

shacl12-sparql/index.html
```diff
@@ -1315,6 +1317,79 @@ <h3>Select Expressions</h3>
...
+				</p>
+				<p class="syntax">
+					<span data-syntax-rule="SPARQLExprExpression-syntax-eval">A node in an RDF graph
+					is a <a>well-formed</a> <a>SPARQL expr expression</a> if it is a <a>blank node</a>
+					that has exactly one <a>value</a> for the <a>predicate</a> <code>sh:sparqlExpr</code>
+					and that <a>value</a> is a <a>literal</a> with <a>datatype</a> <code>xsd:string</code>.</span>
+					<span data-syntax-rule="SPARQLExprExpression-syntax-prefixes">A <a>well-formed</a> <a>SPARQL expr expression</a> can have at most one <a>value</a> for the property
+					<code>sh:prefixes</code> and this value is an <a>IRI</a> or a <a>blank node</a>.</span><br/><br/>
+					<span data-syntax-rule="SPARQLExprExpression-template">Let <code>$EXPR$</code> be the <a>value</a> of <code>sh:eval</code> and <code>$PREFIXES$</code>
+					be the SPARQL prefixes block resulting from the <a href="#sparql-prefixes">prefix handling rules</a> using the value of <code>sh:prefixes</code>,
```
```suggestion
					be the SPARQL prefixes block resulting from the <a href="#sparql-prefixes">prefix handling rules</a> using the value of <code>sh:prefixes</code>;
```

TallTed (changes requested):
Small fixes.

TallTed (approved):

shacl12-test-suite/tests/sparql/property/manifest.ttl
```diff
@@ -4,7 +4,7 @@
 
 <>
 	a mf:Manifest ;
-	rdfs:label "Tests converted from http://datashapes.org/sh/tests/tests/sparql/property" ;
+	mf:include <property-eval-001.ttl> ;
```
YoucTagh: The file name has changed:

```suggestion
	mf:include <property-sparqlExpr-001.ttl> ;
```

YoucTagh (approved):

shacl12-sparql/index.html
```diff
@@ -1315,6 +1317,79 @@ <h3>Select Expressions</h3>
 					</div>
 				</aside>
 			</section>
+
+			<section id="SPARQLExprExpression">
+				<h3>SPARQL Expr Expressions</h3>
+				<p>
+					A <a>node expression</a> that has a <a>value</a> for <code>sh:sparqlExpr</code> is called a <dfn>SPARQL expr expression</dfn> with the <a>function name</a>
```
ajnelson-nist, line 1324: `sh:sparqlExpr` is not in `shacl.ttl`.  Was that intentional?

HolgerKnublauch: It is, in sh:SPARQLExprExpression-sparqlExpr which is a sh:Parameter declaration. From a SHACL perspective there is no need to declare it globally, if you mean that, as it's only used as predicate.

Issue #128, opened 2020-12-18
sec 5.1 Missing prefix in example query


GeorgFerdinandSchneider:
Dear SHACL team,

In the [example ](https://www.w3.org/TR/shacl/#sparql-constraints-example) provided to describe a SPARQL-constraint, their might be a prefix missing. Some implementations report an error as the query only fires to the triple store if the prefix is registered. Up to know only tested in Stardog 7.4.3.

Currently in the document:

```turtle
ex:LanguageExampleShape
	a sh:NodeShape ;
	sh:targetClass ex:Country ;
	sh:sparql [
		a sh:SPARQLConstraint ;   # This triple is optional
		sh:message "Values are literals with German language tag." ;
		sh:prefixes ex: ;
		sh:select """
			SELECT $this (ex:germanLabel AS ?path) ?value
			WHERE {
				$this ex:germanLabel ?value .
				FILTER (!isLiteral(?value) || !langMatches(lang(?value), "de"))
			}
			""" ;
	] .
```	

Potential correction:
	
```turtle
ex:LanguageExampleShape
	a sh:NodeShape ;
	sh:targetClass ex:Country ;
	sh:sparql [
		a sh:SPARQLConstraint ;   # This triple is optional
		sh:message "Values are literals with German language tag." ;
		sh:prefixes ex: ;
		sh:select """
		        PREFIX ex: <http://example.com/ns#>
			SELECT $this (ex:germanLabel AS ?path) ?value
			WHERE {
				$this ex:germanLabel ?value .
				FILTER (!isLiteral(?value) || !langMatches(lang(?value), "de"))
			}
			""" ;
	] .
```	
Best

Georg

HolgerKnublauch:
I am aware that many users don't understand the correct syntax of prefixes in SHACL-SPARQL. The incomplete example that you highlight may have contributed. I should have added a sentence to clarify that the given snippet needs to be merged with the prefix declaration from https://www.w3.org/TR/shacl/#sparql-prefixes (this is 5.2.1 "Prefix Declarations for SPARQL Queries")

As this spec is finished and not editable, we typically leave such suggestions open as GitHub issues, until a new revision is created in the future.

PR #356, opened 2025-04-10
#128: Added links to prefix declaration

HolgerKnublauch:
This hopefully clarifies how ex: is used to declare prefixes indirectly, not as "ex:" itself.


ashleysommer (approved):
LGTM.

PR #362, opened 2025-04-14
#276: Unify `ex` namespace usage in tests

### LINKED ISSUE #276: Namespace of `ex:` in tests
I just ran `grep -R '@prefix ex:'` and noticed that the test IRIs are either similar to this...

```
./tests/sparql/pre-binding/unsupported-sparql-001.ttl:@prefix ex: <http://datashapes.org/sh/tests/sparql/pre-binding/unsupported-sparql-001.test#> .
```

... or this:

```
./tests/core/node/xone-duplicate-data.ttl:@prefix ex: <http://example.org/shacl-test/> .
```

Should these be unified, in domain and path style?  The former scopes tests' IRIs nicely, but the latter avoids introducing a domain outside of W3C.

YoucTagh:
This is a PR to unified the use of `ex` in tests

* I changed `@prefix ex: <http://...` to `@prefix ex: <http://example.com/ns#> .`
* I changed  
  ```	
  sh:declare [
        sh:namespace "..."^^xsd:anyURI ;
        sh:prefix "ex" ;
      ] ;
  ```
  to:
  ```	
  sh:declare [
        sh:namespace "http://example.com/ns#"^^xsd:anyURI ;
        sh:prefix "ex" ;
      ] ;
  ```
* I changed `PREFIX ex ...` to `http://example.com/ns#`

Closes https://github.com/w3c/data-shapes/issues/276

### DIFF
--- a/shacl12-test-suite/tests/core/complex/personexample.ttl
+++ b/shacl12-test-suite/tests/core/complex/personexample.ttl
@@ -1,5 +1,5 @@
 @prefix dash: <http://datashapes.org/dash#> .
-@prefix ex: <http://datashapes.org/sh/tests/core/complex/personexample.test#> .
+@prefix ex: <http://example.com/ns#> .
 @prefix mf: <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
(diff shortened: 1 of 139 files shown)


ajnelson-nist:
I noticed a lot of `diff` noise unrelated to the in-scope changes of this PR.  It looks like line endings weren't consistent between files - some were DOS endings, some not.

Dear reviewers: How are we handling line endings repository-wide?  Does this PR's initial commit need to be force-pushed to minimize its change to the in-scope effects?  I find it unfortunate to lose `git blame` value from line ending issues.

YoucTagh:
I believe it is also possible to use `.gitattributes`, but I haven't used it before.

ajnelson-nist (commented):
(Commenting using the Review mechanism - same sentiment as in my initial comment on this PR.)

I would prefer this PR not alter every line of every file due to line ending changes, unless the group agrees to add some mechanism to prevent a future global change from happening accidentally.  `.gitattributes` would be fine if someone could get it working.  I've personally been unlucky with that file, so I'll refrain from adding it myself, but won't object to someone else adding it.

Issue #311, opened 2025-03-06
Support Node Expressions as constraint parameters


HolgerKnublauch:
The Core spec currently has a red TODO item to allow node expressions in most of the constraint parameters such as sh:datatype or sh:minCount. All of these values currently need to be constants and cannot be computed depending on the context like node expressions could.

This ticket here is to apply this generalization to the Core document and to add test cases and a couple of examples from the SPARQL document.

HolgerKnublauch:
I am not sure whether this should apply to all constraint components. My concern is that if we allow them in too many place (esp sh:class and sh:datatype), the static analysis of shapes graphs becomes too difficult and SHACL becomes less useful. Other constraint components like sh:maxInclusive are much more suitable and less on the critical path.

Also, the implementation burden becomes larger if we support this flexibility in too many places.

With this in mind, here is the list of constraint components with suggested yes or no to allow node expressions.

- sh:class NO - should remain limited to IRIs
- sh:datatype NO - should remain limited to IRIS, and cannot think of a use case to support different datatypes depending on the focus node
- sh:nodeKind NO - cannot think of use case
- sh:minCount YES e.g. to make some properties mandatory if certain conditions are true
- sh:maxCount PERHAPS, does anyone have use cases? Otherwise there are benefits in having sh:maxCount 1 predictably for all values of a property
- sh:min/max/Ex/Inclusive YES
- sh:minLength/maxLength PERHAPS - can anyone see use cases though?
- sh:pattern - YES, e.g. different zip code format depending on country
- sh:singleLine - NO - cannot think of use case
- sh:languageIn - PERHAPS - use case?
- sh:uniqueLang - NO - cannot think of use case
- sh:equals, disjoint, lessThan, lessThanOrEquals - NO as values need to be URIs right now
- sh:not, and, or, xone - NO as the values are shapes and shape lists
- sh:node - PERHAPS, e.g. to select a different shape for ex:address depending on ex:country
- sh:property - NO - SHACL would become too unpredictable
- sh:qualifiedValueShape - NO, sh:qualifiedMinCount/MaxCount PERHAPS
- sh:closed - NO
- sh:hasValue - YES
- sh:in - YES

How do other people feel?

ajnelson-nist:
Is there currently a way sketched to use node expressions to devise new IRI-nodes in the data graph?

If so, are there conditions under which a node expression used on one of these constraints could cause an infinite loop?  Maybe `sh:in`?

TallTed:
[...]

PR #364, opened 2025-04-14
#311: Don't support Node Expressions as constraint parameters in Core

HolgerKnublauch:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-311/shacl12-core/index.html)

This PR only covers Core - removing the previously suggested allowance of node expressions as constraint parameters. Several people had voiced concerns, while others were in general favour. The proposal here is to instead leave this to the Node Expressions document, as a more flexible optional dialect of SHACL.

(Even if this gets approved, I will wait until enough people had a chance to disagree).

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2949,7 +2949,7 @@ ex:SomeClass-alternativePathExample
 				i.e. the <a>object</a> of the <a>triple</a> in the <a>shapes graph</a> where the <a>subject</a> is the <a>shape</a>
 				and the <a>predicate</a> is the <a>parameter</a> (such as <code>sh:class</code>).
 			</p>
-			<p>
+			<p class="issue">
 				At the time of writing, the intent of the WG is to define a dialect of SHACL outside of SHACL Core in which the term
 				<a>parameter value</a> also allows <a>node expressions</a>.
 				Note that not all constraint components use the term <a>parameter value</a> but instead refer to the term <a>value</a>.
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2391,7 +2391,6 @@ ex:SomeClass-alternativePathExample
 				<li>At <a href="#property-shapes"><code>sh:values</code> and <code>sh:defaultValue</code></a> to derive the value nodes of a property shape.</li>
 				<li>At <a href="#targetNode"><code>sh:targetNode</code></a> to dynamically compute the targets of a shape.</li>
 				<li>At <a href="#ExpressionConstraintComponent"><code>sh:expression</code></a> to validate nodes against a condition.</li>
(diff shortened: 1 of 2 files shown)


ajnelson-nist (approved):

YoucTagh (approved):

bergos (approved):

robert-david:
I think the meaning of `activeGraph` is not explained in https://raw.githack.com/w3c/data-shapes/issue-311/shacl12-core/index.html#dfn-node-expression.

HolgerKnublauch:
@robert-david ok makes sense. I have attempted to clarify that active graph. The idea is that SHACL Core does not really depend on the active graph, but each Node Expression (e.g. in the SHACL-SPARQL document) should reference it when it defines what queries need to be executed.

afs:
This can just be "graph" in `evalExpr(expr, activeGraph, scope)`.

The current SHACL definition of "active graph" only says it is called "the active graph".

> activeGraph is a [graph](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-graph) [link to RDF concepts] , called the active graph.

If you want to keep the terminology, then define "activeGraph" as being the graph that the node expression has access to during evaluation and say that the initial (i.e. first, top) activeGraph is the data graph. "active graph" is SPARQL terminology so a different name for a different concept might be helpful. "focus graph" would follow SHACL style, otherwise a sentence like "the focus graph of the node expression is the active graph of the dataset" is going to be odd.

HolgerKnublauch:
@afs good suggestion. I have renamed active graph to focus graph. I am doing all this in https://github.com/w3c/data-shapes/pull/370 which is a different PR from this one here, because that overlapped otherwise.

HolgerKnublauch:
I would appreciate if any concerns about this could be raised before this Friday. I will then merge the branch into main.

shacl12-core/index.html
```diff
@@ -2946,6 +2945,19 @@ <h2>Core Constraint Components</h2>
 				<code>$paramName</code> where <code>paramName</code> is the part of the parameter's <a>IRI</a> after the <code>sh:</code> namespace.
 				For example, the textual definition of <code>sh:ClassConstraintComponent</code> refers to the value of
 				<code>sh:class</code> using the variable <code>$class</code>.
+				In SHACL Core, the term <dfn data-lt="parameter value|parameter values">parameter value</dfn> means the <a>value</a> of a parameter,
+				i.e. the <a>object</a> of the <a>triple</a> in the <a>shapes graph</a> where the <a>subject</a> is the <a>shape</a>
+				and the <a>predicate</a> is the <a>parameter</a> (such as <code>sh:class</code>).
+			</p>
+			<p>
```
```suggestion
			<p class="issue">
```

PR #366, opened 2025-04-17
#164: Added hint on execution order for sh:and and sh:or

### LINKED ISSUE #164: Shortcut evaluation semantics for `sh:and`, `sh:or`, `sh:xone`
https://www.w3.org/TR/shacl/#AndConstraintComponent says of `sh:and`:
> Note that although sh:and has a SHACL list of shapes as its value, the order of those shapes does not impact the validation results.

I suggest to add this:
> For implementations that use shortcut evaluation semantics, the order may impact the efficiency of validation. It is recommended to put earlier in the list constraints that are easier to evaluate, or are more likely to fail.

A similar consideration applies to `sh:or` where we can add:
> ... or are more likely to succeed.

A similar consideration applies to `sh:xone`. I think (but I'm not sure) we can add
> ... or are more likely to succeed (the first two constraints that succeed will cause the overall shape to fail)

HolgerKnublauch:
I didn't do sh:xone because that's much less clear, and also much less used.

Closes #164

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-164/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -4707,6 +4707,8 @@ ex:NotExampleShape
 					<p>
 						Note that although <code>sh:and</code> has a <a>SHACL list</a> of shapes as its value,
 						the order of those shapes does not impact the validation results.
+						For implementations that use shortcut evaluation semantics, the order may impact the efficiency of validation.
+						It is recommended to put earlier in the list constraints that are easier to evaluate, or are more likely to fail.
 					</p>
 					<p>
 						The following example illustrates the use of <code>sh:and</code> in a shape to specify the condition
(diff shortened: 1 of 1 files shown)


ajnelson-nist (approved):
Non-Editor feedback: I appreciate these changes, and I agree with not putting remarks on `sh:xone`.  Arguments of "Try these shapes first because they'll match *faster*" could be conditional on "Well, in the less-common cases that you're matching this one shape 1/3 of the day down the `sh:xone` list, this other shape 3/4 of the way down the `sh:xone` list is the most likely confusion point."  So, general guidance for shortcut evaluations for `sh:xone` seem impractical to suggest, but the reminders on shortcut evaluation added in this PR could be good reminders for shape authors.

YoucTagh (approved):

PR #368, opened 2025-04-21
#300: Added support for RDF 1.2 triple terms

### LINKED ISSUE #300: Support for RDF 1.2 triple terms
Triples used as RDF terms:
https://www.w3.org/TR/rdf12-concepts/#section-triple-terms

Reifiying triples - also known as "occurences" have syntactic support in Turtle and wil have in JSON-LD and RDF/XML.
https://www.w3.org/TR/rdf12-turtle/#reifying-triples

Annotation support (syntactic support)
https://www.w3.org/TR/rdf12-turtle/#annotation-syntax

N-Triples does not have syntactic support
https://www.w3.org/TR/rdf12-n-triples/#triple-terms

bergos:
This PR adds support for RDF 1.2 triple terms, as discussed in #300.

Since this is the first RDF 1.2 feature PR, I want to discuss in the next meeting how to handle RDF 1.2 features before merging it.

Closes #300

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-300/shacl12-core/index.html#ReifiableByShapeConstraintComponent)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5691,7 +5691,6 @@ ex:HandShape
 					<p>
 						<code>sh:reifierShape</code> can be used to link a <a>property shape</a> with one or more <a>node shapes</a>.
 						Any <a>reifier</a> must conform to these node shapes.
-						If a property shape contains a <code>sh:ReifierShapeConstraintComponent</code>, <code>sh:path</code> values are constrained to IRIs.
 					</p>
 
 					<div class="parameters">Parameters:</div>
@@ -5704,7 +5703,10 @@ ex:HandShape
 							<td><code>sh:reifierShape</code></td>
 							<td>
 								The <a>node shape</a> that a reifier for this triple must conform to.
-								<span data-syntax-rule="reifierShape-node">The values of <code>sh:reifierShape</code> must be <a>well-formed</a> <a>node shapes</a>.</span>
+								<span data-syntax-rule="reifierShape-node">
+									The values of <code>sh:reifierShape</code> must be <a>well-formed</a> <a>node shapes</a>.
+									If a value for <code>sh:reifierShape</code> is given, <code>sh:path</code> values are constrained to IRIs.
+								</span>
 							</td>
 						</tr>
 						<tr>
@@ -5722,7 +5724,7 @@ ex:HandShape
 						<div class="def-text-body" data-validator="ReifierShape">
 							Let <code>t</code> be the <a>triple term</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
 							For each <a>reifier</a> for the <a>triple term</a> <code>t</code>, a failure MUST be produced if validating the <a>reifier</a> against the <a>node shape</a> <code>$reifierShape</code> with the <a>reifier</a> as <a>focus node</a> produces a <a>failure</a>.
-							Otherwise, if the <a>reifiers</a> of <code>t</code> do not conform to <code>$reifierShape</code>, there is a validation result with <code>t</code> as <code>sh:value</code>.
+							For each <a>reifier</a> <code>t</code> that does not conform to <code>$reifierShape</code>, there is a validation result with <code>t</code> as <code>sh:value</code>.
 						</div>
 					</div>
 
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -495,6 +495,8 @@
 						<dfn data-cite="rdf12-concepts#dfn-literal" data-lt="literal|literals">literal</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-datatype" data-lt="datatype">datatype</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-blank-node" data-lt="blank node|blank nodes">blank node</dfn>,
+						<dfn data-cite="rdf12-concepts#dfn-triple-term" data-lt="triple term|triple terms">triple term</dfn>,
+						<dfn data-cite="rdf12-concepts#dfn-reifier" data-lt="reifier">reifier</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-node" data-lt="node|nodes">node</dfn> of an RDF graph,
 						<dfn data-cite="rdf12-concepts#dfn-rdf-term" data-lt="term|terms">RDF term</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-subject" data-lt="subject|subjects">subject</dfn>,
@@ -5679,6 +5681,98 @@ ex:HandShape
 		"@id": "ex:Hand"
 	}
 }</pre>
+							</div>
+						</div>
+					</aside>
+				</section>
+
+				<section id="ReifiableByShapeConstraintComponent">
+					<h4>sh:reifiableBy, sh:reificationRequired</h4>
+					<p>
+						<code>sh:reifiableBy</code> can be used to link a <a>property shape</a> with one or more <a>node shapes</a>.
+						Any reified statement must conform to these node shapes.
+					</p>
+
+					<div class="parameters">Parameters:</div>
+					<table class="term-table">
+						<tr>
+							<th>Property</th>
+							<th>Summary and Syntax Rules</th>
+						</tr>
+						<tr>
+							<td><code>sh:reifiableBy</code></td>
+							<td>
+								The <a>node shape</a> that the reified statement must conform to.
+								<span data-syntax-rule="reifiableBy-node">The values of <code>sh:reifiableBy</code> must be <a>well-formed</a> <a>node shapes</a>.</span>
+							</td>
+						</tr>
+						<tr>
+							<td><code>sh:reificationRequired</code></td>
+							<td>
+								This is an <a>optional parameter</a> of <code>sh:ReifiableByConstraintComponent</code>.
+								If set to <code>true</code> there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
+								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifiableBy">
+							Let <code>t</code> be the <a>triple terms</a> of the triples obtained from the last traversal step when following the <code>$path</code> <a>SHACL property path</a> of the <a>property shape</a> from the <a>focus node</a> to the <a>value node</a>.
+							For each reified statement for the <a>triple terms</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
+						</div>
+					</div>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION of sh:reificationRequired</div>
+						<div class="def-text-body" data-validator="ReificationRequired">
+							If <code>$reificationRequired</code> is set to <code>true</code>, a failure MUST be produced when there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a>.
+						</div>
+					</div>
+
+					<aside class="example">
+						<div class="shapes-graph">
+							<div class="turtle">
+ex:ProvenanceShape
+  a sh:NodeShape ;
+  sh:property [
+    sh:path ex:date ;
+    sh:datatype xsd:date ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path ex:author ;
+    sh:nodeKind sh:IRI ;
+    sh:maxCount 1 ;
+  ] .
+
+ex:PersonShape
+  a sh:NodeShape ;
+  sh:targetClass ex:Person ;
+  sh:property ex:PersonShape-age .
+
+ex:PersonShape-age
+  a sh:PropertyShape ;
+  sh:path ex:age ;
+  sh:datatype xsd:integer ;
+  sh:maxCount 1 ;
+  sh:reifiableBy ex:ProvenanceShape ;
+  sh:reificationRequired true .
+							</div>
+							<div class="jsonld">
+								<pre class="jsonld"></pre>
+							</div>
+						</div>
+						<div class="data-graph">
+							<div class="turtle">
+ex:Bob ex:age 23 {|
+  ex:date "2019-12-05"^^xsd:date .
+  ex:author ex:Claire
+|}.
+							</div>
+							<div class="jsonld">
+								<pre class="jsonld"></pre>
 							</div>
 						</div>
 					</aside>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5711,7 +5711,7 @@ ex:HandShape
 							<td><code>sh:reificationRequired</code></td>
 							<td>
 								This is an <a>optional parameter</a> of <code>sh:ReifiableByConstraintComponent</code>.
-								If set to <code>true</code> there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
+								If set to <code>true</code>, there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
 								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
 							</td>
 						</tr>
@@ -5721,7 +5721,7 @@ ex:HandShape
 						<div class="def-header">TEXTUAL DEFINITION</div>
 						<div class="def-text-body" data-validator="ReifiableBy">
 							Let <code>t</code> be the <a>triple term</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
-							For each reified statement for the <a>triple term</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
+							For each reified statement for the <a>triple term</a> <code>t</code>, a failure MUST be produced if validating the reified statement against the <code>$reifiableBy</code> <a>node shape</a> with the <a>reifier</a> as <a>focus node</a> produces a <a>failure</a>.
 							Otherwise, if the reified statements of <code>t</code> do not conform to <code>$reifiableBy</code>, there is a validation result with <code>t</code> as <code>sh:value</code>.
 						</div>
 					</div>
@@ -5729,7 +5729,7 @@ ex:HandShape
 					<div class="def def-text">
 						<div class="def-header">TEXTUAL DEFINITION of sh:reificationRequired</div>
 						<div class="def-text-body" data-validator="ReificationRequired">
-							If <code>$reificationRequired</code> is set to <code>true</code>, and when there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a> there is a validation result with <code>t</code> as <code>sh:value</code>.
+							If <code>$reificationRequired</code> is set to <code>true</code> and there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a>, there is a validation result with <code>t</code> as <code>sh:value</code>.
 						</div>
 					</div>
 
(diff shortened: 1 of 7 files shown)


shacl12-core/index.html
```diff
@@ -5683,6 +5685,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifiableBy">
+							Let <code>t</code> be the <a>triple terms</a> of the triples obtained from the last traversal step when following the <code>$path</code> <a>SHACL property path</a> of the <a>property shape</a> from the <a>focus node</a> to the <a>value node</a>.
+							For each reified statement for the <a>triple terms</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
```
HolgerKnublauch: Note the terminology here: I believe you mean validation result, not failure. Validation result is something like sh:Violation. Failure is used when the validation produces something like an exception.  The same applies in the next text definition.

shacl12-core/index.html
```diff
@@ -5683,6 +5685,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+								If set to <code>true</code> there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
+								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifiableBy">
+							Let <code>t</code> be the <a>triple terms</a> of the triples obtained from the last traversal step when following the <code>$path</code> <a>SHACL property path</a> of the <a>property shape</a> from the <a>focus node</a> to the <a>value node</a>.
```
HolgerKnublauch: We would need to exactly define what the last traversal step is. This may be tricky. What about limiting this to property shapes with an IRI or sh:inversePath of an IRI as sh:path? This would clearly define the triple.

shacl12-test-suite/tests/core/node/reifiableBy-001.ttl
```diff
@@ -0,0 +1,66 @@
...
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix sht: <http://www.w3.org/ns/shacl-test#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+ex:InvalidResource1
+  ex:propertyA "invalid" {|
+      ex:properrtyB false ;
```
afs:
```suggestion
      ex:propertyB false ;
```
or, if intended, this is too subtle a difference.

shacl12-test-suite/tests/core/node/reifiableBy-001.ttl
```diff
@@ -0,0 +1,66 @@
...
+  sh:targetNode ex:InvalidResource1 ;
+  sh:targetNode ex:ValidResource1 ;
+.
+ex:TestShape-propertyA
+  sh:path ex:propertyA ;
+  sh:reifiableBy ex:ReifyShape ;
+.
+ex:ValidResource1
+  ex:propertyA "valid" {|
+      ex:properrtyB true ;
```
afs:
```suggestion
      ex:propertyB true ;
```

shacl12-core/index.html
```diff
@@ -5683,6 +5685,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
 						</div>
 					</aside>
 				</section>
+
+				<section id="ReifiableByShapeConstraintComponent">
+					<h4>sh:reifiableBy, sh:reificationRequired</h4>
+					<p>
+						<code>sh:reifiableBy</code> can be used to link a <a>property shape</a> with one or more <a>node shapes</a>.
+						Any reified statement must conform to these node shapes.
```
afs: `{| |}` is surface syntax for the triples.

I think SHACL needs to define validation for the basic forms then work up from there. This PR seems to be for an annotation usage pattern but that isn't the only way to use RDF 1.2.

Starting with:

```
ex:ValidResource1
  ex:propertyA "valid" {|
      ex:propertyB true ;
  |} .
```
`{| |}` is syntax to both reify a triple and provide triple about the reification. It is only syntax and is not retained in the RDF graph.

```
ex:ValidResource1  ex:propertyA  "valid" .
<< ex:ValidResource1 ex:propertyA "valid" >> ex:propertyB  true .
```
and `<< >>` (an occurence - a _use_ of a triple term) is in turn surface syntax  for `rdf:reifies`
so the whole thing three triples:

```
ex:ValidResource1  ex:propertyA  "valid" .
_:b0    rdf:reifies   <<( ex:ValidResource1 ex:propertyA "valid" )>>;
        ex:propertyB  true .
```

`<<( ... )>>` is a _triple term_.

I think the intent is to apply a shape to the structure and also the triples insided `{| |}`.
Is that right?

Only Turtle/TriG have this annotation syntax.

Data does not have to use `{| |}` and it does not exist graph / triple store.

Sometimes, there may not be an asserted triple:
```
<< :s :p :o ~ _:reif1 >> :added "2025-04-01" .
<< :s :p :o ~ _:reif1 >> :deleted "2025-05-01" .
```
which is **3** triples:
```
_:b0    rdf:reifies  <<( :s :p :o )>>;
        :added       "2025-04-01" .
        rdf:reifies  <<( :s :p :o )>>; ## Duplicate - same as the first triple.
        :deleted     "2025-05-01" .
```
Note the shared blank node.

shacl12-core/index.html
bergos: Yes, that would make more sense. I aligned the wording to the other shape-based and logical constraint components. All of them use the wording "produces a failure". Should we also change them?

shacl12-core/index.html
bergos: I would like to support complex paths, but it's a good idea to limit it to IRI property paths in the initial version and create an issue to extend it later.

shacl12-core/index.html
bergos: Yes, it's for the use case where an asserted triple can be used to traverse into the triple term. The text definition says that the reifier will become the focus node for that validation.

Based on the discussion in #300, I thought we don't need to cover the use case without asserted triples. Node Expressions, or SPARQL-based targets can be used for that use case.

shacl12-core/index.html
```diff
@@ -2064,7 +2064,9 @@ <h3>Property Shapes</h3>
 					that is the <a>subject</a> of a <a>triple</a> that has <code>sh:path</code> as its <a>predicate</a>.</span>
 					<span data-syntax-rule="path-maxCount">A shape has at most one <a>value</a> for <code>sh:path</code>.</span>
 					<span data-syntax-rule="path-node">The <a>value</a> of <code>sh:path</code> in a property shape is a <a>well-formed</a>
-					<a>SHACL property path</a>.</span>
+					<a>SHACL property path</a>. If a property shape contains a <code>sh:ReifiableByConstraintComponent</code>,
+					<code>sh:path</code> values are constrained to IRIs.</span>
```
HolgerKnublauch: I would move this into the sh:reifiableBy definition itself, e.g. "The $path is an IRI" or "The sh:path MUST be an IRI". Whenever any constraint definition is ill-formed, the whole shapes graph is invalid and thus can produce a failure. I believe it's better to bundle this rule where it belongs because there are already other exceptions on sh:path, e.g. sh:values can only be used for IRIs, and moving it into sh:reifiableBy keeps the document more modular.

shacl12-core/index.html
HolgerKnublauch: To me the logical components look correct. They basically forward any failure that happens during the evaluation of the nested shapes, but they normally return a validation result (e.g. sh:Violation) when the nodes violate the shape. So I think there is still a sentence missing that a validation result is produced when the validation of the reification does not conform to the given shape.

shacl12-core/index.html
```diff
@@ -5683,6 +5687,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+								If set to <code>true</code> there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
+								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifiableBy">
+							Let <code>t</code> be the <a>triple terms</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
```
```suggestion
							Let <code>t</code> be the <a>triple term</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
```

shacl12-core/index.html
```diff
@@ -5683,6 +5687,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+							For each reified statement for the <a>triple terms</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
```
```suggestion
							For each reified statement for the <a>triple term</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
```

shacl12-core/index.html
```diff
@@ -5683,6 +5687,98 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+						</div>
+					</div>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION of sh:reificationRequired</div>
+						<div class="def-text-body" data-validator="ReificationRequired">
+							If <code>$reificationRequired</code> is set to <code>true</code>, a failure MUST be produced when there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a>.
```
afs: There can be zero or more groups of triples 

```
## Reifier, no triple with that as subject
:s :p :o1 ~:r .  
## Same as :s :p :o1 ~:r  {} . 
```

```
## Two reifiers, each with triples.
:s :p :o2 {| ex:signedOffBy "A" |}
          {| ex:signedOffBy "B" |} .
```
In this example, `reificationRequired` is more count-like. This includes 0 -- the first example above -- here the reifier is allocated for later use.

afs: > Node Expressions, or SPARQL-based targets can be used for that use case.

How? SPARQL-based constraints can do many things but what would node expressions look like?

General observation: Pushing things onto node expressions may work technically but they aren't necessarily a friendly way to express things. If the WG decides to publish core as a REC ASAP, then core is then fixed and unchangeable. The WG needs to be certain that pushing out things out to phase2 actually ends up with the best outcome.

shacl12-core/index.html
bergos: Sorry, I missed that part. Now it should be correct.

shacl12-core/index.html
bergos: We can handle that change in a separate PR. That's why I've moved the discussion back to the [issue](https://github.com/w3c/data-shapes/issues/300#issuecomment-2826634593).

shacl12-core/index.html
```diff
@@ -5683,6 +5685,100 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+							<td>
+								The <a>node shape</a> that the reified statement must conform to.
+								<span data-syntax-rule="reifiableBy-node">The values of <code>sh:reifiableBy</code> must be <a>well-formed</a> <a>node shapes</a>.</span>
+							</td>
+						</tr>
+						<tr>
+							<td><code>sh:reificationRequired</code></td>
+							<td>
+								This is an <a>optional parameter</a> of <code>sh:ReifiableByConstraintComponent</code>.
+								If set to <code>true</code> there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
```
```suggestion
								If set to <code>true</code>, there must be at least one reification value for the focus node/path combination in the <a>data graph</a>.
```

shacl12-core/index.html
```diff
@@ -5683,6 +5685,100 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+								<span data-syntax-rule="reificationRequired-datatype">The values of <code>sh:reificationRequired</code> in a shape are literals with datatype <code>xsd:boolean</code>.</span>
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifiableBy">
+							Let <code>t</code> be the <a>triple term</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
+							For each reified statement for the <a>triple term</a> <code>t</code>, a failure MUST be produced if the validation of the reified statement with the <a>reifier</a> as <a>focus node</a> against the <a>node shape</a> <code>$reifiableBy</code> produces a <a>failure</a>.
```
```suggestion
							For each reified statement for the <a>triple term</a> <code>t</code>, a failure MUST be produced if validating the reified statement against the <code>$reifiableBy</code> <a>node shape</a> with the <a>reifier</a> as <a>focus node</a> produces a <a>failure</a>.
```

shacl12-core/index.html
```diff
@@ -5683,6 +5685,100 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+							Otherwise, if the reified statements of <code>t</code> do not conform to <code>$reifiableBy</code>, there is a validation result with <code>t</code> as <code>sh:value</code>.
+						</div>
+					</div>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION of sh:reificationRequired</div>
+						<div class="def-text-body" data-validator="ReificationRequired">
+							If <code>$reificationRequired</code> is set to <code>true</code>, and when there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a> there is a validation result with <code>t</code> as <code>sh:value</code>.
```
```suggestion
							If <code>$reificationRequired</code> is set to <code>true</code> and there is no reified statement for the <a>triple term</a> <code>t</code> in the <a>data graph</a>, there is a validation result with <code>t</code> as <code>sh:value</code>.
```

TallTed (changes requested):
A few changes, which I hope improve clarity and readability.

afs:
This PR is motivated to the annotation pattern/syntax in RDF 1.2 Turtle.

Annotation hides the reifier although the reifier can be explicitly given.

The property shape object of `sh:reifiableBy` is applied to asserted triples whose subject is the reifier.

`sh:reifiableBy` could instead be named `sh:describedBy` or `sh:tripleDescribedBy` or `sh:annotatedBy`.

With explicit reifier:
```
:s :p :o ~:r {| ex:signedOffBy "A" |}
```

then the name `sh:reifiableBy` could be used to get the reifier (node shape).

TallTed:
> The property shape object of `sh:reifiableBy` is applied to asserted triples whose subject is the reifier.
>
> `sh:reifiableBy` could instead be named `sh:describedBy` or `sh:tripleDescribedBy` or `sh:annotatedBy`.

I must be missing something somewhere. Why not `sh:reifiedBy`?

HolgerKnublauch:
@TallTed the name probably was copied from dash:reifiableBy. The reason is that reifiedBy would mean that reification is required, while reifableBy means that if a reification exists then it must have the given shape.

TallTed:
> `sh:reifableBy` means that if a reification exists then it must have the given shape

Fair enough.

afs:
I think Ted has a point. 

To get to the focus node for the annotations (data graph traversal), there must be a connection between the reifier (blank node or the `:r` above) and the current triple as a triple term. This is the `rdf:reifies` triple(s) with this focus triple as a triple term in the object position.

The annotation syntax checking resides in `ex:ProvenanceShape`.

The focus node of `ex:ProvenanceShape` is the reifier.  
If it has no other constraints, it becomes a test for the existence of `rdf:reifies`.
(implicitly it has `sh:property [ sh:path rdf:reifies ]`)

Suggestion: rename `sh:reifiableBy` as 

`sh:reifierShape` -- "The [node shape] that a reifier for this triple must conform to."

```
ex:PersonShape-age
  a sh:PropertyShape ;
  sh:path ex:age ;
  sh:datatype xsd:integer ;
  sh:maxCount 1 ;
  sh:reifierShape ex:ProvenanceShape ;
  .
```

```
ex:ProvenanceShape
    sh:property [ 
      sh:path rdf:reifies ;
      sh:nodeKind sh:IRI ; ## e.g. requires the reifier to be an IRI
    ] ;
    sh:property [
      sh:path ex:author ;
      sh:nodeKind sh:IRI ;
      sh:maxCount 1 ;
    ] .
] .
```

There can be multiple reifiers, each with an annotation blocks in the syntax. I'm not sure what that is going to look like.

bergos:
I've renamed `sh:reifiableBy` to `sh:reifierShape`. Also, "reified statement" has been replaced with "reifier" to match the RDF Concepts specification.

@afs
Did you mean the inverse of `rdf:reifies`?

```turtle
ex:ProvenanceShape
    sh:property [ 
      sh:path rdf:reifies ; # inverse?
      sh:nodeKind sh:IRI ; ## e.g. requires the reifier to be an IRI
    ]
] .
```

bergos:
@afs @TallTed @HolgerKnublauch Can you please review the latest changes?

shacl12-core/index.html
```diff
@@ -5683,6 +5685,100 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
 						</div>
 					</aside>
 				</section>
+
+				<section id="ReifierShapeShapeConstraintComponent">
+					<h4>sh:reifierShape, sh:reificationRequired</h4>
+					<p>
+						<code>sh:reifierShape</code> can be used to link a <a>property shape</a> with one or more <a>node shapes</a>.
+						Any <a>reifier</a> must conform to these node shapes.
+						If a property shape contains a <code>sh:ReifierShapeConstraintComponent</code>, <code>sh:path</code> values are constrained to IRIs.
```
HolgerKnublauch: The term "contains" is unclear here and I think it would be better to say "if a property shape has a <a>value</a> for <code>sh:reifierShape</code>...

HolgerKnublauch: Also I believe this is a formal constraint and needs to be moved into a data-syntax-rule, i.e. under the parameter definition of sh:reifierShape.

shacl12-core/index.html
```diff
@@ -5683,6 +5685,100 @@ <h4>sh:qualifiedValueShape, sh:qualifiedMinCount, sh:qualifiedMaxCount</h4>
...
+							</td>
+						</tr>
+					</table>
+
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="ReifierShape">
+							Let <code>t</code> be the <a>triple term</a> (<a>focus node</a>,  <code>$path</code>, <a>value node</a>).
+							For each <a>reifier</a> for the <a>triple term</a> <code>t</code>, a failure MUST be produced if validating the <a>reifier</a> against the <a>node shape</a> <code>$reifierShape</code> with the <a>reifier</a> as <a>focus node</a> produces a <a>failure</a>.
+							Otherwise, if the <a>reifiers</a> of <code>t</code> do not conform to <code>$reifierShape</code>, there is a validation result with <code>t</code> as <code>sh:value</code>.
```
HolgerKnublauch: "if the reifiers" is unclear: do you mean any of them or all of them, some etc. Better write it like "For each reifier <code>t</code> that does not conform..."

shacl12-vocabularies/shacl.ttl
```diff
@@ -1120,6 +1120,40 @@ sh:qualifiedValueShapesDisjoint
...
+
+sh:ReifierShapeConstraintComponent-reificationRequired
+	a sh:Parameter ;
+	sh:path sh:reificationRequired ;
+	sh:datatype xsd:boolean ;
+	rdfs:isDefinedBy sh: .
+
+sh:reifierShape
+	a rdf:Property ;
+	rdfs:label "reifiable by"@en ;
```
HolgerKnublauch: -> "reifier shape"

HolgerKnublauch (approved):

HolgerKnublauch:
I have approved this for now, but I expect further changes will happen in follow-up tickets. It is important though to get this into the main branch to have a starting point.

TallTed (approved):
Good enough for merge. Iterate for further improvement.

HolgerKnublauch:
Just a small observation here. When you define a sh:reifierShape constraint that is sh:closed, it will likely flag a constraint violation because the subject will hold a rdf:reifies triple. In those cases, sh:closed needs to be combined with sh:ignoredProperties ( rdf:reifies ).

PR #370, opened 2025-04-22
#314: Made focusNode an explicit param of exprEval, SPARQL-based expressions now pre-bind all scope vars

### LINKED ISSUE #314: Add sh:SPARQLFunction
SHACL-AF include vocabulary to declare new (SPARQL) functions: https://w3c.github.io/shacl/shacl-af/#functions

The SHACL-SPARQL 1.2 spec should include this feature. In TopBraid we have supported this for many years and it has proven to be very useful for encapsulating reusable query snippets.

HolgerKnublauch:
Two changes in this PR:

- Instead of having focusNode as part of the scope, I have pulled it out into an explicit special variable. This is cleaner because this variable really has a special status (e.g. it's called this in SPARQL) and it usually has a value (except in sh:targetNode).
- The SPARQL-based node expressions now accept the scope vars as parameters. This makes it possible to define user-defined (SPARQL) functions through custom node expression types. For map-based NEe, these would be simple name-value pairs for the input parameters. For list-based NEs, these could be arg1, arg2, arg3 etc. The NE document can specify the details, but the SPARQL document does not need to change further to accommodate the use cases.

Closes #314

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1230,10 +1230,12 @@ ex:hasLang
 					<p>
 						The <a>output nodes</a> of a <a>select expression</a> are the list <code>resultNodes</code> consisting of exactly the bindings of the (only)
 						variable that is projected from the SELECT clause.
-						If present in the <a>scope</a>, the value of the scope variable <code>focusNode</code> MUST be <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
+						The value of <code>focusNode</code> is <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
+						The values of each scope variable is <a>pre-bound</a> as a SPARQL variable with the same name and value.
+						A <a>failure</a> is produced when one of the scope variables is called <code>this</code>.
 						<br/>
 						<br/>
-						<code>evalExpr(expr, activeGraph, scope) -> resultNodes</code>
+						<code>evalExpr(expr, activeGraph, focusNode, scope) -> resultNodes</code>
 					</p>
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
@@ -1341,10 +1343,12 @@ ex:Bernd
 					<p>
 						The <a>output nodes</a> of an <a>SPARQL expr expression</a> are the list <code>resultNodes</code> consisting of exactly the bindings of the (only)
 						variable that is projected from the SELECT clause of the <code>select</code> query as defined <a href="#syntax-rule-SPARQLExprExpression-template">above</a>.
-						If present in the <a>scope</a>, the value of the scope variable <code>focusNode</code> MUST be <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
+						The value of <code>focusNode</code> is <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
+						The values of each scope variable is <a>pre-bound</a> as a SPARQL variable with the same name and value.
+						A <a>failure</a> is produced when one of the scope variables is called <code>this</code>.
 						<br/>
 						<br/>
-						<code>evalExpr(expr, activeGraph, scope) -> resultNodes</code>
+						<code>evalExpr(expr, activeGraph, focusNode, scope) -> resultNodes</code>
 					</p>
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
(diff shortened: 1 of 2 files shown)


afs:
Changes in 

[shacl12-core, section Node Expressions](https://raw.githack.com/w3c/data-shapes/issue-314/shacl12-core/index.html#node-expression-evaluation)

[shacl12-core, section sh:expessions](https://raw.githack.com/w3c/data-shapes/issue-314/shacl12-core/index.html##ExpressionConstraintComponent)

[shacl12-sparql, section SPARQL expressions](https://raw.githack.com/w3c/data-shapes/issue-314/shacl12-sparql/index.html##sparql-node-expressions)

HolgerKnublauch:
Interestingly I was also trying to add links to the githack renderings, but they were (for me) still caching an earlier version, and didn't show the second commit. So I elected not to include the link. But it renders fine now...

YoucTagh (approved):

shacl12-sparql/index.html
```diff
@@ -1230,13 +1230,13 @@ <h3>Select Expressions</h3>
 					<div class="def-header">EVALUATION OF SELECT EXPRESSIONS</div>
 					<p>
 						The <a>output nodes</a> of a <a>select expression</a> are the list <code>resultNodes</code> consisting of exactly the bindings of the (only)
-						variable that is projected from the SELECT clause when the query is evaluated against the <a>active graph</a>.
+						variable that is projected from the SELECT clause when the query is evaluated against the <a>focus graph</a>.
```
```suggestion
						variable that is projected from the <code>SELECT</code> clause when the query is evaluated against the <a>focus graph</a>.
```

shacl12-sparql/index.html
```diff
@@ -1344,13 +1344,13 @@ <h3>SPARQL Expr Expressions</h3>
 					<p>
 						The <a>output nodes</a> of an <a>SPARQL expr expression</a> are the list <code>resultNodes</code> consisting of exactly the bindings of the (only)
 						variable that is projected from the SELECT clause of the <code>select</code> query as defined <a href="#syntax-rule-SPARQLExprExpression-template">above</a>
```
```suggestion
						variable that is projected from the <code>SELECT</code> clause of the <code>select</code> query as defined <a href="#syntax-rule-SPARQLExprExpression-template">above</a>
```

TallTed (changes requested):
Tweaks for clarity. I feel strongly that SPARQL keywords (like `SELECT`) should be presented within `<code>` elements.

afs:
Not sure where to put comment but if this is @HolgerKnublauch's active PR ...

> sh:expression	The [node expression](https://raw.githack.com/w3c/data-shapes/issue-314/shacl12-core/index.html#dfn-node-expression) that must return true.

Node expressions return lists.

> does not return true as one of its [output nodes](https://raw.githack.com/w3c/data-shapes/issue-314/shacl12-core/index.html#dfn-output-nodes),

Is that the right behaviour? Wouldn't it be better to make it the conjunction of the list so any violation is a violation. That stops the case where an accidental mistake adding  value when only one is expected can make it pass.

`sh:expression` is already being called multiple times for each value node so better would be to require a single valued expression.

HolgerKnublauch:
I have no strong opinion on sh:expression. I believe in the original draft it had to return exactly [ true ]. However, I have the feeling that when a NE returns multiple values, the general interpretation should be "or". For example, sh:class ( ex:Class1 ex:Class2 ) should be satisfied if either class is present among the types.

One approach moving forward is to move this constraint component into the NE document and let it evolve there. So when people have more experience with NEs, we can adjust the semantics accordingly. It doesn't really need to be in Core.

afs:
> sh:class ( ex:Class1 ex:Class2 )

As in node expression `[ sh:class ( ex:Class1 ex:Class2 ) ]` or the property value of the result of a single node-expr returning a list for sh:class?

As a property shape, it's "and"-like `[ sh:path ... ; sh:class ex:Class1 ; sh:class ex:Class2 ]`

For `sh:expression` the safe choice is to define the requirement to be exactly `[ true ]`.

Others are not naturally an "or":

* `sh:targetNode` is a foreach.
* aggregates
* `sparql:lessThan` - "and" is more natural
* `sh:minInclusive`, `sh:minCount` would be strange if an "or" because then one value overrides another

What would 
```
[ sh:add ( 
    [ :function1 (...) ] 
    [ :function2 (...) ] 
    ) ]
```
be?  a list of N*M results?

`sh:add` and the rest would need to be implemented, with defined error handling, and not simply be the same as `sparql:add` which is 2 numbers only.

Maybe nodes expression evaluation should be defined to return "node or list of nodes" and node-expr functions declare which they are.

HolgerKnublauch:
Ok, this change to sh:expression is fine for me. Changed to [ true ] on the branch.

HolgerKnublauch:
@TallTed could you unblock the merge of this PR?

TallTed (approved):
Sorry, didn't realize application of my suggestions didn't auto-unblock.

PR #372, opened 2025-05-02
#338: Generalized sh:deactivated to node expressions

### LINKED ISSUE #338: Add support for Node Expressions as sh:deactivated parameter
It should be possible to dynamically deactivate shapes using Node Expressions as a parameter for the `sh:deactivated`. The allowed object values must be extended, and the expected return type must be defined.

HolgerKnublauch:
Closes #338

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-338/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1967,7 +1967,8 @@ ex:MyInstance
 					</p>
 					<p><em>The remainder of this section is informative.</em></p>
 					<p>
-						In SHACL Core, the only valid values for <code>sh:deactivated</code> are the (constant) node expressions
+						In SHACL Core, the only valid values for <code>sh:deactivated</code> are the 
+						<a href="#LiteralExpression">constant literal node expressions</a>
 						<code>true</code> and <code>false</code>.
 					</p>
 					<p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1956,13 +1956,20 @@ ex:MyInstance
 					<h4>Deactivating a Shape</h4>
 					<p class="syntax">
 						<span data-syntax-rule="deactivated-maxCount">Shapes can have at most one value for the property <code>sh:deactivated</code>.</span>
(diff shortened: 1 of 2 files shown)


shacl12-core/index.html
```diff
@@ -1956,13 +1956,20 @@ <h4>Declaring Messages for a Shape</h4>
...
 					<p>
-						A shape that has the <a>value</a> <code>true</code> for the property <code>sh:deactivated</code> is called <dfn data-lt="deactivate">deactivated</dfn>.
+						Let <code>expr</code> be the <a>value</a> of <code>sh:deactivated</code> in a <a>shape</a>.
+						If <code>evalExpr(expr, <a>data graph</a>, <a>focus node</a>, {})</code> produces <code>true</code> as its only
+						<a>output node</a>, the shape is called <dfn data-lt="deactivate">deactivated</dfn>.
 						All RDF terms <a>conform</a> to a deactivated shape.
 					</p>
 					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						In SHACL Core, the only valid values for <code>sh:deactivated</code> are the (constant) node expressions
```
bergos:
```suggestion
						In SHACL Core, the only valid values for <code>sh:deactivated</code> are constant literal node expressions
```

We don't have a definition or established wording yet, but that's what I would call it at the moment: constant literal/named node/term node expression.

bergos (approved):
See the one comment, otherwise LGTM

shacl12-core/index.html
HolgerKnublauch: Yes, good point (also similar to Andy's comment). I have added an explicit link to the definition of constant literals.

There is no consistent notion of constant node expressions yet, in part because we haven't decided on adding triple terms as a potential third kind of constants. I assume this should be added? If yes, is this already a ticket?

And yes, quite possibly the terminology for single output nodes can be improved. For the Core document it seems to be sufficient for now, but maybe this should go into the NE spec.

PR #392, opened 2025-05-26
#393 clean up shacl shacl shape

### LINKED ISSUE #393: SHACL-SHACL shape does not restrict severities to relevant classes
The SHACL-SHACL shape does not currently restrict the severities to `sh:Info`, `sh:Warning` and `sh:Violation`

jeswr:
~Closes #393 by `sh:in ( sh:Error sh:Warning sh:Info )` of the `sh:severity` property.~

Makes the following changes:
 - Applies `sh:minInclusive` constraint to appropriate properties such as `sh:minCount`
 - Makes the target class of `sh:qualifiedValueShape` a `sh:NodeShape` and sets `sh:node shsh:ShapeShape`

### DIFF
--- a/shacl/shacl-shacl.ttl
+++ b/shacl/shacl-shacl.ttl
@@ -260,7 +260,6 @@ shsh:ShapeShape
 		sh:path sh:qualifiedValueShape ;
 		sh:maxCount 1 ;                 # multiple-parameters
 		sh:node shsh:ShapeShape ;       # qualifiedValueShape-node
-		sh:class sh:NodeShape ;         # qualifiedValueShape-class
 	] ;
 	sh:property [
 		sh:path sh:qualifiedValueShapesDisjoint ;
--- a/shacl/shacl-shacl.ttl
+++ b/shacl/shacl-shacl.ttl
@@ -110,7 +110,6 @@ shsh:ShapeShape
 		sh:path sh:severity ;
 		sh:maxCount 1 ;                 # severity-maxCount
 		sh:nodeKind sh:IRI ;            # severity-nodeKind
-		sh:in ( sh:Error sh:Warning sh:Info ) ; # severity-in
 	] ;
 	sh:property [
 		sh:path sh:message ;
--- a/shacl/shacl-shacl.ttl
+++ b/shacl/shacl-shacl.ttl
@@ -132,7 +132,6 @@ shsh:ShapeShape
 	sh:property [
 		sh:path sh:closed ;
 		sh:datatype xsd:boolean ;       # closed-datatype
-		sh:in ( true false ) ;          # closed-in
 		sh:maxCount 1 ;                 # multiple-parameters
 	] ;
 	sh:property [
--- a/shacl/shacl-shacl.ttl
+++ b/shacl/shacl-shacl.ttl
@@ -119,7 +119,6 @@ shsh:ShapeShape
 		sh:path sh:deactivated ;
 		sh:maxCount 1 ;                 # deactivated-maxCount
 		sh:datatype xsd:boolean ;       # deactivated-datatype
-		sh:in ( true false ) ;          # deactivated-in
 	] ;
 
 	sh:property [
--- a/shacl/shacl-shacl.ttl
+++ b/shacl/shacl-shacl.ttl
@@ -110,6 +110,7 @@ shsh:ShapeShape
 		sh:path sh:severity ;
 		sh:maxCount 1 ;                 # severity-maxCount
 		sh:nodeKind sh:IRI ;            # severity-nodeKind
+		sh:in ( sh:Error sh:Warning sh:Info ) ; # severity-in
 	] ;
 	sh:property [
 		sh:path sh:message ;
@@ -118,7 +119,8 @@ shsh:ShapeShape
 	sh:property [
 		sh:path sh:deactivated ;
 		sh:maxCount 1 ;                 # deactivated-maxCount
-		sh:in ( true false ) ;          # deactivated-datatype
+		sh:datatype xsd:boolean ;       # deactivated-datatype
+		sh:in ( true false ) ;          # deactivated-in
 	] ;
 
 	sh:property [
@@ -132,6 +134,7 @@ shsh:ShapeShape
 	sh:property [
 		sh:path sh:closed ;
 		sh:datatype xsd:boolean ;       # closed-datatype
+		sh:in ( true false ) ;          # closed-in
 		sh:maxCount 1 ;                 # multiple-parameters
 	] ;
 	sh:property [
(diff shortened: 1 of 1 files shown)


shacl/shacl-shacl.ttl
```diff
@@ -110,6 +110,7 @@ shsh:ShapeShape
 		sh:path sh:severity ;
 		sh:maxCount 1 ;                 # severity-maxCount
 		sh:nodeKind sh:IRI ;            # severity-nodeKind
+		sh:in ( sh:Error sh:Warning sh:Info ) ; # severity-in
```
jeswr:
```suggestion
```

Noting resolution to #393  that "any IRI can be used" for `sh:severity`

shacl/shacl-shacl.ttl
```diff
@@ -242,15 +248,19 @@ shsh:ShapeShape
...
 		sh:path sh:qualifiedMinCount ;
 		sh:datatype xsd:integer ;       # qualifiedMinCount-datatype
 		sh:maxCount 1 ;                 # multiple-parameters
+		sh:minInclusive 0 ;             # qualifiedMinCount-minInclusive
 	] ;
 	sh:property [
 		sh:path sh:qualifiedValueShape ;
 		sh:maxCount 1 ;                 # multiple-parameters
+		sh:node shsh:ShapeShape ;       # qualifiedValueShape-node
+		sh:class sh:NodeShape ;         # qualifiedValueShape-class
```
ajnelson-nist: This `ClassConstraint` should not be here.  There are two reasons I can think of.

The first is support of a relaxed writing style: The object of `sh:qualifiedValueShape` is not required to have an `rdf:type`.  It should only be conformant with `shsh:ShapeShape` (which I'd assumed had been here all along, but thank you for adding it).  If `sh:class sh:NodeShape` is added, then `a sh:NodeShape` must have some `rdf:type`.

The second is that it is theoretically possible in SHACL 1.0 to have the object of `sh:qualifiedValueShape` be a `sh:PropertyShape`.  I'm a little challenged to think of an example at this hour of the morning.

Did you mean to constrain `sh:qualifiedValueShape`'s objects to being `shsh:NodeShapeShape` instead of `shsh:ShapeShape`?  If so, think that would be a scope expansion needing its own Issue.

shacl/shacl-shacl.ttl
```diff
@@ -118,7 +118,8 @@ shsh:ShapeShape
 	sh:property [
 		sh:path sh:deactivated ;
 		sh:maxCount 1 ;                 # deactivated-maxCount
-		sh:in ( true false ) ;          # deactivated-datatype
+		sh:datatype xsd:boolean ;       # deactivated-datatype
+		sh:in ( true false ) ;          # deactivated-in
```
ajnelson-nist: Every constraint added is a constraint calculated in review.  That is, there is a CPU cost to adding the extra constraint.  So, I think it would be better to have one of `sh:datatype xsd:boolean` OR `sh:in ( true false )`, but not both.

You'd caught another spot where the `-datatype` syntax point was already encoded instead of the `-in` syntax point.  Let's just pick one.

I lean towards `-datatype` because it obligates fewer "Set" or "List" object constructions for SHACL engine implementers.  Any other opinions?

ajnelson-nist: Edited prior comment's last sentence to disambiguate SHACL _engine_ implementers vs. shape authors.

shacl/shacl-shacl.ttl
```diff
@@ -182,6 +184,7 @@ shsh:ShapeShape
 		sh:path sh:maxCount ;
 		sh:datatype xsd:integer ;       # maxCount-datatype
 		sh:maxCount 1 ;                 # maxCount-maxCount
+		sh:minInclusive 0 ;              # maxCount-minInclusive
```
ajnelson-nist, line 185: I was about to note that this seemed redundant with some encoding, but, no, the datatype is `xsd:integer`, so we don't actually have a floor encoded.  I'm fine with this addition, especially since the alternative of using a `xsd:nonNegativeInteger` datatype would induce a lot of extra keyboard work for shape authors.

ajnelson-nist (commented):
One comment for the group (+line 122), one question for @jeswr (+line 263), before I approve.

shacl/shacl-shacl.ttl
jeswr: I also lean to `-datatype` for all instances. @HolgerKnublauch are you happy for me to remove all instances of `sh:in (true false)` as a commit in this PR.

shacl/shacl-shacl.ttl
jeswr:
```suggestion
```

Noting that "The object of sh:qualifiedValueShape is not required to have an rdf:type." (https://github.com/w3c/data-shapes/pull/392/files#r2111636649)

jeswr: > The first is support of a relaxed writing style: The object of sh:qualifiedValueShape is not required to have an rdf:type.

Good point - removed in https://github.com/w3c/data-shapes/pull/392/commits/6d0290f5c23fb6060add55f25c8bf42298e195f4

> Did you mean to constrain sh:qualifiedValueShape's objects to being shsh:NodeShapeShape instead of shsh:ShapeShape? 

The [example given in the 1.0 spec](https://www.w3.org/TR/shacl/#QualifiedValueShapeConstraintComponent) is indeed a property shape - the introduction of the class as a `sh:NodeShape` as opposed to a Node or Property Shape was just me being sloppy - I shouldn't do late night spec editing!

shacl/shacl-shacl.ttl
jeswr:
```suggestion
```

shacl/shacl-shacl.ttl
```diff
@@ -132,6 +133,7 @@ shsh:ShapeShape
 	sh:property [
 		sh:path sh:closed ;
 		sh:datatype xsd:boolean ;       # closed-datatype
+		sh:in ( true false ) ;          # closed-in
```
jeswr:
```suggestion
```

shacl/shacl-shacl.ttl
ajnelson-nist: Whoops, apologies, I read the page from the bottom-up.  Sorry, @HolgerKnublauch , I missed this line asking for you to weigh in & committed the changes just within the PR.

ajnelson-nist (approved):
I'm fine approving as-is.

Merging should wait for someone else's approval.

shacl/shacl-shacl.ttl
HolgerKnublauch: Right, the implicit design assumption in SHACL is that we only ever use xsd:integer and xsd:decimal for numbers. Anything else can be IMHO better expressed with sh:minInclusive etc.

shacl/shacl-shacl.ttl
HolgerKnublauch: The range of xsd:boolean (unfortunately) also includes "0"^^xsd:boolean and "1"^^xsd:boolean. So in cases where we really only want true and false it makes sense to combine sh:datatype with sh:in.

HolgerKnublauch:
I think it is fine to merge but I am not a reviewer here nor do I have time to fully look into the details. The SHACL-SHACL namespace could be finished and reviewed late in the release cycle. Anything else are intermediate informal versions.

PR #396, opened 2025-06-05
#173: Support `sh:message`, `sh:severity`, and `sh:deactivated` per constraint triple 

### LINKED ISSUE #173: Support `sh:message`, `sh:severity`, and `sh:deactivated` per constraint triple
RDF 1.2 will provide some syntax for attaching statements to other statements.

Currently, `sh:message` and `sh:severity` need to be placed in shapes. But shapes can contain multiple constraint triples, and not all of them should receive the same message and severity.

For example

    ex:MyClass-myProperty
        a sh:PropertyShape ;
        sh:path ex:myProperty ;
        sh:maxCount 1 .
    
    ex:MyClass-myProperty-minCount
        a sh:PropertyShape ;
        sh:path ex:myProperty ;
        sh:minCount 1 ;
        sh:severity sh:Warning ;
        sh:message "There SHOULD be at least one value here" .

It should become possible to use RDF 1.2 reification or quoted triples to do something like this

    ex:MyClass-myProperty
        a sh:PropertyShape ;
        sh:path ex:myProperty ;
        sh:maxCount 1 ;
        sh:minCount 1 {| 
            sh:severity sh:Warning ;
            sh:message "There SHOULD be at least one value here" ;
        |} .

This would avoid having to introduce artificial shapes just to encapsulate warning or severity.

We have implemented this in TopBraid for several years now and didn't run into complaints from users.

The same mechanism would also be great for `sh:deactivated` (we didn't implement that yet).

HolgerKnublauch:
Closes #173

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-173/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2020,7 +2020,6 @@ ex:MyShape
 	a sh:NodeShape ;
 	<span class="target-can-be-skipped">sh:targetNode ex:MyInstance ;</span>
 	sh:property [
-		# The default severity here is sh:Violation
 		sh:path ex:myProperty ;
 		sh:maxLength 10 {|
 			sh:message "Too many characters"@en ;
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1980,11 +1980,11 @@ ex:MyShape
 				</section>
 				
 				<section id="message">
-					<h4>Declaring Messages for a Shape</h4>
+					<h4>Declaring Messages for a Shape or Constraint</h4>
 					<p class="syntax">
 						Shapes can have values for the property <code>sh:message</code>.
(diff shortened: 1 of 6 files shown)


shacl12-core/index.html
```diff
@@ -1936,35 +1949,97 @@ <h4>Declaring the Severity of a Shape</h4>
...
+						The following example is a variation where the message is declared using reification.
+					</p>
+					<aside class="example">
+						<div class="shapes-graph">
+							<div class="turtle">
+ex:MyShape
+	a sh:NodeShape ;
+	<span class="target-can-be-skipped">sh:targetNode ex:MyInstance ;</span>
+	sh:property [
+		# The default severity here is sh:Violation
```
bergos: I guess it was copied by mistake:

```suggestion
```

bergos (approved):
I added one suggestions. Everything else looks good.

YoucTagh (approved):

PR #395, opened 2025-06-03
#160: Allow lists at sh:class and sh:datatype

HolgerKnublauch:
I have not yet added the JSON-LD to the samples - is my understanding correct that the generation of that will be automated?

Closes #160

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-160/shacl12-core/index.html)

Issue #160, opened 2024-12-04
Multivalued or union sh:dataType


VladimirAlexiev:
Some literal (attribute) props naturally admit a variety of datatypes,  eg
- dc:date, dct:created etc can be xsd:date, xsd:dateTime, xsd:dateTimeStamp
- s:name can be xsd:string, rdf:langString 
- ex:age can be byte, int, integer, and their nonNegative variants (Inspired by #142)
- sh:order can be integer or decimal (#125)

SHACL can express this with sh:or property shapes, but it gets tedious. 
Can we allow multiple values of sh:dataType?

Alternatively, we could think of some synthetic  "union datatypes", eg 
- sh:dateOrDateTimeStamp
- sh:stringOrLangString (rdf:PlainLiteral means the same)
- sh:natural, sh:nonNegativeNatural, etc

We actually have the first two in Ontotext SOML (GraphQL), which in SHACL is expanded to sh:or

HolgerKnublauch:
The union datatype approach works quite well and comes out of the box in the dash namespace

https://datashapes.org/dash.html#union-datatypes

Adding multiple sh:datatypes would be inconsistent with how the rest of SHACL is working - no other constraint component takes multiple values of the same property.

I agree however that processing this through sh:or is not elegant. The old WG had considered something like sh:datatypeIn which would take an enumeration but rejected this in order to keep the language smaller.

afs:
A union datatype has consequences for data exchange though.  The import and use of such data must now consider all possibilities or be "by value".

Adding as a library, rather than changing SHACL-core is better. Then a system can publish which libraries it is using, which is a contract it can adhere to even if it is not using all features at one point in time.

In other words - shapes files declare feature sets, whether by imports or a new, specific property.

ajnelson-nist:
Is it possible to consider datatype _hierarchy_ to address some of the concerns in this thread and related issues?

For instance, quoting XSD 1.1:

> The dateTimeStamp datatype is [·derived·](https://www.w3.org/TR/xmlschema11-2/#dt-derived) from [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime) by giving the value required to its [explicitTimezone](https://www.w3.org/TR/xmlschema11-2/#f-tz) facet.

Specifically, because of the term "derived," should a `DatatypeConstraint` requiring `xsd:dateTime` accept `xsd:dateTimeStamp`, without the need to resort to `sh:or`?

[...]

PR #403, opened 2025-06-11
Issue 391: support for collections

### LINKED ISSUE #391: feat: better support for collections
It would be nice to have the constraint `sh:listOf` that could be used as follows (i.e. used to state that the object should be a collection with contents satisfying given constraints):

```ttl
	sh:property [
		sh:path ex:speakerOrder ;
		sh:listOf [
                  sh:nodeKind IRI ;
                ] ;
               sh:maxCount 1 
	] .
```

If not feasible to add to the core spec in 1.2 - then we may wish to still have shorthand in SHACL-C which gets expanded to something that looks like http://www.w3.org/ns/shacl-shacl#ListShape .

jeswr:
Closes #391

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html)


jeswr:
Marking as draft whilst cleaning up the HTML to ask the following questions:
 - Should we allow `sh:NodeShapes` to be the *subject* of `sh:memberShape`? I've currently disallowed this.
 - Is there a normative reference for well-formed lists that I should point towards

HolgerKnublauch:
I would allow it for NodeShapes too. Your definition already uses the term value node, and for node shapes that just means the focus node. Are you hesitating because it may require additional test cases?

And on RDF lists, why not just reference <a>SHACL Lists</a> within the same document?

shacl12-core/index.html
```diff
@@ -2700,7 +2701,7 @@ <h3>Conformance Checking</h3>
 					<a>conforms</a> to a given <a>shape</a>, and <code>false</code> otherwise.
 				</p>
 				<p id="conformance-nested">
-					Note that some <a>constraint components</a> of SHACL Core (e.g., those of <code>sh:not</code>, <code>sh:or</code> and <code>sh:node</code>)
+					Note that some <a>constraint components</a> of SHACL Core (e.g., those of <code>sh:not</code>, <code>sh:or</code>, <code>sh:node</code> and <code>sh:memberShape</code>)
```
```suggestion
					Note that some <a>constraint components</a> of SHACL Core (e.g., those of <code>sh:not</code>, <code>sh:or</code>, <code>sh:node</code>, and <code>sh:memberShape</code>)
```

shacl12-core/index.html
```diff
@@ -5507,6 +5508,185 @@ <h4>sh:property</h4>
...
+					<table class="term-table">
+						<tr>
+							<th>Property</th>
+							<th>Summary and Syntax Rules</th>
+						</tr>
+						<tr>
+							<td><code>sh:memberShape</code></td>
+							<td>
+								The shape that all members of the <a href="#syntax-rule-SHACL-list">SHACL list</a> must conform to.
+								<span data-syntax-rule="memberShape-node">The value of <code>sh:memberShape</code> must be a <a>well-formed</a> <a>node shapes</a>.</span>
```
```suggestion
								<span data-syntax-rule="memberShape-node">The value of <code>sh:memberShape</code> must be a <a>well-formed</a> <a>node shape</a>.</span>
```

TallTed (changes requested):
tiny

shacl12-core/index.html
```diff
@@ -2700,7 +2701,7 @@ <h3>Conformance Checking</h3>
 					<a>conforms</a> to a given <a>shape</a>, and <code>false</code> otherwise.
 				</p>
 				<p id="conformance-nested">
-					Note that some <a>constraint components</a> of SHACL Core (e.g., those of <code>sh:not</code>, <code>sh:or</code> and <code>sh:node</code>)
+					Note that some <a>constraint components</a> of SHACL Core (e.g., those of <code>sh:not</code>, <code>sh:or</code>, <code>sh:node</code>, and <code>sh:memberShape</code>)
```
ajnelson-nist: _(Non-blocking suggestion improvement, with follow-on message I suspect will migrate to an Issue.)_

Now that this line's highlighted, I just noticed a missing symmetry.  If this parenthetical is meant to be exhaustive, shouldn't `sh:and` and `sh:xone` be here too?

Going a bit further, so we can pare back:  The list of "shape-expecting constraint parameters" is on line 2645, and 8 items long.

I think we can shorten some English here and remove the parenthetical.  Couldn't we just say:

> Note that all shape-expecting constraint parameters [link to L2645] of SHACL Core rely on conformance checking.

I suggest that because, unless I've forgotten something, all 8 of those constraint components rely on conformance checking.

ajnelson-nist: Also, having written the above, I got distracted by "constraint parameter" and "constraint component" being used so close to one another with word-positioning that hints that they're synonymous, but IIRC that's not the case.  Line 2645 uses the former, the list links use the latter.

Glancing at the SVG in #380 , there's a class for both.  `sh:Parameter` specializes `sh:PropertyShape`.  `sh:ConstraintComponent` specializes `sh:Parameterizable`.  Reviewing the Turtle reminds me that they appear in pretty different places, but it's not quick to review & remind myself.

shacl12-core/index.html
```diff
@@ -5507,6 +5508,185 @@ <h4>sh:property</h4>
...
+						<tr>
+							<th>Property</th>
+							<th>Summary and Syntax Rules</th>
+						</tr>
+						<tr>
+							<td><code>sh:memberShape</code></td>
+							<td>
+								The shape that all members of the <a href="#syntax-rule-SHACL-list">SHACL list</a> must conform to.
+								<span data-syntax-rule="memberShape-node">The value of <code>sh:memberShape</code> must be a <a>well-formed</a> <a>node shape</a>.</span>
+								<span data-syntax-rule="memberShape-maxCount">A shape has at most one value for <code>sh:memberShape</code>.</span>
```
ajnelson-nist: Is there a firm design reason to cap `sh:memberShape` at 1?  I could imagine a use case wanting to reuse multiple `sh:NodeShape`s that are also used outside the current `sh:PropertyShape`.

ajnelson-nist: Ah, please excuse the `sh:PropertyShape` remark at the end, I forgot there was a design permitance on `sh:NodeShape`s too.

shacl12-core/index.html
```diff
@@ -5507,6 +5508,185 @@ <h4>sh:property</h4>
...
+					}
+				]
+			}
+		}
+	]
+}</pre>
+							</div>
+						</div>
+						<div class="results-graph">
+							<div class="turtle">
```
ajnelson-nist, line 5828: ReSpec/HTML double-checking question - I see the above block needed `<div class="json"><pre class="jsonld">...`.  Turtle blocks don't need the same?  I think I see a subtle font difference in the rendered document.

shacl12-test-suite/tests/core/node/manifest.ttl
```diff
@@ -23,6 +23,7 @@
 	mf:include <maxExclusive-001.ttl> ;
 	mf:include <maxInclusive-001.ttl> ;
 	mf:include <maxLength-001.ttl> ;
+	mf:include <memberShape-001.ttl> ;
```
ajnelson-nist, line 29: I see the new files `memberShape-00{1,2,3,4}.ttl`; but, only `...001.ttl` is listed.  Missed?

shacl12-test-suite/tests/core/node/memberShape-001.ttl
```diff
@@ -0,0 +1,94 @@
...
+  rdf:type sht:Validate ;
+  rdfs:label "Test of sh:memberShape on node shape 001" ;
+  mf:action [
+      sht:dataGraph <> ;
+      sht:shapesGraph <> ;
+    ] ;
+  mf:result [
+      rdf:type sh:ValidationReport ;
+      sh:conforms "false"^^xsd:boolean ;
+      sh:result [
```
ajnelson-nist, line 86: Isn't `ex:Rex` supposed to be in the result?

shacl12-test-suite/tests/core/node/manifest.ttl
ajnelson-nist: Whoops, please disregard, I mixed up the folder.

shacl12-test-suite/tests/core/node/memberShape-001.ttl
```diff
@@ -0,0 +1,94 @@
...
+
+# Valid person list with all Person instances
+ex:validList
+  rdf:type ex:PersonList ;
+  rdf:value (
+      ex:Alice
+      ex:Bob
+    ) ;
+  rdfs:label "Valid Person List" ;
+.
```
ajnelson-nist: Shouldn't this be spelled:

```turtle
ex:validList
  rdf:type ex:PersonList ;
  rdf:first ex:Alice ;
  rdf:rest (
    ex:Bob
  ) ;
  rdfs:label "Valid Person List" ;
.
```

`rdf:value` is a property, but has no meaning, per [5.4.3 or RDF 1.2 today](https://www.w3.org/TR/2025/WD-rdf12-schema-20250224/#ch_value).

I forget if you would also need to assert `ex:PersonList rdfs:subClassOf rdf:List`.

shacl12-test-suite/tests/core/property/memberShape-001.ttl
```diff
@@ -0,0 +1,97 @@
...
+  rdf:type sht:Validate ;
+  rdfs:label "Test of sh:memberShape at property shape 001" ;
+  mf:action [
+      sht:dataGraph <> ;
+      sht:shapesGraph <> ;
+    ] ;
+  mf:result [
+      rdf:type sh:ValidationReport ;
+      sh:conforms "false"^^xsd:boolean ;
+      sh:result [
```
ajnelson-nist, line 68: The blank Charlie node might need to be given a `_:b0` style name if it's supposed to be linked as `sh:value`.

shacl12-test-suite/tests/core/property/memberShape-002.ttl
```diff
@@ -0,0 +1,101 @@
...
+  rdf:type ex:Person ;
+  rdfs:label "Alice" ;
+.
+
+ex:Bob
+  rdf:type ex:Person ;
+  rdfs:label "Bob" ;
+.
+
+ex:Charlie
```
ajnelson-nist: Better to sub `ex:Rex` here for consistency with the new `node/` test.

shacl12-test-suite/tests/core/property/memberShape-003.ttl
```diff
@@ -0,0 +1,86 @@
...
+  rdfs:label "Bob" ;
+.
+
+ex:Person
+  rdf:type rdfs:Class ;
+  rdf:type sh:NodeShape ;
+  rdfs:label "Person" ;
+.
+
+# Agenda with invalid list structure (doesn't terminate with rdf:nil)
```
ajnelson-nist: Is the SHACL test suite meant to exercise base matters of RDF syntax?  I would expect this test to fail for non-SHACL reasons related to parsing the graph.

ajnelson-nist:
I won't log a "Request Changes" review in case that mucks up process (especially as a non-Editor on the doc), but I do urge working through Resolving all of my comment threads except maybe the first one that I said was non-blocking.

Also, thanks for updating the SHACL-SHACL shapes.

shacl12-core/index.html
jeswr: > Is there a firm design reason to cap `sh:memberShape` at 1?

If we enable multiple then we need to define whether we want all (i.e. `sh:and`) or some (i.e. `sh:or`) `?memberShape`s to be satisfied for the validation to pass; and implementations need to support this.

Equivalent behaviour can be achieved by having those writing shapes define a `?memberShape` that uses `sh:or` or `sh:and` within the shape to achieve the same behaviour. Granted, this is slightly more verbose for those writing shapes.

HolgerKnublauch: All constraint components only allow a single value per constraint parameter. So having two sh:memberShapes would just execute them individually, resulting in AND semantics. For OR semantics, wrap them into sh:or.

shacl12-core/index.html
jeswr: Change made in https://github.com/w3c/data-shapes/pull/403/commits/180b418d55c3995347a4625cd4a7a588b8b85c78 - I think the wording "shape-expecting constraint parameters" is correct noting the [following from the spec](https://www.w3.org/TR/shacl/#constraints)

> A constraint component is an [IRI](https://www.w3.org/TR/shacl/#dfn-iri). Each constraint component has one or more mandatory parameters, each of which is a property. Each constraint component has zero or more optional parameters, each of which is a property. The parameters of a constraint component are its mandatory parameters plus its optional parameters.

shacl12-core/index.html
jeswr: I'm following what other examples in the document are doing.

If this is problematic from a ReSpec/HTML perspective then I'd suggest opening and issue as this would need to be a change applied document wide.

jeswr:
@ajnelson-nist Thanks for reviewing the tests - I've gone back and put together simpler test cases to improve readability. I had used copilot to quickly put together the tests whilst the PR was in draft. They are now human crafted and checked.

> I would allow it for NodeShapes too. Your definition already uses the term value node, and for node shapes that just means the focus node. Are you hesitating because it may require additional test cases?

This is now allowed; I was struggling to think of a case where it'd be used in practise - but I can think of a few including `sh:targetObjectsOf ex:speakerOrder`.

> And on RDF lists, why not just reference SHACL Lists within the same document?

Good point - now using SHACL lists; I'd forgotten they were defined in the doc.

shacl12-vocabularies/shacl.ttl
```diff
@@ -899,7 +899,7 @@ sh:minLength
 sh:MemberShapeConstraintComponent
 	a sh:ConstraintComponent ;
 	rdfs:label "Member shape constraint component"@en ;
-	rdfs:comment "Can be used to specify constraints on the members of a given list, assuming that the given sh:property has rdf:Lists as values. A violation is reported for each member of the list that does not comply with the constraints specified by the given shape."@en ;
+	rdfs:comment "Can be used to specify constraints on the members of a given SHACL list. A violation is reported for each member of the list that does not comply with the constraints specified by the given shape. A violation is reported if the value is an invalid SHACL list."@en ;
```
```suggestion
	rdfs:comment "Can be used to specify constraints on the members of a given SHACL list. A violation is reported for each member of the list that does not comply with the constraints specified by the given shape. A violation is reported if the value is not a valid SHACL list."@en ;
```

jeswr:
@TallTed I think I've addressed all of your requested changes - could you please update your review so this isn't blocked from merging.

@ajnelson-nist Could I ask you to confirm your approval with the current state of this PR.

shacl12-core/index.html
```diff
@@ -5507,6 +5508,180 @@ <h4>sh:property</h4>
...
+[	a sh:ValidationReport ;
+	sh:conforms false ;
+	sh:result [
+		a sh:ValidationResult ;
+		sh:resultSeverity sh:Violation ;
+		sh:focusNode ex:agenda2 ;
+		sh:resultPath ex:speakerOrder ;
+		sh:value "Charlie" ;
+		sh:resultMessage "List member does not have the required node kind sh:IRI." ;
+		sh:sourceConstraintComponent sh:MemberShapeConstraintComponent ;
```
ajnelson-nist: I'm mulling over these lines.

The documentation prior to here says there should be information provided about the non-conformant list member.

`sh:sourceConstraintComponent` points at `sh:MemberShapeConstraintComponent`, but misses a degree of recursion that the component violated is `sh:NodeKindComponent` (apologies if I typo'd that IRI just now).  `sh:resultMessage` is bearing the weight of describing what component ultimately caught an issue.

I think the practice used with `sh:node` should be replicated here.  IIRC, there would be a contained `sh:ValidationResult` describing how the shape linked by `sh:node` triggered - is it by `sh:description`?  I have a meeting in a few minutes, apologies for sparsity of details.

Also, I realized since your update that is a necessary exercise to put two non-conformant list members (say, `"Charlie"` and `"Deborah"`).  I have a hunch this will reveal an important logistics issue to assist with data-curing - what happens if a list has, say, a hundred members and twenty are non-conformant?

jeswr: > I think the practice used with sh:node should be replicated here.

Check - will have a go replicating.

> I have a hunch this will reveal an important logistics issue to assist with data-curing - what happens if a list has, say, a hundred members and twenty are non-conformant?

The choice was made to be consistent with the rest of the specification. e.g. with `sh:property` the requirement is that "For each [value node](https://www.w3.org/TR/shacl/#dfn-value-nodes) v: A [failure](https://www.w3.org/TR/shacl/#dfn-failure) must be produced" and the same logistics concern applies if there are 100 objects of a property (e.g. `ex:contact`) of which 20 are non-conformant.

We could provide the additional information of the index at which the violation occurred - I am not sure how useful this is.

ajnelson-nist: > > I think the practice used with sh:node should be replicated here.
> 
> Check - will have a go replicating.

I suspect what will end up happening is the visual display of a violation via `sh:node` will look like a special case of `sh:memberShape`.

> 
> > I have a hunch this will reveal an important logistics issue to assist with data-curing - what happens if a list has, say, a hundred members and twenty are non-conformant?
> 
> The choice was made to be consistent with the rest of the specification. e.g. with `sh:property` the requirement is that "For each [value node](https://www.w3.org/TR/shacl/#dfn-value-nodes) v: A [failure](https://www.w3.org/TR/shacl/#dfn-failure) must be produced" and the same logistics concern applies if there are 100 objects of a property (e.g. `ex:contact`) of which 20 are non-conformant.
> 
> We could provide the additional information of the index at which the violation occurred - I am not sure how useful this is.

I suspect the index would not be helpful.

I've sketched an example showing current behavior of `sh:node`, going in another direction too - what if there are two constraints in the value of `sh:memberShape`?

Example graph, SHACL and Data together:

```turtle
@prefix ex: <http://example.org/kb/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .

ex:IsIRIAndPersonShape
	a sh:NodeShape ;
	sh:class ex:Person ;
	sh:nodeKind sh:IRI ;
	.

ex:AgendaShape
	a sh:NodeShape ;
	sh:targetClass ex:Agenda ;
	sh:property [
		sh:path ex:speaker ;
		sh:node ex:IsIRIAndPersonShape ;
	] .


ex:Alice
	a ex:Person ;
	.

ex:agenda2 a ex:Agenda ;
	ex:speaker ex:Alice ;
	ex:speaker ex:Bob ;
	ex:speaker "Charlie" .
```

Example output, according to a SHACL engine:

```
@prefix ex: <http://example.org/kb/> .
@prefix sh: <http://www.w3.org/ns/shacl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

[] a sh:ValidationReport ;
    sh:conforms false ;
    sh:result [ a sh:ValidationResult ;
            sh:detail [ a sh:ValidationResult ;
                    sh:focusNode ex:Bob ;
                    sh:resultMessage "Value does not have class ex:Person" ;
                    sh:sourceConstraintComponent sh:ClassConstraintComponent ;
                    sh:sourceShape ex:IsIRIAndPersonShape ;
                    sh:value ex:Bob ] ;
            sh:focusNode ex:agenda2 ;
            sh:resultMessage "Value does not conform to Shape ex:IsIRIAndPersonShape. See details for more information." ;
            sh:resultPath ex:speaker ;
            sh:sourceConstraintComponent sh:NodeConstraintComponent ;
            sh:sourceShape _:nc462657021324c38a78ea78943b3216cb1 ;
            sh:value ex:Bob ],
        [ a sh:ValidationResult ;
            sh:detail [ a sh:ValidationResult ;
                    sh:focusNode "Charlie" ;
                    sh:resultMessage "Value does not have class ex:Person" ;
                    sh:sourceConstraintComponent sh:ClassConstraintComponent ;
                    sh:sourceShape ex:IsIRIAndPersonShape ;
                    sh:value "Charlie" ],
                [ a sh:ValidationResult ;
                    sh:focusNode "Charlie" ;
                    sh:resultMessage "Value is not of Node Kind sh:IRI" ;
                    sh:sourceConstraintComponent sh:NodeKindConstraintComponent ;
                    sh:sourceShape ex:IsIRIAndPersonShape ;
                    sh:value "Charlie" ] ;
            sh:focusNode ex:agenda2 ;
            sh:resultMessage "Value does not conform to Shape ex:IsIRIAndPersonShape. See details for more information." ;
            sh:resultPath ex:speaker ;
            sh:sourceConstraintComponent sh:NodeConstraintComponent ;
            sh:sourceShape _:nc462657021324c38a78ea78943b3216cb1 ;
            sh:value "Charlie" ] .

_:nc462657021324c38a78ea78943b3216cb1 sh:node ex:IsIRIAndPersonShape ;
    sh:path ex:speaker .
```

`sh:detail` is the property I should've mentioned, not `sh:description`.

The reason I think the list-index would not be helpful is the node triggering the validation result is already directly mentioned in the result's `sh:focusNode`.

I think it's also helpful for your reference to see what the values are for `sh:sourceConstraintComponent`, esp. where it's `sh:NodeConstraintComponent`.  That looks like a clean swap-spot for `sh: MemberShapeConstraintComponent`.

I leave it up to you whether this goes in the spec doc. or in a test.  The detail level feels more right for clarification via the test suite, but I'm not sure if we need more English to prescribe for the sake of the test suite.

shacl12-test-suite/tests/core/node/memberShape-001.ttl
```diff
@@ -0,0 +1,102 @@
...
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix sht: <http://www.w3.org/ns/shacl-test#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+ex:IRIListShape
+  rdf:type sh:NodeShape ;
+  sh:memberShape [ sh:nodeKind sh:IRI ] ;
+  sh:targetNode ( ex:Alice ex:Bob ) ;
+  sh:targetNode () ;
```
ajnelson-nist, line 19: This target isn't among the expected test results, but looks like it should trigger a violation.

shacl12-test-suite/tests/core/node/memberShape-001.ttl
```diff
@@ -0,0 +1,102 @@
...
+  sh:targetNode _:b1, _:b2, _:b3, _:b4, _:b5 ;
```
ajnelson-nist: `_:b3` isn't among the data.

shacl12-test-suite/tests/core/property/memberShape-001.ttl
```diff
@@ -0,0 +1,69 @@
...
+      ex:Charlie
+    ) ;
+.
+
+# Invalid agenda with a blank node in the list
+ex:list2
+  rdf:type ex:ListSubject ;
+  ex:testProperty (
+      ex:Bob
+      _:bc
```
ajnelson-nist: From having started this conversation, I know why this `_:bc` is here; but the actual (blank) node definition, including `rdfs:label`, isn't currently in the test data.
The test works as-is, but could just use the clarification of the `rdfs:label`.

ajnelson-nist:
> @ajnelson-nist Could I ask you to confirm your approval with the current state of this PR.

Overall, the progress is good.  My (soft-)blockers are:

* There's a validation illustration that I think needs an update.
* `shacl12-test-suite/tests/core/node/memberShape-001.ttl` needs a data fix.
* Another test technically works, but for could use a nit-level data fix.

Again, I won't push the "Request Changes" button in case that mucks up process.

shacl12-test-suite/tests/core/node/memberShape-001.ttl
jeswr: This is a valid SHACL list (it is just `rdf:nil`) so I expect this to pass.

I can see cases where you'd want to make sure the list does have have some elements - I've raised https://github.com/w3c/data-shapes/issues/414 to propose `sh:minCollectionLength` and `sh:maxCollectionLength` as constraints

shacl12-test-suite/tests/core/node/memberShape-001.ttl
jeswr: That is intended. It should raise a violation because a blank node with no properties is an invalid list.

ajnelson-nist:
While looking over SHACL-SHACL updates today, I noticed there's a live use case for this PR within SHACL-SHACL.

#160 / #395 expanded `sh:datatype` and `sh:class` to take lists that otherwise need to behave like a singleton value.

It's a bit debatably scope-creep, but I think `sh:memberShape` could be used on the (blank-node) property shapes on `sh:ShapeShape` for `sh:class` and `sh:property`.  E.g.,

```patch
        ] ;
        sh:property [
                sh:path sh:class ;
-                sh:nodeKind sh:IRI ;            # class-nodeKind
+                sh:or (
+                        [ sh:nodeKind sh:IRI ]
+                        [ sh:memberShape [ sh:nodeKind sh:IRI ] ]
+                ) ;            # class-nodeKind
         ] ;
        sh:property [
                sh:path sh:closed ;
```

Though, thinking about the case (possibly pathologically?) of an IRI-identified `rdf:List`, I'm seeing some vision of a list that could evade SHACL-SHACL's syntax review.

```turtle
ex:Foo
        a rdfs:Class ;
        .

ex:ExampleShape
        sh:class ex:ExampleList ;
        .

ex:ExampleList
        rdf:first "foo" ;
        rdf:rest (
                ex:Foo
        )
        .
```

All the above would also apply for `datatype-nodeKind`.

I'm fine with this being broken out into a separate issue, not blocking this PR, examining 391 and 160 thrown against one another.

shacl12-core/index.html
```diff
@@ -5691,6 +5692,194 @@ <h4>sh:property</h4>
...
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="MemberShape">
+							Let <code>$memberShape</code> be a <a>value</a> of <code>sh:memberShape</code>.
+							Each <a>value node</a> <code>v</code> must be a <a href="#syntax-rule-SHACL-list">SHACL list</a> - a <a>failure</a> MUST be produced if <code>v</code> is not a SHACL list.
+							For each member <code>m</code> of the list <code>v</code>
+							a <a>failure</a> MUST be produced if the <a>conformance checking</a> of <code>m</code> against <code>$memberShape</code> produces a <a>failure</a>.
+							There must be at most one <a>validation result</a> for each list <code>v</code>.
+
+							The `sh:detail` property MUST be used to specify the validation results for each member <code>m</code> of the list that fails conformance checking.
+							The `sh:detail` property MUST also be used to specify when <code>v</code> is not a <a>SHACL list</a>.
```
ajnelson-nist: Nit: Shouldn't sh:detail be in an HTML code element, not a Markdown backtick?

Not-nit: Why the "MUST" for sh:detail "being used to specify the validation results..."?  I think I see that `sh:node` does not have guidance to adapt.  I agree this is at least SHOULD-strength, but what do others think about MUST-strength?

Also, the second line about not-a-SHACL-list feels like it needs illustration.  Can you in-page link to a demonstration?

shacl12-core/index.html
```diff
@@ -5691,6 +5692,194 @@ <h4>sh:property</h4>
...
+						</tr>
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="MemberShape">
+							Let <code>$memberShape</code> be a <a>value</a> of <code>sh:memberShape</code>.
+							Each <a>value node</a> <code>v</code> must be a <a href="#syntax-rule-SHACL-list">SHACL list</a> - a <a>failure</a> MUST be produced if <code>v</code> is not a SHACL list.
+							For each member <code>m</code> of the list <code>v</code>
+							a <a>failure</a> MUST be produced if the <a>conformance checking</a> of <code>m</code> against <code>$memberShape</code> produces a <a>failure</a>.
+							There must be at most one <a>validation result</a> for each list <code>v</code>.
```
ajnelson-nist: Should this end something like "...for each list v and triggering constraint," (indicating cross-product) since `memberShape` could be used to link multiple shapes?

jeswr: Comment addressed in https://github.com/w3c/data-shapes/pull/403/commits/5db4a2729f3036f31b2464a5950abeb67dee1b32

shacl12-core/index.html
jeswr: Addressed in https://github.com/w3c/data-shapes/pull/403/commits/5db4a2729f3036f31b2464a5950abeb67dee1b32

> Also, the second line about not-a-SHACL-list feels like it needs illustration. Can you in-page link to a demonstration?

I've now linked to the [definition of a SHACL List](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#dfn-shacl-lists). Given the that definition is quite well-defined I think it should be clear that the failure is anything that does not satisfy that definition.

TallTed (approved):
lgtm

shacl12-core/index.html
jeswr: > Also, the second line about not-a-SHACL-list feels like it needs illustration. Can you in-page link to a demonstration?

Clarified offline that this was requesting a pointer to example of `sh:detail`. Added in https://github.com/w3c/data-shapes/pull/403/commits/7ff2a9a8c13811779f955788edcbe6811ac4a4a3

jeswr:
Closing as this is now deprecated by #416

Issue #401, opened 2025-06-11
Deactivated constraint in SHACL-SHACL shape should use `datatype`


jeswr:
Currently it is constrained using `sh:in ( true false )` it should be constrained using `sh:datatype boolean`

ajnelson-nist:
Reopening because #402 accidentally modified the wrong SHACL-SHACL file.  The file that should have been modified is `/shacl12-vocabularies/shacl-shacl.ttl`; the file that was modified was `/shacl/shacl-shacl.ttl`, which is SHACL 1.0's `shacl-shacl.ttl`, and my understanding was we weren't modifying that.  I have a meeting soon, so might not file a PR for an hour or two.

_(Edit: Fixed path to file affected in #402 )._

ajnelson-nist:
...PR coming once I figure out what actually happened in commit `3a1573a`.  Something about the squash-merge seems to have removed all of the effects of PR 402?

Meanwhile, I would appreciate a double-check that dates back to #392 - that PR touched SHACL 1.0's `shacl-shacl.ttl` on accident, and none of us caught that it didn't look at SHACL 1.2's.  Should that be fixed?

jeswr:
I think the best option might be to revert by merging https://github.com/w3c/data-shapes/pull/404 (created using the revert option in GH as shown below)

<img width="1041" alt="Image" src="https://github.com/user-attachments/assets/fcd517a0-d851-4e7d-ae74-8949ff570339" />

The issue was that my local version of the `gh-pages` branch wasn't up to date (sorry!) - hence why I was making a change that had already been made in #392. This is the result of running `git log` for the `fix/deactivated-datatype` branch locally.

```bash
commit 2736ec8d901bc7bbb6338322d9f98d6603455a6e (HEAD -> fix/deactivated-datatype, origin/fix/deactivated-datatype)
Author: Jesse Wright <63333554+jeswr@users.noreply.github.com>
Date:   Wed Jun 11 10:26:21 2025 +0100

    fix: constrain deactivated using datatype rather than in

commit 2e8388ea1900899df0d0a9f2fc51ba8170e1a60f
Merge: a0a79ea c6047d6
Author: Yousouf Taghzouti <48333491+YoucTagh@users.noreply.github.com>
Date:   Mon May 26 09:59:24 2025 +0200

    Merge pull request #387 from w3c/issue-191-multiple-shacl-lists
    
    #191: Clarify Logical Constraint Components with Multiple SHACL Lists

[...]

PR #404, opened 2025-06-11
Revert "Issue 401 fix: constrain deactivated using `datatype` rather than `in`

jeswr:
… (#402)"

This reverts commit 3a1573a1bdd664580c9cf0257346cf3ec45cba3a - identified as problematic by @ajnelson-nist in https://github.com/w3c/data-shapes/issues/401#issuecomment-2962499728


HolgerKnublauch (approved):
I have lost track and probably shouldn't have been involved in the shacl-shacl review process in the first place. But if this keeps the ball rolling, here is my approval.

afs:
There don't seem to be any changes  (because 3a1573a1b has none?).
All this PR seems to do is to add an empty commit.

(It does have an appropriate commit number!)

jeswr:
> There don't seem to be any changes (because https://github.com/w3c/data-shapes/commit/3a1573a1bdd664580c9cf0257346cf3ec45cba3a has none?).

In #402 I had branched off an old commit (https://github.com/w3c/data-shapes/commit/785b0f0909c20dc471dd3c9ba69c440da213c001). When squashing this to main it appears to have effectively reverted the changes of subsequent commits; even though those changes did not appear in the file change view of #402.

My expectation is that this PR will restore those changes; even though the file change view of #404 is empty. I will proceed with merging so as to avoid blocking further work.

@HolgerKnublauch @ajnelson-nist if you have permission to do so I strongly encourage you to enable branch protections which require PRs to be up-to-date to merge. These settings should be available at https://github.com/w3c/data-shapes/settings/branches

jeswr:
*phew* it worked - https://github.com/w3c/data-shapes/blob/0f8d872b125c67cf5c3ec9bcf21dc110357c25ca/shacl/shacl-shacl.ttl now appears to be back in the correct state

ajnelson-nist:
> @HolgerKnublauch @ajnelson-nist if you have permission to do so I strongly encourage you to enable branch protections which require PRs to be up-to-date to merge. These settings should be available at https://github.com/w3c/data-shapes/settings/branches

I don't have those permissions.  I'd be fine with those protections going in.  It might be some light friction when multiple PRs merge near one another, but then again, I think those slightly turbulent times are when testing is more needed.

PR #408, opened 2025-06-19
Issue 189: Add `sh:nodeByExpression` Constraint Component

### LINKED ISSUE #189: Enable "Shapes as Data" Paradigm
I've had an idea for a possible extension to SHACL for a while and I'm wondering what others think about it.

Over the past couple years, I have run into several situations where constraints are part of the domain of interest and those constraints should apply to other data in the domain. In those cases, it would be helpful to have shapes be defined as part of *data* instead of at the *schema* level, and it would be helpful if the SHACL engine knew how data were connected to these shapes they should be validated against via *some existing path expressed in domain terminology*.

Doing this would prevent users from needing to extend the ontology/schema to add new constraints. Also, it could prevent the use of metamodeling to accomplish a similar goal, which can get messy and confusing for users.

Here are three generic examples where this feature could potentially be helpful to help convey the idea:

### Example 1
Consider the Function Ontology (https://fno.io/spec/#ontology-abstract). If you look at the documentation for `fno:Parameter` and `fno:Output` they look very similar to `sh:PropertyShape` in spirit, and the class `fno:Function` is therefore like `sh:NodeShape`. It might be useful to use SHACL to validate sure that function arguments and outputs match what is expected based on the function definition.

However, if instances of `fno:Function` were Node Shapes, then there would be no convenient way to configure each `fno:Function` instance to target the right nodes with current SHACL. You'd have to either make each one a class and have the corresponding instances of `fno:Execution` be instances of each (which might tempt the introduction of metamodeling similar to SPIN Functions), write a clunky custom target type using SHACL-AF that wouldn't be supported by all SHACL engines, or use sh:targetNode to connect each `fno:Function` instance to the corresponding `fno:Execution` instances instead of the domain property `fno:executes` (or in addition to it, which would be redundant).

In this case, it would be convenient if each `fno:Execution` could be validated against whatever node it was connected to via `fno:executes`.

### Example 2
Consider some future state of the W3C Data Cube ontology. Data Structure Definitions (https://www.w3.org/TR/vocab-data-cube/#dsd-dsd) and Component Specifications are data in this domain. However, they could be modified to be represented as Node Shapes and Property Shapes respectively such that SHACL could be used to validate that the Observations that are part of DataSets that have that Data Structure Definition actually conform to that structure.

The same challenges exist for trying to validate a `qb:DataStructureDefinition` as a Node Shape as for `fno:Function`; there is no convenient way to configure each `qb:DataStructureDefinition` instance to target the right nodes with current SHACL.

In this case, it would be convenient if each `qb:Observation` could be validated against whatever node it was connected to via the path `qb:dataSet/qb:structure`.

Furthermore, this would allow more fancy data cube behavior more easily, like how shapes are used for datatypes of QB components here: https://docs.allotrope.org/ADF%20Data%20Cube%20Ontology.html (see examples 5 and 11)

### Example 3
Consider the EP-PLAN ontology (https://trustlens.github.io/EP-PLAN/, documentation: https://trustlens.github.io/EP-PLAN/widoco_output/index-en.html), an extension to W3C PROV for capturing in detail the plans that go along with the Activities in PROV. It may be desired to use SHACL to determine whether an activity went according to plan or if some deviation occured. Note that `ep-plan:Step` and `ep-plan:Variable` both could be similar in spirit to `sh:NodeShape`.

The same challenges exist for trying to validate instances of these classes as Node Shapes as for `fno:Function`; there is no convenient way to configure each `ep-plan:Step` and `ep-plan:Variable` instance to target the right nodes with current SHACL.

In this case, it would be convienient if each `ep-plan:Activity` could be validated against whatever node it was connected to via `ep-plan:correspondsToStep` and if each `ep-plan:Entity` could be validated against whatever node it was connected to via `ep-plan:correspondsToVariable`.

### Possible Implementation
I've thought of a few different ways to implement this behavior, but I think the simplest and most efficient way I've thought of so far is to create a new Constraint Component.

This new Constraint Component would function somewhat like the one for `sh:node`. However, instead of specifying the *URI of a Node Shape* that value nodes must also conform to, it specifies a SHACL path using a parameter perhaps called, e.g., `sh:nodesPath`. For each value node for the shape with a value for `sh:nodesPath`, that value node is also validated against *any Node Shape(s) found at the specifed path **from the value node*** (if any resources at that path exist and are Node Shapes).

This would enable the following addition for the Function Ontology in order to validate that all instances of `fno:Execution` conform to any corresponding instance of `fno:Function`:
```turtle
fno:Execution
  sh:nodesPath fno:executes ;
.
```
And this addition for the Data Cube Ontology in order to validate that all instances of `qb:Observation` conform to any corresponding instance of `qb:DataStructureDefinition`:
```turtle
qb:Observation
  sh:nodesPath (
    qb:dataSet
    qb:structure
  ) ;
.
```
And these additions for the EP-PLAN Ontology in order to validate that all instances of `ep-plan:Activity` conform to any corresponding instance(s) of `ep-plan:Step` and that all instances of `ep-plan:Entity` conform to any corresponding instance(s) of `ep-plan:Variable`:
```turtle
ep-plan:Activity
  sh:nodesPath ep-plan:correspondsToStep ;
.
ep-plan:Entity
  sh:nodesPath ep-plan:correspondsToVariable ;
.
```

My main reservation with this approach is that I'm not a huge fan of how if sh:node fails validation, many SHACL engines don't include the nested results via `sh:detail` in their reports, and this constraint would probably function the same way. I hope that more validators would use/take advantage of `sh:detail` in the future in general.

I have added a prototype implementation of this to [this branch in this fork of pyshacl](https://github.com/mgberg/pySHACL/tree/nodesPath) (just because I happen to be the most familiar with the internals of that SHACL engine) and have been playing around with it. Included in [this folder in the repo](https://github.com/mgberg/pySHACL/tree/nodesPath/proposal) is a file with example data and shapes that demonstrates how it works, as well as the output from the modified version of pyshacl (cleaned up a bit for readability).

I'm curious to know what the community thinks of this, both as a concept and also this particular method of implementation.

mgberg:
This is one potential approach to resolve #189.

In the original issue, I discuss an idea for a capability that would enable dynamic computation of a set of node shapes to which a particular value node should should conform. I originally suggested a capability that would behave similarly to `sh:node` except instead of specifying the URI of a node shape that value nodes must also conform to, it specifies a SHACL path; value nodes would also validated against any node shape(s) found at that path from each value node. Refer to #189 for more details, the motivation for this capability, and some example use cases.

After reading the current state of the SHACL 1.2 spec and the expanded use of Node Expressions, I realized it might be possible to add this capability with more expressiveness and less impact than originally proposed by following some of the patterns already appearing in some of the changes from 1.1.

This proposes to generalize `sh:node` to accept not a node shape, but rather a node expression. This change is similar to how `sh:targetNode`, `sh:deactivated`, and `sh:defaultValue` were generalized to accept node expressions. Additionally, this change would function in a manner quite like `sh:expression`. This would enable existing implementations to continue to use IRIs of specific node shapes as values of `sh:node` as prior to 1.2 and also could use more complex Node Expressions (if supported) to dynamically compute sets of node shapes to validate value nodes against using a mechanism similar to `sh:expression`.

Another minor addition is the requirement that validation results generated by `sh:node` must include the shape against which conformance checking failed as the value for `sh:sourceConstraint`. If this was not present, there would be no machine interpretable way to know which dynamically computed shape that caused the validation result. Before this change, the only place the URI of the shape appeared in the example validation results was in the generated value for `sh:resultMessage`.

The primary concern I currently see with this change is something I didn't think about until reviewing the test cases involving `sh:node` after putting together the rest of the PR. This change would prevent using a node shape that is a blank node as the value of `sh:node`, which would be a non-backward compatible change. While it is convenient to use blank nodes node shapes in test cases and examples, I'm not sure how often blank node values of  `sh:node` are actually used in practice. Consequently, I'm not sure if this change would actually be acceptable based on what level of backwards compatibility is expected/required for SHACL 1.2 vs 1.1. Note that in SHACL 1.1 a blank node technically could have been specified as a value of `sh:targetNode` to be validated and or as a value of `sh:defaultValue`, neither of which would be possible in 1.2 when values for these properties would be node expressions, so there are some similar breaking changes being made elsewhere. However, I would assume these examples are almost certainly less likely to occur.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-189/shacl12-core/index.html)


HolgerKnublauch (changes requested):
Hi Matt, you have mentioned the show-stopper problem at the end. We cannot switch to node expressions whenever a constraint component previously allowed blank nodes. Perhaps you need to introduce a new constraint component, and leave sh:node unchanged?

mgberg:
I added this capability as a new constraint component instead as requested, called `sh:nodeByExpression`. I am not particularly attached to that name, so if anyone comes up with something that sounds better, I can definitely change it.

shacl12-core/index.html
```diff
@@ -6074,6 +6075,220 @@ <h4>sh:reifierShape, sh:reificationRequired</h4>
...
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="NodeByExpression">
+							Let <code>$expr</code> be a <a>value</a> of <code>sh:nodeByExpression</code>.
+							For each <a>value node</a> <code>v</code>: perform <a>conformance checking</a> of <code>v</code> against each <a>output node</a>
+							of <code>evalExpr(expr, <a>data graph</a>, v, {})</code> <code>s</code> that is a <a>node shape</a> in the <a>shapes graph</a>.
+							For each conformance check, a <a>failure</a> MUST be produced if the <a>conformance checking</a> of <code>v</code> against <code>s</code> produces a <a>failure</a>.
+							Otherwise, if <code>v</code> does not <a>conform</a> to <code>s</code>,
+							there is a <a>validation result</a> with <code>v</code> as <code>sh:value</code> and a <a>deep copy</a> of <code>s</code> as <code>sh:sourceConstraint</code>.
```
TallTed: Consider applying this indent adjustment throughout the ReSpec source (basically, start all body content lines from position 0 instead of a 7 tab indent; optimally, to be applied across the board via a `tidy` script). I will adjust my suggestion here, if the indent adjustment is not accepted.

Separately, consider the text tweaks. Especially, I think most if not all instances of `conformance checking` should become `conformance check`, throughout the document(s).

```suggestion
Let <code>$expr</code> be a <a>value</a> of <code>sh:nodeByExpression</code>.
For each <a>value node</a> <code>v</code>: perform <a>conformance check</a> of
<code>v</code> against each <a>output node</a> of <code>evalExpr(expr,
<a>data graph</a>, v, {})</code> <code>s</code> that is a <a>node shape</a>
in the <a>shapes graph</a>. For each <a>conformance check</a>, a <a>failure</a>
MUST be produced if the <a>conformance check</a> of <code>v</code> against
<code>s</code> produces a <a>failure</a>. Otherwise, if <code>v</code> does
not <a>conform</a> to <code>s</code>, there is a <a>validation result</a>
with <code>v</code> as <code>sh:value</code> and a <a>deep copy</a> of
<code>s</code> as <code>sh:sourceConstraint</code>.
```

TallTed (changes requested):

mgberg:
@TallTed I accepted your suggestion, but I didn't modify the formatting of body text throughout the remainder of the document. I assume it would be more appropriate to reformat the entire document as a separate PR.

TallTed:
will pick up my remaining suggestions in #411 and #412

TallTed (approved):
Good to go.

shacl12-core/index.html
```diff
@@ -6074,6 +6075,224 @@ <h4>sh:reifierShape, sh:reificationRequired</h4>
...
+						</tr>
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="NodeByExpression">
+Let <code>$expr</code> be a <a>value</a> of <code>sh:nodeByExpression</code>.
+For each <a>value node</a> <code>v</code>: perform <a>conformance check</a> of
+<code>v</code> against each <a>output node</a> of <code>evalExpr(expr,
+<a>data graph</a>, v, {})</code> <code>s</code> that is a <a>node shape</a>
+in the <a>shapes graph</a>. For each <a>conformance check</a>, a <a>failure</a>
```
HolgerKnublauch: Drop the "For each <a>conformance check</a>" as it is redundant and you want the whole validation to fail.

shacl12-core/index.html
```diff
@@ -6074,6 +6075,224 @@ <h4>sh:reifierShape, sh:reificationRequired</h4>
...
+							</td>
+						</tr>
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="NodeByExpression">
+Let <code>$expr</code> be a <a>value</a> of <code>sh:nodeByExpression</code>.
+For each <a>value node</a> <code>v</code>: perform <a>conformance check</a> of
+<code>v</code> against each <a>output node</a> of <code>evalExpr(expr,
+<a>data graph</a>, v, {})</code> <code>s</code> that is a <a>node shape</a>
```
HolgerKnublauch: Drop the "that is a node shape..." sub-sentence because there is no way for the engine to check whether the node really is a well-formed node shape. If it's not a node shape, it likely will simply mean that the validation will succeed. If it's malformed, it will likely cause a failure.

shacl12-core/index.html
```diff
@@ -6074,6 +6075,224 @@ <h4>sh:reifierShape, sh:reificationRequired</h4>
...
+		sh:maxCount 1 ;
+	] .
+
+ex:PersonShape
+	a sh:NodeShape ;
+	<span class="target-can-be-skipped">sh:targetClass ex:Person ;</span>
+	sh:property [   # _:b1
+		sh:path ex:address ;
+		sh:minCount 1 ;
+		sh:node ex:AddressShape ;
```
HolgerKnublauch: Did you mean to use sh:nodeByExpression here? In both cases the example could just be dropped because it is almost identical to the sh:node example. So you could just point at that. The examples would be more interesting after the Node Expression spec is done, but we are not there yet.

shacl12-test-suite/tests/core/node/nodeByExpression-001.ttl
```diff
@@ -0,0 +1,59 @@
...
+  rdf:type ex:TestClass ;
+  rdfs:label "Invalid instance" ;
+.
+ex:TestClass
+  rdf:type rdfs:Class ;
+  rdf:type sh:NodeShape ;
+  rdfs:label "Test class" ;
+  rdfs:subClassOf rdfs:Resource ;
+  # Only using an IRI Expression here because Core doesn't define interesting node expressions
+  sh:node ex:TestNodeShape ;
```
HolgerKnublauch: sh:node -> sh:nodeByExpression

shacl12-test-suite/tests/core/property/nodeByExpression-001.ttl
```diff
@@ -0,0 +1,117 @@
...
+  rdfs:label "Issue" ;
+  rdfs:subClassOf rdfs:Resource ;
+  sh:property ex:Issue-assignedTo ;
+  sh:property ex:Issue-submittedBy ;
+.
+ex:Issue-assignedTo
+  sh:path ex:assignedTo ;
+  sh:class ex:Person ;
+  # Only using an IRI Expression here because Core doesn't define interesting node expressions
+  sh:node ex:AssignedToShape ;
```
HolgerKnublauch: sh:node -> sh:nodeByExpression

HolgerKnublauch (changes requested):

HolgerKnublauch (approved):
Tentatively approving but it would be best to wait with the merge until next week, when it was presented to the meeting for everyone's awareness.

shacl12-core/index.html
```diff
@@ -6128,170 +6127,9 @@ <h4>sh:nodeByExpression</h4>
 						as an <a>IRI expression</a> that evaluates to a set containing the same <a>node shape</a>.
 					</p>
 					<p>
-						In the following example, all values of the property <code>ex:address</code> must fulfill the
-						constraints expressed by the <a>shape</a> <code>ex:AddressShape</code>.
+						For a simple usage example, refer to the example for <a href="#NodeConstraintComponent"><code>sh:node</code></a>,
+						but replace `sh:node` with `sh:nodeByExpression`.
```
TallTed: How many electrons do we save by this "look at the other example"? How much misunderstanding do we expect from people not quite following "replace" instructions? I think we'd be wiser to keep each example distinct, with any such "replace" already done.
```suggestion
						For a simple example of its use, refer to the example for <a href="#NodeConstraintComponent"><code>sh:node</code></a>,
						replacing `sh:node` with `sh:nodeByExpression`.
```

mgberg: > How many electrons do we save by this "look at the other example"? How much misunderstanding do we expect from people not quite following "replace" instructions? I think we'd be wiser to keep each example distinct, with any such "replace" already done.

I can always revert that change if it is decided that's best.

HolgerKnublauch: It would make more sense to have an example of this in the Node Expressions spec, not in Core.

mgberg: It would be great to put an example in the Node Expressions spec or somewhere similar since an example in the core spec essentially could only be the same as `sh:node` except for the addition of `sh:sourceConstraint` in the result.

shacl12-core/index.html
```diff
@@ -6074,6 +6075,62 @@ <h4>sh:reifierShape, sh:reificationRequired</h4>
...
+					</p>
+					<p>
+						Note that `sh:node` and `sh:nodeByExpression` exhibit the same behavior when given a <a>value</a> that is an <a>IRI</a> of a <a>node shape</a>.
+						In this case, `sh:node` directly validates against the specified <a>node shape</a>, whereas `sh:nodeByExpression` interprets the <a>IRI</a>
+						as an <a>IRI expression</a> that evaluates to a set containing the same <a>node shape</a>.
+					</p>
+					<p>
+						For a simple example of its use, refer to the example for <a href="#NodeConstraintComponent"><code>sh:node</code></a>,
+						replacing `sh:node` with `sh:nodeByExpression`.
+					</p>
```
afs, line 6698: An example at this point showing where it is different to `sh:node` would be useful.

mgberg: An example was removed as per the discussion above ([here](https://github.com/w3c/data-shapes/pull/408#discussion_r2166126422) and [here](https://github.com/w3c/data-shapes/pull/408#discussion_r2172542066)).

There are two differences between `sh:nodeByExpression` and `sh:node` within the scope of the core spec:
1. `sh:nodeByExpression` cannot take a node shape that is a blank node as `sh:node` can, as a blank node would be interpreted as a Node Expression.
2. Results generated by `sh:nodeByExpression` additionally include a value for `sh:sourceConstraint`.

The example I originally included was essentially a duplicate of the example for `sh:node` with the constraint component replaced and a value for `sh:sourceConstraint` in the example result. @HolgerKnublauch suggested I remove it and wait to put a more comprehensive example in the Node Expression spec.

I can certainly revert that change if it is desired to have that example back.

afs: Thanks for the explanation.

As a general comment, I'm not sure we are helping users by not including examples that show clear node expressions in Core but that's not for this PR.

(Do people really think there will be common static only SHACL 1.2 implementations?)

HolgerKnublauch:
Yes there will be many pure core implementations, if shacl-sparql is an example to learn from.

Sent from my iPhone

afs:
> Yes there will be many pure core implementations, if shacl-sparql is an example to learn from.

So there will be a profile of SHACL which is "1.2-basic" - no node expressions except constants?

HolgerKnublauch:
SHACL Core is just exactly what is written in the SHACL Core spec, i.e. no NEs. I guess that's the default profile. Then the other documents define additional profiles.

ajnelson-nist:
> SHACL Core is just exactly what is written in the SHACL Core spec, i.e. no NEs. I guess that's the default profile. Then the other documents define additional profiles.

@HolgerKnublauch : This implies to me that SHACL-SHACL would have specialized files for each profile.  If one were to use "Strictly" SHACL-Core, they would validate with a graph that disallows certain things like you mentioned, and that validation graph would import a general SHACL-SHACL graph.  Is this consistent with what you're thinking?  I'll log an Issue if so.

HolgerKnublauch:
I thought the SHACL Profiles work will suggest a framework for this. Maybe it's a combination of graphs that are referencing each other via owl:imports, with shacl-shacl-core sitting as the base.

mgberg:
@HolgerKnublauch, @afs, @TallTed: are there any other changes that should be made to this PR, or can it be merged?

shacl12-core/index.html
simonstey: > (Do people really think there will be common static only SHACL 1.2 implementations?)

@afs I mean there are some SHACL 1.0 ones that don't support SPARQL-based constraints such as https://github.com/zazuko/rdf-validate-shacl#limitations (on which https://shacl-playground.zazuko.com/ is based) for example.

> There are two differences between `sh:nodeByExpression` and `sh:node` within the scope of the core spec:
> 
>     1. `sh:nodeByExpression` cannot take a node shape that is a blank node as `sh:node` can, as a blank node would be interpreted as a Node Expression.
> 
>     2. Results generated by `sh:nodeByExpression` additionally include a value for `sh:sourceConstraint`.

@mgberg I like this description as well.. can't we add this as well to the text?

> 
> The example I originally included was essentially a duplicate of the example for `sh:node` with the constraint component replaced and a value for `sh:sourceConstraint` in the example result. [@HolgerKnublauch](https://github.com/HolgerKnublauch) suggested I remove it and wait to put a more comprehensive example in the Node Expression spec.
> 
> I can certainly revert that change if it is desired to have that example back.

I don't think one can have _too much_ examples, esp. if they help clearing things up. 
+1 to adding it back

TallTed:
> @HolgerKnublauch, @afs, @TallTed: are there any other changes that should be made to this PR, or can it be merged?

I believe all my requested changes have been made.

afs (approved):

shacl12-core/index.html
mgberg: @simonstey I modified the description in the spec to include those points explicitly, and I also added the example back.

PR #409, opened 2025-06-22
feat: #192 add severity levels that are not violations

### LINKED ISSUE #192: Add severity levels that are not violations
`sh:Info`, `sh:Warning`, and `sh:Violation` are defined as violations. It would be good to have severity levels not defined as violations. They could show detailed information on which constraints have been processed.

`shacl-engine` already uses two new levels in its own namespace:

- `Debug`: Shows that a constraint has been processed.
- `Trace`: For property shapes without constraints showing the traversal path.

bergos:
This PR closes #192 "Add severity levels that are not violations". It contains the following changes:

- Adds `sh:Trace` and `sh:Debug` severity levels that aren't violations.
- Allows validation engines to treat `sh:Info` and `sh:Warning` as non-violating.
- Changes the definition of `sh:conforms` of being true when there are no results to being true if there are no results with a severity level that is a violation.

A rule based on RDF classes for sh:conforms would have been preferred, but since it's possible to dynamically set sh:Info and sh:Warning being a violation, that is not possible.

Allowing validation engines to treat levels as non-violating but keeping the default makes this a non-breaking change.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-192-severity-levels/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1740,6 +1740,14 @@ ex:Bob ex:livesIn ex:NewYork .
 							<th>Description</th>
 						</tr>
 						<tr>
+							<td><code>sh:Trace</code></td>
+							<td>A trace message that is not a violation.</td>
+						</tr>
+						<tr>
+							<td><code>sh:Debug</code></td>
+							<td>A debug message that is not a violation.</td>
+						</tr>
+						<tr>
 							<td><code>sh:Info</code></td>
 							<td>A non-critical constraint violation indicating an informative message.</td>
 						</tr>
@@ -1752,6 +1760,10 @@ ex:Bob ex:livesIn ex:NewYork .
 							<td>A constraint violation.</td>
 						</tr>
 					</table>
+					<p>
+						Validation engines may treat <code>sh:Info</code> and <code>sh:Warning</code> as non-violating based on
+						options passed to the engine. By default, they must be treated as violations.
+					</p>
 					<p>
 						In addition to declaring severities per shape, the property <code>sh:severity</code> can also be used
 						on a <a>reifier</a> for a triple where the <a>shape</a> is the <a>subject</a> and one of the <a>parameters</a>
@@ -2917,7 +2929,7 @@ ex:SomeClass-alternativePathExample
 						<p>
 							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph has exactly one value for the property <code>sh:conforms</code> and the value is of datatype <code>xsd:boolean</code>.
 							It represents the outcome of the <a>conformance checking</a>.
-							The value of <code>sh:conforms</code> is <code>true</code> if and only if the <a>validation</a> did not produce any <a>validation results</a>, and <code>false</code> otherwise.
+							The value of <code>sh:conforms</code> is <code>true</code> if and only if the <a>validation</a> did not produce any <a>validation results</a> with a severity level representing a violation, and <code>false</code> otherwise.
 						</p>
 					</section>
 					<section id="result">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -1761,8 +1761,8 @@ ex:Bob ex:livesIn ex:NewYork .
 						</tr>
 					</table>
 					<p>
-						Validation engines may treat <code>sh:Info</code> and <code>sh:Warning</code> as non-violating based on
-						options passed to the engine. By default, they must be treated as violations.
+						Validation engines MAY treat <code>sh:Info</code> and <code>sh:Warning</code> as non-violating based on
+						options passed to the engine. By default, they are treated as violations.
 					</p>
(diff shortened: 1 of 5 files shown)


shacl12-core/index.html
```diff
@@ -1752,6 +1760,10 @@ <h4>Declaring the Severity of a Shape or Constraint</h4>
 							<td>A constraint violation.</td>
 						</tr>
 					</table>
+					<p>
+						Validation engines may treat <code>sh:Info</code> and <code>sh:Warning</code> as non-violating based on
```
HolgerKnublauch: I believe you should use MAY instead of may.

shacl12-core/index.html
```diff
@@ -1752,6 +1760,10 @@ <h4>Declaring the Severity of a Shape or Constraint</h4>
...
+						options passed to the engine. By default, they must be treated as violations.
```
HolgerKnublauch: ... they are treated as violations.

(This avoids the special semantics of "must")

HolgerKnublauch (changes requested):
This is a good addition, and it's good to relax the interpretation of conforms. However, there are other places that now need to be changed too. A quick look revealed the section "Conformance Checking", which formally defines the term "conforms". This needs to take the additional flexibility into account. Other places that rely on "conforms" such as sh:node then don't need to be adjusted further.

bergos:
@HolgerKnublauch I also rechecked the spec and found only the "Conformance Checking" section that needed an update. The change is included in the latest commit.

HolgerKnublauch (approved):

YoucTagh (approved):

bergos:
@ajnelson-nist I've added the test cases for `sh:Debug` and `sh:Trace`, as discussed in our call.

shacl12-core/index.html
```diff
@@ -1752,6 +1760,10 @@ <h4>Declaring the Severity of a Shape or Constraint</h4>
 							<td>A constraint violation.</td>
 						</tr>
 					</table>
+					<p>
+						Validation engines MAY treat <code>sh:Info</code> and <code>sh:Warning</code> as non-violating based on
+						options passed to the engine. By default, they are treated as violations.
+					</p>
```
ajnelson-nist, line 1766: I think my one last hangup on this PR is this escape and the lack of demonstration of a "test configuration record".

The tests @bergos added, 4 and 5, are fine demonstrations of a validation result set with conformance of `true` and the non-violation levels.  (Thank you, Thomas.)

Can the test suite support a new test 6, mostly duplicative of 4 or 5, but instead using `sh:Warning` and a conformance of `true`?  How would test suite conformance be reproducible?

I feel like another property in the `ValidationReport` would be helpful, that logs whether its `sh:conformance` value was set to "excuse" `sh:Info` or `sh:Warning` severity results from tripping `sh:conformance` into `false`.  `[] a sh:ValidationReport ; sh:excuses sh:Warning ; ...`?  `sh:permits`, `sh:permitsSeverity`, or `sh:excusesSeverity` might work too.

bergos: @ajnelson-nist I understand, and I also would like to add tests for it, but there are two problems I tried to avoid:

- The feature is optional, and as an implementor, I would like to pick only those tests that match the features supported by the implementation to see if any tests fail. That means the tests must be tagged to test a specific optional feature.
- The feature is controlled by a parameter given to the engine. A data structure for the parameters would be required that can be added to the test to configure the engine.

I've created an issue #431 to discuss that topic, as these two problems are not specific to this PR. If it's OK, I would merge this PR now and use the issue to track the missing tests?

ajnelson-nist: This feels like a pretty significant design decision:

A. Make a general runtime-configuration data structure.
B. Represent a specific runtime-configuration value having an impact on a top-level `sh:ValidationReport` feature.

To do B, do we need to do A?

This particular case feels like it's toeing a line I hadn't appreciated before dividing "optional behavior" from "optional feature".

More practically speaking, looking at the surround text, I think this section needs a bit more tweaking.  We currently have this scale:

| Severity | Violation? |
| --- | --- |
| Trace | No |
| Debug | No |
| Info | Non-critically yes |
| Warning | Non-critically yes |
| Violation | Yes |

Should the "Validation engines MAY treat ..." paragraph describe "conformance," rather than whether each `sh:ValidationResult` Boolean-is-or-is-not a violation?

ajnelson-nist: We'll need to address this, but after lengthy discussion on today's call, I believe it unlikely we'll forget to.  Please do leave the conversation "unresolved" in the GitHub interface so we can still see this later.

ajnelson-nist (approved):
Approving, aware that we'll need to visit later how to address `sh:Info` and `sh:Warning` optionally influencing `sh:conforms`.

PR #416, opened 2025-06-28
Issue 414: list constraints

### LINKED ISSUE #414: Extend collection validation features
In https://github.com/w3c/data-shapes/pull/403 we implement support for #391 (`sh:memberShape`). As discussed in https://github.com/w3c/data-shapes/pull/403#discussion_r2169534020 we may wish to support some other constraints on collections - including:
 - `sh:minCollectionLength` and `sh:maxCollectionLength` (I think we SHOULD support this)
 - `sh:uniqCollection` to check that all elements in the list are unique (I think we SHOULD support this)
 - `sh:memberCompare` which operates similarly to [property pair constraints](https://www.w3.org/TR/shacl/#core-components-property-pairs) by comparing adjacent elements; e.g. to enforce numeric ordering. (I am on the fence about supporting this)

I would like to land https://github.com/w3c/data-shapes/pull/403 before implementing these if we choose to proceed.

jeswr:
Closes #414. This PR implements `sh:minListLength`, `sh:maxListLength` and `sh:uniqList`.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/feat/list-constraints/shacl12-core/index.html)

---

My interpretation of the following from the definition of [2.1.1 Constraints, Parameters and Constraint Components](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#constraints) is that because the `sh:MemberShapeConstraintComponent` now has multiple parameters; there can only be one occurrence of `sh:memberShape` in a given shape - and I have updated the `sh:maxCount` accordingly.

> Some constraint components declare only a single parameter. For example [sh:ClassConstraintComponent](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#ClassConstraintComponent) has the single parameter sh:class. These parameters may be used multiple times in the same shape, and each [value](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#dfn-values) of such a parameter declares an individual [constraint](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#dfn-constraints). The interpretation of such declarations is conjunction, i.e. all constraints apply. The following example specifies that the values of ex:customer have to be [SHACL instances](https://raw.githack.com/w3c/data-shapes/feat/memberShape/shacl12-core/index.html#dfn-shacl-instance) of both ex:Customer and ex:Person.

---

⚠️⚠️ ⚠️ 
This is currently targeted to #403 branch for a smaller diff in the GH UI.  #403 should be merged and this targeted to `gh-pages` rather than merging this branch into #403


HolgerKnublauch (changes requested):
I think you have linked the new constraint parameters to the same constraint component for memberShape. I believe it would be better to introduce separate constraint components for each of them. It makes the definition more modular but, more importantly, it is perfectly sensible to use for example sh:minListLength without a sh:memberShape constraint.

shacl12-core/index.html
```diff
@@ -5715,21 +5715,59 @@ <h4>sh:memberShape</h4>
...
+								An optional <a>boolean</a> that specifies whether the members of the <a href="#syntax-rule-SHACL-list">SHACL list</a> must be unique.
+							</td>
+						</tr>
 					</table>
 					<div class="def def-text">
 						<div class="def-header">TEXTUAL DEFINITION</div>
 						<div class="def-text-body" data-validator="MemberShape">
 							Let <code>$memberShape</code> be a <a>value</a> of <code>sh:memberShape</code>.
 							Each <a>value node</a> <code>v</code> must be a <a href="#syntax-rule-SHACL-list">SHACL list</a> - a <a>failure</a> MUST be produced if <code>v</code> is not a SHACL list.
+							For each <a>value node</a> <code>v</code> that is a valid SHACL list, a <a>failure</a> MUST be produced for each violation of the following conditions:
```
HolgerKnublauch: Note that FAILURE is not the right word here. It should be "validation result". Failure is used when a node shape fails to evaluate due to some low-level problem. Normal constraint violations are validation results.

shacl12-core/index.html
```diff
@@ -5715,21 +5715,59 @@ <h4>sh:memberShape</h4>
...
+							<ul>
+								<li>
+									If <code>sh:minListLength</code> is present, the number of members in the list <code>v</code> MUST be greater than or equal to the value of <code>sh:minListLength</code>.
+								</li>
+								<li>
+									If <code>sh:maxListLength</code> is present, the number of members in the list <code>v</code> MUST be less than or equal to the value of <code>sh:maxListLength</code>.
+								</li>
+								<li>
+									If <code>sh:uniqList</code> is present and has the value <code>true</code>, the members of the list <code>v</code> MUST be unique.
```
HolgerKnublauch: I still don't like the name uniqList, and would prefer uniqueMembers. The abbreviation uniq is not really necessary for such a short word, and not the list is unique but its members are.

jeswr:
> I think you have linked the new constraint parameters to the same constraint component for memberShape. I believe it would be better to introduce separate constraint components for each of them. It makes the definition more modular but, more importantly, it is perfectly sensible to use for example sh:minListLength without a sh:memberShape constraint.

Makes sense! Following this approach; I expect similarly to `sh:memberShape` - there should be an error if `sh:minListLength` is declared and the focus node is not a valid SHACL list. This would mean that if both `sh:memberShape` and `sh:minListLength` are declared - and the focus node is not a valid list then both constraints would independently trigger validation results which look like https://github.com/w3c/data-shapes/blob/536ce96fe3c6f1a1c774f569292a281d198aac8d/shacl12-test-suite/tests/core/node/memberShape-001.ttl#L121-L138

I guess we can expect similar behaviour if there are multiple `sh:memberShape` declarations in #403

Is this the behaviour you would expect for validation results?

HolgerKnublauch:
I wouldn't worry too much about malformed lists. Usually people use tools to produce them, including editing TTL source code. So if they are reported twice, this shouldn't be the end of the world.

jeswr:
I also saw a comment about lists not being valid objects of `sh:targetNode` but it's now not appearing in the GH UI - was this comment deleted because the statement doesn't hold, or is GH doing weird things?

HolgerKnublauch:
Yeah my fault. I deleted the comment because I understood later what you did.

shacl12-core/index.html
jeswr: I have fixed this, aligned the textual definition with the other textual definitions in the document and moved discussion of `sh:detail` to the informative section in doing so.

jeswr:
I think this is good to go now - I suggest we merge this to `gh-pages` and close #403 as I have applied changes to the `sh:memberShape` definition in this PR including moving it to the new [List Constraint Component](https://raw.githack.com/w3c/data-shapes/feat/list-constraints/shacl12-core/index.html#core-components-list) section and addressing https://github.com/w3c/data-shapes/pull/416#discussion_r2173705517

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -226,7 +227,24 @@ shsh:ShapeShape
...
+		sh:minInclusive 0 ;             # minListLength-minInclusive
+	] ;
+	sh:property [
+		sh:path sh:maxListLength ;
+		sh:datatype xsd:integer ;       # maxListLength-datatype
+		sh:maxCount 1 ;                 # maxListLength-maxCount
+		sh:minInclusive 0 ;             # maxListLength-minInclusive
+	] ;
+	sh:property [
+		sh:path sh:uniqList ;
```
HolgerKnublauch: sh:uniqList -> sh:uniqueMembers

shacl12-core/index.html
```diff
@@ -4517,6 +4520,462 @@ <h4>sh:uniqueLang</h4>
...
+						<div class="def-text-body" data-validator="MemberShape">
+							Let <code>$memberShape</code> be a <a>parameter value</a> for <code>sh:memberShape</code>.
+							Each <a>value node</a> <code>v</code> must be a <a href="#syntax-rule-SHACL-list">SHACL list</a> - if <code>v</code> is not a SHACL list there is a <a>validation result</a>.
+							If any member <code>m</code> of the list <code>v</code> does not <a>conform</a> to <code>$memberShape</code>, there is a <a>validation result</a>. There must be at most one <a>validation result</a> for each list <code>v</code>.
+						</div>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						Each member <code>m</code> of a value node <code>v</code> that does not conform to the <code>$memberShape</code> should be reported as a separate <code>sh:detail</code> in the <a>validation result</a> for <code>v</code>.
+						If the list <code>v</code> is not a valid <a href="#syntax-rule-SHACL-list">SHACL list</a>, this should also be reported as a <code>sh:detail</code> in the <a>validation result</a> for <code>v</code>.
```
HolgerKnublauch: I think this needs to be changed. If the list is not valid, it should become a separate top-level violation and it shouldn't even attempt to validate the individual members anymore. This applies to all List-based constraint components IMHO, including length.

shacl12-core/index.html
```diff
@@ -4517,6 +4520,462 @@ <h4>sh:uniqueLang</h4>
...
+							</td>
+						</tr>
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="UniqueMembers">
+							Let <code>$uniqueMembers</code> be a <a>parameter value</a> for <code>sh:uniqueMembers</code>.
+							Each <a>value node</a> <code>v</code> must be a <a href="#syntax-rule-SHACL-list">SHACL list</a> - if <code>v</code> is not a SHACL list there is a <a>validation result</a>.
+							If <code>$uniqueMembers</code> is <code>true</code> and the list <code>v</code> has duplicate members,
+							there is a <a>validation result</a>. There must be at most one <a>validation result</a> for each list <code>v</code>.
```
HolgerKnublauch: The sentence "There must be at most one validation result..." should be removed because it is redundant and may not apply in cases where the same value node is reached by different, violated constraints.

shacl12-test-suite/tests/core/node/uniqueMembers-001.ttl
```diff
@@ -0,0 +1,86 @@
...
+      sht:dataGraph <> ;
+      sht:shapesGraph <> ;
+    ] ;
+  mf:result [
+      rdf:type sh:ValidationReport ;
+      sh:conforms "false"^^xsd:boolean ;
+      sh:result [
+          rdf:type sh:ValidationResult ;
+          sh:focusNode _:b1 ;
+          sh:resultMessage "Value does not conform to Shape ex:ListShape. See details for more information." ;
```
HolgerKnublauch: Wrong message - no ex:ListShape here.

shacl12-vocabularies/shacl.ttl
```diff
@@ -907,13 +907,76 @@ sh:MemberShapeConstraintComponent-memberShape
...
+	rdfs:isDefinedBy sh: .
+
+sh:UniqueMembersConstraintComponent
+	a sh:ConstraintComponent ;
+	rdfs:label "Unique list members constraint component"@en ;
+	rdfs:comment "Specifies that all members of a SHACL list must be unique. A violation is reported if the value is not a valid SHACL list."@en ;
+	sh:parameter sh:UniqueMembersConstraintComponent-uniqueMembers ;
+	rdfs:isDefinedBy sh: .
+
+sh:MemberShapeConstraintComponent-uniqueMembers
```
HolgerKnublauch: Wrong URI - needs to be UniqueMembers...

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
```diff
@@ -0,0 +1,60 @@
...
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix sht: <http://www.w3.org/ns/shacl-test#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+ex:ListShape
+  rdf:type sh:NodeShape ;
+  sh:maxListLength 2 ;
+  # Satisfies all constraints
+  sh:targetNode ( 1 2 ), rdf:nil ;
```
HolgerKnublauch: So my initial comment (that I later deleted) turned out to be correct. This doesn't work. When you use a blank node such as ( 1 2) as targetNode, it will be evaluated as a node expression, which means it will try to walk the rdf:List and then use each list member as target node. To work around this for the test cases, I think all lists need to be URI nodes.

HolgerKnublauch (changes requested):
I am in the middle of implementing all this in Java and found various issues.

shacl12-vocabularies/shacl-shacl.ttl
jeswr:
```suggestion
		sh:path sh:uniqueMembers ;
```

shacl12-vocabularies/shacl.ttl
jeswr:
```suggestion
sh:UniqueMembersConstraintComponent-uniqueMembers
```

shacl12-test-suite/tests/core/node/uniqueMembers-001.ttl
jeswr: I think this is the right message `ex:ListShape` is the NodeShape defined at the top of this file which requires unique memebrs

shacl12-core/index.html
jeswr: Changed in https://github.com/w3c/data-shapes/pull/416/commits/ca91c2c71ce6526995d39a5ba2eca744ba2d9d39. Length based constraints were already top-level violations so I don't believe any changes are needed to those tests - unless there is an expectation that there are distinct values for some of the result properties.

shacl12-core/index.html
jeswr: Resolved in https://github.com/w3c/data-shapes/commit/ca91c2c71ce6526995d39a5ba2eca744ba2d9d39

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
jeswr: Resolved in https://github.com/w3c/data-shapes/pull/416/commits/5c72e8eb8a891b008e8c99c54dbdac60407a51ec

jeswr:
I have addressed the issues found - is there an(other?) open source implementation that'd be useful for me to implement the changes against (@bergos' JS shacl-engine perhaps) in case there are any more errors to catch.

P.S. I am conferencing for the next 3 days so will be slow to respond

shacl12-test-suite/tests/core/property/maxListLength-001.ttl
```diff
@@ -0,0 +1,85 @@
...
+      rdf:type sh:ValidationReport ;
+      sh:conforms "false"^^xsd:boolean ;
+      sh:result [
+          rdf:type sh:ValidationResult ;
+          sh:focusNode ex:person3 ;
+          sh:resultMessage "List has length 3, but maximum allowed length is 2." ;
+          sh:resultPath ex:hobbies ;
+          sh:resultSeverity sh:Violation ;
+          sh:sourceConstraintComponent sh:MaxListLengthConstraintComponent ;
+          sh:sourceShape ex:PersonShape ;
```
HolgerKnublauch: The sourceShape must be the property shape, not the node shape. To implement this it would be best to use URIs for the property shapes. Same issue below and in the minListLength and uniqueMembers tests.

shacl12-vocabularies/shacl.ttl
```diff
@@ -907,13 +907,76 @@ sh:MemberShapeConstraintComponent-memberShape
...
+	rdfs:label "Minimum list length constraint component"@en ;
+	rdfs:comment "Specifies the minimum length of a SHACL list. A violation is reported if the value is not a valid SHACL list."@en ;
+	sh:parameter sh:MinListLengthConstraintComponent-minListLength ;
+	rdfs:isDefinedBy sh: .
+
+sh:MinListLengthConstraintComponent-minListLength
+	a sh:Parameter ;
+	sh:path sh:minListLength ;
+	sh:datatype xsd:integer ;
+	sh:maxCount 1 ;
```
HolgerKnublauch: Unnecessary: parameters are always maxCount 1

shacl12-vocabularies/shacl.ttl
```diff
@@ -907,13 +907,76 @@ sh:MemberShapeConstraintComponent-memberShape
...
+	rdfs:label "Maximum list length constraint component"@en ;
+	rdfs:comment "Specifies the maximum length of a SHACL list. A violation is reported if the value is not a valid SHACL list."@en ;
+	sh:parameter sh:MaxListLengthConstraintComponent-minListLength ;
+	rdfs:isDefinedBy sh: .
+
+sh:MaxListLengthConstraintComponent-maxListLength
+	a sh:Parameter ;
+	sh:path sh:maxListLength ;
+	sh:datatype xsd:integer ;
+	sh:maxCount 1 ;
```
HolgerKnublauch: Unnecessary: parameters are always maxCount 1

shacl12-vocabularies/shacl.ttl
```diff
@@ -907,13 +907,76 @@ sh:MemberShapeConstraintComponent-memberShape
...
+	rdfs:label "Unique list members constraint component"@en ;
+	rdfs:comment "Specifies that all members of a SHACL list must be unique. A violation is reported if the value is not a valid SHACL list."@en ;
+	sh:parameter sh:UniqueMembersConstraintComponent-uniqueMembers ;
+	rdfs:isDefinedBy sh: .
+
+sh:UniqueMembersConstraintComponent-uniqueMembers
+	a sh:Parameter ;
+	sh:path sh:uniqueMembers ;
+	sh:datatype xsd:boolean ;
+	sh:maxCount 1 ;
```
HolgerKnublauch: Unnecessary: parameters are always maxCount 1

shacl12-vocabularies/shacl.ttl
```diff
@@ -907,13 +907,76 @@ sh:MemberShapeConstraintComponent-memberShape
...
+	sh:path sh:minListLength ;
+	sh:datatype xsd:integer ;
+	sh:maxCount 1 ;
+	rdfs:isDefinedBy sh: .
+
+sh:MaxListLengthConstraintComponent
+	a sh:ConstraintComponent ;
+	rdfs:label "Maximum list length constraint component"@en ;
+	rdfs:comment "Specifies the maximum length of a SHACL list. A violation is reported if the value is not a valid SHACL list."@en ;
+	sh:parameter sh:MaxListLengthConstraintComponent-minListLength ;
```
HolgerKnublauch: Typo: minListLength -> maxListLength

HolgerKnublauch (changes requested):

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
```diff
@@ -0,0 +1,64 @@
...
+  mf:action [
+      sht:dataGraph <> ;
+      sht:shapesGraph <> ;
+    ] ;
+  mf:result [
+      rdf:type sh:ValidationReport ;
+      sh:conforms "false"^^xsd:boolean ;
+      sh:result [
+          rdf:type sh:ValidationResult ;
+          sh:focusNode rdf:nil ;
```
HolgerKnublauch: sh:focusNode and also the sh:value must be ex:list1

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
```diff
@@ -0,0 +1,64 @@
...
+          sh:resultMessage "List has length 3, but maximum allowed length is 2." ;
```
HolgerKnublauch: Test cases should not include the sh:resultMessage because that is up to the implementation, and locale-specific. Right now, my test engine would fail on this, so I guess others may break too.

shacl12-vocabularies/shacl.ttl
jeswr:
```suggestion
```

shacl12-vocabularies/shacl.ttl
jeswr:
```suggestion
```

shacl12-vocabularies/shacl.ttl
jeswr:
```suggestion
```

shacl12-vocabularies/shacl.ttl
jeswr:
```suggestion
	sh:parameter sh:MaxListLengthConstraintComponent-maxListLength ;
```

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
jeswr:
```suggestion
          sh:focusNode ex:list1 ;
```

shacl12-test-suite/tests/core/node/maxListLength-001.ttl
```diff
@@ -0,0 +1,64 @@
...
+          sh:resultSeverity sh:Violation ;
+          sh:sourceConstraintComponent sh:MaxListLengthConstraintComponent ;
+          sh:sourceShape ex:ListShape ;
```
jeswr, line 50:
```suggestion
          sh:sourceShape ex:ListShape ;
          sh:value ex:list1 ;
```

shacl12-test-suite/tests/core/property/maxListLength-001.ttl
jeswr: Resolved in https://github.com/w3c/data-shapes/pull/416/commits/99ca14f632dba3ca9f623d708864dd8071f375b3

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -74,7 +74,8 @@ shsh:ShapeShape
 		sh:ignoredProperties, sh:in, sh:languageIn, sh:lessThan, sh:lessThanOrEquals, sh:maxCount, sh:maxExclusive,
 		sh:maxInclusive, sh:maxLength, sh:memberShape, sh:minCount, sh:minExclusive, sh:minInclusive, sh:minLength, sh:node, sh:nodeKind,
 		sh:not, sh:or, sh:pattern, sh:property, sh:qualifiedMaxCount, sh:qualifiedMinCount, sh:qualifiedValueShape,
-		sh:qualifiedValueShape, sh:qualifiedValueShapesDisjoint, sh:qualifiedValueShapesDisjoint, sh:uniqueLang, sh:xone ;
+		sh:qualifiedValueShape, sh:qualifiedValueShapesDisjoint, sh:qualifiedValueShapesDisjoint, sh:uniqueLang, sh:xone ,
+		sh:minListLength, sh:maxListLength, sh:uniqList ;
```
jeswr:
```suggestion
		sh:minListLength, sh:maxListLength, sh:uniqueMembers ;
```

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -226,7 +227,24 @@ shsh:ShapeShape
...
+	sh:property [
+		sh:path sh:maxListLength ;
+		sh:datatype xsd:integer ;       # maxListLength-datatype
+		sh:maxCount 1 ;                 # maxListLength-maxCount
+		sh:minInclusive 0 ;             # maxListLength-minInclusive
+	] ;
+	sh:property [
+		sh:path sh:uniqueMembers ;
+		sh:datatype xsd:boolean ;       # uniqList-datatype
+		sh:maxCount 1 ;                 # uniqList-maxCount
```
jeswr:
```suggestion
		sh:datatype xsd:boolean ;       # uniqueMembers-datatype
		sh:maxCount 1 ;                        # uniqueMembers-maxCount
```

HolgerKnublauch (approved):

jeswr:
Thanks for the fixes @HolgerKnublauch

In case needed as future provenance for others: note that https://github.com/w3c/data-shapes/pull/416/commits/58daff683e4d0c3c200018051082235fac217154 makes it such that if a list is malformed - the `sh:value` should always be the head of the list - _not_ the first malformed node.

PR #417, opened 2025-06-28
Issue 415: Use `sh:memberShape` in SHACL SHACL

### LINKED ISSUE #415: Apply `sh:memberShape` to SHACL-SHACL
`sh:memberShape` can be used in shacl-shacl to:
 - Replace the existing custom list constraints
 - Make the constraints more specific https://github.com/w3c/data-shapes/pull/403#issuecomment-3009809900

I will implement this once #403 is landed

jeswr:
Closes #415

### DIFF
--- a/shacl12-vocabularies/shacl-shacl.ttl
+++ b/shacl12-vocabularies/shacl-shacl.ttl
@@ -74,7 +74,7 @@ shsh:ShapeShape
 		sh:ignoredProperties, sh:in, sh:languageIn, sh:lessThan, sh:lessThanOrEquals, sh:maxCount, sh:maxExclusive,
 		sh:maxInclusive, sh:maxLength, sh:memberShape, sh:minCount, sh:minExclusive, sh:minInclusive, sh:minLength, sh:node, sh:nodeKind,
 		sh:not, sh:or, sh:pattern, sh:property, sh:qualifiedMaxCount, sh:qualifiedMinCount, sh:qualifiedValueShape,
-		sh:qualifiedValueShape, sh:qualifiedValueShapesDisjoint, sh:qualifiedValueShapesDisjoint, sh:uniqueLang, sh:xone ,
+		sh:qualifiedValueShape, sh:qualifiedValueShapesDisjoint, sh:qualifiedValueShapesDisjoint, sh:uniqueLang, sh:xone,
 		sh:minListLength, sh:maxListLength, sh:uniqueMembers ;
 
 	sh:targetObjectsOf sh:memberShape ; # memberShape-node
@@ -123,14 +123,16 @@ shsh:ShapeShape
 		sh:maxCount 1 ;                 # deactivated-maxCount
 		sh:datatype xsd:boolean ;       # deactivated-datatype
 	] ;
-
 	sh:property [
 		sh:path sh:and ;
-		sh:node shsh:ListShape ;        # and-node
+		sh:memberShape shsh:ShapeShape ;   # and-memberShape
 	] ;
 	sh:property [
 		sh:path sh:class ;
-		sh:nodeKind sh:IRI ;            # class-nodeKind
+		sh:or (
+			[ sh:nodeKind sh:IRI ]
+			[ sh:memberShape [ sh:nodeKind sh:IRI ] ]
+		) ;            # class-nodeKind
 	] ;
 	sh:property [
 		sh:path sh:closed ;
@@ -139,16 +141,16 @@ shsh:ShapeShape
 	] ;
 	sh:property [
 		sh:path sh:ignoredProperties ;
-		sh:node shsh:ListShape ;        # ignoredProperties-node
-		sh:maxCount 1 ;                 # multiple-parameters
-	] ;
-	sh:property [
-		sh:path ( sh:ignoredProperties [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
-		sh:nodeKind sh:IRI ;            # ignoredProperties-members-nodeKind
+		sh:memberShape [ sh:nodeKind sh:IRI ] ; # languageIn-members-datatype
+		sh:minListLength 1 ;                    # languageIn-members-datatype
+		sh:maxCount 1 ;                         # multiple-parameters
 	] ;
 	sh:property [
 		sh:path sh:datatype ;
-		sh:nodeKind sh:IRI ;            # datatype-nodeKind
+		sh:or (
+			[ sh:nodeKind sh:IRI ]
+			[ sh:memberShape [ sh:nodeKind sh:IRI ] ]
+		) ;            # datatype-nodeKind
 		sh:maxCount 1 ;                 # datatype-maxCount
 	] ;
 	sh:property [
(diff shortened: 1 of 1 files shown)


shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -137,16 +141,16 @@ shsh:ShapeShape
...
 	sh:property [
 		sh:path sh:ignoredProperties ;
-		sh:node shsh:ListShape ;        # ignoredProperties-node
-		sh:maxCount 1 ;                 # multiple-parameters
-	] ;
-	sh:property [
-		sh:path ( sh:ignoredProperties [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
-		sh:nodeKind sh:IRI ;            # ignoredProperties-members-nodeKind
+		sh:memberShape [ sh:nodeKind sh:IRI ] ; # ignoredProperties-memberShape
+		sh:minListLength 1 ;                    # ignoredProperties-minListLength
```
ajnelson-nist: Added in #416 .

jeswr: This is defining a constraint on `sh:ignoredProperties` which is not added in #416. It is not defining the `sh:memberShape` or `sh:minListLength` constraint which is what is done in #416. (I assume this was already observed which is why the comment is resolved).

ajnelson-nist: Apologies, I thought I highlighted just the predicate, but apparently I tagged the whole triple.  I understand 416 added `sh:minListLength`, I'd just missed that 416 had been merged, making the concept available.

So, we're on the same page for this line.  All's well.

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -161,17 +163,15 @@ shsh:ShapeShape
 	] ;
 	sh:property [
 		sh:path sh:in ;
-		sh:maxCount 1 ;                 # in-maxCount		
-		sh:node shsh:ListShape ;        # in-node
+		sh:maxCount 1 ;                     # in-maxCount
+		sh:minListLength 1 ;                # in-minListLength
+		sh:memberShape [ a sh:NodeShape ] ; # in-memberShape
```
jeswr:
```suggestion
```

Removing as `minListLength` can now be declared without needing to have `sh:memberShape` present

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -346,11 +346,7 @@ shsh:ShapesListShape
 	sh:targetObjectsOf sh:and ;         # and-members-node
 	sh:targetObjectsOf sh:or ;          # or-members-node
 	sh:targetObjectsOf sh:xone ;        # xone-members-node
-	sh:property [
-		sh:path ( [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
-		sh:node shsh:ShapeShape ;
-	] .
-
+	sh:memberShape shsh:ShapeShape .
```
jeswr, line 349:
```suggestion
	sh:targetObjectsOf sh:memberShape ;
	sh:memberShape shsh:ShapeShape .
```

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -161,17 +163,14 @@ shsh:ShapeShape
...
 		sh:path sh:languageIn ;
-		sh:maxCount 1 ;                 # languageIn-maxCount
-		sh:node shsh:ListShape ;        # languageIn-node
-	] ;
-	sh:property [
-		sh:path ( sh:languageIn [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
-		sh:datatype xsd:string ;        # languageIn-members-datatype
+		sh:maxCount 1 ;                             # languageIn-maxCount
+		sh:minListLength 1 ;                        # languageIn-minListLength
+		sh:memberShape [ sh:datatype xsd:string ] ; # languageIn-memberShape
```
ajnelson-nist: The document-anchor reference should revert to `languageIn-members-datatype`; `languageIn-memberShape` doesn't exist in the current Core document's source.

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -295,7 +294,7 @@ shsh:ShapeShape
 	] ;
 	sh:property [
 		sh:path sh:xone ;
-		sh:node shsh:ListShape ;        # xone-node
+		sh:memberShape shsh:ShapeShape ; # xone-memberShape
```
ajnelson-nist, line 297: Interesting.  The test has caught up with an uncaught corner case of the spec.

[Section 4.6.4](https://www.w3.org/TR/shacl/#XoneConstraintComponent) of the 1.0 spec., consistent with today's draft state, reads:

> `sh:xone` specifies the condition that each value node conforms to _exactly one_ of the provided shapes.

However, the remainder of that section doesn't talk about whether `sh:xone` can or can't be empty.  This shape in SHACL-SHACL is consistent with that lack of detail, but the validation for a user with any targeted nodes would always fail, because there isn't exactly one match.

Thinking about it a bit, this seems like it would be a backwards-incompatible change to add a `sh:minListLength 1` here that fails a SHACL-SHACL run, because the `sh:xone` subject could be defined with no targets and sneak in as valid against SHACL-SHACL of SHACL 1.0.

Thoughts on whether we try fixing this now, or adding a warning-severity shape in SHACL-SHACL?

ajnelson-nist (commented):
I found one minor thing to fix, and highlighted a spec. incongruity.  I'd like feedback on what to do about the incongruity before I approve.

shacl12-vocabularies/shacl-shacl.ttl
jeswr: I would suggest adding text to the 1.2 spec requiring at least one shape - and reflecting that here with a minListLength of 1.

shacl12-vocabularies/shacl-shacl.ttl
jeswr: Good point - note that for suggested changes like this one you can propose the suggestion directly in the GH UI (see screenshot below) which enables them to be directly committed :)

image

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -139,16 +141,16 @@ shsh:ShapeShape
...
 	sh:property [
 		sh:path sh:ignoredProperties ;
-		sh:node shsh:ListShape ;        # ignoredProperties-node
-		sh:maxCount 1 ;                 # multiple-parameters
-	] ;
-	sh:property [
-		sh:path ( sh:ignoredProperties [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
-		sh:nodeKind sh:IRI ;            # ignoredProperties-members-nodeKind
+		sh:memberShape [ sh:nodeKind sh:IRI ] ; # ignoredProperties-memberShape
+		sh:minListLength 1 ;                    # ignoredProperties-minListLength
```
jeswr:
```suggestion
		sh:memberShape [ sh:nodeKind sh:IRI ] ; # languageIn-members-datatype
		sh:minListLength 1 ;                    # languageIn-members-datatype
```

shacl12-vocabularies/shacl-shacl.ttl
ajnelson-nist:
```suggestion
		sh:memberShape [ sh:datatype xsd:string ] ; # languageIn-members-datatype
```

ajnelson-nist: **Thank you!**  That button has hidden from me very well in plain sight for far too long.

shacl12-vocabularies/shacl-shacl.ttl
ajnelson-nist: Drafting an explanatory comment, I now think this is better left delegated to a follow-on Issue.  Please leave this thread unresolved for reference.

ajnelson-nist (approved):

PR #422, opened 2025-07-14
Fixed a test case

HolgerKnublauch:


jeswr (approved):
Don't know how I managed to miss that ... thanks!

Note that a new `.DS_Store` file is added in this PR as well. Please remove before merging if this is not intended.

PR #425, opened 2025-07-14
Fixed sourceConstraintComponent and applied more modern ex: namespace

HolgerKnublauch:
I ran into those while implementing the support for this new constraint type in TopBraid.


ajnelson-nist (approved):
Reviewing as non-Editor: Looks good to me.

mgberg (commented):
Looks good to me.

PR #427, opened 2025-07-15
Issue 426

### LINKED ISSUE #426: SPARQL: Drop support for $shapesGraph and $currentShape as pre-bound variables
https://www.w3.org/TR/shacl/#sparql-constraints-prebound stated that SPARQL engines MAY support the optional variables $shapesGraph and $currentShape. This is reinforced by test case shapesGraph-001.ttl (which is incorrect because not all implementations need to support these variables).

I suggest we delete these variables altogether for 1.2. Having an optional feature such may cause Failures is not a good idea. Also I don't think this was used much by anyone. The original idea was to have them in the spec so that all Core constraint components can be expressed via SPARQL, but that didn't work anyway.

HolgerKnublauch:
Closes #426

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-426/sparql12-sparql/index.html)


shacl12-sparql/index.html
```diff
@@ -1059,42 +1058,13 @@ <h3>Validation with SPARQL-based Constraints</h3>
 					</div>
 				</div>
 				<section id="sparql-constraints-prebound">
-					<h4>Pre-bound Variables in SPARQL Constraints ($this, $shapesGraph, $currentShape)</h4>
+					<h4>The pre-bound Variable $this in SPARQL Constraints</h4>
```
```suggestion
					<h4>Pre-bound variable $this in SPARQL Constraints</h4>
```

afs:
There is no rendered document.

Not just this PR -- this one is relative simple but even so there are changes separated by 100's of lines.

Reviewing generally is made harder by the lack of a rendered diff. 

Looking at people's reviewing shows that looking at changes only through a git diff is error prone. It does not give the context of the surrounding unchanged text.

afs (approved):

shacl12-sparql/index.html
afs: Preferred change.

afs:
Please put the change in the git log message so "git log" is useful.

PR #432, opened 2025-07-17
Initial contribution to define Node Expressions library

HolgerKnublauch:
This is an initial contribution from me to the Node Expressions document, introducing the definitions of the supported Node Expression types that were mostly already present in SHACL-AF. This isn't complete but hopefully a starting point that gets the ball rolling. Some details are missing but we mainly need to add examples. I copied some of the latest JS support code and styles from the Core spec.

I have added myself as editor too, if that's OK.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/hk-node-expr-start/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -486,6 +486,8 @@
 						<dfn data-cite="shacl12-core#dfn-shacl-instance" data-lt="shacl instance">SHACL instance</dfn>,
 						<dfn data-cite="shacl12-core#dfn-shacl-subclass" data-lt="shacl subclass">SHACL subclass</dfn>,
 						<dfn data-cite="shacl12-core#dfn-shacl-type" data-lt="shacl type">SHACL type</dfn>,
+						<dfn data-cite="shacl12-core#dfn-shacl-list" data-lt="shacl list|shacl lists">SHACL list</dfn>,
+						<dfn data-cite="shacl12-core#dfn-members" data-lt="members">members</dfn>,
 						<dfn data-cite="shacl12-core#dfn-well-formed" data-lt="well-formed">well-formed</dfn>,
 						as defined in the SHACL 1.2 Core specification [[!shacl12-core]].
 					</div>
@@ -573,6 +575,7 @@
 				Such blank nodes are only <a>well-formed</a> if they are not the <a>subject</a> of any other
 				triples, or where any of these properties are used more than once.
 				Also, the tables may list SHACL constraints that the property values are required to conform to.
+				In the tables, mandatory properties are rendered in <b>bold face</b>.
 			</p>
 			<section id="library-constants">
 				<h3>Constant Node Expressions</h3>
@@ -640,7 +643,7 @@
 								</thead>
 								<tbody>
 									<tr>
-										<td><code>sh:var</code></td>
+										<td><b><code>sh:var</code></b></td>
 										<td>
 											<code>sh:datatype xsd:string</code><br />
 											<code>sh:minLength 1</code></li>
@@ -694,7 +697,51 @@ ex:Person-loves
 
 				<section id="ListExpression">
 					<h3>List Expressions</h3>
-					<p class="todo">TODO</p>
+					<p class="syntax">
+						<span data-syntax-rule="ListExpression-syntax">
+							A <a>blank node</a> that is the <a>subject</a> of the following properties
+							is called a <dfn>list expression</dfn> with the <a>function name</a> <code>sh:ListExpression</code>:
+							<table class="term-table">
+								<thead>
+									<th>Property</th>
+									<th>Constraints</th>
+									<th>Description</th>
+								</thead>
+								<tbody>
+									<tr>
+										<td><b><code>rdf:first</code></b></td>
+										<td>
+										</td>
+										<td>
+											The first member of the list.
+										</td>
+									</tr>
+									<tr>
+										<td><b><code>rdf:rest</code></b></td>
+										<td>
+											Must be a <a>well-formed</a> <a>SHACL list</a>.
+										</td>
+										<td>
+											The rest of the list, e.g. <code>rdf:nil</code>.
+										</td>
+									</tr>
+								</tbody>
+							</table>
+						</span>
+					</p>
+					<div class="def" id="ListExpression-evaluation">
+						<div class="def-header">EVALUATION OF LIST EXPRESSIONS</div>
+						<p>
+							The <a>output nodes</a> of a <a>list expression</a> are the <a>members</a> of the <a>SHACL list</a>.
+						</p>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						Note that <code>rdf:nil</code> itself is not a <a>list expression</a> because it will be interpreted
+						as a <a>IRI expression</a>.
+						As a result, all <a>well-formed</a> list expressions have at least one <a>member</a>.
+					</p>
+					<p class="todo">TODO: Add example</p>
 				</section>
 
 				<section id="PathExpression">
@@ -711,9 +758,9 @@ ex:Person-loves
 								</thead>
 								<tbody>
 									<tr>
-										<td><code>sh:path</code></td>
+										<td><b><code>sh:path</code></b></td>
 										<td>
-											must be a <a>well-formed</a> <a>SHACL property path</a>
+											Must be a <a>well-formed</a> <a>SHACL property path</a>.
 										</td>
 										<td>
 											The path to get the values from.
@@ -722,7 +769,7 @@ ex:Person-loves
 									<tr>
 										<td><code>sh:nodes</code></td>
 										<td>
-											optional, must be a <a>well-formed</a> <a>node expression</a>
+											Optional, must be a <a>well-formed</a> <a>node expression</a>.
 										</td>
 										<td>
 											A node expression producing the <a>focus nodes</a>, defaulting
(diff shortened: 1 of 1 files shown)


afs:
Presumably:
"Basic Node" => "Blank Node"

I think it would be helpful to have section that is the definitions of terminology and how the different forms work. This is in more detail than in SHACL Core and picks up on #322.

Then have a section that is a library reference. The entries in the library section can say a simple "this is a <insert ref>" to pick the style they are, and proceed with the specifics and examples. It is also a framework for custom node expressions without needing to extend the code of an executor.

The rules/inference work will benefit if there are definitions for the features it needs, which, below are first and foremost list argument functions (single valued), some, not all, single valued named argument expressions. Maybe aggregates (a later update to inference because they are beyond datalog and need some technical work to assess the impact).

To better reflect #322:

```
2. Node Expressions Model
2.1 Constants
  2.1.1 IRI
  2.1.2 Literal
2.2 Variable : `[ sh:var "name" ]`
2.3 Functions
  2.3.1 List parameters functions (single valued)
           Arguments are evaluated then passed to the function
  2.3.2 Named argument expressions (these may take expression arguments)
    2.3.2.1 single valued named argument expressions
            aggregates (list of one)
    2.3.2.2 list-valued List-valued named argument expressions
         distinct, intersection/union/difference, limit/offset, orderBy

3. Node Expressions Library
   One inner section per concrete description
3.1 List Argument functions
3.2 Named argument expressions
   3.2.1 Aggregates
   3.2.3 List-valued expressions
3.3 Path expressions (multi-valued)
3.4 Specials
  3.4.1 sh:if - lazy evaluation of arguments but otherwise is a Named argument expression
  3.4.2 sh:path - the argument has an interpretation as a property path.
    Some SPARQL are "special" - Logical AND, logical OR, COALESCE if these are available.
```

Are aggregates a fixed set?  For example, there are other statistical aggregates like standard deviation and sample of the population variants which is why there is a specific form (2.3.2.1).

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+				Such blank nodes are only <a>well-formed</a> if they are not the <a>subject</a> of any other
+				triples, or where any of these properties are used more than once.
+				Also, the tables may list SHACL constraints that the property values are required to conform to.
+				In the tables, mandatory properties are rendered in <b>bold face</b>.
+			</p>
+			<section id="library-constants">
+				<h3>Constant Node Expressions</h3>
+				<p>
+					The <a>node expression functions</a> in this section were already introduced in the SHACL Core
+					specificiation but are repeated here to keep this document self-contained.
```
```suggestion
					The <a>node expression functions</a> in this section were introduced in the SHACL Core
					specification but are repeated here to keep this document self-contained.
```

TallTed (changes requested):
Note to self: continue reviewing at line 595

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+				</p>
+				
+				<section id="IRIExpression">
+					<h3>IRI Expressions</h3>
+					<p>
+						A <a>node expression</a> that is a <a>IRI</a> is called an <dfn>IRI expression</dfn> with the <a>function name</a>
```
```suggestion
						A <a>node expression</a> that is an <a>IRI</a> is called an <dfn>IRI expression</dfn> with the <a>function name</a>
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+							If <code>sh:nodes</code> is not given, <code>nodes</code> is the list consistent of exactly the <a>focus node</a>.
+							Let <code>N</code> be the nodes produced by <code>evalExpr(nodes, focusGraph, focusNode, scope)</code>.
+							The <a>output nodes</a> of the <a>path expression</a> are the list of <a>nodes</a> produced by concatenating
+							the <a>value nodes</a> of the <code>path</code> at each <a>node</a> in <code>N</code>.
+							<span class="todo">TODO: Clarify if those can contain duplicates.</span>
+						</p>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						The following example illustrates the use of a <a>path expression</a> to compute the values
```
```suggestion
						The following example illustrates the use of a <a>path expression</a> to compute the value
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+										<td>
+											A node expression that must return <code>true</code> as its only <a>output node</a>
+											so that the <code>sh:then</code> branch is returned, otherwise <code>sh:else</code>.
+										</td>
+									</tr>
+									<tr>
+										<td><code>sh:then</code></td>
+										<td>
+											A <a>well-formed</a> <a>node expression</a>.
+											Optional, if <code>sh:else</code> exists.
```
```suggestion
											Optional but at least one of <code>sh:then</code> or <code>sh:else</code> is required.
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+										</td>
+										<td>
+											The <a>node expression</a> that is returned when the <code>sh:if</code> evaluated to <code>[true]</code>.
+										</td>
+									</tr>
+									<tr>
+										<td><code>sh:else</code></td>
+										<td>
+											A <a>well-formed</a> <a>node expression</a>.
+											Optional, if <code>sh:then</code> exists.
```
```suggestion
											Optional but at least one of <code>sh:then</code> or <code>sh:else</code> is required.
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+								</thead>
+								<tbody>
+									<tr>
+										<td><b><code>sh:if</code></b></td>
+										<td>
+											A <a>well-formed</a> <a>node expression</a>.
+										</td>
+										<td>
+											A node expression that must return <code>true</code> as its only <a>output node</a>
+											so that the <code>sh:then</code> branch is returned, otherwise <code>sh:else</code>.
```
```suggestion
											A node expression that must return <code>true</code> or <code>false</code> as its only <a>output node</a>.
											If the result is <code>true</code>, the <code>sh:then</code> branch is returned, otherwise <code>sh:else</code>.
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+								</tbody>
+							</table>
+						</span>
+					</p>
+					<div class="def" id="IntersectionExpression-evaluation">
+						<div class="def-header">EVALUATION OF INTERSECTION EXPRESSIONS</div>
+						<p>
+							Let <code>members</code> be the <a>members</a> of the <a>value</a> of <code>sh:intersection</code> in the <a>intersection expression</a>.
+							The <a>output nodes</a> of the <a>intersection expression</a> are the list of <a>nodes</a> that are in all <a>output nodes</a>
+							for each <a>node expression</a> <code>NE</code> in <code>members</code>, using <code>evalExpr(NE, focusGraph, focusNode, scope)</code>.
```
```suggestion
							The <a>output nodes</a> of the <a>intersection expression</a> are the <a>nodes</a> that form the intersection of the <a>output nodes</a>
							produced by each <a>node expression</a> <code>NE</code> in <code>members</code>, using <code>evalExpr(NE, focusGraph, focusNode, scope)</code>.
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+							</table>
+						</span>
+					</p>
+					<div class="def" id="UnionExpression-evaluation">
+						<div class="def-header">EVALUATION OF UNION EXPRESSIONS</div>
+						<p>
+							Let <code>members</code> be the <a>members</a> of the <a>value</a> of <code>sh:union</code> in the <a>union expression</a>.
+							The <a>output nodes</a> of the <a>union expression</a> are the concatenation of all <a>output nodes</a>
+							for each <a>node expression</a> <code>NE</code> in <code>members</code>, using <code>evalExpr(NE, focusGraph, focusNode, scope)</code>.
+							The order is preserved, evaluating the <code>members</code> from left to right and keeping the order of each list.
```
```suggestion
							The order is preserved, evaluating the <code>members</code> from left to right and keeping the order of each set of output nodes.
```

shacl12-node-expr/index.html
```diff
@@ -592,13 +555,779 @@ <h3>Document Conventions</h3>
...
+							</table>
+						</span>
+					</p>
+					<div class="def" id="OffsetExpression-evaluation">
+						<div class="def-header">EVALUATION OF OFFSET EXPRESSIONS</div>
+						<p>
+							Let <code>offset</code> be the <a>value</a> of <code>sh:offset</code>
+							and <code>nodes</code> be the <a>value</a> of <code>sh:nodes</code> in the <a>offset expression</a>.
+							Let <code>N</code> be the <a>output nodes</a> of <code>evalExpr(nodes, focusGraph, focusNode, scope)</code>.
+							The <a>output nodes</a> of the <a>offset expression</a> are the a>nodes</a> in <code>N</code>
```
```suggestion
							The <a>output nodes</a> of the <a>offset expression</a> are the <a>nodes</a> in <code>N</code>
```

recalcitrantsupplant (commented):
Rephrasing suggestions and typo fixes

shacl12-node-expr/index.html
HolgerKnublauch: In fact it MAY return anything else except true as well, e.g. the empty list. I need to reformulate this better, but it's not just false.

HolgerKnublauch:
@afs thanks for your input, which I have applied in parts.

- I didn't move sh:var into its own section because they are also just blank nodes.
- I didn't yet separate the library between list and named argument categories, because so far all of them are named argument style and I am not convinced the syntax should be the defining categorization. Also the distinction between "specials" like IF and COALESCE doesn't seem very clear and is specific to each function, not a general pattern.

We can easily re-arrange them as we collect more experience with the definitions.

shacl12-node-expr/index.html
```diff
@@ -618,6 +608,42 @@ <h3>Literal Expressions</h3>
...
+			<h2>Node Expressions Library</h2>
+			<p>
+				This section defines all <a>node expression functions</a> that are built into SHACL engines
+				that implement this specification.
+			</p>
+			<p>
+				The syntax definitions of node expression functions that are based on <a>blank nodes</a>
+				typically use a table of properties that these blank nodes can or must have.
+				Such blank nodes are only <a>well-formed</a> if they are not the <a>subject</a> of any other
+				triples, or where any of these properties are used more than once.
```
TallTed: Trying to make this sentence gramatically sensible.
```suggestion
				Such blank nodes are only <a>well-formed</a> when they are not the <a>subject</a> of any other
				triples, and when none of these properties is used more than once.
```

shacl12-node-expr/index.html
```diff
@@ -618,6 +608,42 @@ <h3>Literal Expressions</h3>
...
+				Also, the tables may list SHACL constraints that the property values are required to conform to.
```
```suggestion
				The tables may also list SHACL constraints with which the property values are required to conform.
```

recalcitrantsupplant (approved):
approving as initial spec

TallTed (approved):

afs:
The section "List Expressions" is something different to one of the two kind of expression in #322.

We need a form that is `[ :predicate SHACL_List ]`. That is what I understood as a "List parameters function".
e.g `[ :add ( 1 2) ]`, `[ :add ( 1 [ sh:var "X"] ) ]`. 

The spec text defines `( ???? )` or the tail thereof. Not sure what that is, is it trying to be list valued?

(
Or do you really want lisp s-expressions! `( :add 1 2 )`. The `[ :function (args) ]` syntax is IMO more natural because it suggestions  "function(1, 2)".
)

I think that the definition should also be closed - the blank node can't the subject of any other predicate.

```
[ :add (1 2) ; :multiply ( 4 5 ) ]
```
and 

```
[ :add (1 2) ; :add ( 6 7 ) ]
```
should be are bad syntax.

HolgerKnublauch:
@afs yes this is another topic from list parameter expressions. Those are introduced in this PR:

https://github.com/w3c/data-shapes/pull/441

and I believe capture the essence of what you are asking for here.

afs:
It's multivalued and the example isn't a function.

> I didn't move sh:var into its own section because they are also just blank nodes.

`[ sh:var "name"]` is special because it's evaluation is context sensitive. It will need to be called out as such.

PR #441, opened 2025-07-25
Node Expression Intro and Syntax overview chapters

HolgerKnublauch:
This is again meant as a starting point. It will be IMHO easier to discuss details and next steps when the basic outline is present.

Note that the rendering produces a couple of broken references, but those will disappear when my parallel PR is merged into main. Some examples refer to sh:filterShape and sh:instancesOf which are not yet merged in.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/hk-node-expr-syntax/shacl12-node-expr/index.html)

@afs do these definitions reflect your understanding of list-vs-named parameters too?

PR #446, opened 2025-07-28
Late updates from #441

TallTed:


HolgerKnublauch (approved):
Perfect. I appreciate the double effort and will merge it in next, as it is uncontroversial editorial changes only.

Issue #431, opened 2025-07-17
Tag tests for optional features and define engine parameters for tests


bergos:
If we define tests for optional features, those tests will fail if the implementation doesn't support the features. It would be necessary to tag the tests for optional features for filtering.

Some tests may require running an engine with specific parameters. The parameters must be attached to the tests for automatic testing.

Do we want to support such tests? Which features would be affected (current, upcoming, past), and which tests?

- #409 feat: #192 add severity levels that are not violations
  - treat sh:Info and sh:Warning non-violating based on options passed to the engine.
- #426 SPARQL: Drop support for $shapesGraph and $currentShape as pre-bound variables

ajnelson-nist:
Another use-case for this Issue has come up in #185 , whose resolution ( #447 ) adds this paragraph (words copied from current unmerged state):

> SHACL processors SHOULD offer a parameter <code>subClassOfInShapesGraph</code> that, if set to <code>true</code>, should alter the definition of <a>SHACL Type</a> so that the <code>rdfs:subClassOf</code> triples are queried from the <a>shapes graph</a> in addition to the <a>data graph</a>.

@bergos , your configuration suggestion seems like a way to support a test for this SHOULD, in response to @TallTed 's [comment](https://github.com/w3c/data-shapes/pull/447#issuecomment-3150802441), quoted here:

> SHOULD statements are as deserving of tests as MUST statements. Failing a SHOULD test should not result in failure of conformance report, but should result in reporting that failure of SHOULD such that deployers can differentiate between two implementations that both satisfy all MUST statements but satisfy different (whether counts or instances) SHOULD statements.

PR #447, opened 2025-07-29
#185: Added subClassOfInShapesGraph parameter

### LINKED ISSUE #185: Where should my ontology go? Data graph versus shapes graph
Originally posed over at <https://github.com/w3c/data-shapes/issues/155>; also see the comments by others over there.

### Observation

According to the SHACL standard, two graphs are relevant for validation: the data graph and the shapes graph. The ontology should be part of the **data graph**:

> The data graph is expected to include all the ontology axioms related to the data and especially all the rdfs:subClassOf triples in order for SHACL to correctly identify class targets and validate Core SHACL constraints.

This seems counter-intuitive to me, since I associate the ontology more with the shapes graph. For example, a shapes graph can `owl:import` an ontology.

### Example

To illustrate my unease, let's take the following **data graph**:

```turtle
prefix id: <https://example.com/>
prefix foaf: <http://xmlns.com/foaf/0.1/>

id:john a foaf:Person.
```

And the following **shapes graph**:

```turtle
prefix foaf: <http://xmlns.com/foaf/0.1/>
prefix sh: <http://www.w3.org/ns/shacl#>

[] sh:targetClass foaf:Agent;
   sh:property
     [ sh:minCount 1;
       sh:path foaf:name ].
```

Adding the following **ontology graph** is crucial, otherwise we cannot invalidate the data graph which is missing a `foaf:name` statement:

```turtle
prefix foaf: <http://xmlns.com/foaf/0.1/>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>

foaf:Person rdfs:subClassOf foaf:Agent.
```

### Use case

I have a specific use case where this comes up: in TriplyETL we stream though the instance data. The stream passes along millions of small data graphs. For each of these data graphs, we have to add the ontology before the data graph can be validated in-stream. In this use case, it makes more sense to add the ontology to the shapes graph once, and use that same shapes graph to validate all data graphs that pass by.

### Expected

I expect either of the following:

- The SHACL standard requires that the ontology graph is added to the shapes graph, and not to the data graph.
- The SHACL standard requires that the ontology graph is added to the data graph and/or the shapes graph.

HolgerKnublauch:
Closes #185

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-185/shacl12-core/index.html)

Added Andy and Ashley as reviewers too, to check if this is implementable from their engines point of view. The main difference is that instead of

```
?instance rdf:type/rdfs:subClassOf $class .
```

it would become

```
?instance rdf:type ?type .
GRAPH $shapesGraph {
    $type rdfs:subClassOf* ?class
}
```

And it's an optional feature...


shacl12-core/index.html
```diff
@@ -2702,6 +2710,27 @@ <h3>Data Graph</h3>
...
+					of <a>SHACL type</a> to determine whether a <a>node</a> is a <a>SHACL instance</a> of a given class.
+					By default, this is determined by looking up <code>rdfs:subClassOf</code> and <code>rdf:type</code> <a>triples</a>
+					in the <a>data graph</a>.
+					However, in some cases this is impractical as <code>rdfs:subClassOf</code> triples are often stored as part
+					of the class/shape definitions and not the instance data.
+				</p>
+				<p>
+					SHACL processors SHOULD offer a parameter <code>subClassOfInShapesGraph</code> that, if set to <code>true</code>
+					should alter the definition of <a>SHACL Type</a> so that the <code>rdfs:subClassOf</code> triples are queried
+					from the <a>shapes graph</a> instead of the <a>data graph</a>.
```
jeswr:
```suggestion
					from the <a>shapes graph</a> in addition to the <a>data graph</a>.
```

Why omit consideration of `rdfs:subClassOf` statements in the data graph when this mode is enabled?

jeswr:
At risk of overcomplicating things - would it be worth `inferenceDataInShapesGraph` rather than `subClassOfInShapesGraph`; and using this feature to enable inclusion of other data for inference rules in the shapes graph?

HolgerKnublauch:
@jeswr there is already sh:entailment to activate other inferencing. And it is under complete control of the user which graph to pass into the engine as shapes graph. That graph may have been pre-processed with OWL to infer additional rdfs:subClassOf axioms.

As the above IMHO doesn't need a formal parameter to the engine, I wanted to however spell out the new flag because it does need a parameter as it changes how some terms (SHACL Instance) behave.

shacl12-core/index.html
HolgerKnublauch: That COULD work too. I went with the simpler solution because querying a union graph would be more complex, and make optimizations harder. If we know that we only look in the shapes graph, all class closures could be pre-computed into HashSets, instead of having to walk the hierarchy over and over.

Also, what about undesirable cases where some subClassOf triples are in the data graph and others in the shapes graph. That sounds like a mess to me.

HolgerKnublauch: Are there realistic scenarios where people want to mix and match?

jeswr: > Are there realistic scenarios where people want to mix and match?

I can imagine some cases e.g. having an upper level ontology in your shapes graph and then lower level ontologies in your data graph. I have no data on whether these appear in the wild.

afs:
> SHACL implementations can be parameterized

Where's the list of all parameters?

The whole "implementation controls" area seems to sit uneasily between implementation and standardization.

afs:
Not sure if the discussion is best here or on #185.

Added a general comment on the issue -- https://github.com/w3c/data-shapes/issues/185#issuecomment-3134014995

HolgerKnublauch:
I am not aware of other flags or parameters for the engine, but they could be grouped into an appendix if we define more.

YoucTagh (approved):
I am approving this pull request (PR) because I favour using *instead of* over *in addition to*, as the latter would add more complexity. Also, if we gather other flags or parameters, we can add them in an appendix in another PR.

afs:
No test?

HolgerKnublauch:
@afs correct, no test because the feature is optional ("SHOULD") and we don't have any mechanism in the test case framework to set this flag.

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
 				</p>
 			</section>
+
+			<section id="subClassOfInShapesGraph">
+				<h3>Graph for rdfs:subClassOf Triples</h3>
+				<p>
+					Some features of SHACL (such as 
+					<a href="#ClassConstraintComponent"></a>, 
+					<a href="#targetClass"></a> and 
+					<a href="#implicit-targetClass"></a>) rely on the notion
```
ajnelson-nist: Are these textless anchors intentional?

shacl12-core/index.html
ajnelson-nist: I can see use cases coming from defining higher-level category classes that join together classes in the data graph.

Use case:

* Data graph goes as ontologically far as classifying `ex:Cat` and `ex:Dog` and putting them under `ex:Mammal`.
* Shapes graph includes an overarching class `ex2:WarmBloodedAnimal`.

Use case:

* Data graph is a union of credit card receipts and cash receipts.
* Shapes graph includes an overarching class `ex3:FinancialTransaction`.

I think mix-and-match should be supported by default.  I'm late coming to this PR and Issue, and a little curious about the motivation for it being "Opt-in" to look for `rdfs:subClassOf` in the shapes graph.  Just the union-performance matter?

shacl12-core/index.html
```diff
@@ -1545,6 +1548,11 @@ <h4>Class-based Targets (sh:targetClass)</h4>
 								</div>
 							</div>
 						</aside>
+						<p>
+							Note that the <code>rdfs:subClassOf</code> triples may be queried from the <a>shapes graph</a>,
+							see <a href="#subClassOfInShapesGraph"></a>, in which case the <code>rdfs:subClassOf</code> triple
+							from the example above would not be required in the <a>data graph</a>.
```
```suggestion
							Note that the <code>rdfs:subClassOf</code> triples may be queried from the <a>shapes graph</a>
							(see <a href="#subClassOfInShapesGraph"></a>) in which case the <code>rdfs:subClassOf</code> triple
							from the example above would not be required to be in the <a>data graph</a>.
```

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
 					several related ontologies, pass all of them to the SHACL processor (together or one by one), do not rely on <code>owl:imports</code> links.
 				</p>
 			</section>
+
+			<section id="subClassOfInShapesGraph">
+				<h3>Graph for rdfs:subClassOf Triples</h3>
+				<p>
+					Some features of SHACL (such as 
+					<a href="#ClassConstraintComponent"></a>, 
+					<a href="#targetClass"></a> and
```
```suggestion
					<a href="#targetClass"></a>, and 
```

shacl12-core/index.html
TallTed: Yes. They are making use of a ReSpec feature that auto-links to `<dfn>` instances elsewhere in the document.

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
+				<p>
+					Some features of SHACL (such as 
+					<a href="#ClassConstraintComponent"></a>, 
+					<a href="#targetClass"></a> and 
+					<a href="#implicit-targetClass"></a>) rely on the notion
+					of <a>SHACL type</a> to determine whether a <a>node</a> is a <a>SHACL instance</a> of a given class.
+					By default, this is determined by looking up <code>rdfs:subClassOf</code> and <code>rdf:type</code> <a>triples</a>
+					in the <a>data graph</a>.
+					However, in some cases this is impractical as <code>rdfs:subClassOf</code> triples are often stored as part
+					of the class/shape definitions and not the instance data.
```
```suggestion
					However, this is impractical in some cases, as <code>rdfs:subClassOf</code> triples are often stored as part
					of the class and/or shape definitions and not the instance data.
```

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
+				</p>
+				<p>
+					SHACL processors SHOULD offer a parameter <code>subClassOfInShapesGraph</code> that, if set to <code>true</code>
```
```suggestion
					SHACL processors SHOULD offer a parameter <code>subClassOfInShapesGraph</code> that, if set to <code>true</code>,
```

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
+					should alter the definition of <a>SHACL Type</a> so that the <code>rdfs:subClassOf</code> triples are queried
+					from the <a>shapes graph</a> instead of the <a>data graph</a>.
+					The <code>rdf:type</code> triples are always expected in the <a>data graph</a>.
```
```suggestion
					The <code>rdf:type</code> triples are always expected to be in the <a>data graph</a>.
```

TallTed (changes requested):

TallTed:
@HolgerKnublauch — SHOULD statements are as deserving of tests as MUST statements. Failing a SHOULD test should not result in failure of conformance report, but should result in *reporting* that failure of SHOULD such that deployers can differentiate between two implementations that both satisfy all MUST statements but satisfy different (whether counts or instances) SHOULD statements.

shacl12-core/index.html
HolgerKnublauch: I have no opposition to generalizing this flag to be the union of data and shapes graph. There seem to be use cases that people need, so why not, even if it can cause some potential performance penalties.

HolgerKnublauch:
Ok, the semantics of the flag have been changed to be the union of shapes and data graphs.

Please re-check and let me know any remaining concerns.

shacl12-core/index.html
```diff
@@ -586,6 +586,9 @@ <h3>Terminology</h3>
 					<div>
 						The <dfn data-lt="type|types|SHACL type">SHACL types</dfn> of an <a>RDF term</a> in an <a>RDF graph</a> is the set of its <a>values</a> for <code>rdf:type</code> in the
 						<a>graph</a> as well as the <a>SHACL superclasses</a> of these <a>values</a> in the <a>graph</a>.
+						Note that some SHACL implementations can be parameterized so that the <code>rdfs:subClassOf</code> triples
+						that determine the <a>SHACL subclasses</a> may be queried from the <a>shapes graph</a> instead of the <a>data graph</a>.
```
ajnelson-nist:
```suggestion
						that determine the <a>SHACL subclasses</a> may be queried from the <a>shapes graph</a> in addition to the <a>data graph</a>.
```

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
+				<p>
+					Some features of SHACL (such as 
+					<a href="#ClassConstraintComponent"></a>, 
+					<a href="#targetClass"></a>, and 
+					<a href="#implicit-targetClass"></a>) rely on the notion
+					of <a>SHACL type</a> to determine whether a <a>node</a> is a <a>SHACL instance</a> of a given class.
+					By default, this is determined by looking up <code>rdfs:subClassOf</code> and <code>rdf:type</code> <a>triples</a>
+					in the <a>data graph</a>.
+					However, this is impractical in some cases, as <code>rdfs:subClassOf</code> triples are often stored as part
+					of the class and/or shape definitions and not the instance data.
```
ajnelson-nist:
```suggestion
					However, this is insufficient in some cases, as <code>rdfs:subClassOf</code> triples are often stored as part
					of the class and/or shape definitions and not the instance data.
```

shacl12-core/index.html
```diff
@@ -2717,6 +2725,27 @@ <h3>Data Graph</h3>
...
+				</p>
+				<p>
+					SHACL processors SHOULD offer a parameter <code>subClassOfInShapesGraph</code> that, if set to <code>true</code>,
+					should alter the definition of <a>SHACL Type</a> so that the <code>rdfs:subClassOf</code> triples are queried
+					from the <a>shapes graph</a> in addition to the <a>data graph</a>.
```
ajnelson-nist, line 2745: Non-blocking remark, anyone feel free to Resolve this thread: #431 may open an avenue for writing tests for this.

shacl12-core/index.html
```diff
@@ -7936,9 +7965,11 @@ <h2>Changes between SHACL 1.0 Core and SHACL 1.2 Core</h2>
 				<li>Moved SPARQL-based validators from Core to an Appendix of SHACL-SPARQL; see <a href="https://github.com/w3c/data-shapes/issues/271">Issue 271</a></li>
 				<li>Added the new constraint component <a href="#ExpressionConstraintComponent"><code>sh:expression</code></a>; see <a href="https://github.com/w3c/data-shapes/issues/357">Issue 357</a></li>
 				<li>Added the new constraint component <a href="#NodeByExpressionConstraintComponent"><code>sh:nodeByExpression</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/408">Issue 408</a></li>
+				<li>Added new <a href="#core-components-list">List constraint components</a>, see <a href="https://github.com/w3c/data-shapes/issues/391">Issue 391</a> and <a href="https://github.com/w3c/data-shapes/issues/414">Issue 414</a></li>
```
ajnelson-nist, line 7968: Wait, is this changelog line part of this PR?

ajnelson-nist (commented):
This overall looks good to me.  I left a few minor suggestions, but didn't see anything meriting blocking with the "Request changes" review.

I think I'm now understanding @afs 's question on parameters.  We have `subClassOfInShapesGraph` in this PR, and potentially one representing conformance-disallowance depending on how #453 shakes out.

Definitely if we have a second parameter to accompany `subClassOfInShapesGraph`, a table, section, or appendix should go in for indexing specification-defined parameters.

I could probably be swayed to wait to sketch that table/section, but request if the decision is to wait that an Issue be posted to track it.

afs:
> @afs correct, no test because the feature is optional ("SHOULD") and we don't have any mechanism in the test case framework to set this flag.

Tests are more than just compliance. They also perform the role of another explanation of the feature. For a large system such as SHACL, tests are valuable implementation resources.

e.g. where would constraints be affected when this optional feature is activated?

TallTed (approved):

shacl12-core/index.html
HolgerKnublauch: Yes I sneaked that in because we missed it before and it's not worth creating another ticket just for that...

jeswr (approved):

bergos (approved):

PR #451, opened 2025-07-29
#429: Have SHACL-SHACL warn about unsatisfiable usage of `sh:in` and `sh:xone`

### LINKED ISSUE #429: `sh:xone` and list length
_Originally raised in [this comment](https://github.com/w3c/data-shapes/pull/417#discussion_r2205907515)_, excerpted here:

> [Section 4.6.4](https://www.w3.org/TR/shacl/#XoneConstraintComponent) of the 1.0 spec., consistent with today's draft state, reads:
> 
> > `sh:xone` specifies the condition that each value node conforms to _exactly one_ of the provided shapes.
> 
> However, the remainder of that section doesn't talk about whether `sh:xone` can or can't be empty.  This shape in SHACL-SHACL is consistent with that lack of detail, but the validation for a user with any targeted nodes would always fail, because there isn't exactly one match.
> 
> Thinking about it a bit, this seems like it would be a backwards-incompatible change to add a `sh:minListLength 1` here that fails a SHACL-SHACL run, because the `sh:xone` subject could be defined with no targets and sneak in as valid against SHACL-SHACL of SHACL 1.0.

Should SHACL 1.2 do anything about empty `sh:xone` lists?  We have a few options:

* Update the spec to say `sh:xone` _must_ have a member, and add a `sh:minListLength 1` constraint to SHACL-SHACL with `sh:Violation` severity.
* Update the spec to say `sh:xone` _should_ have a member, and add a `sh:minListLength 1` constraint to SHACL-SHACL with a lesser severity.
* Update the spec to say `sh:xone` _should_ have a member, and do not update SHACL-SHACL.

If updating the spec, this is my first sketch of wording for the `sh:xone` spec. section, added to the `span`s of `data-syntax-rule`s a new rule `xone-members-minCount`:

> There must be at least one member of such list.

If updating SHACL-SHACL to report less-than-`sh:Violation` severity, then we could use `sh:Warning` for the shape authors that they're about to cause trouble for their shape users.  #192 goes into other severities if we end up thinking a non-violation level would be better.

A PR for this Issue should be built on:

- [x] #417
- [x] #409 (if group decides on non-violation severity)

ajnelson-nist:
Closes #429 .

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/{issue-429-list-lengths-xone-in}/shacl12-core/index.html)

This patch revises part of the implementation in PR 417.

* `in-minListLength` was not added to the Core document, but it was referenced in SHACL-SHACL.  Verbiage is added to Core in line with the text drafted for Issue 429.
* `xone-memberShape` was added as a reference in SHACL-SHACL, but doesn't exist in the Core document.  `xone-members-node` is the anchor in the Core document.  `xone-members-node` is now used.
* The `PropertyShape` for `sh:in` set a minimum required length of 1, which after discussion in Issue 429 appears to be a backward- incompatible change.  A shape paralleling the original suggestion for `sh:xone` is added to house that new restriction at the same warning- level severity.

An implementation style point: The property shapes housing `sh:minListLength` are given IRIs after sketching a test (`.../in-003.ttl`) and assuming it would be more desirable to have an IRI-identified shape in the results graph, rather than a blank node.

A testing style point: Rather than running SHACL-SHACL over the whole of a test, shapes from SHACL-SHACL are copied into some focused tests (`.../{in,xone}-003.ttl`).


jeswr (approved):
LGTM - apologies for the delay in reviewing

YoucTagh:
@ajnelson-nist, can this PR be merged?

shacl12-core/index.html
```diff
@@ -5899,6 +5899,7 @@ <h4>sh:xone</h4>
 								A <a>SHACL list</a> of shapes to validate the value nodes against.
 								<span data-syntax-rule="xone-node">Each value of <code>sh:xone</code> in a shape is a <a>SHACL list</a>.</span>
 								<span data-syntax-rule="xone-members-node">Each <a>member</a> of such list must be a <a>well-formed</a> <a>shape</a>.</span>
+								<span data-syntax-rule="xone-minListLength">There should be at least one <a>member</a> of such list.</span>
```
TallTed: If these are not intended to be [RFC2119 keywords](https://www.w3.org/guide/manual-of-style/#RFC), it's best to not use them, even in lowercase, to avoid common confusion.
```suggestion
								<span data-syntax-rule="xone-members-node">Each <a>member</a> of such list <em title="MUST in RFC 2119 context" class="rfc2119">MUST</em> be a <a>well-formed</a> <a>shape</a>.</span>
								<span data-syntax-rule="xone-minListLength">Each such list <em title="SHOULD in RFC 2119 context" class="rfc2119">SHOULD</em> have at least one <a>member</a>.</span>
```

shacl12-core/index.html
```diff
@@ -7299,6 +7300,7 @@ <h4>sh:in</h4>
 								A <a>SHACL list</a> that has the allowed values as <a>members</a>.
 								<span data-syntax-rule="in-node">Each value of <code>sh:in</code> in a shape is a <a>SHACL list</a>.</span>
 								<span data-syntax-rule="in-maxCount">A shape has at most one value for <code>sh:in</code>.</span>
+								<span data-syntax-rule="in-minListLength">There should be at least one <a>member</a> of such list.</span>
```
```suggestion
								<span data-syntax-rule="in-minListLength">Each such list <em title="SHOULD in RFC 2119 context" class="rfc2119">SHOULD</em> have at least one <a>member</a>.</span>
```

ajnelson-nist: @TallTed - I'm taking your suggestion, but updating the syntax to use the CSS class form I see in today's document state.

```suggestion
								<span data-syntax-rule="in-minListLength">Each such list <em class="rfc2119">SHOULD</em> have at least one <a>member</a>.</span>
```

shacl12-core/index.html
ajnelson-nist: @TallTed - I'm taking your suggestion, but updating the style to today's document form.

```suggestion
								<span data-syntax-rule="xone-members-node">Each <a>member</a> of such list <em class="rfc2119">MUST</em> be a <a>well-formed</a> <a>shape</a>.</span>
								<span data-syntax-rule="xone-minListLength">Each such list <em class="rfc2119">SHOULD</em> have at least one <a>member</a>.</span>
```

ajnelson-nist:
PR caught up.  Merging with @jeswr 's approval and @TallTed 's wording updates.

Issue #220, opened 2025-01-25
Provide RDFS entailment


afs:
_Originally posted by @ajnelson-nist in [#215](https://github.com/w3c/data-shapes/issues/215#issuecomment-2604853612)_

> A use case:
> 
> SHACL inferencing alone should be able to implement RDFS entailment as described in [RDF 1.1 Semantics, Section 9.2.1](https://www.w3.org/TR/rdf11-mt/#patterns-of-rdfs-entailment-informative), both in full, and in selected application of the expansion rules.
> 
> The motivation for _selected_ application of rules is that some of the rules lead to statements like `"example" a rdfs:Resource .` which is true under full RDF semantics but a problem for data stores or concrete syntaxes that expect literals to not appear in the subject position.
> 
> An analogous use case exists for OWL expansion, but I keep misplacing my bookmark to the OWL document that provides similar entailment tables.

Issue #219, opened 2025-01-24
Inferencing, `sh:entailment`, and strategies for describing entailment satisfaction


ajnelson-nist:
This is an expansion of a use case suggested in [this comment](https://github.com/w3c/data-shapes/issues/215#issuecomment-2604853612). The summary is that there may be value in defining a "partner" predicate to `sh:entailment` for custom entailment regimes.

Suppose I have a shapes graph that I expect to run against a knowledge graph (ABox & TBox) with the expectation of RDFS inferencing having taken place before the shapes start reviewing.

My understanding is much of RDFS and OWL entailment can be accomplished with what is in the 2017 state of the [SHACL-AF document](https://w3c.github.io/data-shapes/shacl-af/), by my current understanding of the mutual coverage of SPARQL `CONSTRUCT` and SHACL-AF expressions.  As an example, I have a script that runs parts of RDFS entailment by running a sequence of CONSTRUCT queries implementing some of the rules from [here](https://www.w3.org/TR/rdf11-mt/#patterns-of-rdfs-entailment-informative).  [This CONSTRUCT query](https://github.com/ajnelson-nist/CASE-Examples-QC/blob/ef322dffdffa95cb28d1cb6e6fd13358265b96b3/src/entail.py#L49-L63) entails types from `rdfs:domain`.

Separate thought, which will converge: `sh:entailment`, given in the 2017 state of the SHACL document, [Section 1.5](https://www.w3.org/TR/2017/REC-shacl-20170720/#shacl-rdfs), specifies how a shapes graph can require a certain entailment regime, and what should happen if the SHACL processor doesn't self-declare as supporting the requested entailment regime.

So, this shapes graph ...

```turtle
@prefix ex: <http://example.org/shapes/> .

<http://example.org/shapes>
  a owl:Ontology ;
  sh:entailment <http://www.w3.org/ns/entailment/RDFS> ;
  .
```

... is supposed to trigger a failure from the processor if the processor doesn't declare support for `http://www.w3.org/ns/entailment/RDFS`.

Coming together: Should part of the Inferencing working group's output be a graph implementing RDFS entailment, so that any processor could have a "default" ruleset to follow or use for consistency checking with their own implementation of RDFS entailment?

A benefit would be that the inferencing group could use this a way to test specification behaviors around `sh:entailment`, like non-standard IRIs that might be used in less broadly adopted shape graphs.
[...]

PR #452, opened 2025-07-29
Initial content for 'Inference Rules'

afs:
First pass at content for SHACL rules.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html)

~While this is a draft pull request, the important concerns are structure and design.~

### DIFF
--- a/shacl12-inf-rules/index.html
+++ b/shacl12-inf-rules/index.html
@@ -1,147 +1,210 @@
 <!DOCTYPE html>
 <html lang="en-US">
-	<head>
-		<meta charset="utf-8" />
-		<title>SHACL 1.2 Inference Rules</title>
-		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
-		<script class="remove">
-			var respecConfig = {
-				localBiblio: {},
+  <head>
+    <meta charset="utf-8" />
+    <title>SHACL 1.2 Inference Rules</title>
+    <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+    <script class="remove">
+    
+      function prepareSyntaxRules() {
+        document.querySelectorAll("[data-syntax-rule]").forEach(element => {
+          let ruleId = element.getAttribute("data-syntax-rule");
+          let tr = document.createElement("tr");
+          tr.classList.add("syntax-rule-tr");
+          let td1 = document.createElement("td");
+          td1.classList.add("syntax-rule-id");
+          let a = document.createElement("a");
+          a.classList.add("syntax-rule-id-a");
+          a.href = "#syntax-rule-" + ruleId;
+          a.textContent = ruleId;
+          td1.appendChild(a);
+          let td2 = document.createElement("td");
+          td2.innerHTML = element.innerHTML;
+          tr.appendChild(td1);
+          tr.appendChild(td2);
 
-				specStatus: "ED",
-				group: "wg/data-shapes",
-				github: "w3c/data-shapes",
-				//preProcess:     [  ],
-				edDraftURI: "https://w3c.github.io/data-shapes/shacl-inf-rules/",
-				shortName: "shacl12-inf-rules",
-				copyrightStart: "2025",
-				//xref: ["RDF12-CONCEPTS", "SHACL-CORE"],
-				xref: ["RDF12-CONCEPTS"],
+          // Replace <dfn> elements inside `tr` with <a> elements
+          tr.querySelectorAll("dfn").forEach(dfn => {
+              let a = document.createElement("a");
+              a.textContent = dfn.textContent;
+              dfn.replaceWith(a);
+          });
 
-				// (fix and) RemoveMe
-				lint: { "no-unused-dfns": false },
+          document.getElementById("syntax-rules-table").appendChild(tr);
+          element.setAttribute("id", "syntax-rule-" + ruleId);
+        });
+      }
+    
+      function prepareValidators() {
+        document.querySelectorAll("[data-validator]").forEach(element => {
+          let validatorId = element.getAttribute("data-validator") + "ConstraintComponent";
+          let tr = document.createElement("tr");
+          tr.classList.add("validator-tr");
+          let td = document.createElement("td");
+          tr.appendChild(td);
+          let a = document.createElement("a");
+          a.classList.add("validator-id-a");
+          a.href = `#validator-${validatorId}`;
+          a.textContent = `sh:${validatorId}`;
+          td.appendChild(a);
+          td.innerHTML += ": " + element.innerHTML;
+          document.getElementById("validators-table").appendChild(tr);
+          element.id = "validator-" + validatorId;
+        });
+      }
 
-				editors: [
-					{
-						name: "Robert David",
-						company: "Ontotext",
-						mailto: "<TBD>", // TODO
-						w3cid: "97894"
-					},
-					{
-						name: "David Habgood",
-						company: "KurrawongAI",
-						mailto: "<TBD>", // TODO
-						w3cid: "164266"
-					},
+      document.addEventListener("DOMContentLoaded", () => {
+        // Replace code div blocks in shapes, data, and results graph with tabs
+        for (const graph of document.querySelectorAll(".shapes-graph, .data-graph, .results-graph")) {
+          const tabs = document.createElement("aside");
+          tabs.classList.add("ds-selector-tabs");
+          graph.firstElementChild.classList.add("selected");
 
-					{
-						name: "Andy Seaborne",
-						company: "Apache Software Foundation",
-						mailto: "<TBD>", // TODO
-						w3cid: "29909"
-					},
-					{
-						name: "Simon Steyskal",
-						company: "Siemens AG",
-						companyURL: "https://siemens.com",
-						mailto: "simon.steyskal@siemens.com",
-						w3cid: "73545"
-					}
-				]
-			};
-		</script>
-		<style>
-			pre {
-				tab-size: 3;
-				-moz-tab-size: 3; /* Code for Firefox */
-				-o-tab-size: 3; /* Code for Opera */
-				word-wrap: normal;
-			}
+          for (const child of graph.children) {
+            child.classList.add("tab");
+          }
 
-			th {
-				text-align: left;
-			}
-			table.rule {
-				background-color: #ebebe0;
-			}
-			table.rule td {
-				text-align: center;
-			}
-			td.up {
-				border-bottom: 1px solid black;
-			}
+          tabs.append(...graph.children);
+          graph.append(tabs)
+        }
 
-			td {
-				vertical-align: top;
-			}
+        // Generate buttons for the selection logic
+        for (const tabs of document.querySelectorAll(".ds-selector-tabs")) {
+          const selectors = document.createElement("div");
+          selectors.classList.add("selectors");
+          
+          // Check if shaclc div exists in this tab group
+          const hasShaclc = tabs.querySelector(".shaclc");
+          
+          selectors.innerHTML = `
+            <button class="selected" data-selects="turtle">Turtle</button>
+            <button data-selects="jsonld">JSON-LD</button>
+            ${hasShaclc ? '<button data-selects="shaclc">SHACL-C</button>' : ''}
+          `
 
-			.algorithm {
-				background: #fafafc;
-				border-left-style: solid;
-				border-left-width: 0.5em;
-				border-color: #c0c0c0;
-				margin-bottom: 16px;
-				padding: 8px;
-			}
+          tabs.prepend(selectors);
+        }
 
-			.arg {
-				font-weight: bold;
-				color: #000080;
-			}
+        // Add example button selection logic
+        for (const button of document.querySelectorAll(".ds-selector-tabs .selectors button")) {
+          button.onclick = () => {
+            const ex = button.closest(".ds-selector-tabs");
+            ex.querySelector("button.selected").classList.remove("selected");
+            ex.querySelector(".selected").classList.remove("selected");
+            button.classList.add('selected');
+            ex.querySelector("." + button.dataset.selects).classList.add("selected");
+          }
+        }
+      });
+      
+      var respecConfig = {
+        localBiblio: {},
 
-			.def {
-				background: #fcfcfc;
-				border-left-style: solid;
-				border-left-width: 0.5em;
-				border-color: #c0c0c0;
-				margin-bottom: 16px;
-			}
+        specStatus: "ED",
+        group: "wg/data-shapes",
+        github: "w3c/data-shapes",
+        //preProcess:     [  ],
+        //@@
+        edDraftURI: "https://w3c.github.io/data-shapes/shacl-inf-rules/",
+        //shortName: "shacl12-inf-rules",
+        copyrightStart: "2025",
+        // SHACL12-CORE?
+        xref: ["RDF12-CONCEPTS", "SPARQL12-QUERY", "SHACL12-CORE", "SHACL12-SPARQL", "SHACL12-NODE-EXPR" ] ,
+        lint: { "no-unused-dfns": false },
 
-			.def-sparql {
-			}
+        editors: [
+          {
+            name: "Robert David",
+            company: "Ontotext",
+            //mailto: 
+            w3cid: "97894"
+          },
+          {
+            name: "David Habgood",
+            company: "KurrawongAI",
+            // mailto: 
+            w3cid: "164266"
+          },
 
-			.def-sparql-body {
-				margin-top: 0px;
-				margin-bottom: 0px;
-			}
+          {
+            name: "Andy Seaborne",
+            company: "Apache Software Foundation",
+            w3cid: "29909"
+          },
+          {
+            name: "Simon Steyskal",
+            company: "Siemens AG",
+            companyURL: "https://siemens.com",
+            //mailto: "simon.steyskal@siemens.com",
+            w3cid: "73545"
+          }
+        ],
 
-			.def-text {
-			}
+        //previousPublishDate: "2017-07-20",
+        //previousMaturity: "REC",
+        //prevRecShortname: "shacl",
+        //prevRecURI: "https://www.w3.org/TR/2017/REC-shacl-20170720/",
 
-			.def-text-body {
-			}
+        localBiblio: {
+          "shacl12-shacl-shacl": {
+            title: "SHACL 1.2 SHACL-SHACL",
+            href: "https://w3c.github.io/data-shapes/shacl12-shacl-shacl/",
+            status: "ED",
+            publisher: "W3C",
+          }
+        }
+        
+      };
+    </script>
+    <style>
+      pre {
+        tab-size: 3;
+        -moz-tab-size: 3; /* Code for Firefox */
+        -o-tab-size: 3; /* Code for Opera */
+        word-wrap: normal;
+      }
 
-			.def-header {
-				color: #a0a0a0;
-				font-size: 16px;
-				padding-bottom: 8px;
-			}
+      th {
+        text-align: left;
+      }
+      table.rule {
+        background-color: #ebebe0;
+      }
+      table.rule td {
+        text-align: center;
+      }
+      td.up {
+        border-bottom: 1px solid black;
+      }
 
-			.diagram-class {
-				border: 1px solid black;
-				border-radius: 4px;
-				width: 360px;
-			}
+      td {
+        vertical-align: top;
+      }
 
-			.diagram-class-name {
-				font-size: 16px;
-				font-weight: bold;
-				text-align: center;
-			}
+      .algorithm {
+        background: #fafafc;
+        border-left-style: solid;
+        border-left-width: 0.5em;
+        border-color: #c0c0c0;
+        margin-bottom: 16px;
+        padding: 8px;
+      }
 
-			.diagram-class-properties {
-				border-top: 1px solid black;
-			}
+      .arg {
+        font-weight: bold;
+        color: #000080;
+      }
 
-			.diagram-class-properties-start {
-				padding: 8px;
-			}
+      .def {
+        background: #fcfcfc;
+        border-left-style: solid;
+        border-left-width: 0.5em;
+        border-color: #c0c0c0;
+        margin-bottom: 16px;
+      }
 
-			.diagram-class-properties-section {
-				border-top: 1px dashed #808080;
-				padding: 8px;
-			}
+      .def-sparql {
+      }
 
 			.focus-node-selected {
 				color: blue;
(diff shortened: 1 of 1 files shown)


HolgerKnublauch:
Great start! I am assuming there will be some (RDF) vocabulary like sh:rule that links those rules to shapes, so that the target mechanisms come into the mix?

afs:
> Great start! I am assuming there will be some (RDF) vocabulary like sh:rule that links those rules to shapes, so that the target mechanisms come into the mix?

That's the plan - it's in [gist sketch](https://gist.github.com/afs/2547eac679b1411acc078959ac2b3d31#attaching-rules-to-shapes).

Targets `sh:targetClass` (and its implicit form), `sh:targetSubjectsOf` and `sh:targetObjectsOf` are equivalent to triple patterns so make that syntax logically map to a generator of `$this` -- `$this a :Class`, `$this :property ??`, `?? :property $this` -- and can be made to work for forward and backward implementation.

`sh:targetNodes` with node expressions may need limiting to the node expressions evaluated over the base graph, not triples currently being inferred. Node expressions are opaque as to what they depend on so backward chaining breaks and forward chaining can only work by blindly re-evaluating.

Similarly `sh:target [ a sh:SPARQLTarget ;` - though often that can be triple patterns.

So for regularity, maybe all targets should be "base graph only" generators of `$this` - run the rule once at the start of general evaluation.

recalcitrantsupplant:
Following from https://gist.github.com/afs/2547eac679b1411acc078959ac2b3d31#attaching-rules-to-shapes

>> Is there intention/scope for parameterisation, e.g. with sh:parameter, or must all inputs be in the data graph or provided inline as facts with DATA? 

> Where do the parameters come from?

Not from the datagraph - the idea would just be to define what must be substituted prior to the rules being valid & executable. An implementation would then provide some mechanism to supply the arguments. In the example below I've made up sh:argument to define what needs to be substituted, and sh:argval for where it's substituted in the rule. So ex:city would need to be supplied prior to running the rule.

> What a happens if a rule is applied multiple times?

I believe if the params must be substituted once at the start this would be safe (?)

```
ex:DensityRule
   sh:argument [ sh:path ex:city ; sh:nodeKind sh:IRI ] ;
   sh:rule [
       sh:head ( 
           [ sh:subject [ sh:argval ex:city ] ;
             sh:predicate ex:density ;
             sh:object [ sh:var "density" ] 
           ] 
       ) ;
       sh:body ( 
           [ sh:subject [ sh:argval ex:city ] ;
             sh:predicate ex:area ;
             sh:object [ sh:var "area" ]
           ]
           [ sh:subject [ sh:argval ex:city ] ;
             sh:predicate ex:population ;
             sh:object [ sh:var "pop" ]
           ]
           [ sh:bind [ 
               sh:var "density" ;
               sh:expression [ sparql:divide ( [ sh:var "pop" ] [ sh:var "area" ] ) ]
           ]]
       ) 
   ] .
```

HolgerKnublauch:
What are the plans with the existing SPARQL-based rules from SHACL-AF? We do have some users and I guess some people will find it easier to incrementally use their SPARQL strings instead of the new syntax.

Related: are there any scenarios where SPARQL-based rules can express more than the new, native SHACL rules?

afs:
> What are the plans with the existing SPARQL-based rules from SHACL-AF? We do have some users and I guess some people will find it easier to incrementally use their SPARQL strings instead of the new syntax.

Isn't that a question of what your engine chooses to support? I hope there isn't a conflict anywhere.

Some AF-rules can lead to problems (e.g. a CONTRUCT can increment a number ... which, if repeated applied, is an infinite loop).

I believe the best approach is to define the inf-rules then look at the where AF features fit in. 

Triple rules look like that can be mapped although we are now in the situation where there are multiple ways of doing that only one of which is inf-rules.

> Related: are there any scenarios where SPARQL-based rules can express more than the new, native SHACL rules?

SPARQL-based rules can go bad and using order to control means its a question about current usage. I don't think that can be answered except by seeing some actual rules.

Could we have issues labelled them as "UCR" and "Inference"?

afs:
> some mechanism to supply the arguments

@recalcitrantsupplant - We seem to have multiple ways to achieve same/similar tasks. How do you see this fitting with `sh:defaultValue`?

HolgerKnublauch:
I think I can live with just supporting them (SPARQL rules) in our platform even if the standard excludes them. I just wanted to get your opinion. No ticket needed on that.

A good sample use cases for your work would be to support OWL RL...

afs:
> A good sample use cases for your work would be to support OWL RL...

The rules are given:
https://www.w3.org/TR/owl2-profiles/#Reasoning_in_OWL_2_RL_and_RDF_Graphs_using_Rules
and also:
https://www.w3.org/TR/rif-owl-rl/

HolgerKnublauch:
Yes, and I still have a SPIN version of that somewhere in the product code... Yet it would be awesome to have this as a community project.

afs:
@robert-david, @simonstey, @recalcitrantsupplant --

How do you want to process with this? Do you think it needs more material before becoming the initial content, especially as that's going to be editor-draft-published to `shacl12-inf-rules`?

ajnelson-nist:
@afs @HolgerKnublauch I'd posted #220 (well, the comment ported to that Issue) because of the ability of Rules to do inferencing.  220's about RDFS, but #219 gets into an in-common subset of inferencing specifications between RDFS and OWL-RL.

219 may be a better Issue to continue on the topic of whether there's an example RDFS or OWL-RL Rules set the working-group is interested in.  Holger, I'm not sure what you meant by community project - something external to the data-shapes group, or instead produced by the data-shapes group?

afs:
RDFS and OWL-RL are themselves defined by rules sets.

There may be better ways to implement than the usual published rules as been noted.

The proposed rule language has a `TRANSITIVE` declaration so that engines can, for example, implement transitive without general rules recursion or have [custom implementation](https://www.cs.ox.ac.uk/people/boris.motik/pubs/hmh18modular-materialisation.pdf) of transitive properties for efficiency. The other declarations are for inverse and symmetric properties.

At the moment, we don't have a document with the base fundamentals we can make an editors draft.

HolgerKnublauch:
I have meanwhile created a SHACL-AF version of the OWL RL rules, all in SPARQL. This is based on earlier work that I did for SPIN in TopBraid Composer. If anyone is interested I can send a copy.

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+     <p>Rules:</p>
+<pre>
+RULE { ?x :childOf ?y } WHERE { ?y :fatherOf ?x }
+RULE { ?x :childOf ?y } WHERE { ?y :motherOf ?x }
+
+RULE { ?x :descendedFrom ?y } WHERE { ?x :childOf ?y }
+RULE { ?x :descendedFrom ?y } WHERE { ?x :childOf ?z . ?z :childOf ?y }
+</pre>
+
+      <p>will conclude that: :X is the child of :a and :B. and that :X is descendedFrom :C<p>
```
TallTed:
```suggestion
      <p>will conclude that: `:X` is the `:childOf` `:A` and `:B`, and that `:X` is `:descendedFrom` `:C`<p>
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+          <dt><dfn data-lt="data-block">data block</dfn></dt>
+          <dd>
+            A <em>data block</em> is a set of triples.
+            These form extra facts that are included in the inference process.
+          </dd>
+          
+          <dt><dfn>triple template</dfn></dt>
+          <dd>
+            A <em>triple template</em> is 3-tuple where each element is either
+            a variable or an RDF term (including being a triple term).
```
TallTed:
```suggestion
            A <em>triple template</em> is a 3-tuple where each element is either
            a variable or an RDF term (which might be a triple term, as of SHACL 1.2).
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            [=Triple templates=] appear in the [=head=] of a [=rule=].
+          </dd>
+
+          <dt><dfn>triple pattern</dfn></dt>
+          <dd>
+            A <em>triple pattern</em> is 3-tuple where each element is either a
+            variable, or an RDF term  (including being a triple term).
```
TallTed:
```suggestion
            A <em>triple pattern</em> is a 3-tuple where each element is either a
            variable or an RDF term (which might be a triple term, as of SHACL 1.2).
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            that evaluates to true or false.
+            [=Condition expressions=] appear in the [=body=] of a [=rule=].
+          </dd>
+
+          <dt><dfn>assignment</dfn></dt>
+          <dd>
+            An <em>assignment</em> is a pair of a <a>variable</a>, called the
+            <dfn>assignment variable</dfn>, and an <a>expression</a>, called the
+            <dfn>assignment expression</dfn>.
+            [=Assignments=] appear in the [=body=] of a [=rule=].
```
TallTed, line 793:
```suggestion
            An <em>assignment</em> is a pair of a <a>variable</a> and an 
            <a>expression</a>, called the <dfn>assignment variable</dfn>
            and <dfn>assignment expression</dfn>, respectively.
            [=Assignments=] appear in the [=body=] of a [=rule=].
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            A <em>rule head</em> is a sequence where each element
+            of the sequence is a [=triple template=].
+          </dd>
+          
+          <dt>
+            <dfn data-lt="body|rule body">rule body</dfn>
+          </dt>
+          <dd>
+            A <em>rule body</em> is a sequence where each element
+            of the sequence is a [=triple pattern=], a [=condition expression=] or an [=assignment=].
```
TallTed:
```suggestion
            of the sequence is a [=triple pattern=], a 
            [=condition expression=], or an [=assignment=].
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+          <dd>
+            A <em>rule set</em> is a collection of zero or more rules.
+          </dd>
+        </dl>
+
+        <p>
+          In a [=triple pattern=] or a [=triple template=],
+          position 1 of the tuple is informally called the <em>subject</em>;
+          position 2 is informally called the <em>predicate</em>
+          position 3 is informally called the <em>object</em>.
```
TallTed:
```suggestion
          position 1 of the tuple is informally called the <em>subject</em>,
          position 2 is informally called the <em>predicate</em>, and
          position 3 is informally called the <em>object</em>.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+        <section id="wellformed">
+          <h3>Well-formedness Conditions</h3>
+          <p>
+            Well-formedness is a set of conditions on abstract syntax of
+            shapes rules.  Together, these rules ensure that a [=variable=] in a
+            [=head=] of a rule has a value defined in the [=body=] of the rule; 
+            that each variable in an condition expression or assignment 
+            expression has a value at the point of evaluation, and that each
+            assignment in a rule only introduces a new variable, 
+            not used earlier in the rule body.
```
TallTed:
```suggestion
            Well-formedness is a set of conditions on the abstract syntax of
            shapes rules.  Together, these rules ensure that a [=variable=] in the
            [=head=] of a rule has a value defined in the [=body=] of the rule; 
            that each variable in an condition expression or assignment 
            expression has a value at the point of evaluation; and that each
            assignment in a rule introduces a new variable, 
            not used earlier in the rule body.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+          </p>
+          <p>
+            A [=rule=] is a <dfn>well-formed rule</dfn> if all the following
```
TallTed:
```suggestion
            A [=rule=] is a <dfn>well-formed rule</dfn> if all of the following
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            conditions are met:
+          </p>
+          <ul>
+            <li>For every [=variable=] appearing in a [=triple templates=]
+              of the [=head=] of the [=rule=],
+              there is one or more occurrences of a [=variable=] 
+              of the same name in the [=triple patterns=] in the [=body=], 
+              or in an [=assignment=] in the body, or both.
```
TallTed:
```suggestion
            <li>For every [=variable=] appearing in a [=triple template=]
              of the [=head=] of the [=rule=],
              a [=variable=] 
              of the same name occurs one or more times in the
              [=triple patterns=] and/or the [=assignments=] in the body.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            </li>
+            <li>
+              For every [=variable=] in an [=expression=] at position <em>i</em>
+              of the [=body=], there is a corresponding [=variable=] of the same
+              name occuring in a [=triple pattern=], or occurring as an
+              [=assignment variable=], at a position <em>j</em>, where <em>j
+              &lt; i</em>.
```
TallTed: It seems likely that <em>j</em> should be ***greater than*** <em>i</em>, not _less_.
```suggestion
              For every [=variable=] in an [=expression=] at position <em>i</em>
              of the [=body=], a corresponding [=variable=] of the same
              name occurs in a [=triple pattern=] or as an
              [=assignment variable=] at position <em>j</em>, where <em>j</em>
              is less than <em>i</em>.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            </li>
+            <li>
+              An [=assignment variable=] is used in only one [=assignment=] in the [=body] of the [=rule=].
```
TallTed:
```suggestion
              Each [=assignment variable=] is used in only one [=assignment=] in
              the [=body] of the [=rule=].
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            </li>
+            <li>
+              An [=assignment variable=] at position <em>i</em> of a [=rule body=] does not
+              occur in any [=triple pattern=] at position <em>j</em> where <em>j &lt; i</em>.
```
TallTed:
```suggestion
              No [=assignment variable=] at position <em>i</em> of a [=rule body=]
              occurs in any [=triple pattern=] at position <em>j</em>,
              where <em>j</em> is less than <em>i</em>.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            </li>
+            <li>
+              For every [=variable=] in an [=assignment expression=] at position <em>i</em>,
+              there is a corresponding variable in a triple pattern at position 
+              <em>j</em> where <em>j &lt i</em>,
+              or there is an [=assignment variable=] in an [=assignment=] at position 
+              <em>j</em> where <em>j &lt i</em>.
```
TallTed: Again, it seems that `j` _should be_ greater than `i`, here.
```suggestion
              For every [=variable=] in an [=assignment expression=] at position <em>i</em>,
              there is a corresponding variable in a triple pattern,
              or an [=assignment variable=] in an [=assignment=],
              at position 
              <em>j</em>, where <em>j</em> is less than <em>i</em>.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+            </li>
+          </ul>
+          <p>
+            A [=rule set=] is "well-formed" if and only all the [=rules=] of the
+            rule set are "well-formed".
```
TallTed:
```suggestion
            A [=rule set=] is "well-formed" if and only if all of the [=rules=] of the
            rule set are "well-formed".
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+
+    <section id="concrete-syntax">
+      <h2>Concrete Syntax forms for Shapes Rules</h2>
+      <ul>
+        <li><a href="rdf-rules-syntax">RDF Rules Syntax</a></li>
+        <li><a href="compact-rules-syntax">Compact Rules Syntax></a></li>
+      </ul>
+
+      <p class="note">
+        The compact syntax has an equivalent RDF syntax form.<br/>
```
TallTed:
```suggestion
        The compact syntax has an equivalent RDF syntax form.
```

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1880 @@
...
+        <p>Well -formed ness:</p>
+        <ul>
+          <li>All RDF list are well-formed</li>
+          <li>exactly one of subject - predicate - object per body of head
+            element</li>
+          <li>Well-formed, single-valued,list-argument node expressions
+          <li>well-formed abstract syntax
+        </ul>
+
+        <p>Describe how the abtract model maps to triples??.
```
TallTed: Based on the abbreviated text here, I'm not reviewing the rest of this PR, as it seems clear that the rest remains "draft".

TallTed (changes requested):

shacl12-inf-rules/index.html
afs: Changed to "which might be".
(due to a merge conflict (some other PR changed the CSS) I am editing locally).

No need for the "as of SHACL 1.2". (1) RDF 1.2 and (2) this document is new at SHACL 1.2.

shacl12-inf-rules/index.html
afs: I prefer the original as it more closely connects the part of the assignment wth its `<dfn>`.

shacl12-inf-rules/index.html
afs: Correct.

The whole document is initial content so we can starting publishing editors' working drafts.

shacl12-inf-rules/index.html
afs: I prefer to talk about "occurrences" at this point.

shacl12-inf-rules/index.html
afs: The variable must be given a value (i.e. be defined by triple pattern or assignment - j) before used in a condition (position i) so j is less than i.

Would you prefer `i is greater than j`?

shacl12-inf-rules/index.html
afs: Same as above.

shacl12-inf-rules/index.html
afs: Would you prefer `i is greater than j`?

shacl12-inf-rules/index.html
afs: Would you prefer i is greater than j?

recalcitrantsupplant:
> @robert-david, @simonstey, @recalcitrantsupplant --
> 
> How do you want to process with this? Do you think it needs more material before becoming the initial content, especially as that's going to be editor-draft-published to `shacl12-inf-rules`?

4.1 could be cleaned up perhaps - I'm not sure what all of the things there are referring to.
Otherwise, these two topics come to mind but as I understand the process they do not need to happen prior to publishing a draft(?):
1. stratification
2. parameterisation similar to the questions in https://github.com/w3c/data-shapes/issues/498

Would it be worth the editors meeting now to discuss these and any other sections?

shacl12-inf-rules/index.html
recalcitrantsupplant: Could we switch i & j in these, for example:
> For every [variable](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-variable) in an [expression](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-expression) at position j of the [body](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-body), there is a corresponding [variable](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-variable) of the same name occuring in a [triple pattern](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-triple-pattern) at a position i, or occurring as an [assignment variable](https://raw.githack.com/w3c/data-shapes/initial-infer-rules/shacl12-inf-rules/index.html#dfn-assignment-variable) at a position i, where i < j. </code>

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1883 @@
...
+          position 3 is informally called the <em>object</em>.
+        </p>
+
+        <section id="wellformed">
+          <h3>Well-formedness Conditions</h3>
+          <p>
+            Well-formedness is a set of conditions on the abstract syntax of
+            shapes rules.  Together, these rules ensure that a [=variable=] in the
+            [=head=] of a rule has a value defined in the [=body=] of the rule; 
+            that each variable in an condition expression or assignment
```
recalcitrantsupplant, line 837:
```suggestion
            that each variable in a condition expression or assignment 
```

simonstey:
Given we introduced `shnex` as prefix for all things node expressions, and (for now)  `shnex-sparql` do we want/should have a dedicated prefix for rules as well?

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1883 @@
...
+          </dt>
+          <dd>
+            A <em>rule body</em> is a sequence where each element
+            of the sequence is a [=triple pattern=], a 
+            [=condition expression=], or an [=assignment=].
+          </dd>
+          
+          <dt><dfn>rule set</dfn></dt>
+          <dd>
+            A <em>rule set</em> is a collection of zero or more rules.
```
recalcitrantsupplant: Does a rule set also include zero or more data blocks, to align better with the grammar? If so, should `sh:ruleSet` also be something like `sh:rules`

simonstey:
@afs 
> > @robert-david, @simonstey, @recalcitrantsupplant --
> > How do you want to process with this? Do you think it needs more material before becoming the initial content, especially as that's going to be editor-draft-published to `shacl12-inf-rules`?

It would certainly benefit from at least an initial Introduction draft. Primarily, to clarify how this document relates to the other specs (core, node-expr in particular) and what it aims to achieve.

Other than that, I think it's in a very good initial state and (imo) can be published as a first draft to get it into the "publishing cycle".

recalcitrantsupplant (approved):

afs:
> Given we introduced `shnex` as prefix for all things node expressions, and (for now) `shnex-sparql` do we want/should have a dedicated prefix for rules as well?

Yes - introducing `shnex`post-dates this content. `shnex-sparql` is new to me.

afs:
> It would certainly benefit from at least an initial Introduction draft. Primarily, to clarify how this document relates to the other specs (core, node-expr in particular) and what it aims to achieve.
> 
> Other than that, I think it's in a very good initial state and (imo) can be published as a first draft to get it into the "publishing cycle".

@simonstey - are you saying that an introduction is required in your opinion before merging to gh-pages?

shacl12-inf-rules/index.html
afs: Usually names are "i" then "j" as they appear in text.

I've made the change to switch the condition -- `i > j`.

We can come back to rewording the entire condition at some point.

shacl12-inf-rules/index.html
afs: Currently changed to `i > j`.

afs:
> 4.1 could be cleaned up perhaps - I'm not sure what all of the things there are referring to.

Made into editors notes.

afs:
> 1. stratification
> 2. parameterisation similar to the questions in [Support User-Defined Node Expression Functions; Parametrization, "Partial Evaluation" #498](https://github.com/w3c/data-shapes/issues/498)
> 
> Would it be worth the editors meeting now to discuss these and any other sections?

Yes - a meeting would be a good idea.

shacl12-inf-rules/index.html
afs: Data blocks added.

At the moment, they are at the same level as rules:

```ttl
:ruleSet-1
  rdf:type srl:RuleSet;
  srl:data (
    <<( :s :p :o )>>
  );
  srl:rules (
  ...
  );
.
```

then as you have it -- `srl:ruleSet` (property) would attach a rule set to a subject.

afs: The use of triple terms means application/domain data is not asserted as true in the rule set.

shacl12-inf-rules/index.html
TallTed: I'm glad I understood what was originally intended, and so caught the errors/typos in the existing text.

I can live with the `>` symbol instead of the `is greater than` wording. I do fear it will be less comprehensible to readers, especially those for whom English is a second or later language, which is why I recommended this change.

shacl12-inf-rules/index.html
TallTed: As above. I prefer the words to the symbol, but I can live with the symbol.

shacl12-inf-rules/index.html
TallTed: The original does more closely tie each `<a>` with its `<dfn>`. However, it obscures the pairings. It's a tradeoff. I'll live with your preference, but I wonder whether there's a good way to gather outside opinions?

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1890 @@
...
+</pre>
+
+      <section id="rdf-rules-syntax">
+        <h3>RDF Rules Syntax</h3>
+
+        <div class="ednote">
+          <p>Well-formed ness:</p>
+          <ul>
+            <li>All RDF list are well-formed</li>
+            <li>exactly one of subject - predicate - object per body of head
```
TallTed: To make it less likely to be interpreted as having a period where I'm adding a comma.
```suggestion
            <li>exactly one of subject - predicate - object, per body of head
```

shacl12-inf-rules/index.html
afs: It's early days.

shacl12-inf-rules/index.html
afs: Change made.

However, blocking merging for something in an editors note is taking up time.

TallTed (approved):
My tweaks thus far are done.

I think we'll need a better wording for `exactly one of subject - predicate - object, per body of head element`, possibly starting from `exactly one each of subject, predicate, and object, per body of head element`...

simonstey:
@afs 
> > It would certainly benefit from at least an initial Introduction draft. Primarily, to clarify how this document relates to the other specs (core, node-expr in particular) and what it aims to achieve.
> > Other than that, I think it's in a very good initial state and (imo) can be published as a first draft to get it into the "publishing cycle".
> 
> @simonstey - are you saying that an introduction is required in your opinion before merging to gh-pages?

no, but it would certainly benefit from having one.

maybe something along the lines of:

> Introduction
> 
> This document introduces the concept of **Inference Rules** for SHACL 1.2, a mechanism for deriving new RDF triples from existing data using declarative rules defined in shapes graphs. This extends SHACL’s capabilities beyond validation, enabling reasoning and data enrichment.
> 
> This document complements other SHACL 1.2 specifications, such as SHACL Core, by defining the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint validation, the Inference Rules specification provides a standardized way to express and evaluate rules that generate new data.

simonstey (approved):

afs:
@simonstey - thanks - I'll put that text in.

We have to find a way to position rules. For me, Core is closely tied to "constraints" (history) whereas there is a larger "shapes" concept and rules is a new area.

HolgerKnublauch:
Right, the positioning is important. Node expressions and sh:values in particular can also be used to derive new triples, so this is not unique for the rules spec.

The core spec has this intro section on relationship to RDFS and entailment regimes, and that should certainly point at the rules work.

afs:
@HolgerKnublauch - question for you.

Do `sh:values` and `sh:defaultValue` provide anything that can not be written in rules (ignoring the fact that it may be unnatural and burdensome to write it that way)?

HolgerKnublauch:
If rules can execute arbitrary node expressions then I don't see why they couldn't cover everything. In our own platform we have some algorithms that perform both types of inferences together, e.g. to assert anything implicit.

The values nodes produced by sh:values and defaultValue are, however, used for constraint checking and to pre-populate forms.

afs:
Not arbitrary node expressions in arbitrary position: ones that touch the data graph (NX style aggregates), ones that evaluate to lists with >1 member (unless used only as seeds) or whatever comes along in the future (it's a open world).

shacl12-inf-rules/index.html
```diff
@@ -160,444 +223,1902 @@
...
+
+      <section id="rdf-rules-syntax">
+        <h3>RDF Rules Syntax</h3>
+
+        <div class="ednote">
+          <p>Well-formed ness:</p>
+          <ul>
+            <li>All RDF list are well-formed</li>
+            <li>exactly one of subject - predicate - object, per body of head
+              element</li>
```
TallTed, line 965:
```suggestion
            <li>exactly one each of subject, predicate, and object, per body of head
              element</li>
```

PR #268, opened 2025-02-23
#177: Added sh:singleLine

HolgerKnublauch:

Issue #449, opened 2025-07-29
SHACL-SHACL's `shsh:ShapeShape` should stop line-wrapping property links


ajnelson-nist:
These targeting statements within `shsh:ShapeShape` have undergone some churn lately:

https://github.com/w3c/data-shapes/blob/9c4f352e942426dbfd00634c45a6c99a958de155/shacl12-vocabularies/shacl-shacl.ttl#L72-L78

They are likely to undergo more as more properties are added.

It is becoming a bit error-prone and time-consuming to try managing this like a paragraph with line-wraps.  I would like to instead use the style of the following lines, and sorted.

https://github.com/w3c/data-shapes/blob/9c4f352e942426dbfd00634c45a6c99a958de155/shacl12-vocabularies/shacl-shacl.ttl#L80-L84

Then, inserting a new predicate into the middle doesn't induce a line re-wrap and difficult Git merge conflicts.

PR coming momentarily.  It will have some friction with any in-flight PRs modifying those code sections, but in my opinion it will be easier to address those merge conflicts.

nicholascar:
Rather than the predicate & object per line, if we just used a longturtle serialization, which also sorts, then we will get a better Git result. 

Using [RDFLib](https://github.com/rdflib/rdflib/) v 7.1.4, we get:

```
    sh:targetObjectsOf
        sh:memberShape ,
        sh:node ,
        sh:not ,
        sh:property ,
        sh:qualifiedValueShape ;
```
Here's the file, longturtle serialised, but with a txt file extension to keep GitHub happy:

* [shacl-shacl.txt](https://github.com/user-attachments/files/21499470/shacl-shacl.txt)

I don't know whether other Turtle serializers do as much new line forcing as RDFLib does with longturtle.

HolgerKnublauch:
From a versioning point of view I find it better to be redundant and make each line self-contained. Makes it easier to delete lines without affecting adjacent lines. So ideal is

```
ex:Subject
    ex:prop1 ex:value1 ;
    ex:prop1 ex:value2 ;
    ex:prop2 ex:value3 ;
.
```

(Note the dot at the extra line)

ajnelson-nist:
@nicholascar - While working at the standards level, I would prefer to remain as tool-agnostic as possible.

[...]

PR #463, opened 2025-08-04
#177: Update SHACL-SHACL to review `sh:singleLine` usage

ajnelson-nist:
This PR is a follow-on to #268, and just updates SHACL-SHACL to test syntax for #177 .

### DIFF
--- a/shacl12-vocabularies/shacl-shacl.ttl
+++ b/shacl12-vocabularies/shacl-shacl.ttl
@@ -108,10 +108,10 @@ shsh:ShapeShape
 	sh:targetSubjectsOf sh:qualifiedMinCount ;
 	sh:targetSubjectsOf sh:qualifiedValueShape ;
 	sh:targetSubjectsOf sh:qualifiedValueShapesDisjoint ;
+	sh:targetSubjectsOf sh:singleLine ;
 	sh:targetSubjectsOf sh:uniqueLang ;
 	sh:targetSubjectsOf sh:uniqueMembers ;
 	sh:targetSubjectsOf sh:xone ;
-  sh:targetSubjectsOf sh:singleLine ;
 
 	sh:targetObjectsOf sh:memberShape ; # memberShape-node
 	sh:targetObjectsOf sh:node ;        # node-node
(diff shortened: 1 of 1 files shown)


jeswr (approved):

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -111,6 +111,7 @@ shsh:ShapeShape
 	sh:targetSubjectsOf sh:uniqueLang ;
 	sh:targetSubjectsOf sh:uniqueMembers ;
 	sh:targetSubjectsOf sh:xone ;
+  sh:targetSubjectsOf sh:singleLine ;
```
TallTed: Should alignment of this line not match those above? I still hate tabs for this purpose, but it doesn't make sense to indent one line with two-spaces when the rest of its stanza has one-tab indents.
```suggestion
	sh:targetSubjectsOf sh:singleLine ;
```

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -111,6 +111,7 @@ shsh:ShapeShape
...
 
 	sh:targetObjectsOf sh:memberShape ; # memberShape-node
```
TallTed: I note that `sh:targetSubjectsOf` precedes `sh:targetObjectsOf`, so again, these documents can only be sorted one predicate at a time.

shacl12-vocabularies/shacl-shacl.ttl
ajnelson-nist: Now addressed - tab-indented and sorted, consistent with #449 .

shacl12-vocabularies/shacl-shacl.ttl
ajnelson-nist: Unfortunately, that is an artifact of terminology choice sorting differently from triple-component order, set in motion from the original RDF specification and I think beyond the scope of this group to address.

ajnelson-nist:
Having had 2 weeks for review, 1 approval, 1 set of responded-to remarks, increased test coverage, and IMO low complexity, I'll merge this PR now.

PR #467, opened 2025-08-05
minor editorial change to shacl-shacl.ttl

VladimirAlexiev:


jeswr (approved):

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -23,11 +23,11 @@ shsh:
 shsh:ListShape
 	a sh:NodeShape ;
 	rdfs:label "List shape"@en ;
-	rdfs:comment "A shape describing well-formed RDF lists. Currently does not check for non-recursion. This could be expressed using SHACL-SPARQL."@en ;
+	rdfs:comment "A shape describing well-formed RDF lists. Currently does not check for non-recursion (this could be expressed using SHACL-SPARQL)."@en ;
```
TallTed, line 26: Checking for absence almost always means checking for presence, and reporting that no presence was found. It seems better to just check for presence.
```suggestion
	rdfs:comment "A shape describing well-formed RDF lists. Currently does not check for recursion. (This could be expressed using SHACL-SPARQL.)"@en ;
```

PR #470, opened 2025-08-05
Fix UML diagram style

YoucTagh:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-166-fix-uml-style/shacl12-core/index.html#constraints-section)


VladimirAlexiev:
@YoucTagh Looks good, but https://github.com/w3c/data-shapes/pull/471 looks better :-) and we don't need to hardcode a width

VladimirAlexiev (approved):

PR #492, opened 2025-08-11
Issue 481 - move `nodeByExpression` and `expression` to Node Expr spec and give better examples

### LINKED ISSUE #481: move `nodeByExpression, expression` to Node Expr spec and give better examples
https://raw.githack.com/w3c/data-shapes/master/shacl12-core/index.html#example-52 :
```ttl
ex:PersonShape
	a sh:NodeShape ;
	sh:targetClass ex:Person ;
	sh:property [   # _:b1
		sh:path ex:address ;
		sh:minCount 1 ;
		sh:nodeByExpression ex:AddressShape ;
	] .
```
This is not a good example because instead of `sh:nodeByExpression`, we can use a plain `sh:node` (as in https://raw.githack.com/w3c/data-shapes/master/shacl12-core/index.html#example-48).

@mgberg @afs @HolgerKnublauch 
Can we come up with a better example?

I see in one of the commits of https://github.com/w3c/data-shapes/pull/408
>  Only using an IRI Expression here because Core doesn't define interesting node expressions

So I guess you want to keep Core self-contained and not use expressions except the trivial ones in https://raw.githack.com/w3c/data-shapes/master/shacl12-core/index.html#node-expressions. I think there are 2 better alternatives:
1. That example could use a more complex expression, refer to the Node Expressions spec, and mention that supporting such expressions by Core-conforming engines is not required
2. Or I could argue that `sh:nodeByExpression` itself should not be part of Core but of Node Expressions. 
   - @mgberg can you confirm that the evaluation result of `sh:node` and `sh:nodeByExpression` on IRIs/literals is the same?

Now I think that 2 is the better option: If an engine doesn't support non-trivial Node Expressions, why should it support `sh:nodeByExpression`?

HolgerKnublauch:
Closes #481

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-481/shacl12-node-expr/index.html)


HolgerKnublauch:
Note there seems to be something wrong with the W3C specs right now, as all references to external specs are broken. As a result, the preview is full of errors (but the same happens to SHACL Core right now, so this doesn't seem to be our problem).

robert-david (approved):

recalcitrantsupplant (approved):

PR #499, opened 2025-08-11
Issue 490

### LINKED ISSUE #490: Move Node Expressions into their own namespace
As discussed in https://github.com/w3c/data-shapes/issues/484#issuecomment-3170494909 at #484 we should move the built-in node expression functions (except sparql:) into their own namespace and graph. I suggest to use shnex as prefix and namespace

`http://www.w3.org/ns/shnex#`

The next-best alternative would be expr but I am worried this is too generic, and may already be used by some projects. With shnex this is very unlikely to clash and uniquely identifies the SHACL foundation.

HolgerKnublauch:
Closes #490
* #490
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-490/shacl12-node-expr/index.html)


afs:
(added a bullet point to the PR description so the title text shows up here)

HolgerKnublauch:
May I get a PR approval here because there are other tickets that will be impacted by this change. We should get this in this week to not delay other work.

shacl12-node-expr/index.html
```diff
@@ -937,7 +935,7 @@ <h3>Var Expressions</h3>
 								</thead>
 								<tbody>
 									<tr>
-										<td><b><code>sh:var</code></b></td>
```
simonstey, line 940: so both `sh:var` and `sh:this` became `shnex:var`?

shacl12-node-expr/index.html
```diff
@@ -2185,8 +2183,8 @@ <h3>sh:expression</h3>
             [ 
                 sparql:ucase (
                     [
-                        sh:path ( ex:country ex:code )
-                        sh:nodes [ sh:var "focusNode" ]
+                        shnex:path ( ex:country ex:code )
+                        shnex:nodes [ sh:var "focusNode" ]
```
simonstey:
```suggestion
                        shnex:nodes [ shnex:var "focusNode" ]
```

shacl12-node-expr/index.html
HolgerKnublauch: Right, we could not keep sh:this because any URI node is treated as a Constant Node Expression.

shacl12-node-expr/index.html
```diff
@@ -2185,8 +2183,8 @@ <h3>sh:expression</h3>
             [
```
simonstey, line 2183: the line above also still has `sh:var`

simonstey: (for some reason I wasn't able to comment directly on the line before that one 🙄

simonstey (approved):

PR #509, opened 2025-08-14
#505: Added section on Dynamic SHACL, including two examples

### LINKED ISSUE #505: Dynamic SHACL
This idea was discussed before in this group, and here https://www.datashapes.org/dynamic.html

The Node Expressions document should have a section on a dialect of SHACL in which
- the values of all constraint parameters that cannot take blank nodes may be node expressions
- those node expressions are evaluated before validation, to compute the actual constraints in the shapes graph

This is an entirely optional feature that some engines may support. We cannot make it mandatory due to its complexity and unpredictability for static analysis.

HolgerKnublauch:
Closes #505

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-505/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -2729,7 +2729,7 @@ ex:QueenslandAddress1
 					We want to express that the valid values of <code>ex:state</code> depend on the value of <code>ex:country</code>
 					at the given focus node.
 					For example, the valid values for country <code>ex:USA</code> would be <code>( "AL" "AK" "AZ" ... )</code>
-					while values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
+					while valid values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
 					This fact can be represented as part of the data:
 				</p>
 				<aside class="example" title="The valid states can be attached to each country">
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -481,6 +481,7 @@
 						<dfn data-cite="shacl12-core#dfn-conform" data-lt="conform|conforms">conform</dfn>,
 						<dfn data-cite="shacl12-core#dfn-conformance-check" data-lt="conformance check">conformance check</dfn>,
 						<dfn data-cite="shacl12-core#dfn-failure" data-lt="failure|failures">failure</dfn>,
+						<dfn data-cite="shacl12-core#dfn-validation">validation</dfn>,
 						<dfn data-cite="shacl12-core#dfn-shacl-instance" data-lt="shacl instance">SHACL instance</dfn>,
 						<dfn data-cite="shacl12-core#dfn-shacl-subclass" data-lt="shacl subclass">SHACL subclass</dfn>,
 						<dfn data-cite="shacl12-core#dfn-shacl-type" data-lt="shacl type">SHACL type</dfn>,
@@ -557,9 +558,8 @@
 			</section>
 		</section>
 
-		<section id="getting-started">
+		<section id="getting-started" class="informative">
 			<h2>Getting started with Node Expressions</h2>
-			<p><em>This section is informative.</em></p>
 			<p>
 				A SHACL <a>shapes graph</a> can declare node expressions as values of various properties where dynamic computation is useful,
 				such as <code>sh:targetNode</code>, <code>sh:values</code> and <code>sh:deactivated</code>.
@@ -2645,6 +2645,142 @@ eg:obs1a
 				</aside>
 			</section>
 			
+		</section>
+
+		<section id="dynamic-shacl">
+			<h2>Dynamic SHACL</h2>
+			<p>
+				This section defines <dfn>Dynamic SHACL</dfn> as a dialect of SHACL that some implementations MAY support.
+			</p>
+			<p>
+				In Dynamic SHACL any <a>parameter</a> of a <a>constraint</a> can be a <a>node expression</a>,
+				excluding those that do <em>not</em> allow <a>blank nodes</a> (such as <code>sh:node</code>)
+				but including those that take <a>SHACL lists</a> as values (such as <code>sh:class</code>, <code>sh:datatype</code> and <code>sh:in</code>).
+				During <a>validation</a>, such <a>node expressions</a> are evaluated in the <a>data graph</a>,
+				using the current <a>focus node</a>.
+				The resulting <a>nodes</a> will be used as <a>parameters</a> for the <a>constraint</a>.
+			</p>
+			<section class="informative">
+				<h3>Example: Dynamic Minimum Age of Presidents</h3>
+				<p>
+					As a use case of <a>Dynamic SHACL</a>, assume we want to express that the legal minimum age
+					of a president is 18 unless the country is USA, where it is 40.
+				</p>
+				<aside class="example" title="Example of Dynamic SHACL using a node expression at sh:minInclusive">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:PresidentShape
+    a sh:NodeShape ;
+	sh:targetClass ex:President ;
+    sh:property ex:PresidentShape-age ;
+.
+ex:PresidentShape-age
+    a sh:PropertyShape ;
+    sh:path ex:age ;
+    <b>sh:minInclusive [
+        shnex:if [
+            sparql:eq (
+                [ shnex:path ex:country ]
+                ex:USA
+            )
+        ]
+        shnex:then 40 ;
+        shnex:else 18 ;
+    ]</b> .
+						</div>
+						<div class="jsonld">
+							<pre class="jsonld">
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+			</section>
+			<section class="informative">
+				<h3>Example: Dynamic Enumerations</h3>
+				<p>
+					As a use case of <a>Dynamic SHACL</a>, assume the following <a>data graph</a>.
+				</p>
+				<aside class="example" title="Example data graph with addresses">
+					<div class="data-graph">
+						<div class="turtle">
+ex:ArizonaAddress1
+    a ex:Address ;
+    ex:street "123 John Muir Ave" ;
+    ex:country ex:USA ;
+    ex:state "AZ" ;
+.
+ex:QueenslandAddress1
+    a ex:Address ;
+    ex:street "123 Bob Katter Cl" ;
+    ex:country ex:Australia ;
+    ex:state "QLD" ;
+.
+						</div>
+						<div class="jsonld">
+							<pre class="jsonld">
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					We want to express that the valid values of <code>ex:state</code> depend on the value of <code>ex:country</code>
+					at the given focus node.
+					For example, the valid values for country <code>ex:USA</code> would be <code>( "AL" "AK" "AZ" ... )</code>
+					while values values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
+					This fact can be represented as part of the data:
+				</p>
+				<aside class="example" title="The valid states can be attached to each country">
+					<div class="data-graph">
+						<div class="turtle">
+ex:Australia
+    a ex:Country ;
+    ex:stateCode "ACT", "NSW", "NT", "QLD", "SA", "TAS", "VIC", "WA" ;
+.
+ex:USA
+    a ex:Country ;
+    ex:stateCode "AL", "AK", "AZ", "AR" ; # ...
+.
+						</div>
+						<div class="jsonld">
+							<pre class="jsonld">
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					Using this extra information, we can now define a <code>sh:in</code> constraint using a <a>Path Expression</a>:
+				</p>
+				<aside class="example" title="Example of Dynamic SHACL using a node expression at sh:in">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:Address
+    a sh:ShapeClass ;
+    sh:property ex:Address-state ;
+.
+ex:Address-state
+    a sh:PropertyShape ;
+    sh:path ex:state ;
+    <b>sh:in [
+        shnex:path ( ex:country ex:stateCode )
+    ]</b> .
+						</div>
+						<div class="jsonld">
+							<pre class="jsonld">
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					During validation, a Dynamic SHACL engine will evaluate the path expression at <code>sh:in</code>
+					and use the resulting nodes as members of the allowed values.
+					So, when the value of <code>ex:country</code> is <code>ex:USA</code>, it will look up the
+					state codes that are linked to <code>ex:USA</code>.
+				</p>
+			</section>
 		</section>
 
 		<section id="security">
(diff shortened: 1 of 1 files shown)


shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+				but including those that take <a>SHACL lists</a> as values (such as <code>sh:class</code>, <code>sh:datatype</code> and <code>sh:in</code>).
+				During <a>validation</a>, such <a>node expressions</a> are evaluated in the <a>data graph</a>,
+				using the current <a>focus node</a>.
+				The resulting <a>nodes</a> will be used as <a>parameters</a> for the <a>constraint</a>.
+			</p>
+			<section class="informative">
+				<h3>Example: Dynamic Minimum Age of Presidents</h3>
+				<p>
+					As a use case of <a>Dynamic SHACL</a>, assume we want to express that the legal minimum age
+					of a president is 18 unless the country is USA, where it is 40.
```
ajnelson-nist:
```suggestion
					of a president is 18 unless the country is USA, where it is 35.
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+    a sh:PropertyShape ;
+    sh:path ex:age ;
+    <b>sh:minInclusive [
+        shnex:if [
+            sparql:eq (
+                [ shnex:path ex:country ]
+                ex:USA
+            )
+        ]
+        shnex:then 40 ;
```
ajnelson-nist:
```suggestion
        shnex:then 35 ;
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+        shnex:else 18 ;
```
ajnelson-nist, line 2689: Would you be open to expanding this example to demonstrate one nesting round of `shnex:else` and a node expression - say, another country with some other minimum age?

HolgerKnublauch: I wanted to keep the first example as simple as possible. The second example is more complex and is how this really should be done - by looking up values in the data instead of hard-coding them in the node expression. Here, it would be better to attach the age to each country, but that would again add another layer of complexity to the example.

ajnelson-nist: I agree with the sentiment on easy introductory examples.  But, the recursive wording at the top of current-4.1.5 ("If Expressions") makes me want just one recursion demonstration.

If I'm the only one that asks for it, I'm fine with this just carrying forward as-is.

shacl12-node-expr/index.html
```diff
@@ -557,9 +558,8 @@ <h3>Document Conventions</h3>
 			</section>
 		</section>
 
-		<section id="getting-started">
+		<section id="getting-started" class="informative">
 			<h2>Getting started with Node Expressions</h2>
-			<p><em>This section is informative.</em></p>
 			<p>
 				A SHACL <a>shapes graph</a> can declare node expressions as values of various properties where dynamic computation is useful,
 				such as <code>sh:targetNode</code>, <code>sh:values</code> and <code>sh:deactivated</code>.
```
```suggestion
				such as <code>sh:targetNode</code>, <code>sh:values</code>, and <code>sh:deactivated</code>.
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
 
+		<section id="dynamic-shacl">
+			<h2>Dynamic SHACL</h2>
+			<p>
+				This section defines <dfn>Dynamic SHACL</dfn> as a dialect of SHACL that some implementations MAY support.
+			</p>
+			<p>
+				In Dynamic SHACL any <a>parameter</a> of a <a>constraint</a> can be a <a>node expression</a>,
+				excluding those that do <em>not</em> allow <a>blank nodes</a> (such as <code>sh:node</code>)
+				but including those that take <a>SHACL lists</a> as values (such as <code>sh:class</code>, <code>sh:datatype</code> and <code>sh:in</code>).
```
```suggestion
				but including those that take <a>SHACL lists</a> as values (such as <code>sh:class</code>,
				<code>sh:datatype</code>, and <code>sh:in</code>).
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+    sh:path ex:age ;
+    <b>sh:minInclusive [
+        shnex:if [
+            sparql:eq (
+                [ shnex:path ex:country ]
+                ex:USA
+            )
+        ]
+        shnex:then 35 ;
+        shnex:else 18 ;
```
TallTed, line 2689:
```suggestion
ex:PresidentShape
    a               sh:NodeShape ;
	sh:targetClass  ex:President ;
    sh:property     ex:PresidentShape-age ;
.
ex:PresidentShape-age
    a               sh:PropertyShape ;
    sh:path         ex:age ;
    <b>sh:minInclusive [
        shnex:if [
            sparql:eq (
                [ shnex:path ex:country ]
                ex:USA
            )
        ]
        shnex:then 35 ;
        shnex:else 18 ;
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+    ex:street "123 John Muir Ave" ;
+    ex:country ex:USA ;
+    ex:state "AZ" ;
+.
+ex:QueenslandAddress1
+    a ex:Address ;
+    ex:street "123 Bob Katter Cl" ;
+    ex:country ex:Australia ;
+    ex:state "QLD" ;
+.
```
TallTed, line 2719:
```suggestion
ex:ArizonaAddress1
    a          ex:Address ;
    ex:street  "123 John Muir Ave" ;
    ex:country ex:USA ;
    ex:state   "AZ" ;
.
ex:QueenslandAddress1
    a          ex:Address ;
    ex:street  "123 Bob Katter Cl" ;
    ex:country ex:Australia ;
    ex:state   "QLD" ;
.
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+				<aside class="example" title="The valid states can be attached to each country">
+					<div class="data-graph">
+						<div class="turtle">
+ex:Australia
+    a ex:Country ;
+    ex:stateCode "ACT", "NSW", "NT", "QLD", "SA", "TAS", "VIC", "WA" ;
+.
+ex:USA
+    a ex:Country ;
+    ex:stateCode "AL", "AK", "AZ", "AR" ; # ...
```
```suggestion
ex:Australia
    a            ex:Country ;
    ex:stateCode "ACT",
                 "NSW",
                 "NT",
                 "QLD",
                 "SA",
                 "TAS",
                 "VIC",
                 "WA" ;
.
ex:USA
    a            ex:Country ;
    ex:stateCode "AL",
                 "AK",
                 "AZ",
                 "AR" ; # ...
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+ex:Address
+    a sh:ShapeClass ;
+    sh:property ex:Address-state ;
+.
+ex:Address-state
+    a sh:PropertyShape ;
+    sh:path ex:state ;
+    <b>sh:in [
+        shnex:path ( ex:country ex:stateCode )
+    ]</b> .
```
TallTed, line 2779:
```suggestion
ex:Address
    a           sh:ShapeClass ;
    sh:property ex:Address-state ;
.
ex:Address-state
    a           sh:PropertyShape ;
    sh:path     ex:state ;
    <b>sh:in [
        shnex:path ( ex:country ex:stateCode )
    ]</b> .
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,142 @@ <h3>sh:nodeByExpression</h3>
...
+							<pre class="jsonld">
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					During validation, a Dynamic SHACL engine will evaluate the path expression at <code>sh:in</code>
+					and use the resulting nodes as members of the allowed values.
+					So, when the value of <code>ex:country</code> is <code>ex:USA</code>, it will look up the
```
```suggestion
					Thus, when the value of <code>ex:country</code> is <code>ex:USA</code>, it will look up the
```

TallTed (changes requested):

shacl12-node-expr/index.html
HolgerKnublauch: I don't like that to be honest. Spacing makes sense for comments but if we do this here, we would need to change everything and the choice where to place the spaces is rather arbitrary.

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,153 @@ <h3>sh:nodeByExpression</h3>
 			
 		</section>
 
+		<section id="dynamic-shacl">
+			<h2>Dynamic SHACL</h2>
+			<p>
+				This section defines <dfn>Dynamic SHACL</dfn> as a dialect of SHACL that some implementations MAY support.
+			</p>
+			<p>
+				In Dynamic SHACL any <a>parameter</a> of a <a>constraint</a> can be a <a>node expression</a>,
```
```suggestion
				In Dynamic SHACL any <a>parameter</a> of a <a>constraint</a> can be computed using a <a>node expression</a>,
```

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,153 @@ <h3>sh:nodeByExpression</h3>
...
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					We want to express that the valid values of <code>ex:state</code> depend on the value of <code>ex:country</code>
+					at the given focus node.
+					For example, the valid values for country <code>ex:USA</code> would be <code>( "AL" "AK" "AZ" ... )</code>
+					while values values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
```
```suggestion
					while values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
```

recalcitrantsupplant (approved):
two minor comments

shacl12-node-expr/index.html
TallTed: There is some arbitrariness.

*Generally*, I insert optional LFs before any enclosure (braces, brackets, parentheses, etc.) and optional SPACEs to indent each level by 4 spaces. Then, within each line, I indent to make visual columnization of predicate, object, and comment, based on the widest value of the preceding element (i.e., objects begin one space further right than the wides predicate ends; comments begin one space further right than the wides object ends). Similar applies to nesting symbols, where I put the opener as the first non-whitespace character in a line, followed by a space and the first contained element, and I put the closer either on its own line, vertically aligned with the opener, *or* as the last character of the line with the last contained element, with one space preceding.

I was not rigid about these "rules" in my suggestions here; perhaps I should have been.

I should note that the previous suggestion that you reacted more positively to, toward getting opinions from others, indented all predicates and objects as well as comments.

HolgerKnublauch: @TallTed I am not happy when PR requests are blocked on trivial editorial matters. It is perfectly valid to have such discussions and either agree to disagree on matters of taste. But please don't mark these discussions as blockers, but as GitHub comments.

shacl12-node-expr/index.html
```diff
@@ -2647,6 +2647,153 @@ <h3>sh:nodeByExpression</h3>
...
+								TODO
+							</pre>
+						</div>
+					</div>
+				</aside>
+				<p>
+					We want to express that the valid values of <code>ex:state</code> depend on the value of <code>ex:country</code>
+					at the given focus node.
+					For example, the valid values for country <code>ex:USA</code> would be <code>( "AL" "AK" "AZ" ... )</code>
+					while values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
```
TallTed: Keeping with the surrounding content, this change should have been as below:
```suggestion
					while valid values for country <code>ex:Australia</code> would be <code>( "ACT" "NSW" "NT" "QLD" "SA" "TAS" "VIC" "WA" )</code>.
```

shacl12-node-expr/index.html
TallTed: @HolgerKnublauch — 

Please note that there is nothing about the GitHub UI that makes it obvious that "change requests" are "blockers", though they are certainly "requests for change" which should be affirmatively addressed *somehow*.

If I do _not_ mark them as "requests for change", perhaps marking them "comments" as you suggest, they can silently disappear without having been applied (because there's no direct update nor notification of such "resolve conversation" actions), leading to me and others reviewing the same problematic content twice or thrice or more. 

Typically (across some dozens if not hundreds of repos), my suggestions are applied without controversy, and there is no blockage.

When some controversy *is* present, GitHub makes it fairly trivial for a PR author (or anyone, really) to open a new issue based on any change request (or any other comment) — look to the `...` at the upper-right of any comment on an issue or PR. Opening such an issue is reasonable justification to mark the original conversation about the change request as resolved without the suggested change having been applied, and with the requested change not being lost.

shacl12-node-expr/index.html
TallTed: Closing a change request without comment is *not* a good way to "resolve conversation". Please do not do so in future. If you disagree with a change request, either reply to it in place, or use the `...` menu at upper-right to `Reference in a new issue`.

(I find that GitHub's default `Reference in a new issue` boilerplate is sub-optimal, and so I edit it when I do this, but the default is functional if imperfect, so it's fine to just take it as they give it.)

shacl12-node-expr/index.html
TallTed: Regarding the specifics of this change request, I feel that it falls into the [Priority of Constituencies](https://www.w3.org/TR/design-principles/#priority-of-constituencies), as it makes comprehension (substantially, I think) easier for the consuming reader at some (small, I think) cost to the Authors/Editors of the documents we're producing. You may disagree with this or any other specific suggestion. To my mind, each such controversial suggestion should then become a question for the WG writ large, whose decisions on such questions become implementation instructions for the Editors.

shacl12-node-expr/index.html
HolgerKnublauch: I think this needs to be escalated to the WG chairs or discussed in the next meeting. You are literally arguing about adding spaces in a non-normative example and the PR is still blocked because of that. May I suggest, if you feel this is important enough, *you* raise a general issue on formatting of all examples - not only this one here - and then a policy can be agreed upon. IMHO it is far more important that we get the specifications correct, and formatting can be fixed later, just before publishing for example.

simonstey (approved):

shacl12-node-expr/index.html
afs:
```
ex:PresidentShape
    a               sh:NodeShape ;
```
I don't think this is an improvement.

simonstey: maybe it looks better in the browser? 🤔

shacl12-node-expr/index.html
HolgerKnublauch: Raised as #528

shacl12-node-expr/index.html
HolgerKnublauch: BTW I used the ... context menu to create a new issue from this, yet there is no apparent backlink created here, so it needs to be done by hand: #528

shacl12-node-expr/index.html
TallTed: followup in https://github.com/w3c/data-shapes/issues/528

shacl12-node-expr/index.html
TallTed: https://github.com/w3c/data-shapes/issues/528

TallTed (approved):
followup in https://github.com/w3c/data-shapes/issues/528

Issue #498, opened 2025-08-11
Support User-Defined Node Expression Functions; Parametrization, "Partial Evaluation"


HolgerKnublauch:
Not sure if this was captured by another ticket but we had discussed the ability for SHACL users to define new Node Expression functions that are calling other functions under the hood. This would basically be parameterized functions that use sh:var to access input values in certain places.

afs:
Example?

afs:
Is this related to [currying functions](https://en.wikipedia.org/wiki/Currying)?
Or function definition?

```
# [ :myInc (3) ] ==> 4
:myInc sh:defun [
    sh:args ( [ sh:var "x" ] ) ;
    sh:functionBody [ sparql:plus ( [ sh:var "x" ] 1 ) ] ]
  ] .
```

taking inspiration from other places.

afs:
Collecting related issues and comments:

* #498
* #484
* #502
* https://github.com/w3c/data-shapes/pull/452#issuecomment-3173150839

Feel free to edit and add bullet points.

VladimirAlexiev:
I posted an issue **NodeExp composability / partial evaluation / parametrization** and Holger suggested I move it here.

(In that issue Andy pointed to: https://github.com/w3c/data-shapes/pull/452#issuecomment-3173150839, which is the last bullet above).

Below are some vague ruminations on
- How composable are NodeExprs?
- Are they amenable to partial evaluation?
- How parameterizable are they?
- Is there some Combinator or Lambda calculus hidden behind them?

SPIN and SHACL 1.0 Constraint Components do this sort of stuff, but I don't see them in the current draft https://raw.githack.com/w3c/data-shapes/master/shacl12-node-expr/index.html

**Use case 1**: Composition
(Inspired by @HolgerKnublauch example in [#484](https://github.com/w3c/data-shapes/issues/484#issuecomment-3170494909))
`shnex:filterShape` filters a nodeset by a constraint. Assume a lot of the time that will be a single `sh:property` constraint.
- Will it be possible to define a custom shortcut like
`ex:filterProperty [...]` that expands to `shnex:filterShape [sh:property [ ... ]]`

**Use case 2**: Parametrization
(Inspired by https://raw.githack.com/w3c/data-shapes/master/shacl12-node-expr/index.html#EstonianCompanyShapeExample)
- How to parameterize a generic shape "Organization in Country" with `<type>, <country>`
[...]

PR #511, opened 2025-08-15
#455: Integrate SPARQL Functions

### LINKED ISSUE #455: Support all SPARQL functions as Node Expressions
As discussed in #367 there is an RDF/Turtle file at https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl that enumerates all SPARQL functions and establishes URIs for them. Those should all become List Parameter Functions in SHACL.

For example:

```
[ sparql:plus ( 38 4 ) ]
```

would be produce the result of 38 + 4 as a list with one RDF node.

The Node Expressions document could contain an auto-populated HTML table listing all these functions with their description and URI, and then a general description of how this table is used by SHACL engines.

simonstey:
Closes #455

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-455/shacl12-node-expr/index.html)


simonstey:
Do we want to have the SPARQL function sections being populated dynamically (from https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl ) or just updated on demand?

HolgerKnublauch:
I am afraid the proper process would be:

1) We create a new TTL file that owl:imports the base file
2) In that file we add triples that are useful for rendering, e.g.:
    a) grouping by category
    b) deep links into the SPARQL spec
3) Then the content can be created dynamically, if we have a TTL parser available.
4) Maybe the TTL will be accepted by the sparql main document too, but I am not sure about the update cycle.

This is of course a lot of work but will be paying off on the long term because the info will be machine readable, e.g. for syntax-directed editors.

BTW the diff is huge - what happened?

simonstey:
> I am afraid the proper process would be:
> 
>     1. We create a new TTL file that owl:imports the base file
> 
>     2. In that file we add triples that are useful for rendering, e.g.:
>        a) grouping by category
>        b) deep links into the SPARQL spec
> 
>     3. Then the content can be created dynamically, if we have a TTL parser available.
> 
>     4. Maybe the TTL will be accepted by the sparql main document too, but I am not sure about the update cycle.
> 
> 
> This is of course a lot of work but will be paying off on the long term because the info will be machine readable, e.g. for syntax-directed editors.
> 
> BTW the diff is huge - what happened?

I messed up the formatting.. will fix that next

afs:
> 2. deep links into the SPARQL spec

If contributed back to RDF & SPARQL WG, I'm sure that could get incorporated into sparql-ns.

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,901 @@
...
+@prefix shnex: <http://www.w3.org/ns/shnex#> .
+@prefix shnex-sparql: <http://www.w3.org/ns/shnex-sparql#> .
+@prefix sparql:  <http://www.w3.org/ns/sparql#> .
+
+shnex-sparql:
+	a owl:Ontology ;
+	rdfs:label "W3C SHACL SPARQL Functions Vocabulary"@en ;
+	rdfs:comment "This vocabulary defines SPARQL functions for use with SHACL."@en ;
+	sh:declare [
+		sh:prefix "shnex" ;
```
HolgerKnublauch: Should declare sparql instead

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,901 @@
...
+# the SPARQL local name (e.g., sparql:str -> #func-str). Where a function had a
+# more specific anchor in the source, the same target was used. (This comment
+# documents the fragment pattern used for constructed links.)
+
+# Imported SPARQL function terms (rdf:type, rdfs:label, dc:description, shnex:category, rdfs:isDefinedBy)
+# Category values: algebraic, numeric, string, date, aggregate, other
+
+sparql:plus
+	a sparql:Function ;
+	rdfs:label "Plus"@en ;
```
HolgerKnublauch: I would use lower case for all these labels.

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,901 @@
...
+	dc:description "This operator adds two numeric expressions and returns their sum."@en ;
```
HolgerKnublauch: I would use rdfs:comment - no need to introduce a new namespace just for that.

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,901 @@
...
+# as <http://www.w3.org/TR/sparql12-query/#func-LOCALNAME> where LOCALNAME is
+# the SPARQL local name (e.g., sparql:str -> #func-str). Where a function had a
+# more specific anchor in the source, the same target was used. (This comment
+# documents the fragment pattern used for constructed links.)
+
+# Imported SPARQL function terms (rdf:type, rdfs:label, dc:description, shnex:category, rdfs:isDefinedBy)
+# Category values: algebraic, numeric, string, date, aggregate, other
+
+sparql:plus
+	a sparql:Function ;
```
HolgerKnublauch: Pending our final design, I believe this would become sh:ListParameterExpressionFunction.

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,901 @@
...
+	rdfs:label "Plus"@en ;
+	dc:description "This operator adds two numeric expressions and returns their sum."@en ;
+	shnex:category "algebraic" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-plus> ;
+	shnex-sparql:example """[ sparql:plus ( 38 4 ) ]""" ;
```
HolgerKnublauch, line 75: Another nice property here would be infixOperator "+" that can be used to convert into SPARQL syntax.

shacl12-vocabularies/shnex-sparql.ttl
simonstey: this I took directly from https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl. Should I make them all `a sh:ListParameterExpressionFunction` instead?

shacl12-vocabularies/shnex-sparql.ttl
simonstey: addressed in https://github.com/w3c/data-shapes/pull/511/commits/753311e3dec073ec49740accf58ea07e3087f25c

shacl12-vocabularies/shnex-sparql.ttl
simonstey: addressed in https://github.com/w3c/data-shapes/pull/511/commits/530b33dc38d504091a4ed2bc4bd521855418f680

shacl12-vocabularies/shnex-sparql.ttl
simonstey: addressed in https://github.com/w3c/data-shapes/pull/511/commits/530b33dc38d504091a4ed2bc4bd521855418f680

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: I believe yes, but we may change the final design later. We could also leave it like this and make sparql:Function a subclass of sh:ListParameterExpressionFunction. So for now, maybe just leave it.

shacl12-vocabularies/shnex-sparql.ttl
simonstey: added infix property in https://github.com/w3c/data-shapes/pull/511/commits/fdfad059690e0b0d31b8a4deb10b8d12092dbc62 and added it to selected functions according to https://www.w3.org/TR/sparql12-query/#OperatorMapping

HolgerKnublauch: Excellent job, mate!

afs:
SPARQL functions work on arguments that are single RDF terms.  In `[ sparql:agg-count ( ?values ) ]` there isn't a term for `?values`. (The pain of lists not being first class in RDF.)

SPARQL Aggregates tied into the way `GROUP BY` works (e.g. `SUM(?X+1)`) to produce sets - the design means the lists are solely within aggregates a don't leak out to general evaluation. SPARQL Aggregates are not directly usable as node expressions.

shnex has it's own aggregates and that's probably the way to go - I think that defining an adaption layer to get SPARQL aggregates to work is not that different from shnex's aggregates so there is no advantage to SPARQL aggregates.

shnex aggregates working on lists which integrates with the way shnex is e.g. `sh:values`.

`[ sparql:coalesce ( ?optionalValue "default" ) ]` is n-ary including a "no result" error.
It is also special because of it's error handling - ditto `logical-and`, and `logical-or`.

I don't see that node expressions can have variables other than the environment : https://github.com/w3c/data-shapes/issues/498#issuecomment-3193749433

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+			sparql:FunctionalForm
+		)
+	] ;
+	rdfs:range xsd:string ;
+	.
+
+shnex-sparql:infixOperator
+	a rdf:Property ;
+	rdfs:label "infix operator"@en ;
+	rdfs:comment "Indicates the infix operator symbol used in SPARQL syntax for this function, enabling conversion from functional form to infix notation."@en ;
```
afs, line 48: Where is infix used in node expressions?

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	a sparql:Function ;
+	rdfs:label "plus"@en ;
+	rdfs:comment "This operator adds two numeric expressions and returns their sum."@en ;
+	shnex-sparql:category "algebraic" ;
+	shnex-sparql:infixOperator "+" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
+	shnex-sparql:example """[ sparql:plus ( 38 4 ) ]""" ;
+	.
+
+sparql:subtract
```
afs, line 78: This has the same URI as the entry in `/ns/sparql.ttl`.

But both graphs have triples with subject `sparql:subtract` and properties `rdfs:comment` and `rdfs:isDefinedBy` with different object.

It would be better to only include the SHACL Node Expression additions.

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: It is not but the information is useful to capture IMHO, to produce human-readable syntax.

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: Yes, if necessary create a fresh property to avoid clashes.

shacl12-vocabularies/shnex-sparql.ttl
afs: That would also need infix precedence.

shacl12-node-expr/index.html
```diff
@@ -1,415 +1,747 @@
...
+				filterDiv.innerHTML = `<label for="sparql-category-filter">Filter by category: </label>`;
+				const select = document.createElement('select');
+				select.id = 'sparql-category-filter';
+				select.innerHTML = `<option value="All">All</option>` + categories.map(c => `<option value="${escapeHtml(c)}">${escapeHtml(c)}</option>`).join('');
+				filterDiv.appendChild(select);
+	
+				// Table
+				const table = document.createElement('table');
+				table.id = 'sparql-functions-table';
+				table.setAttribute('summary','List of SPARQL functions with description, spec link and category');
```
```suggestion
				table.setAttribute('summary','List of SPARQL functions with description, spec link, and category');
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	rdfs:comment "This form returns the first non-error, non-unbound value from a sequence of expressions."@en ;
+	shnex-sparql:category "other" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-coalesce> ;
+	shnex-sparql:example "[ sparql:coalesce ( ?optionalValue \"default\" ) ]" ;
+	.
+
+sparql:filter-exists
+	a sparql:FunctionalForm ;
+	rdfs:label "filter exists"@en ;
+	rdfs:comment "This form tests whether a given pattern exists for each solution."@en ;
```
TallTed:
```suggestion
	rdfs:comment "This form tests whether a given pattern exists in each solution."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	shnex-sparql:category "other" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-filter-exists> ;
+	shnex-sparql:example "[ sparql:filter-exists { ?this foaf:name ?name } ]" ;
+	.
+
+sparql:filter-not-exists
+	a sparql:FunctionalForm ;
+	rdfs:label "filter not exists"@en ;
+	rdfs:comment "This form tests whether a given pattern does not exist for each solution."@en ;
```
TallTed:
```suggestion
	rdfs:comment "This form tests that a given pattern does not exist in each solution."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	shnex-sparql:example """[ sparql:langMatches ( "en-US" "en" ) ]""" ;
+	.
+
+sparql:regex
+	a sparql:Function ;
+	rdfs:label "regex"@en ;
+	rdfs:comment "This function tests whether a string matches a regular expression pattern, optionally with a specified flag (e.g., i for case-insensitive)."@en ;
+	shnex-sparql:category "string" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-regex> ;
+	shnex-sparql:example """[ sparql:regex ( "hello123" "[0-9]+" ) ]""" ;
```
TallTed, line 591: Where does the "specified flag" go? It seems this should have either multiple examples, or one that includes the optional "specified flag".

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	rdfs:comment "This function returns the month component of an xsd:dateTime or xsd:date."@en ;
+	shnex-sparql:category "date" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-month> ;
+	shnex-sparql:example """[ sparql:month ( "2023-12-25T10:30:00"^^xsd:dateTime ) ]""" ;
+	.
+
+sparql:day
+	a sparql:Function ;
+	rdfs:label "day"@en ;
+	rdfs:comment "This function returns the day component of an xsd:dateTime or xsd:date."@en ;
```
TallTed, line 699:
```suggestion
	rdfs:comment "This function returns the day (1–31) component of an xsd:dateTime or xsd:date."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	rdfs:comment "This function returns the year component of an xsd:dateTime or xsd:date."@en ;
+	shnex-sparql:category "date" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-year> ;
+	shnex-sparql:example """[ sparql:year ( "2023-12-25T10:30:00"^^xsd:dateTime ) ]""" ;
+	.
+
+sparql:month
+	a sparql:Function ;
+	rdfs:label "month"@en ;
+	rdfs:comment "This function returns the month component of an xsd:dateTime or xsd:date."@en ;
```
TallTed, line 689:
```suggestion
	rdfs:comment "This function returns the month component (1–12) of an xsd:dateTime or xsd:date."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,957 @@
...
+	rdfs:comment "This function returns the minute component (0–59) of an xsd:dateTime value."@en ;
+	shnex-sparql:category "date" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-minutes> ;
+	shnex-sparql:example """[ sparql:minutes ( "2023-12-25T10:30:00"^^xsd:dateTime ) ]""" ;
+	.
+
+sparql:seconds
+	a sparql:Function ;
+	rdfs:label "seconds"@en ;
+	rdfs:comment "This function returns the second component (0–60, including leap seconds) of an xsd:dateTime value."@en ;
```
TallTed: If this is "including leap seconds", the range must reach 61 (possibly higher; I do not know the maximum number of leap seconds that might be applied in any given year).
```suggestion
	rdfs:comment "This function returns the seconds component (0 ≤ seconds < 60, including leap seconds) of an xsd:dateTime value."@en ;
```

afs: This counting from zero.
`60` would be the leap  second.

But `xsd:dateTime` does not support leap seconds.
"a decimal number greater than or equal to 0 and less than 60."
Discussion: https://www.w3.org/TR/xmlschema11-2/#d-t-values

afs:
s/second/seconds/ to reduce the ambiguity of "the second component of".

shacl12-vocabularies/shnex-sparql.ttl
simonstey:
```suggestion
	rdfs:comment "This function returns the month component (1-12) of an xsd:dateTime or xsd:date."@en ;
```

shouldn't it be 1-12 then?

shacl12-vocabularies/shnex-sparql.ttl
simonstey:
```suggestion
	rdfs:comment "This function returns the day (1-31) component of an xsd:dateTime or xsd:date."@en ;
```
shouldn't it be 1-31 then?

shacl12-vocabularies/shnex-sparql.ttl
simonstey: in which case it should actually be:

```suggestion
	rdfs:comment "This function returns the second (0–59) component of an xsd:dateTime value."@en ;
```

shacl12-node-expr/index.html
simonstey: addressed in https://github.com/w3c/data-shapes/pull/511/commits/5479885720df5e81608f9016dfd64a98250a6110

shacl12-vocabularies/shnex-sparql.ttl
simonstey: addressed in  https://github.com/w3c/data-shapes/pull/511/commits/5479885720df5e81608f9016dfd64a98250a6110

shacl12-vocabularies/shnex-sparql.ttl
simonstey: addressed in  https://github.com/w3c/data-shapes/pull/511/commits/5479885720df5e81608f9016dfd64a98250a6110

shacl12-vocabularies/shnex-sparql.ttl
simonstey: > That would also need infix precedence.

which means? 🤔

shacl12-vocabularies/shnex-sparql.ttl
simonstey: @afs 
> This has the same URI as the entry in /ns/sparql.ttl.

what has the same URI? `shnex-sparql:example` ?

shacl12-vocabularies/shnex-sparql.ttl
simonstey:
```suggestion
	shnex-sparql:example """[ sparql:regex ( "HeLLo123" "^hello" "i" ) ]""" ;
```

shacl12-vocabularies/shnex-sparql.ttl
afs: `59.5` is legal.

simonstey: so `[0,60)` then? 

what about removing the value range altogether?
```suggestion
	rdfs:comment "This function returns the seconds component of an xsd:dateTime value."@en ;
```

afs: > what about removing the value range altogether?

+1

HolgerKnublauch:
@simonstey this will be a great contribution, but the PR has been in draft state and unchanged for several weeks. Do you need input/help to finish it?

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -247,7 +247,7 @@ sparql:filter-exists
 sparql:filter-not-exists
 	a sparql:FunctionalForm ;
 	rdfs:label "filter not exists"@en ;
-	rdfs:comment "This form tests whether a given pattern does not exist for each solution."@en ;
+	rdfs:comment "This form tests that a given pattern does not exist in each solution."@en ;
```
TallTed: Upon re-reading, I'm not sure what is meant by this.

Is it testing that each (i.e., _every_) solution does not contain the given pattern?

Or that _some_ solution(s) do not contain the given pattern?

Or something else?

Probably, some rephrasing such as I've done here should be applied.

shacl12-vocabularies/shnex-sparql.ttl
TallTed: Updated my suggestion with `0 ≤ seconds < 60`

afs: Remove the range. The function returns the seconds component of a valid xsd:dateTime(Stamp).

simonstey:
@HolgerKnublauch 

> @simonstey this will be a great contribution, but the PR has been in draft state and unchanged for several weeks. Do you need input/help to finish it?

whoops, sry.. I'll have this PR merge-ready by next Monday!

simonstey:
@HolgerKnublauch 
I rebased the document and fixed various HTML/merge commit issues. Should be good to go now. fyi, currently I only render the SPARQL functions as a table rather than fully fledged sections.

HolgerKnublauch:
Thanks a lot, Simon! I have made mostly cosmetic updates to your branch, which are hopefully OK.

We need to decide what to do with aggregates though. They are very different from the other functions as the don't work as in SPARQL: In a SPARQL query you would use them with a variable while in Node Expressions, the processing would work best if they receive another node expression that produces the aggregate values.

This however does not match how the other functions here work, which receive only individual values as arguments, never a list of values. Also, the shnex namespace already contains aggregates.

I suggest we delete the aggregates from our sparql: namespace.

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -0,0 +1,958 @@
...
+	shnex-sparql:category "other" ;
+	shnex-sparql:infixOperator "<" ;
+	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
+	shnex-sparql:example """[ sparql:less-than ( 5 10 ) ]""" ;
+	.
+
+sparql:greater-than-or-equal
+	a sparql:Function ;
+	rdfs:label "greater than or equal"@en ;
+	rdfs:comment "This operator tests whether the first RDF term is greater or equal to the second RDF term."@en ;
```
TallTed:
```suggestion
	rdfs:comment "This operator tests whether the first RDF term is greater than or equal to the second RDF term."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
TallTed:
```suggestion
	rdfs:comment "This function returns the seconds component of a valid xsd:dateTime or xsd:dateTimeStamp value."@en ;
```

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: I guess Andy's comment was about the subjects such as sparql:subtract. Both documents use rdfs:comment and rdfs:isDefinedBy. But IMHO those values should align. So if we in the SHACL WG find incorrect comments, the same fix should be applied to the /ns/sparql.ttl

Having said this, I don't think we need to owl:import the /ns/sparql.ttl graph so it will unlikely affect anyone.

In the interest of moving forward, I'll mark this as resolved but please reopen if anyone disagrees.

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: Note: these comments were copied from the original source at https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl and should be addressed there.

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: Note: these comments were copied from the original source at https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl and should be addressed there.

shacl12-vocabularies/shnex-sparql.ttl
HolgerKnublauch: This functional form does not fit into the node expression execution model, so it was removed alongside filter-exists, in and not-in. Please reopen a fresh issue if anyone disagrees.

HolgerKnublauch (approved):
To me it's good enough for now. Further refinements can and will happen in fresh tickets, when needed.

simonstey:
@robert-david

robert-david (approved):

robert-david:
@simonstey should we specify what is returned for each of the sparql functions to streamline the descriptions? Currently, sometimes it is stated and sometimes not. For some functions, it might not be clear.

shacl12-vocabularies/shnex-sparql.ttl
robert-david: @afs @simonstey i think this is still open?

HolgerKnublauch: The infix predicate is optional and technical and shouldn't block this PR. We can turn it into a fresh issue if we want to pursue it forward.

PR #512, opened 2025-08-15
Rearranged node expression (meta) classes and added the current funct…

### LINKED ISSUE #504: Declare all Node Expression types in TTL
All built-in node expression types such as shnex:path and shnex:filterShape need proper declarations in the TTL vocabulary, at least the properties.

HolgerKnublauch:
…ions to shnex.ttl

sh:Expression/NodeByExpression and annotations were just moved around within the document

This is quite geeky stuff due to the meta-levels: 

- The metaclasses such as sh:NamedParameterExpressionFunction
- The actual node expression functions are then instances of sh:NamedParameterExpression, these declare the parameters such as sh:filterName and sh:nodes
- The instances of those are the blank nodes (with values for sh:filterShape), although their rdf:type is usually never asserted

I hope I got these metalevels correct, so please double check anyone :)

Closes #504

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-504/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-vocabularies/shacl.ttl
+++ b/shacl12-vocabularies/shacl.ttl
@@ -1011,26 +1011,6 @@ sh:node
 	rdfs:isDefinedBy sh: .
 
 
-sh:NodeByExpressionConstraintComponent
-	a sh:ConstraintComponent ;
-	rdfs:label "Node by expression constraint component"@en ;
-	rdfs:comment "A constraint component that can be used to verify that all value nodes conform to the node shape(s) produced by the given node expression."@en ;
-	sh:parameter sh:NodeByExpressionConstraintComponent-nodeByExpression ;
-	rdfs:isDefinedBy sh: .
-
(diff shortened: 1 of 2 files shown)


TallTed (commented):
No obvious issues, to my eye.

HolgerKnublauch:
I would appreciate *some* approval of this PR because it is blocking other PRs that also require edits to the shnex.ttl file. @recalcitrantsupplant @simonstey @robert-david

recalcitrantsupplant (approved):
Looks good to me. My understanding here is the spec defined named parameter functions are both instances of sh:NamedParameterExpressionFunction, and also subclass sh:NamedParameterExpression, which enables explicit parameter linking from the key parameters (sh:Parameter), to the function using rdfs:domain.

HolgerKnublauch:
Thank you. It is possible that we revisit the metaclasses, but at least we now have the URIs and parameter declarations in TTL form.

PR #513, opened 2025-08-16
#498: Started with examples

### LINKED ISSUE #498: Support User-Defined Node Expression Functions; Parametrization, "Partial Evaluation"

HolgerKnublauch:
Closes #498

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-498/shacl12-node-expr/index.html#custom-node-expressions)


recalcitrantsupplant:
Looking at the first example, is there a way the link between the function call and the function (`ex:average`) can be made more explicit?

Something like:

```turtle
ex:CompanyShape-averageIncome
    a sh:PropertyShape ;
    sh:path ex:averageIncome ;
    sh:datatype xsd:decimal ;
    sh:values [
        shnex:call ex:AverageExpression ;
        ex:average [
            shnex:path ( ex:employee ex:income )
        ]
    ]
```

Otherwise I assume an implementation knows it is a function not an IRI based on the syntax (+ possibly a function register).

HolgerKnublauch:
Yes, for all these expression functions, there can be a triple such as `ex:average rdfs:domain ex:AverageExpression` and then `[ ... rdf:type ex:AverageExpression ]` which IMHO makes a lot of sense. We can state that `rdf:type` triples are optional but allowed.

HolgerKnublauch:
I have implemented the named parameter functions in our product and it appears to work. I am yet to confirm the list parameter syntax but don't expect problems.

robert-david (approved):
5.3 Arg Expressions should provide an example.

PR #524, opened 2025-08-24
#178: Added sh:someValue

### LINKED ISSUE #178: Add something similar to `dash:hasValueWithClass`
OWL has this `owl:someValuesFrom` construct that is quite popular to declare relationships between classes. To express this in SHACL, one would need to create a verbose `qualifiedValueShape` with `minCount 1` and a `sh:class` in an inner node shape.

I suggest to add something like [`dash:hasValueWithClass`](https://datashapes.org/constraints.html#HasValueWithClassConstraintComponent) to SHACL Core.

### LINKED ISSUE #485: why are `sh:minListLength sh:maxListLength sh:uniqueMembers` "shape-expecting"?
https://raw.githack.com/w3c/data-shapes/master/shacl12-core/index.html#shapes-recursion
>The following properties are the so-called shape-expecting constraint parameters
...
sh:minListLength
sh:maxListLength
sh:uniqueMembers

I think these 3 should not be in the list

HolgerKnublauch:
Also closing #485 while I was there

Closes #178
Closes #485

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-178/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2853,9 +2853,7 @@ ex:SomeClass-alternativePathExample
 						<li><a href="#QualifiedValueShapeConstraintComponent">sh:qualifiedValueShape</a></li>
 						<li><a href="#NodeConstraintComponent">sh:node</a></li>
 						<li><a href="#MemberShapeConstraintComponent">sh:memberShape</a></li>
-						<li><a href="#MinListLengthConstraintComponent">sh:minListLength</a></li>
-						<li><a href="#MaxListLengthConstraintComponent">sh:maxListLength</a></li>
-						<li><a href="#UniqueMembersConstraintComponent">sh:uniqueMembers</a></li>
+						<li><a href="#SomeValueConstraintComponent">sh:someValue</a></li>
 						<li><a href="#XoneConstraintComponent">sh:xone</a></li>
 					</ul>
 					<p>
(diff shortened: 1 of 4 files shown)


shacl12-core/index.html
```diff
@@ -6413,6 +6411,98 @@ <h4>sh:property</h4>
...
+									The <a>shape</a> that at least one of the <a>value nodes</a> needs to conform to.
+									<span data-syntax-rule="someValue-shape">The values of <code>sh:someValue</code> in a shape must be <a>well-formed</a> <a>shapes</a>.</span>
+								</td>
+							</tr>
+						</tbody>
+					</table>
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="Node">
+							Let <code>$someValue</code> be a <a>value</a> of <code>sh:someValue</code>.
```
ajnelson-nist, line 6449: Does `$someValue` need to be listed in the table of pre-bound SPARQL variables?  SHACL-SPARQL Section 2.3.2 today.

https://github.com/w3c/data-shapes/blob/7d2d4c28864eea174c5f46eba7e0f0b92c890ea9/shacl12-sparql/index.html#L1061-L1062

ajnelson-nist: Less-durable link to today's rendered SHACL-SPARQL document:

https://www.w3.org/TR/shacl12-sparql/#sparql-constraints-variables

shacl12-core/index.html
```diff
@@ -6413,6 +6411,98 @@ <h4>sh:property</h4>
...
+							A <a>failure</a> MUST be produced if the <a>conformance checking</a> of any <a>value node</a> against
+							<code>$someValue</code> produces a <a>failure</a>, unless at least one of the <a>value nodes</a> conforms to <code>$someValue</code>.
```
ajnelson-nist, line 6451: Is this right?  If the existence condition by some node is satisfied, a **failure** is overridden?  Is this conflating failure and violation?

ajnelson-nist: Oh, I see, the informative paragraph explains this.  This feels ... fragile, at first consideration.

shacl12-core/index.html
```diff
@@ -6413,6 +6411,98 @@ <h4>sh:property</h4>
...
+						</div>
+						<div class="shaclc"><pre class="shaclc">TODO</pre></div>
+					</aside>
+					<p>
+						<code>sh:someValue</code> can be regarded as syntactic sugar for a combination of
+						<code>sh:qualifiedValueShape</code> and <code>sh:qualifiedMinCount 1</code>.
+					</p>
+					<p>
+						Also note that <code>sh:someValue</code> mostly makes sense when used in <a>property shapes</a>.
+						In <a>node shapes</a>, <code>sh:someValue</code> is equivalent to using <code>sh:node</code> or <code>sh:property</code>.
```
ajnelson-nist, line 6507: I think we need to dig into this a bit more semantically.  If I use `sh:someValue` on a node shape, could I be expressing that I just want there to be at least one conformant node in the graph?  "Equivalent to using `sh:node`" feels like a _universal_ constraint rather than an _existential_ constraint.

```turtle
ex:DuckFarmerShape2
    a sh:NodeShape ;
    sh:someValue [
        sh:property [
            sh:path ex:tendsAnimal ;
            sh:class ex:Duck ;
        ]
    ] .
```

vs.

```turtle
ex:DuckFarmerShape3
    a sh:NodeShape ;
    sh:node [
        sh:property [
            sh:path ex:tendsAnimal ;
            sh:class ex:Duck ;
        ]
    ] .
```

`ex:DuckFarmerShape2` and `ex:DuckFarmerShape3` are not equivalent.

ajnelson-nist (commented):
I'm not ACKing or NACKing yet.  I think there's one change to consider (`someValue` on a `NodeShape`), and some mulling for the group to do about the failure handling.  On the one hand, short-circuiting is important for performance; on the other hand, masking failures does not feel immediately right, to me at least.

shacl12-core/index.html
HolgerKnublauch: Failures are very rare, so I don't think this should block performance optimizations. Because right now, it can stop evaluating once it has found a match, otherwise it will always be N validations. Not a show stopper, in either case though, as long as the feature can go in at all. We could, for example, accept the current draft with a note that we welcome feedback.

shacl12-core/index.html
HolgerKnublauch: No, this has nothing to do with SPARQL here. I only used $someValue because I think we should do that everywhere. I started actually doing this with $focusNode in the revised pathValues definition, because the name otherwise clashed with the variable focusNode.

HolgerKnublauch:
In fact, why shouldn't it be OK to ignore failures. The condition here is that just one of the items has to conform. Whether others cannot be validated doesn't matter as long as one positive example was found.

shacl12-core/index.html
HolgerKnublauch: Can you construct an example of data where these two are NOT equivalent? In my opinion they have the same effect. The value nodes of a node shape are the focus node itself (list of 1). If we state that at least one of them needs to conform to the given shape (using sh:someValue) then we basically say that *all* one of them need to conform.

TallTed:
It might be clearer to say "The condition here is that <strike>just</strike> _at least_ one of the items has to conform."

shacl12-core/index.html
bergos: @HolgerKnublauch the informative part matches `sh:or`, but there is a mismatch in the textual definition. That should be aligned. Your definition for `sh:someValue` matches the informative parts, and probably also the reality of current implementations. I can confirm that shacl-engine skips any validations for `sh:or` after the first successful validation, unless debug mode is enabled.

bergos (approved):
The textual definition of `sh:or` should be aligned. That can be done in this PR or separate PR.

shacl12-core/index.html
HolgerKnublauch: Recorded as #553 which should be handled independently of this one here.

I will merge this PR into master after the next meeting, unless I hear otherwise.

YoucTagh (approved):

PR #525, opened 2025-08-25
#514: Renamed shnex:path to shnex:pathValues and switched from shnex:nodes to shnex:focusNode

### LINKED ISSUE #514: Rename shnex:path to shnex:pathValues and change its shnex:nodes to shnex:focusNode
Discussed here https://github.com/w3c/data-shapes/issues/484#issuecomment-3194163351

The first suggestion is to rename shnex:path to shnex:pathValues.

The second is to replace shnex:nodes with shnex:focusNode. This is to modify the focus node, which defaults to the currently active one. The current design allows multiple nodes and I guess few people will understand what that means and use cases are likely rather light. It also raises questions such as how are these values concatenated into a single list. It would make the language more predictable if there is always just one focus node.

HolgerKnublauch:
Closes #514

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-514/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -600,8 +600,8 @@ ex:EstonianCompanyShape
 			<p>
 				The following diagram illustrates how this node expression is interpreted, from a logical point of view.
 				During validation, a SHACL processor will determine the <a>target</a> nodes of the shape
-				by evaluating the <a>filterShape expression</a>.
-				However, the filterShape expression first evaluates its input expression, which is specified via <code>sh:nodes</code>
+				by evaluating the <a>filter shape expression</a>.
+				However, the filter shape expression first evaluates its input expression, which is specified via <code>sh:nodes</code>
 				and is an <a>instancesOf expression</a>.
 				This will produce all instances of the given class, <code>ex:Company</code>.
 				The <code>shnex:filterShape</code> is then applied to all of these instances, to keep only the companies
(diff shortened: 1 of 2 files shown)


recalcitrantsupplant (approved):
No issues

robert-david (approved):

shacl12-node-expr/index.html
```diff
@@ -859,14 +859,14 @@ <h3>List Parameter Functions</h3>
 	sh:values <b>[
 		ex:coalesce (
 			[
-				# This is a path expression that is expected to return zero or one values
-				shnex:path ex:fullName ;
+				# This is a path values expression that is expected to return zero or one values
+				shnex:pathValues ex:fullName ;
```
TallTed, line 863: I think this may be the root of the difference between `pathValues` and `pathValue`. "zero or one values" leads to the troublesomely indeterminate expansion of "zero values or one value". `zero pathValues` reads better than `zero pathValue`, while `one pathValue` reads better than `one pathValues`. (`no pathValue` and `no pathValues` both read OK.)

It's too bad parentheses are problematic in predicate identifiers, e.g., `shnex:pathValue(s)` and "zero or one value(s)". I think these might be acceptable to all, and perhaps the problems are small enough to likewise accept.

Similar discussion applies to `focus node`, `focus nodes`, and `focus node(s)`.

HolgerKnublauch: I am usually in favor of singular names for RDF properties, so (has) child over children. But here this is similar to sh:values which also produces multiple values as a list. I hope I can proceed here, as this has two approvals now and other branches are waiting for this change. If you still feel this issue is critical enough, feel free to open a new ticket.

TallTed: Fair enough.

PR #531, opened 2025-08-27
Issue 490 images updated

### LINKED ISSUE #490: Move Node Expressions into their own namespace

HolgerKnublauch:
Closes #490

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-490-images/shacl12-node-expr/index.html)


robert-david (approved):

TallTed:
If at all possible, all current `png` images should be replaced with `svg` images. Among other reasons, `svg` are far more friendly to visual assistance tools, about which we're sure to hear much during accessibility review.

PR #549, opened 2025-08-28
Profiles of and Profiling with starts

nicholascar:
A start on the two main Profiling sections.

See https://raw.githack.com/w3c/data-shapes/pro-of-sh-sects/shacl12-profiling/index.html

### DIFF
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -813,7 +813,7 @@
           Base profiles of SHACL are profiles that correspond to the individual SHACL 1.2 specifications listed in the <a href="#specifications">Specifications</a> section, above.
         </p>
         <p>
-          Together, these Base Profiles cover all of SHACL and, when used together in the <a href="#union-profile">Union Profile</a>, they can validate that any graph is valid SHACL 1.2 data.
+          Together, these Base Profiles cover all of SHACL and, when used together in the <a href="#union-profile">Union Profile</a>, they can validate whether any graph is valid SHACL 1.2 data.
         </p>
       </section>
       <section id="non-base-profiles">
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -868,7 +868,7 @@
           A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement some, and potentially all the specification's rules, but no other rules, in SHACL.
         </p>
         <p>
-          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but do not provide a mechanism for data validation, such as a SHACL validator.
         </p>
         <div class="note">
           <p>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -888,12 +888,12 @@
 		<section id="security" class="appendix informative">
 			<h2>Security and Privacy Considerations</h2>
 			<p>
-				Like most RDF-based technologies, SHACL processors may operate on graphs that are combined
-				from various sources.  Some applications may have an open "linked data" architecture and dynamically
+				Like most RDF-based technologies, SHACL processors may operate on graphs that are assembled
+				from various sources.  Some applications may have an open "linked data" ("LD") architecture and dynamically
 				assemble RDF triples from sources that are outside an organization's network of trust.
 				Since RDF allows anyone to add statements about any resource, triples may modify the originally
-				intended semantics of shape definitions or nodes in a data graph and thus lead to misleading results.
-				Protection against this (and the following) scenario can be achieved by only using trusted
+				intended semantics of shape definitions or nodes in a data graph and thus feed into misleading results.
+				Protection against this (and the following) scenario is achievable by using only trusted
 				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
 				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
 			</p>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -872,7 +872,7 @@
         </p>
         <div class="note">
           <p>
-            While a null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
+            While null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
           </p>
           <p>
             An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constraints <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement those constraints in SHACL too.
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -819,7 +819,7 @@
       <section id="non-base-profiles">
         <h3>Non-base Profiles</h3>
         <p>
-          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, and also other profiles of parts of SHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
+          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, as well as other profiles of parts of SHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
         </p>
         <section id="union-profile">
           <h4>Union Profile</h4>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -850,7 +850,7 @@
           Referencing [[[dx-prof]]]'s <a href="https://www.w3.org/TR/dx-prof/#resource-roles-vocab">vocabulary of <em>Resource Role Instances</em></a>, it seems clear that SHACL graphs can be applied to the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a> within a profile and perhaps also to the role of <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>, since SHACL can be used to declare Node and Property Shapes just as OWL can be used to declare Classes and Properties.
         </p>
         <p>
-          SHACL resources can also document the constraints they implement validation shapes for, therefore they can also perform the <a href="https://www.w3.org/TR/dx-prof/#Role:constraints">Constraints</a> role.
+          SHACL resources can also document the constraints they implement validation shapes for, therefore they can also be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:constraints">Constraints</a> role.
         </p>
         <p>
           They could also be used for the <a href="https://www.w3.org/TR/dx-prof/#Role:mapping">Mapping</a> role, if SHACL Rules were implemented to map/transform data from one model to another.
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -844,7 +844,7 @@
           A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement potentially many and all the specification's rules, and no other rules, in SHACL.
         </p>
         <p>
-          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or just a written document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
         </p>
         <div class="note">
           <p>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -844,7 +844,7 @@
           SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the total set of things that a profile designers may wish to do, for example, they may want to include documentation about why and how the profile was created or provide new model elements - classes, predicates etc. - within an extended schema.
         </p>
         <p>
-          Since SHACL cannot be used for all possible profile parts, profile designers must look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is recommended for this.
+          Since SHACL cannot be used for all possible profile parts, profile designers need to look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is expected to be used for this.
         </p>
         <p>
           Referencing [[[dx-prof]]]'s <a href="https://www.w3.org/TR/dx-prof/#resource-roles-vocab">vocabulary of <em>Resource Role Instances</em></a>, it seems clear that SHACL graphs can perform the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a> within a profile and perhaps also <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>, since SHACL can be used to declare Node and Property Shapes just as OWL can declare Classes and Properties.
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -898,9 +898,9 @@
 				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
 			</p>
       <p>
-        When creating profiles of other specifications - base specifications - profile creators need to ensure that
-        their constraints are not in violation of the base specification's rules. If any were so, and if only the
-        profile but not the base specification was used to check for data validity, by accident or by design, invalid
+        When creating profiles of other specifications, profile creators need to ensure that
+        their constraints do not violate those specification's rules. If any did so, and if only the
+        profile's rules, but not the specification's rules, were used to check for data validity, by accident or by design,
         data could be wrongly calculated to be valid. This could lead to accidental data release or use, potentially
         introducing security issues.
       </p>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -851,7 +851,7 @@
             While a null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
           </p>
           <p>
-            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to the [[[prov-0]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constrains <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement these constrains in SHACL too.
+            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to the [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constrains <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement these constrains in SHACL too.
           </p>
         </div>
       </section>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -841,7 +841,7 @@
       <section id="profile-parts">
         <h3>Profile Part Roles</h3>
         <p>
-          SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the total set of things that a profile designers may wish to do, for example, they may want to include documentation about why and how the profile was created or provide new model elements - classes, predicates etc. - within an extended schema.
+          SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the complete set of things that a profile designer may wish to do; for example, they may want to include documentation about why and how the profile was created, or provide new model elements (classes, predicates, etc.) within an extended schema.
         </p>
         <p>
           Since SHACL cannot be used for all possible profile parts, profile designers need to look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is expected to be used for this.
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -853,7 +853,7 @@
           SHACL resources can also document the constraints they implement validation shapes for, therefore they can also be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:constraints">Constraints</a> role.
         </p>
         <p>
-          They could also be used for the <a href="https://www.w3.org/TR/dx-prof/#Role:mapping">Mapping</a> role, if SHACL Rules were implemented to map/transform data from one model to another.
+          They could also be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:mapping">Mapping</a> role, if SHACL Rules are implemented to transform data from one model to another.
         </p>
         <p>
           SHACL lists of values required for use with a model could perform the <a href="https://www.w3.org/TR/dx-prof/#Role:vocabulary">Vocabulary</a> role.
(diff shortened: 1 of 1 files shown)


shacl12-profiling/index.html
```diff
@@ -804,15 +804,57 @@ <h3>Document Conventions</h3>
...
+          Base profiles of SHACL are profiles that correspond to the individual SHACL 1.2 specifications listed in the <a href="#specifications">Specifications</a> section, above.
+        </p>
+        <p>
+          Together, these Base Profiles cover all of SHACL and, when used together in the <a href="#union-profile">Union Profile</a>, they can validate that any graph is valid SHACL 1.2 data.
+        </p>
+      </section>
+      <section id="non-base-profiles">
+        <h3>Non-base Profiles</h3>
+        <p>
+          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, and also other profiles of parts of sHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
```
ajnelson-nist:
```suggestion
          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, and also other profiles of parts of SHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
```

shacl12-profiling/index.html
```diff
@@ -804,15 +804,57 @@ <h3>Document Conventions</h3>
...
+			<p>
+        This section describes how to profile specifications with SHACL.
+      </p>
+      <section id="null-profiles">
+        <h3>Creating Null Profiles</h3>
+        <p>
+          A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement potentially many and all the specification's rules, and no other rules, in SHACL.
+        </p>
+        <p>
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or just a written document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
```
ajnelson-nist:
```suggestion
          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
```

shacl12-profiling/index.html
```diff
@@ -804,15 +804,57 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <div class="note">
+          <p>
+            While a null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
+          </p>
+          <p>
+            An example of a non-SHACL null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to [[[prov-o]]] data. Implementations of those constrains <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation.
```
ajnelson-nist: I don't think PROV-CONSTRAINTs vs. PROV-O is a good example of a _null_ profile.  It uses some temporal concepts that are not defined in PROV-O, such as "precedes," which might find an aligned property in OWL-Time [per this discussion](https://github.com/w3c/sdw/issues/1430).

I _do_ think PROV-CONSTRAINTS is an example of a profile of PROV-O, just not a "null" one.

nicholascar: Oh, ok. I thought that PROV-CONSTRIANTS didn't implement any constraints beyond those in the PROV models but I haven't checked this for a while. Are you sure "precedes" etc aren't implied in the PROV models? If they are then it would be a null profile. Will investigate.

Do you want to contribute your null profile of DCAT? You could put it in shacl-resources and we could indicate it as an example of a Null Profile implemented in SHACL

ajnelson-nist: I'm happy to discuss potential null profiles, but think we'd need to check a few assumptions on a call.

On "Precedes":

No PROV documents (I just looked through the 12) define an IRI for "precedes".

"precedes" is used in definitions as a temporal term, though without definition itself, in PROV-O and PROV-DM.  It's used in other documents with the same or lesser formality, or in examples pertaining to other non-temporal sequencing.

PROV-CONSTRAINTS defines "precedes" and "strictly precedes" [here](https://www.w3.org/TR/2013/REC-prov-constraints-20130430/#event-ordering-constraints), and is the only REC to do so.

PROV-SEM defines "precedes" and "strictly precedes" [here](https://www.w3.org/TR/2013/NOTE-prov-sem-20130430/#precedes-and-strictly-precedes) and uses it in many proofs.  However, this is a NOTE, and IRIs are not part of this document's purpose or outcomes.

So, because of PROV-CONSTRAINTS' formulations around the defined but unencoded predicate "precedes," a SHACL profile of PROV-O to conform to PROV-CONSTRAINTS would need to include a few unprovided ontology concepts (precedes, strictly precedes, and their inverses "follows").  It feels to me like a profile that needs to define new predicates would not be a "Null" profile.

nicholascar: I don't really think this is a problem: why should every constraint on a model be realized in specialized predicates? 

If you look at the implementation of the test for precedence cycles [here](https://github.com/pgroth/prov-check/blob/master/provcheck/provconstraints.py#L28-L35), you can see it uses a made-up predicate that is inserted into the graph by a number of `INSERT` queries first [here](https://github.com/pgroth/prov-check/blob/master/provcheck/provconstraints.py#L39-L523) calculated using timing and logical information from PROV-O data. So pure PROV-O data is the only needed input. We can have the same in SHACL using SHACL rules to do what the `INSERT` queries do.

I have added more info to the note to ensure the issue of the constraints testing for things implied but not stated in PROV-O is known.

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+			<p>
+        This section describes how to profile specifications with SHACL.
+      </p>
+      <section id="profile-parts">
+        <h3>Profile Part Roles</h3>
+        <p>
+          SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the total set of things that a profile designers may wish to do, for example, they may want to include documentation about why and how the profile was created or provide new model elements - classes, predicates etc. - within an extended schema.
+        </p>
+        <p>
+          Since SHACL cannot be used for all possible profile parts, profile designers must look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is recommended for this.
```
ajnelson-nist: > ... is recommended for this.

How much group review is required to put the word "Recommended" into this text?

ajnelson-nist: Is this whole profiling document going to frame itself as an application of dx-prof?  If so, I think we could bypass the discussion around whether the group is making any kind of recommendation.

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        <h3>Creating Null Profiles</h3>
+        <p>
+          A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement some, and potentially all the specification's rules, but no other rules, in SHACL.
+        </p>
+        <p>
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
+        </p>
+        <div class="note">
+          <p>
+            While a null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
```
ajnelson-nist:
```suggestion
            While null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+          </p>
+          <p>
+            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to the [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constrains <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement these constrains in SHACL too.
```
ajnelson-nist: A few nits:

```suggestion
            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constraints <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement those constraints in SHACL too.
```

ajnelson-nist: Separate from the grammar nits, I think this sentence at the end should be dropped:

 > It would be possible to implement those constraints in SHACL too.

I've been doing work around PROV-CONSTRAINTS for a while, and my current thinking is that not all of the time logic available to PROV-CONSTRAINTS and PROV-O can be done in SHACL or SPARQL alone.  One matter is, PROV-O uses `xsd:dateTime` in its literals, not `xsd:dateTimeStamp`.  Time zones are optional.  So, this leads to cases that can't be checked without supplemental information, such as the second example below:

```turtle
kb:Entity-1
  prov:generatedAtTime "2020-01-01T14:00"^^xsd:dateTime ;
  prov:invalidatedAtTime "2020-01-20T13:59"^^xsd:dateTime ;
  .
kb:Entity-2
  prov:generatedAtTime "2020-01-01T14:00"^^xsd:dateTime ;
  prov:invalidatedAtTime "2020-01-01T13:59"^^xsd:dateTime ;
  .
```

`kb:Entity-1` can be seen as consistent with PROV-CONSTRAINTS because no time zone situation lets the `prov:Generation` happen after the `prov:Invalidation`.  `kb:Entity-2` can be seen as consistent **if** the context outside these triples happens to include that the generation happened in one time zone and the invalidation happened 59 minutes later the next time zone over.  E.g., borrowing from the examples of `prov:atLocation`:

```turtle
kb:Entity-2
  prov:qualifiedGeneration [
    prov:atlocation <http://example.org/kb/Denver> ;  # UTC-7 in January
  ] ;
  prov:qualifiedInvalidation [
    prov:atlocation <http://example.org/kb/Seattle> ;  # UTC-8 in January
  ] ;
  .
```

I'm not sure how feasible it is to implement this in SHACL alone, even with the current working-group's development.  Some application logic would need to know PROV's vocabulary well enough to interpret locations, time zones for locations, and timezone-aware timestamps to compute temporal ordering.  Temporal ordering would provide an alternative implementation (or a source for imputed `:precedes` triples) of the constraint-in-English "Generation of an entity precedes its invalidation," supplementing the link-based implementation in [Constraint 36](https://www.w3.org/TR/prov-constraints/#generation-precedes-invalidation_text).

ajnelson-nist (commented):
Not ACKing or NACKing yet - this mostly looks good, but there is one sentence I think should be dropped, and a document-relationship to dx-prof worth clarifying.

shacl12-profiling/index.html
ajnelson-nist: Thank you for the text updates.  I'm pretty much content on this Conversation.  My remaining remark on feasibility in SHACL/SPARQL alone is now on a Conversation attached to the PROV Note box.

YoucTagh:
I believe this is a good starting point. I will validate it once we have discussed the concerns raised by @ajnelson-nist above.

shacl12-profiling/index.html
nicholascar: > How much group review is required to put the word "Recommended" into this text?

> Is this whole profiling document going to frame itself as an application of dx-prof? If so, I think we could bypass the discussion around whether the group is making any kind of recommendation.

The underlying issue for both these points is that DX PROF is only a Note, not a Rec. If it was a Rec, then of course we could/should recommend it as it would be the W3C's recommended way of working.

(I don't think there's any issue with DX PROF that prevents it from becoming a Rec as the content has proved useful and uncontroversial for a number of years, but within the Dataset Exchange WG, we encountered internal conflicts that ran out the clock on WG time to get it through the system)

So perhaps we should remove the word "Recommended" but then I'm uncertain as to how strongly all this work should be framed in relation to DX PROF. 

We can't directly update DX PROF to get it accepted as a Rec for now, so we will have to decide how strongly framed we are in relation to it. I think we have to accept notions of profiles having parts with roles to move ahead with shat SHACL can do - what roles it's performing - so if we don't just accept them directly from PROF, we will have to either re-implement the logic here or appeal to a conceptual logic from first principles that PROF could be seen as just a peer implementation of.

YoucTagh (approved):

shacl12-profiling/index.html
ajnelson-nist: "Recommended" seems outside of our ability to write here.

If this document (`shacl12-profiling`) gets revised to "Build atop" DX PROF with DX PROF as a Note, then is this document also limited to being a Note?

shacl12-profiling/index.html
ajnelson-nist: Unresolving conversation.  It was probably a technical accident waiting to happen from me putting two unrelated discussions atop one suggestion.  Apologies for the confusion.
I still think we should work through whether the last sentence should be dropped.

shacl12-profiling/index.html
TallTed: Note that there is also a `must` in this paragraph. It's best to avoid both `must` and `recommended` (among other RFC2119 keywords, even when not allcaps) in NOTEs and other non-normative content. Note that I've not wrapped these lines, so GitHub should present my suggested changes in proper highlighted style.
```suggestion
          Since SHACL cannot be used for all possible profile parts, profile designers need to look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is expected to be used for this.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+      <p>
+        This section describes profiles of the total SHACL 1.2 specification and how to create other profiles of SHACL.
       </p>
+      <section id="base-profiles">
+        <h3>Base Profiles</h3>
+        <p>
+          Base profiles of SHACL are profiles that correspond to the individual SHACL 1.2 specifications listed in the <a href="#specifications">Specifications</a> section, above.
+        </p>
+        <p>
+          Together, these Base Profiles cover all of SHACL and, when used together in the <a href="#union-profile">Union Profile</a>, they can validate that any graph is valid SHACL 1.2 data.
```
TallTed:
```suggestion
          Together, these Base Profiles cover all of SHACL and, when used together in the <a href="#union-profile">Union Profile</a>, they can validate whether any graph is valid SHACL 1.2 data.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+      </section>
+      <section id="non-base-profiles">
+        <h3>Non-base Profiles</h3>
+        <p>
+          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, and also other profiles of parts of SHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
```
TallTed:
```suggestion
          Non-base profiles of SHACL are profiles that do not correspond directly to a SHACL 1.2 specification. These profiles include supersets of the specification-derived base profiles, such as the <a href="#union-profile">Union Profile</a> defined below, as well as other profiles of parts of SHACL, such as <a href="#profile-x">Profile X</a> that contains SHACL elements that, when used for validation, require no more than XXX computational complexity.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
 		<section id="profiling-with-shacl">
 			<h2>Profiling with SHACL</h2>
-			<p>Content.</p>
+			<p>
+        This section describes how to profile specifications with SHACL.
+      </p>
+      <section id="profile-parts">
+        <h3>Profile Part Roles</h3>
+        <p>
+          SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the total set of things that a profile designers may wish to do, for example, they may want to include documentation about why and how the profile was created or provide new model elements - classes, predicates etc. - within an extended schema.
```
TallTed:
```suggestion
          SHACL is a language that can implement constraints and inference rules for RDF data. These two things are not the complete set of things that a profile designer may wish to do; for example, they may want to include documentation about why and how the profile was created, or provide new model elements (classes, predicates, etc.) within an extended schema.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <p>
+          Since SHACL cannot be used for all possible profile parts, profile designers must look to specifications outside the <a href="#specifications">SHACL family of specifications</a> for guidance on a profile's total set of parts and how to relate them to one another. [[[dx-prof]]] is recommended for this.
+        </p>
+        <p>
+          Referencing [[[dx-prof]]]'s <a href="https://www.w3.org/TR/dx-prof/#resource-roles-vocab">vocabulary of <em>Resource Role Instances</em></a>, it seems clear that SHACL graphs can perform the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a> within a profile and perhaps also <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>, since SHACL can be used to declare Node and Property Shapes just as OWL can declare Classes and Properties.
```
TallTed: It's unfortunate that dx-prof used the wrong words for most if not all of its roles. `Role:validation` *should be* `Role:validator` or _perhaps_ `Task:validation`, though that would mean broader and deeper changes to dx-prof.
```suggestion
          Referencing [[[dx-prof]]]'s <a href="https://www.w3.org/TR/dx-prof/#resource-roles-vocab">vocabulary of <em>Resource Role Instances</em></a>, it seems clear that SHACL graphs can be applied to the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a> within a profile and perhaps also to the role of <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>, since SHACL can be used to declare Node and Property Shapes just as OWL can be used to declare Classes and Properties.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <p>
+          SHACL resources can also document the constraints they implement validation shapes for, therefore they can also perform the <a href="https://www.w3.org/TR/dx-prof/#Role:constraints">Constraints</a> role.
```
TallTed:
```suggestion
          SHACL resources can also document the constraints they implement validation shapes for, therefore they can also be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:constraints">Constraints</a> role.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <p>
+          They could also be used for the <a href="https://www.w3.org/TR/dx-prof/#Role:mapping">Mapping</a> role, if SHACL Rules were implemented to map/transform data from one model to another.
```
TallTed:
```suggestion
          They could also be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:mapping">Mapping</a> role, if SHACL Rules are implemented to transform data from one model to another.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <p>
+          SHACL lists of values required for use with a model could perform the <a href="https://www.w3.org/TR/dx-prof/#Role:vocabulary">Vocabulary</a> role.
```
TallTed:
```suggestion
          SHACL lists of values required for use with a model could be applied to the <a href="https://www.w3.org/TR/dx-prof/#Role:vocabulary">Vocabulary</a> role.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        <h3>Creating Null Profiles</h3>
+        <p>
+          A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement some, and potentially all the specification's rules, but no other rules, in SHACL.
+        </p>
+        <p>
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
+        </p>
+        <div class="note">
+          <p>
+            While null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
```
TallTed:
```suggestion
            While null profiles of specifications can be created using mechanism other than SHACL, here we focus only on the use of SHACL.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+      <section id="hierarchies">
+        <h3>Profile Hierarchies</h3>
+      </section>
+      <section id="null-profiles">
+        <h3>Creating Null Profiles</h3>
+        <p>
+          A "null" profile of a specification, created using SHACL, is a profile of that specification in which SHACL is used to implement some, and potentially all the specification's rules, but no other rules, in SHACL.
+        </p>
+        <p>
+          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but which do not provide a mechanism for data validation, such as a SHACL validator.
```
TallTed:
```suggestion
          The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a model specification, perhaps an OWL model or only a natural-language document of a model, but do not provide a mechanism for data validation, such as a SHACL validator.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+        </p>
+        <div class="note">
+          <p>
+            While null profiles of specifications can be created using mechanism other than SHACL, here we focus on the use of SHACL only.
+          </p>
+          <p>
+            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, for ordering of temporal entities. Implementations of those constraints <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL allowing for RDF data validation. It would be possible to implement those constraints in SHACL too.
```
TallTed:
```suggestion
            An example of a non-SHACL validator for RDF data that acts as a null profile is the W3C's [[[prov-constraints]]] which provides a list of constraints that apply to provenance data formulated according to [[[prov-o]]]. [[[prov-constraints]]] implements no constraints beyond those stated or implied in [[[prov-o]]] and the conceptual [[[prov-dm]]], however it does include tests for things that the models do not explicitly model but whose proper use requires, e.g., for ordering of temporal entities. Implementations of those constraints <a href="https://github.com/pgroth/prov-check/tree/master">have been made</a> as Python scripts that execute SPARQL, allowing for RDF data validation. It is also possible to implement those constraints in SHACL.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+			<h2>Security and Privacy Considerations</h2>
+			<p>
+				Like most RDF-based technologies, SHACL processors may operate on graphs that are combined
+				from various sources.  Some applications may have an open "linked data" architecture and dynamically
+				assemble RDF triples from sources that are outside an organization's network of trust.
+				Since RDF allows anyone to add statements about any resource, triples may modify the originally
+				intended semantics of shape definitions or nodes in a data graph and thus lead to misleading results.
+				Protection against this (and the following) scenario can be achieved by only using trusted
+				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
+				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
```
TallTed:
```suggestion
				Like most RDF-based technologies, SHACL processors may operate on graphs that are assembled
				from various sources.  Some applications may have an open "linked data" ("LD") architecture and dynamically
				assemble RDF triples from sources that are outside an organization's network of trust.
				Since RDF allows anyone to add statements about any resource, triples may modify the originally
				intended semantics of shape definitions or nodes in a data graph and thus feed into misleading results.
				Protection against this (and the following) scenario is achievable by using only trusted
				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
```

shacl12-profiling/index.html
```diff
@@ -804,29 +804,106 @@ <h3>Document Conventions</h3>
...
+			</p>
+      <p>
+        When creating profiles of other specifications - base specifications - profile creators need to ensure that
+        their constraints are not in violation of the base specification's rules. If any were so, and if only the
+        profile but not the base specification was used to check for data validity, by accident or by design, invalid
+        data could be wrongly calculated to be valid. This could lead to accidental data release or use, potentially
+        introducing security issues.
```
TallTed: I don't understand what the `— based specifications —` is supposed to be doing. Is this referring back to `profiles` or `specifications`? Could this first sentence be rewritten as `...of other (base) specifications, profile creators...`?
```suggestion
        When creating profiles of other specifications, profile creators need to ensure that
        their constraints do not violate those specification's rules. If any did so, and if only the
        profile's rules, but not the specification's rules, were used to check for data validity, by accident or by design,
        data could be wrongly calculated to be valid. This could lead to accidental data release or use, potentially
        introducing security issues.
```

shacl12-profiling/index.html
nicholascar: > It's unfortunate that dx-prof used the wrong words for most if not all of its roles. `Role:validation` _should be_ `Role:validator` or _perhaps_ `Task:validation`, though that would mean broader and deeper changes to dx-prof.

Yes, I think you are correct.

But, we may have a chance to fix this as I've asked for dx-prof to be added to the list of specifications that the proposed renewed DXWG will update and move to Rec.

Issue tracked at https://github.com/w3c/dx-prof/issues/52

shacl12-profiling/index.html
nicholascar: I've removed the problematic sentence. It's not required

PR #552, opened 2025-08-29
#551: Clarified evalExpr parameters

### LINKED ISSUE #551: Definition of sh:targetNode does not fully specify evaluation of node expressions
The current textual definition of sh:targetNode looks like a left-over from previous definitions of node expressions, and it needs to properly use evalExpr and clarify what parameters go in, such as the focus node.

HolgerKnublauch:
Closes #551

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-551/shacl12-core/index.html)


YoucTagh (approved):
I fixed the typo in the rendering URI so that it points to the correct page.

Issue #183, opened 2024-01-31
Need support for RDF datasets (i.e., graphs, quads)


TallTed:
Came up in https://github.com/w3c/vc-data-model/pull/1416#issuecomment-1915680267

afs:
Some thoughts: https://afs.github.io/shacl-datasets.html

hmottestad:
Here is what we came up with for RDF4J:

```
[
      a rsx:DataAndShapesGraphLink;
      rsx:shapesGraph <http://example.com/ns#shapesGraph1>;
      rsx:dataGraph ex:dataGraph, ex:dataGraph2;
]
```

There can be multiple shape graphs and multiple data graphs. The validator will union the shapes graph and parse those, the validate the shapes against the union of the data graphs.

A good thing about this approach is that users can split up their shapes into multiple graphs, then select some of those graphs to validate certain parts of the data. An example could be a `ex:relaxedShapesGraph` and a `ex:strictShapesGraph` where some data graphs will need to follow only the relaxed shapes while others need to follow both the strict and the relaxed shapes. A second example could be where there are common vocabularies in separate graphs `ex:internalVocabularies` and `ex:commonVocabularies` and we want to validate `ex:shapesGraph1` together with the vocabulary graphs, but separate from `ex:shapesGraph2`.

In RDF4J users have to specify which named graphs can contain shapes. The ShaclSail will only parse shapes from those graphs. This allows users to validate untrusted data without having to worry that the data could contain shapes that would affect the validation. We also then require that any `rsx:DataAndShapesGraphLink` nodes be the shapes graphs. An `rsx:DataAndShapesGraphLink` needs to fully reside in a named graph, if it is split between two named graphs then we will ignore it.

https://github.com/eclipse-rdf4j/rdf4j/blob/2306c8b5b3f1d333e1259bb72ea1841b521c9deb/core/sail/shacl/src/test/resources/test-cases/minCount/unionDataset/shacl.trig#L19

HolgerKnublauch:
@gkellogg I think this is what you were asking for today?

VladimirAlexiev:
https://github.com/Sveino/Inst4CIM-KG/issues/140 is another user need.

https://github.com/w3c-cg/awesome-semantic-shapes/tree/VladimirAlexiev-patch-1#specifications sec "SHACL on Graphs" lists all approaches that I've heard about:

[...]

PR #555, opened 2025-09-01
added `sh:conformanceDisallows`, updated definition of conformance

### LINKED ISSUE #453: Definition of `sh:conforms` / handling of non-violating validation results
We extended the specification to allow giving a parameter to the engine to change the severity level when a result is a violation. Now there are violating and non-violating results. Only the violating result will change the `sh:conforms` to false.

That led to two follow-up issues:
1. The validation report doesn't contain enough information to explain which severity level had an impact on the `sh:conforms` value.
2. There is now the severity level `sh:Violation` and violation as an attribute for severity levels, which can be set with a parameter to the engine. Before, the specification defined `sh:Info` and `sh:Warning` already as violations. That was confusing, but consistent in that way, that all results have been violations. Now that's no longer the case.

Ideally, we find a way to extend the validation report data structure to contain enough information to be able to reproduce the result of `sh:conforms`, and the same definition avoids the problem of dealing with the violation attribute.

There have been multiple proposals:
- A property `sh:conformanceDisallows` could be added with a list of severity levels, which will change the `sh:conforms` to false if a result of that level shows up. e.g.: `[] sh:conformanceDisallows (sh:Violation sh:Warning) .`
- `sh:order` could be attached to the levels, and a threshold could be written to the report where every level above that threshold has an impact on `sh:conforms`.
- Results without impact on `sh:conforms` could be added with a different property `sh:informativeResult`. The existing definition that any result attached with `sh:result` leads to a `sh:conforms` equals false could be kept.

In our last weekly call, it was already mentioned that the solution with `sh:order` could cause problems if the level is changed afterwards. That could be a change of `sh:order` but also deprecating the level.

The last solution, using the property `sh:informativeResult`, would be the one with the smallest impact, but I tend towards the first solution. The `sh:conformanceDisallows` property, just for the severity levels, would give us the option to add other properties to group the results based on other attributes. That could be again the severity level, but selected for a different use case, like what should be shown in the UI. With the `sh:informativeResult`, we would be somehow settled on how results will be grouped.

bergos:
This PR adds the `sh:conformanceDisallows` property and updates the definition of the conformance check. The examples are updated to cover the non-default severity levels. The section for the `sh:conforms` property contained a redundant definition of the conformance check that was removed.

Closes #453 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html)


shacl12-core/index.html
```diff
@@ -2897,9 +2897,15 @@ <h4>Handling of Recursive Shapes</h4>
...
-					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level representing a violation and no <a>failure</a>
+					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level of the set of disallowed levels and no <a>failure</a>
 					has been reported by it.
 				</p>
+        <p>
+          The set of disallowed severity levels is defined as the values of the property
+          <code>sh:conformanceDisallows</code> in the report, or if no triple with the property
+          <code>sh:conformanceDisallows</code> is present in the report, <code>sh:Violation</code>,
+          <code>sh:Warning</code>, and <code>sh:Info</code> are set as defaults.
+        </p>
```
HolgerKnublauch: (Editorial: inconsistent indentation - tabs vs spaces?)

shacl12-core/index.html
```diff
@@ -2897,9 +2897,15 @@ <h4>Handling of Recursive Shapes</h4>
 				<h3>Conformance Checking</h3>
 				<p>
 					A <a>focus node</a> <dfn data-lt="conform|conformance">conforms</dfn> to a <a>shape</a> if and only if
-					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level representing a violation and no <a>failure</a>
+					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level of the set of disallowed levels and no <a>failure</a>
 					has been reported by it.
 				</p>
+        <p>
+          The set of disallowed severity levels is defined as the values of the property
+          <code>sh:conformanceDisallows</code> in the report, or if no triple with the property
```
HolgerKnublauch: "in the report" is a bit too vague, better spell out what the subject is of the triples is.

shacl12-core/index.html
```diff
@@ -3012,9 +3055,16 @@ <h5>Conforms (sh:conforms)</h5>
...
 							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph has exactly one value for the property <code>sh:conforms</code> and the value is of datatype <code>xsd:boolean</code>.
 							It represents the outcome of the <a>conformance checking</a>.
-							The value of <code>sh:conforms</code> is <code>true</code> if and only if the <a>validation</a> did not produce any <a>validation results</a> with a severity level representing a violation, and <code>false</code> otherwise.
 						</p>
 					</section>
+          <section id="conformanceDisallows">
+            <h5>Conformance Disallow Set (sh:conformanceDisallows)</h5>
+            <p>
+              Each SHACL instance of <code>sh:ValidationReport</code> in the results graph can have one or more values
+              for the property <code>sh:conformanceDisallows.</code> All values combined define the set of disallowed
```
HolgerKnublauch: The dot should be moved outside of the </code> block.

shacl12-core/index.html
```diff
@@ -3012,9 +3055,16 @@ <h5>Conforms (sh:conforms)</h5>
...
+              severity levels. The values must be valid severity levels.
```
HolgerKnublauch: What does the last sentence entail - how can a severity level be non valid? I think you can delete the sentence and just say they need to be IRIs, because anyone can invent their own levels, can't they?

YoucTagh (commented):
Would it make sense to add some test cases as well?

shacl12-core/index.html
bergos: I've changed the wording.

HolgerKnublauch (approved):

VladimirAlexiev:
@bergos I searched for "disallowed" in your rendered version, and I don't understand something:
how would `ValidationReport` know which levels to disallow? Isn't that some info that the shape designer or invoker should specify? Where?

UPD: I've read https://github.com/w3c/data-shapes/issues/453 and in particular https://github.com/w3c/data-shapes/issues/453#issuecomment-3153463912 (which is the basis of the voting). You say at the start:
> We extended the specification to allow giving a parameter to the engine to change the severity level when a result is a violation.

- Where is that in the spec? It should be linked from the sections mentioning "disallowed"
- "Parameter to the engine" (provided dynamically) should not be the only way.
This should relate to eg `sh:shapesGraph` and `rsx:DataAndShapesGraphLink` as mentioned in #183

bergos:
@VladimirAlexiev You are right. I revisited **severity** section and added a note on how the conformance disallow set is defined.

bergos:
@YoucTagh Tests would require a solution for issue #431.

shacl12-core/index.html
```diff
@@ -2897,9 +2893,15 @@ <h4>Handling of Recursive Shapes</h4>
 				<h3>Conformance Checking</h3>
 				<p>
 					A <a>focus node</a> <dfn data-lt="conform|conformance">conforms</dfn> to a <a>shape</a> if and only if
-					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level representing a violation and no <a>failure</a>
+					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level of the set of disallowed levels and no <a>failure</a>
```
ajnelson-nist:
```suggestion
					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any <a>validation results</a> with a severity level of the set of disallowed levels and no <a>failure</a>
```

shacl12-core/index.html
```diff
@@ -3012,7 +3051,18 @@ <h5>Conforms (sh:conforms)</h5>
 						<p>
 							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph has exactly one value for the property <code>sh:conforms</code> and the value is of datatype <code>xsd:boolean</code>.
 							It represents the outcome of the <a>conformance checking</a>.
-							The value of <code>sh:conforms</code> is <code>true</code> if and only if the <a>validation</a> did not produce any <a>validation results</a> with a severity level representing a violation, and <code>false</code> otherwise.
+						</p>
+					</section>
+					<section id="conformanceDisallows">
+						<h5>Conformance Disallow Set (sh:conformanceDisallows)</h5>
```
ajnelson-nist: I think eventually this English could use a little grammar adjustment, like "conformance-disallow set," or "conformance disallowance set".  The gist comes across fine for now, so this is a non-blocking point from me fine to revisit after merging.

shacl12-core/index.html
```diff
@@ -3012,7 +3051,18 @@ <h5>Conforms (sh:conforms)</h5>
...
+						<p>
+							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph can have one or more values
+							for the property <code>sh:conformanceDisallows</code>. All values combined define the set of disallowed
+							severity levels. All values of <code>sh:conformanceDisallows</code> must be IRIs.
+						</p>
+						<p>
+							The conformance disallow set is defined by the validation engine.
+							A validation engine MAY provide mechanisms to customize this set.
```
ajnelson-nist: Are these two lines talking about an engine-default behavior that differs from the spec?

We the working-group are proposing the default be violation, warning, info.  So, there's either the default by the Core spec, or a runtime configuration that gets recorded in the `sh:ValidationReport` as a deviation from the default.  I.e., 2 modes - default, or customized.

Is this paragraph suggesting engines can provide alternative defaults?  Should this appear any different in a generated `sh:ValidationReport` than a user providing defaults?  My initial reaction is there should be no difference in the generated `sh:ValidationReport` between an engine providing non-Core-doc defaults and a user requesting something aside from the Core-doc defaults.

I think we should clarify a group decision on this point with a little more text in this paragraph.

shacl12-core/index.html
```diff
@@ -1832,8 +1828,8 @@ <h4>Declaring the Severity of a Shape or Constraint</h4>
 					</span></p>
 					<p><em>The remainder of this section is informative.</em></p>
 					<p>
-						The specific values of <code>sh:severity</code> have no impact on the validation,
-						but MAY be used by user interface tools to categorize validation results.
+						The validation handles the values of <code>sh:severity</code> according to the <a>conformance checking</a>.
```
ajnelson-nist: Non-blocking grammar revision:

```suggestion
						The validation process handles the values of <code>sh:severity</code> according to <a>conformance checking</a>.
```

"The validation process" is used elsewhere in the document.

I think "The conformance checking" should either drop "the" or end "configuration," "setting" or similar.

shacl12-core/index.html
```diff
@@ -2897,9 +2893,15 @@ <h4>Handling of Recursive Shapes</h4>
 				<h3>Conformance Checking</h3>
 				<p>
 					A <a>focus node</a> <dfn data-lt="conform|conformance">conforms</dfn> to a <a>shape</a> if and only if
-					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level representing a violation and no <a>failure</a>
+					the set of result of the <a>validation</a> of the <a>focus node</a> against the <a>shape</a> does not contain any results with a severity level of the set of disallowed levels and no <a>failure</a>
 					has been reported by it.
 				</p>
+				<p>
+					The set of disallowed severity levels is defined as the objects of triples with predicate
+					<code>sh:conformanceDisallows</code> and the validation report as subject.
```
ajnelson-nist:
```suggestion
					<code>sh:conformanceDisallows</code> and the <a>validation report</a> as subject.
```

ajnelson-nist (changes requested):
I only have one blocking remark on a paragraph about engines interacting with defaults.  Either it's a group-discussion matter for a significant behavior in engines providing defaults, or there's just a little English cleanup to clarify that engines can provide some way for users to request different conformance levels for a single validation run.

ajnelson-nist:
> @YoucTagh Tests would require a solution for issue #431.

@bergos , it looks to me like we can actually write a test for this without #431 .  The configuration of the validation run doesn't need to be included here - the _result_ of the call-time configuration is included, by merit of `sh:conformanceDisallows` being present at all.

VladimirAlexiev:
@bergos I think it's still backwards:
- https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#conformance-definition "The set of disallowed severity levels is defined as the objects of triples with predicate sh:conformanceDisallows and the validation report as subject."
But the conformance checking process **outputs** the validation report, so it cannot **take** the "disallowed" definitions from that report!
- instead, it should refer to https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#conformanceDisallows and state that the "disallowed" definition from the engine is
  - used when computing conformance
  - copied into the validation report

shacl12-core/index.html
bergos: I've updated it to your first proposal. Ideally, we are happy with it, and no further changes are required.

shacl12-core/index.html
bergos: I agree, something was missing. I've added a description of the default set. As we don't want breaking changes, the default must be the set of violation, warning, and info. Please feel free to propose some more details if you think it's not yet clear enough.

bergos:
> @bergos , it looks to me like we can actually write a test for this without #431 . The configuration of the validation run doesn't need to be included here - the _result_ of the call-time configuration is included, by merit of `sh:conformanceDisallows` being present at all.

@ajnelson-nist That would only work if we make `sh:conformanceDisallows` mandatory. I don't have a strong opinion on that. I guess my initial thought was that people don't stumble over a missing property in case they read the new spec, but deal with a validation report from an older engine, which would be the user's fault.

bergos:
> * https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#conformance-definition "The set of disallowed severity levels is defined as the objects of triples with predicate sh:conformanceDisallows and the validation report as subject."
> But the conformance checking process **outputs** the validation report, so it cannot **take** the "disallowed" definitions from that report!
> 
> * instead, it should refer to https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#conformanceDisallows and state that the "disallowed" definition from the engine is
>       
>   * used when computing conformance
>   * copied into the validation report

@VladimirAlexiev The set in the validation report is a representation of the set of the engine. I think that's an implementation detail. One could write an implementation that creates the validation report before the validation and uses that to store the set. I think the spec should not cover how an engine keeps intermediate states/values and how they are copied during processing.

shacl12-core/index.html
```diff
@@ -3012,7 +3051,19 @@ <h5>Conforms (sh:conforms)</h5>
...
+						<h5>Conformance-Disallow Set (sh:conformanceDisallows)</h5>
+						<p>
+							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph can have one or more values
+							for the property <code>sh:conformanceDisallows</code>. All values combined define the set of disallowed
+							severity levels. All values of <code>sh:conformanceDisallows</code> must be IRIs.
+						</p>
+						<p>
+							The conformance-disallow set is defined by the validation engine.
+              The default set MUST consist of <code>sh:Violation</code>, <code>sh:Warning</code>, and <code>sh:Info</code>.
+							A validation engine MAY provide mechanisms to customize this set.
```
ajnelson-nist, line 2395: Ah, I figured out what was confusing me.  The last three sentences talk about the total conformance-disallow set, then the default set, and then "mechanisms to customize **this** set."  "This" is ambiguous in the context of the paragraph.

It remains unclear to me whether there's a difference in the default set according to SHACL-Core and a default set according to a validation engine.  Mainly, I'm not convinced there's a need to represent anything in the results graph that isn't the SHACL-Core default set differently, whether it's a user one-off configuration or an engine-level deviation from the Core default-set.

How does this rearrangement look?

```suggestion
					<section id="conformanceDisallows">
						<h5>Conformance-Disallow Set (sh:conformanceDisallows)</h5>
						<p>
							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph MAY have one or more values
							for the property <code>sh:conformanceDisallows</code>. All values of <code>sh:conformanceDisallows</code> MUST be IRIs.
						</p>
						<p>
							All values combined define the set of disallowed
							severity levels.  Presence of any <code>sh:ValidationResult</code> with a severity level in the set of disallowed severity levels MUST result in a <code>sh:conforms</code> value of <code>false</code> on the associated <code>sh:ValidationReport</code> instance.
						</p>
						<p>
							If no values are present in the results graph for the property <code>sh:conformanceDisallows</code>, then a default set MUST be used, comprised of <code>sh:Violation</code>, <code>sh:Warning</code>, and <code>sh:Info</code>.
						</p>
						<p>
							The conformance-disallow set is defined by the validation engine.
							A validation engine MAY provide mechanisms to customize this set.
```

bergos:
@ajnelson-nist I've merged the changes you proposed.

ajnelson-nist:
> Would it make sense to add some test cases as well?

Added in https://github.com/w3c/data-shapes/pull/555/commits/a43a3526344560e59215f3c0ab565d30dd24ca89 , thank you for the reminder.

ajnelson-nist (approved):
Someone else should approve before merge, if not to review the text, then to check the tests I just added.

YoucTagh (commented):
The `ex:PersonShape-age` shape is used in both validation reports, once with severity `sh:Warning` and once with severity `sh:Violation`.
This shape is later defined in the specification as follows: 
```turtle
ex:PersonShape-age;
  a sh:PropertyShape;
  sh:path ex:age;
  sh:datatype xsd:integer;
  sh:maxCount 1;
  sh:reifierShape ex:ProvenanceShape;
  sh:reificationRequired true .
```
If the data graph in [Example 24](https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#example-shape-with-an-iri-as-sh-datatype-property-constraint) is the one used for illustration, I think the validation reports are missing the reification validation results.

ajnelson-nist:
> The `ex:PersonShape-age` shape is used in both validation reports, once with severity `sh:Warning` and once with severity `sh:Violation`. This shape is later defined in the specification as follows:
> 
> ```turtle
> ex:PersonShape-age;
>   a sh:PropertyShape;
>   sh:path ex:age;
>   sh:datatype xsd:integer;
>   sh:maxCount 1;
>   sh:reifierShape ex:ProvenanceShape;
>   sh:reificationRequired true .
> ```
> 
> If the data graph in [Example 24](https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#example-shape-with-an-iri-as-sh-datatype-property-constraint) is the one used for illustration, I think the validation reports are missing the reification validation results.

Ah, whoops.  Good catch, I forgot to search for the shape definition.

One response pertinent to the test design: I *did* keep the shape I wrote consistent between both tests.  What I varied in the test was the presence of `sh:conformanceDisallows`.  Here's the diff between the two test files, posting to just note mootness before line 42:

```patch
28c28
<       <conformance-disallows-001>
---
>       <conformance-disallows-002>
31c31
< <conformance-disallows-001>
---
> <conformance-disallows-002>
33,35c33,35
<   rdfs:label "Test of path sh:conformanceDisallows 001" ;
<   rdfs:comment "conformanceDisallows presence lets sh:conforms be true despite presence of an sh:Warning-severity result.  Test contents and behavior otherwise match conformance-disallows-002."@en ;
<   rdfs:seeAlso <conformance-disallows-002> ;
---
>   rdfs:label "Test of path sh:conformanceDisallows 002" ;
>   rdfs:comment "conformanceDisallows absence causes sh:conforms be false due to presence of an sh:Warning-severity result.  Test contents and behavior otherwise match conformance-disallows-001."@en ;
>   rdfs:seeAlso <conformance-disallows-001> ;
42,43c42
<       sh:conforms "true"^^xsd:boolean ;
<       sh:conformanceDisallows sh:Violation ;
---
>       sh:conforms "false"^^xsd:boolean ;
```

Next step:

The test discrepancy @YoucTagh noted seems to bubble up to the example that I copied.  Is it a problem *in the document* to have two tests in unrelated sections use the same example IRI with different implementations around the example IRI?

I'm fine with adjusting the test, but do we also need to adjust the document?

shacl12-core/index.html
```diff
@@ -3012,7 +3051,24 @@ <h5>Conforms (sh:conforms)</h5>
 						<p>
 							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph has exactly one value for the property <code>sh:conforms</code> and the value is of datatype <code>xsd:boolean</code>.
 							It represents the outcome of the <a>conformance checking</a>.
-							The value of <code>sh:conforms</code> is <code>true</code> if and only if the <a>validation</a> did not produce any <a>validation results</a> with a severity level representing a violation, and <code>false</code> otherwise.
+						</p>
+					</section>
+					<section id="conformanceDisallows">
+						<h5>Conformance-Disallow Set (sh:conformanceDisallows)</h5>
```
```suggestion
						<h5>Conformance-Disallow Set (`sh:conformanceDisallows`)</h5>
```

shacl12-core/index.html
```diff
@@ -3012,7 +3051,24 @@ <h5>Conforms (sh:conforms)</h5>
...
+						<p>
+							Each SHACL instance of <code>sh:ValidationReport</code> in the results graph MAY have one or more values
+							for the property <code>sh:conformanceDisallows</code>. All values of <code>sh:conformanceDisallows</code> MUST be IRIs.
+						</p>
+						<p>
+							All values combined define the set of disallowed
+							severity levels. Presence of any <code>sh:ValidationResult</code> with a severity level in the set of disallowed severity levels MUST result in a <code>sh:conforms</code> value of <code>false</code> on the associated <code>sh:ValidationReport</code> instance.
```
TallTed: just line length
```suggestion
							severity levels. Presence of any <code>sh:ValidationResult</code> with a severity level in the set of
							disallowed severity levels MUST result in a <code>sh:conforms</code> value
							of <code>false</code> on the associated <code>sh:ValidationReport</code> instance.
```

shacl12-core/index.html
```diff
@@ -3012,7 +3051,24 @@ <h5>Conforms (sh:conforms)</h5>
...
+						</p>
+						<p>
+							If no values are present in the results graph for the property <code>sh:conformanceDisallows</code>, then a default set MUST be used, comprised of <code>sh:Violation</code>, <code>sh:Warning</code>, and <code>sh:Info</code>.
```
TallTed: Just line length
```suggestion
							If no values are present in the results graph for the property
							<code>sh:conformanceDisallows</code>, then a default set MUST be used, comprised of
							<code>sh:Violation</code>, <code>sh:Warning</code>, and <code>sh:Info</code>.
```

YoucTagh:
@ajnelson-nist, @bergos  Could we merge this pull request and create a new issue to discuss:

> The test discrepancy @YoucTagh noted seems to bubble up to the example that I copied. Is it a problem in the document to have two tests in unrelated sections use the same example IRI with different implementations around the example IRI?
> I'm fine with adjusting the test, but do we also need to adjust the document?

bergos:
@ajnelson-nist I've merged the changes you proposed.

> @ajnelson-nist, @bergos Could we merge this pull request and create a new issue to discuss:
> 
> > The test discrepancy @YoucTagh noted seems to bubble up to the example that I copied. Is it a problem in the document to have two tests in unrelated sections use the same example IRI with different implementations around the example IRI?
> > I'm fine with adjusting the test, but do we also need to adjust the document?

I don't think it's a problem to reuse IRIs for different examples. I didn't check, but I guess we already have some cases where that is done.

bergos:
@YoucTagh, please approve the PR, if you think it can be merged.

YoucTagh:
@bergos, currently, the first thing a user sees in the validation report section is [Examples 20](https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#example-20) and [21](https://raw.githack.com/w3c/data-shapes/issue-453/shacl12-core/index.html#example-21). In these examples, we immediately present the validation report, which indicates whether the data graph conforms to the SHACL shape graph. The report includes the conformance result (true or false) and the overall validation outcome.

However, when someone unfamiliar with the specification reviews this section for the first time, they will notice that the focus node is `ex:Bob` and the source shape is `ex:PersonShape-age`. Naturally, they will look for these elements in the earlier sections of the document since these are not redefined or clearly referenced. The issue isn’t about reusing IRIs. The validation report we provide does not accurately represent the validation of the data graph and shape graph used in the examples.

If you still don't think this is problematic, I won't block the merge.

PR #556, opened 2025-09-01
#500

### LINKED ISSUE #500: Add node expression function for the empty list
In some cases it is useful to specify an empty list, for example as a dummy input to some other node expression.

Many people will expect to use () for the empty list. However, that is in fact the constant rdf:nil.

I think the "empty" blank node [] should be interpreted as the empty list.

HolgerKnublauch:
Closes #500

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-500/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -923,6 +923,27 @@ ex:Person-displayName
 			<section id="library-basic">
 				<h2>Basic Node Expressions</h2>
 
+				<section id="EmptyExpression">
+					<h3>Empty Expressions</h3>
+					<p class="syntax">
+						<span data-syntax-rule="EmptyExpression-syntax">
+							A <a>blank node</a> that is the <a>subject</a> of no triple is called an <dfn>empty expression</dfn>
+							with the <a>function name</a> <code>shnex:EmptyExpression</code>.
+						</span>
+					</p>
+					<div class="def" id="VarExpression-evaluation">
+						<div class="def-header">EVALUATION OF EMPTY EXPRESSIONS</div>
+						<p>
+							The <a>output nodes</a> of an <a>empty expression</a> is the empty list.
+						</p>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						This node expression function is written in Turtle as <code>[]</code> and must not be confused with
+						the empty SHACL list <code>()</code> which is interpreted as the <a>IRI</a> <code>rdf:nil</code>.
+					</p>
+				</section>
+
 				<section id="VarExpression">
 					<h3>Var Expressions</h3>
 					<p class="syntax">
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -927,7 +927,7 @@ ex:Person-displayName
 					<h3>Empty Expressions</h3>
 					<p class="syntax">
 						<span data-syntax-rule="EmptyExpression-syntax">
-							A <a>blank node</a> that is the <a>subject</a> of no triple is called an <dfn>empty expression</dfn>
+							A <a>blank node</a> that is not the <a>subject</a> of any triple is called an <dfn>empty expression</dfn>
 							with the <a>function name</a> <code>shnex:EmptyExpression</code>.
 						</span>
 					</p>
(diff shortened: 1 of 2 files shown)


shacl12-node-expr/index.html
```diff
@@ -923,6 +923,27 @@ <h2>Node Expressions Library</h2>
 			<section id="library-basic">
 				<h2>Basic Node Expressions</h2>
 
+				<section id="EmptyExpression">
+					<h3>Empty Expressions</h3>
+					<p class="syntax">
+						<span data-syntax-rule="EmptyExpression-syntax">
+							A <a>blank node</a> that is the <a>subject</a> of no triple is called an <dfn>empty expression</dfn>
```
```suggestion
							A <a>blank node</a> that is not the <a>subject</a> of any triple is called an <dfn>empty expression</dfn>
```

recalcitrantsupplant (approved):
looks good

robert-david (approved):

shacl12-node-expr/index.html
```diff
@@ -923,6 +923,27 @@ <h2>Node Expressions Library</h2>
...
+					<div class="def" id="VarExpression-evaluation">
+						<div class="def-header">EVALUATION OF EMPTY EXPRESSIONS</div>
+						<p>
+							The <a>output nodes</a> of an <a>empty expression</a> is the empty list.
+						</p>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						This node expression function is written in Turtle as <code>[]</code> and must not be confused with
+						the empty SHACL list <code>()</code> which is interpreted as the <a>IRI</a> <code>rdf:nil</code>.
```
ajnelson-nist: I find this a bit confusing in a vacuum.  How many empty lists will SHACL as an RDF vocabulary have when the dust settles?

* `rdf:nil` in the role of RDF's empty list.
* `rdf:nil` in the role of "the empty SHACL list."
* `[]` in the role of "the empty [node expression?] list?"

I'm afraid I have failed to follow the instruction to not be confused.

HolgerKnublauch: I don't know how to make this more clear. If you or someone has a suggestion, let me know.

afs: > * `rdf:nil` in the role of RDF's empty list.
> * `rdf:nil` in the role of "the empty SHACL list."

A SHACL list is an [RDF collection](https://www.w3.org/TR/rdf12-schema/#ch_collectionvocab) (RDF List) with a well-formed criterion.

> `[]` in the role of "the empty [node expression?] list?"

A (SHACL) empty expression is syntax (a blank node such that ...) and not a list. 

It evaluates to a sequence (abstract "list"), but not an RDF list. An RDF list is a way to write an abstract "list".

In the text:
"which is interpreted as the [IRI](https://www.w3.org/TR/rdf12-concepts/#dfn-iri) rdf:nil."
==>
"which is the [IRI](https://www.w3.org/TR/rdf12-concepts/#dfn-iri) rdf:nil."

`()` is no more than a Turtle way to write `rdf:nil`.

It is not interpreted (that gets confused with evaluation) - it may be said to "represent" an empty list (in the sense of a value) as does `rdf:nil`.

ajnelson-nist: I think it takes a better handle on the ... "SHACL phase," for lack of a better phrase, that the empty list is in.

In input data and graphs in storage---interactive with the graph-parsing and -storage phases---there are empty lists, and absence of sets.

In the node expression evaluation phase, there are empty lists as in-memory computing objects, and this is comfortable to separate from empty-sets.

Is there something in "phase separation" you could apply here?

shacl12-node-expr/index.html
```diff
@@ -923,6 +923,27 @@ <h2>Node Expressions Library</h2>
...
+					<p class="syntax">
+						<span data-syntax-rule="EmptyExpression-syntax">
+							A <a>blank node</a> that is not the <a>subject</a> of any triple is called an <dfn>empty expression</dfn>
+							with the <a>function name</a> <code>shnex:EmptyExpression</code>.
+						</span>
+					</p>
+					<div class="def" id="VarExpression-evaluation">
+						<div class="def-header">EVALUATION OF EMPTY EXPRESSIONS</div>
+						<p>
+							The <a>output nodes</a> of an <a>empty expression</a> is the empty list.
```
TallTed, line 937: This sentence reads very badly. `output nodes` is plural and calls for a plural verb (i.e., `are`) but then `the empty list` is singular and cannot be sensibly changed to a plural. I think the least disruptive thing to do would be to change the plural `output nodes` to the singular `output node`, which then matches both the singular verb (`is`) and the singular object (`the empty list`).

PR #557, opened 2025-09-01
#501

### LINKED ISSUE #501: Make list optional for list-parameter node expressions that take only one parameter
For example, sparql:abs currently would need to be written as

```
[
    sparql:abs ( [ shnex:path ex:property ] )
]
```

If this was a named-parameter node expression, it would be written as

```
[
    sparql:abs [ shnex:path ex:property ]
]
```

Many users will not know the difference and when to use which notation. Many users will not really be aware that the distinction between list and named parameters even exists.

I think any list parameter functions that take only one argument should support the second syntax without a list. So this would also be legal:

```
[
    sparql:abs -42
]
```

HolgerKnublauch:
Closes #501

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-501/shacl12-node-expr/index.html)

The syntax will hopefully become clearer once we have the SPARQL function section in the document.


afs:
General question: How is testing for node expressions going to be organised?  Tests for both extended shapes and something for the node expressions themselves?

HolgerKnublauch:
I have a ticket to write test cases for all node expression types. In this I will likely introduce a new test case type for the results of a node expression so that they can be developed independently of the rest (such as sh:values, sh:targetNode).

shacl12-node-expr/index.html
```diff
@@ -813,12 +814,34 @@ <h3>Named Parameter Functions</h3>
...
+							Example: <code>[ sparql:now () ]</code>
+						</li>
+						<li>
+							If <code>o</code> is a <a>blank node</a> that is a <a>well-formed</a> <a>SHACL list</a> where all
+							<a>members</a> are <a>well-formed</a> <a>node expressions</a>,
+							then the arguments are the <a>members</a> of that list.<br />
+							Examples: <code>[ sparql:plus ( 38 4 ) ]</code> and <code>[ sparql:abs ( -42 ) ]</code>
+						</li>
+						<li>
+							If the function accepts exactly one parameter (such as <code>sparql:abs</code>)
```
afs: > If the function accepts exactly one parameter ... and is a [well-formed](https://www.w3.org/TR/shacl12-core/#dfn-well-formed) [node expression](https://www.w3.org/TR/shacl12-core/#dfn-node-expression)

This restricts implementation. It requires knowledge of the arity before determining the legality of an expression. 

1. What if the function is withdrawn? Legal syntax become illegal.

2. You can't write a checker without knowing the function - hard for a UI to do checking in the browser as it needs to know the functions at the server side at time of execution.

They are "List Parameter Functions". Allowing literals and URIs raw can be allowed without the known arity condition. 

This section also needs to say it is syntactic shorthand for `[ sparql:abs ( -42 ) ]`. At the moment it could be a different way to call functions --  `[ sparql:abs ( -42 ) ]` and `[ sparql:abs  -42  ]` could be different.

HolgerKnublauch: I think it's not really a pre-condition "if the function accepts exactly one parameter". Instead it could be "if there is exactly one argument (in the function call)".

shacl12-node-expr/index.html
```diff
@@ -813,12 +814,34 @@ <h3>Named Parameter Functions</h3>
...
+							and <code>o</code> is a <a>well-formed</a> <a>node expression</a> then the argument list
+							consists of exactly <code>o</code>.<br />
+							Example: <code>[ sparql:abs -42 ]</code>
```
afs:
```suggestion
							If the function call has one argument,
							and the argument is a <a>well-formed</a> <a>node expression</a>,
							then the argument can given without the list.
							It is equivalent to a function call with a list of one element.
							Example: <code>[ sparql:abs -42 ]</code>, which is equivalent to
							<code>[ sparql:abs ( -42 ) ]</code>.
```
(indent showing up because I copied the tab style)

afs (approved):

robert-david (approved):

PR #566, opened 2025-09-08
Issue 559: Added sh:codeIdentifer annotation property for code generators

### LINKED ISSUE #559: Should we introduce a property ID property for use cases like GraphQL and APIs?
A common use case of shapes is to generate APIs such as GraphQL endpoints or JS or Java classes. The question is how to map from property shapes to valid fields in those languages. Java-style identifiers are a commonly accepted lowest common denominator.

In TopBraid we had to introduce a custom property graphql:name to allow users to specify a Java-style ID for a property shape. This is needed in particular when a property shape uses a path expression. Usually the local name of a property can be used, but it's difficult to construct such names from paths.

Something like sh:propertyID

```
ex:MyShape-myProp
   a sh:PropertyShape ;
   sh:path ( ex:parent ex:income ) ;
   sh:propertyID "parentIncome" ;
   ...
```

```
ex:MyShape-child
   a sh:PropertyShape ;
   sh:name "children" ;
   sh:path [ sh:inversePath ex:parent ] ;
   sh:propertyID "child" ;
   ...
```

Discussed in https://www.linkedin.com/feed/update/urn:li:activity:7367747396455309313?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7367747396455309313%2C7367845601729855488%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287367845601729855488%2Curn%3Ali%3Aactivity%3A7367747396455309313%29

HolgerKnublauch:
Closes #559

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-559/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -7356,7 +7356,7 @@ ex:RainbowPony ex:color ex:Pink .
 				this section covers properties that are ignored by SHACL processors.
 				The use of these so-called <dfn data-lt="non-validating property">non-validating properties</dfn>
 				is entirely optional and is not subject to formal interpretation contracts.
-				They MAY be used for purposes such as form building or predictable printing of RDF files.
+				They MAY be used for purposes such as form building, code generation or predictable printing of RDF files.
 			</p>
 			<section id="name">
 				<h3>sh:name and sh:description</h3>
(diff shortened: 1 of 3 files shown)


shacl12-core/index.html
```diff
@@ -7373,6 +7373,25 @@ <h3>sh:name and sh:description</h3>
 					multiple <a>values</a>, but should only have one <a>value</a> per language tag.
 				</p>
 			</section>
+			<section id="codeIdentifier">
+				<h3>sh:codeIdentifier</h3>
+				<p>
+					Shapes may have one <a>value</a> for <code>sh:codeIdentifier</code>
+					to suggest a name that can be used for a representation of the shape in APIs, query languages
+					and similar programmatic access.
+					The values of <code>sh:codeIdentifier</code> are <a>literals</a> with <a>datatype</a> <code>xsd:string</code>
```
afs: Minor suggestion - because the previous text says "one value", "values" and "are literals" seems to hint otherwise:

```suggestion
					The value of <code>sh:codeIdentifier</code> is a <a>literal</a> with <a>datatype</a> <code>xsd:string</code>
```

shacl12-core/index.html
```diff
@@ -7373,6 +7373,25 @@ <h3>sh:name and sh:description</h3>
...
+					where the string matches the regular expression <code>^[a-zA-Z_][a-zA-Z0-9_]*$</code>.
```
afs, line 7383: Would this be better as a advice rather than a requirement? Some uses may require other characters.

Maybe put with the discussion note below.

shacl12-core/index.html
```diff
@@ -7373,6 +7373,25 @@ <h3>sh:name and sh:description</h3>
...
+				</p>
+				<p>
+					There are no strict rules in SHACL Core how these values should be used.
+					At the time of writing, a typical use case would be to generate GraphQL query
+					schemas from shapes, mapping property shapes to GraphQL fields.
+					By default such schema generators could use the local name of the
+					<code>sh:path</code>, but it should use the <code>sh:codeIdentifier</code> if present.
+					In the case of complex path expressions an explicit <code>sh:codeIdentifier</code> is strongly recommended.
+					Similar requirements exist for API generators, for example in Java, JavaScript or Python.
+				</p>
```
afs: Suggest - feel free to ignore - have a discussion note:

```suggestion
				<p class="note">
					A typical use case would be to generate GraphQL query
					schemas from shapes, mapping property shapes to GraphQL fields.
					By default such schema generators could use the local name of the
					<code>sh:path</code>, but it should use the <code>sh:codeIdentifier</code> if present.
					In the case of complex path expressions an explicit <code>sh:codeIdentifier</code> is strongly recommended.
					Similar requirements exist for API generators, for example in Java, JavaScript or Python.
				</p>
```

bergos (approved):

shacl12-core/index.html
HolgerKnublauch: Does anyone know of a target language where the given (Java) naming convention would produce invalid identifiers? Because I think we should go with the lowest common denominator, i.e. some syntax that works for every language.

If there is a need for different syntaxes, a fresh property can be introduced, e.g. ex:javaScriptIdentifier could allow arbitrary strings.

YoucTagh (approved):

PR #574, opened 2025-09-12
Issue 517

### LINKED ISSUE #517: Shapes for classification of nodes
The focus of SHACL so far has been on validating nodes so that *when we know that a node is targeted by a shape* we can confirm that it conforms to all the conditions laid out in the shape. This is similar to how OWL uses `rdf:type` and `rdfs:subClassOf`: If we know that someone is an instance of `ex:AdultPerson` we can restrict `ex:age` to `>= 18`.

However, the same validation could also be applied on untargeted nodes. This is similar to the "sufficient" conditions expressed using `owl:equivalentClass`, where OWL can express "if we know that some `ex:Person` is `>= 18` years old, then we classify them as `ex:AdultPerson`".

In SHACL this would be useful not only to cover the same use cases (as an "inference") but also for user interfaces. For example, if we know that some `Person` also conforms to an `ex:AdultPersonShape`, the form could show additional input fields for properties that are only available to adults.

The following suggests to introduce some property `sh:mayTargetClass` (or other syntax) to indicate that instances of `Person` should be classified against `AdultPersonShape`.

```
ex:Person
    a sh:ShapeClass ;
    rdfs:subClassOf owl:Thing ;
    sh:property [
        sh:path ex:age ;
        sh:datatype xsd:integer ;
    ] .

ex:AdultPersonShape
    a sh:NodeShape ;
    sh:mayTargetClass ex:Person ;
    sh:property [
        sh:path ex:age ;
        sh:minInclusive 18 ;
        sh:minCount 1 ;
    ] ;
    sh:property [
        sh:path ex:votedFor ;
        sh:class ex:Party ;
    ] .
```

The key is that unlike targets, this means that only SOME of the `Persons` need to conform to the shape, but we want to indicate which nodes are candidates.

If a `Person` fulfills all constraints of `AdultPersonShape`, an engine could also "infer" a triple such as

```
ex:Bob sh:shape ex:AdultPersonShape
```

similar to how an OWL reasoner would infer extra `rdf:type` triples. In fact if the `AdultPersonShape` was also a class, it could infer an extra `rdf:type` triple just like OWL would.

There are many syntactic variations of this idea, above is just a first draft to start the discussion and capture the use case.

HolgerKnublauch:
Closes #517

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-517/shacl12-core/index.html)


shacl12-core/index.html
```diff
@@ -1780,6 +1784,165 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+							because he conforms to the two constraints defined by the <code>sh:targetWhere</code> shape.
+							Based on the <code>sh:class</code> constraint, he is a <a>SHACL instance</a> of <code>ex:Person</code>
+							and his <code>ex:age</code> is 18 or greater.
+							However, as the <code>ex:AdultPerson</code> shape states that all adults must have one value for <code>ex:votedFor</code>,
+							<code>ex:Bob</code> does not conform to <code>ex:AdultPerson</code>.
+						</p>
+						<p class="note">
+							Note that <code>sh:targetWhere</code> can be interpreted as a "definition" providing
+							necessary and sufficient conditions for a shape.
+							It can therefore potentially be used for "classification" tasks, e.g. to collect all nodes
```
```suggestion
							It can therefore potentially be used for "classification" tasks, e.g., to collect all nodes
```

bergos (approved):

shacl12-core/index.html
```diff
@@ -1734,6 +1739,166 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+							<div class="shapes-graph">
+								<div class="turtle">
+ex:PersonShape
+    a sh:NodeShape ;
+    sh:targetClass ex:Person ;
+    sh:property [
+        sh:path ex:age ;
+        sh:datatype xsd:integer ;
+        sh:maxCount 1 ;
+    ].
```
YoucTagh:
```suggestion

```

Can we delete this shape from the example, since it is not being used?

shacl12-core/index.html
```diff
@@ -1734,6 +1739,166 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+
+ex:AdultShape
```
YoucTagh:
```suggestion
ex:AdultPerson
```
In the explanation `ex:AdultPerson` is used instead of `ex:AdultShape`.

shacl12-core/index.html
```diff
@@ -1734,6 +1739,166 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+		},
+		{
+			"@id": "ex:Bob",
+			"@type": "ex:Person",
+			"ex:age": 21
+		}
+	]
+}</pre>
+								</div>
+							</div>
```
YoucTagh:
```suggestion
	
```
Should we delete the JSON-LD example as proposed in #580

shacl12-core/index.html
```diff
@@ -1734,6 +1739,166 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+								<div class="turtle">
+ex:PersonShape
+	a sh:NodeShape .
+								</div>
+								<div class="jsonld">
+									<pre class="jsonld">{
+	"@id": "ex:PersonShape",
+	"@type": "sh:NodeShape"
+}</pre>
+								</div>
```
YoucTagh:
```suggestion
```
Delete example in JSON-LD

shacl12-core/index.html
```diff
@@ -1734,6 +1739,166 @@ <h4>Objects-of targets (sh:targetObjectsOf)</h4>
...
+				"@id": "ex:PersonShape"
+			}
+		},
+		{
+			"@id": "ex:Bob",
+			"@type": "ex:Person"
+		}
+	]
+}</pre>
+								</div>
```
YoucTagh:
```suggestion
```
Delete example in JSON-LD

shacl12-test-suite/tests/core/targets/targetWhere-001.ttl
```diff
@@ -0,0 +1,79 @@
...
+
+ex:PersonShape
+    a sh:NodeShape ;
+    sh:targetClass ex:Person ;
+    sh:property [
+        sh:path ex:age ;
+        sh:datatype xsd:integer ;
+        sh:maxCount 1 ;
+    ] ;
+.
```
YoucTagh:
```suggestion
```
This shape is not used in the test.

shacl12-core/index.html
HolgerKnublauch: Yeah right, it doesn't add much value.

YoucTagh (approved):

HolgerKnublauch:
So: any last-minute concerns about these two new properties, before this gets merged in?

shacl12-vocabularies/shacl.ttl
```diff
@@ -120,6 +128,13 @@ sh:severity
 	rdfs:range sh:Severity ;
 	rdfs:isDefinedBy sh: .
 
+sh:shape
+	a rdf:Property ;
+	rdfs:label "shape"@en ;
+	rdfs:comment "Links from a node in the data graph to the shape(s) that it is expected to conform to."@en ;
```
ajnelson-nist:
```suggestion
	rdfs:comment "Links from a node in the data graph to a shape that it is expected to conform to."@en ;
```

"The" somewhat implies a closed set of shapes.  This is not possible to restrict.

"(s)" somewhat hints a `rdf:List` could be provided.  I don't think that was the intention.

Going to "a shape" is more consistent with the multi-linking possibility that I think the spelling I changed away from meant.

PR #576, opened 2025-09-18
Issue 575: Type metaclass members as classes

### LINKED ISSUE #575: Members of metaclasses should explicitly be classes
This is a follow-on to [PR 512](https://github.com/w3c/data-shapes/pull/512), which, among other things, initiated a metaclass design pattern.  I have no complaints about the pattern's design, but did come across an implementation detail that I think would be helpful to spell more explicitly.

The instances of metaclasses are currently subjects of `rdfs:subClassOf`, but are not themselves `rdfs:Class`es.  There are a few benefits to being explicit that these nodes are `rdfs:Class`es:

* If they are classes, the OWL casting for [Issue 246](https://github.com/w3c/data-shapes/issues/246) can just look for instances of `rdfs:Class`.
* Seeing the extra `rdf:type` on a `rdfs:Class` emphasizes that these are intentionally members of meta-classes; and, likewise, there are meta-classes floating around somewhere.
* There is a reduced chance of confusion from seeing `rdfs:subClassOf` on something that is not explicitly an `rdfs:Class`.

I came across the now-missing `rdfs:Class` statements because a prior state of Issue 246's PR had already done the casting of `sh:SelectExpression` and `sh:SPARQLExprExpression`, but post PR 512, it looked like they'd dropped, and I became curious if they'd migrated to `shnex.ttl`.  They didn't; they'd just lost the `a rdfs:Class` triple I'd been relying on in my `CONSTRUCT` query.

I'll file a PR momentarily to add `a rdfs:Class` to all the subjects of `rdfs:subClassOf`.

ajnelson-nist:
Closes #575 .


ajnelson-nist:
This PR is now caught up and ready for review.  I ran a shapes graph (recorded in Git log) to check for where this pattern applies.  `shnex-sparql.ttl` saw a lot of resultant updates from making the metaclass pattern more explicit.  Nothing further was needed for `shacl-shacl.ttl` or `shacl.ttl`.

HolgerKnublauch (commented):
I am sorry but I don't agree with this change right now. It may be better to wait for input from other people, see the Issue where I explained why I find this triple redundant and we shouldn't repeat ourselves.

ajnelson-nist:
If this PR lacks further support by the end of the May 11 plenary call, I will close this without merging.

Issue #497, opened 2025-08-11
support for light/dark mode


VladimirAlexiev:
The specs should support light/dark mode.

@afs said
> ReSpec allows you to flip dark/light mode. We should support that flipping for diagrams as well.
> I don't know whether ReSpec records somewhere the dark/light setting (so it can be queried with CSS `@media`)

https://github.com/w3c/data-shapes/pull/483 has a discussion of light/dark mode for diagrams, and approaches how to handle that in PlantUML and RR (grammar railroad diagrams).

VladimirAlexiev:
@afs can you give an example of ReSpec light/dark mode?
https://raw.githack.com/w3c/data-shapes/master/shacl12-core/index.html has only light mode: no manual switch, and doesn't respect browser setting "dark mode".
Cheers!

afs:
@VladimirAlexiev -- see the reSpec documentation and the reSpec github repo for issues and current state.

VladimirAlexiev:
Thanks! https://respec.org/docs/#dark-mode

Added to https://raw.githack.com/vladimiralexiev/data-shapes/issue-486-ValidationReport-UML/shacl12-core/index.html#validation-report as a test:

<img width="810" height="841" alt="Image" src="https://github.com/user-attachments/assets/7b06e90c-a190-4785-902c-44d283606793" />

- the main text supports it
- Turtle examples don't
- JSONLD examples do, but at the expense of https://github.com/w3c/data-shapes/issues/477
- Images don't (see https://github.com/w3c/data-shapes/pull/483 discussion)

----

BTW googling "how many people use dark mode?" gave me some staggering numbers:

<img width="592" height="548" alt="Image" src="https://github.com/user-attachments/assets/8b6e7c06-f677-49f7-9e9e-0bc3dbc21452" />

TallTed:
dark mode just bit me. I cannot review #549 usefully, because the contrast between various text colors and their backgrounds (especially but not only the Turtle and JSON-LD code blocks) is far too small. I think that, because we cannot predict whether transition reviewers will be using dark mode or not, we must fix the stylesheets before such transition. Sooner is better, because we also need to be able to read the rendered documents.

HolgerKnublauch:
I have rather quickly merged @smessie 's PR so that dark mode works much better, for everyone's benefit.

There remain a couple of minor issues with it

- The UML diagram
- The .syntax class still has a white background, which somewhat breaks the overall darkness

smessie:
[...]

PR #577, opened 2025-09-22
Moved CSS styling to `shacl12-common/style.css` and improved dark mode support

smessie:
This PR is related to #497 but we'll have to see if this already fully solves the issue or if we want to do some more adjustments before the issue can be closed.

This PR moves the common CSS styling from the `<style>` block to a shared CSS file named `shacl12-common/style.css` for each `shacl12` document. These documents still retain the unmodified styling block as found in the core document.

For the other documents, I leave it to their respective editors to merge the styles and ensure that the resulting styling meets their expectations.

Styling moved to common for:
- shacl12-core
- shacl12-node-expr
- shacl12-profiling
- shacl12-ui

I’ve split this PR into two commits. The first one moves the styling, and the second one makes the necessary adjustments for improved dark mode support.
The latter primarily focuses on the code blocks that weren’t properly styled when using dark mode.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-497-light-dark-mode/shacl12-core/index.html)


HolgerKnublauch (approved):
Nice, thank you for tackling this.

PR #595, opened 2025-10-03
Add a Note for Test Suite Conformance

### LINKED ISSUE #594: Add a Note About the Test Suite Conformance Check
We can add a note in the conformance section to point to the test suite e.g.:

> Implementers can partially check their level of conformance with this specification by successfully passing the test cases of the [SHACL 1.2 Core test suite](https://w3c.github.io/data-shapes/data-shapes-test-suite/). Note, however, that passing all the tests in the test suite does not imply complete conformance to this specification. It only implies that the implementation conforms to the aspects tested by the test suite.

YoucTagh:
Closes #594 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-594-note-test-suite/shacl12-core/index.html#conformance)

### DIFF
--- a/shacl12-common/specifications.html
+++ b/shacl12-common/specifications.html
@@ -26,7 +26,7 @@
 
 			<p class="note">
 				Implementers can partially check their level of conformance with the above specifications by successfully passing
-				the test cases of the <a href="https://w3c.github.io/data-shapes/data-shapes-test-suite/">SHACL 1.2 test suite</a>.
+				the test cases of the <a href="https://github.com/w3c/data-shapes/tree/gh-pages/shacl12-test-suite">SHACL 1.2 test suite</a>.
 				Note, however, that passing all the tests in the test suite does not imply complete conformance to the specifications.
 				It only implies that the implementation conforms to the aspects tested by the test suite.
 			</p>
--- a/shacl12-common/specifications.html
+++ b/shacl12-common/specifications.html
@@ -24,6 +24,12 @@
 				<dd>defines the use of SHACL for profiling data, including SHACL data</dd>
 			</dl>
 
+			<p class="note">
+				Implementers can partially check their level of conformance with the above specifications by successfully passing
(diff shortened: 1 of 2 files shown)


HolgerKnublauch (commented):
The test suite will also cover SPARQL, node expr etc. So would it be better to move this block to an included shared file, or the overview?

ajnelson-nist:
> The test suite will also cover SPARQL, node expr etc. So would it be better to move this block to an included shared file, or the overview?

As this patch is written, the only described tests are for Core.  I'd support making the patch's statement a bit more general and moving the remark to an included shared file.

ajnelson-nist (approved):

shacl12-common/specifications.html
```diff
@@ -24,6 +24,12 @@ <h2>SHACL Specifications</h2>
 				<dd>defines the use of SHACL for profiling data, including SHACL data</dd>
 			</dl>
 
+			<p class="note">
+				Implementers can partially check their level of conformance with the above specifications by successfully passing
+				the test cases of the <a href="https://w3c.github.io/data-shapes/data-shapes-test-suite/">SHACL 1.2 test suite</a>.
```
HolgerKnublauch: That link is for the 1.0 tests. We still need to create a clone of that, which uses https://github.com/w3c/data-shapes/tree/gh-pages/shacl12-test-suite as its tests and has no implementation reports yet.

YoucTagh: Once the clone is ready, I will update the link and complete the merge.

HolgerKnublauch: I believe we need a new ticket here. As substitute for the old test case and implementation report document, we could elect to instead just point at the corresponding test case folder on GitHub, and leave the implementation report for a separate document (or also GitHub page) later.

So if you could simply insert https://github.com/w3c/data-shapes/tree/gh-pages/shacl12-test-suite as link for now and we create a ticket to make this page self-explanatory so that the test case format is described in Git and we can retire and unlink the old document?

HolgerKnublauch (approved):

Issue #534, opened 2025-08-28
[UI] Support for SHACL Property Paths


edmondchuc:
Add support by clarifying when a property path is applicable in viewing and editing scenarios.

Consider scenarios where property paths can be used for querying and how they are applied to user-interface use cases.

DASH Forms provide a couple of examples on how to render forms using property paths, including turning nested forms into a flattened form as a view.

Discussion:
- Which property paths are applicable to which scenarios?
- How are property paths rendered in forms?
- How are property paths used for querying?

Links:
- [SHACL 1.2 Core - SHACL Property Paths](https://www.w3.org/TR/shacl12-core/#property-paths)
- [Form Generation using SHACL and DASH - Path Expressions](https://datashapes.org/forms.html#paths)

edmondchuc:
### Predicate and Inverse Paths
A predicate path is the simplest type of property path for both editing and viewing RDF data constrained by SHACL shapes. It cleanly maps the _focus node_, _predicate path value_ and the _value node_ to the _subject_, _predicate_ and _object_ positions of an RDF triple. This mapping also applies to an inverse path, with the difference that the focus node and value node switch roles, corresponding to the _object_ and _subject_ position of an RDF triple, respectively. Understanding these mappings is essential when performing deletion and addition operations.

### Alternative Paths
Some thoughts on alternative paths and how they could be supported.

Given the following property shape:

```turtle
ex:PropertyShapeAlternativePath
    a sh:PropertyShape ;
    sh:name "author" ;
    sh:path [
        sh:alternativePath ( schema:creator dcterms:creator ex:creator [ sh:inversePath ex:created ] )
    ] .
```

For existing and new statements, the UI should provide a facility to update the predicate to one of the enumerated values in the alternative path, but only if they are a predicate or inverse path. This restriction is necessary for editing because the object value of an `sh:alternativePath` is a SHACL list that allows any well-formed SHACL property path, meaning any of the other SHACL property paths are also valid.

Although redundant, it's allowed in SHACL to nest `sh:alternativePath`. This just flattens to a list of predicate/inverse paths.

Nesting `sh:alternativePath`:

```turtle
ex:PropertyShapeAlternativePath
    a sh:PropertyShape ;
    sh:name "author" ;
    sh:path [
        sh:alternativePath ( schema:creator [
            sh:alternativePath ( dcterms:creator ex:creator )
        ] )
    ] .
```

Flattens out to be:

```turtle
ex:PropertyShapeAlternativePath
[...]

PR #600, opened 2025-10-09
Issue 534: Support for property paths in SHACL UI

smessie:
The document currently lacks screenshots but provides descriptions of the UI. We can add screenshots of potential UIs later, e.g., once we have a first implementation of a SHACL UI library.

Contributes to #534 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-534-ui-property-paths/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -805,6 +805,7 @@ ex:book2 a ex:Book ;
                     <a href="../shacl12-core/#property-path-zero-or-one">zero-or-one paths</a>)
                     in edit mode.
                 </p>
+
                 <p>
                     Implementers are encouraged to support these when their use cases require advanced
                     data navigation or editing patterns, but such support is not mandatory as the complexity of editing
@@ -821,9 +822,23 @@ ex:book2 a ex:Book ;
 
                 <aside class="example" title="Ambiguous complex paths in edit mode">
                     <p>
-                        The following data graph illustrates how a node can be reachable via more than one complex path,
-                        making editing operations ambiguous.
+                        The following shape and data graph illustrate how a node can be reachable via a complex path
+                        expression, making editing operations ambiguous.
                     </p>
+
+                    <div class="shapes-graph">
+                        <div class="turtle">
+ex:CompoundThingShape
+    a sh:NodeShape ;
+    sh:property [
+        sh:path [
+            sh:zeroOrMorePath ex:hasPart
+        ] ;
+        sh:name "All Descendant Parts" ;
+        sh:description "Traverses all parts and sub-parts using a zero-or-more path of hasPart relationships." ;
+    ] .
+                        </div>
+                    </div>
 
                     <div class="data-graph">
                         <div class="turtle">
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -419,6 +419,10 @@
             <td><code>ex:</code></td>
             <td><code>http://example.com/ns#</code></td>
           </tr>
+          <tr>
+            <td><code>dct:</code></td>
+            <td><code>http://purl.org/dc/terms/</code></td>
+          </tr>
         </table>
 
         <p>Within this specification, the following JSON-LD context is used:</p>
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
```diff
@@ -534,8 +539,265 @@ <h2>Core Constraints</h2>
...
+                    and display values, either directly from the focus node or through incoming relationships from other
+                    nodes.
+                </p>
+
+                <aside class="example" title="Predicate and inverse paths in view mode">
+                    <div class="shapes-graph">
+                        <div class="turtle">
+ex:PersonShape
+    a sh:NodeShape ;
+    sh:targetClass ex:Person ;
```
```suggestion
```

shacl12-ui/index.html
```diff
@@ -534,8 +539,265 @@ <h2>Core Constraints</h2>
...
+    a sh:NodeShape ;
+    sh:property [
+        sh:path ( ex:address ex:cityName ) ;
+        sh:name "City" ;
+        sh:datatype xsd:string ;
+    ] .
+
+ex:BookShape
+    a sh:NodeShape ;
+    sh:targetClass ex:Book ;
```
```suggestion
```

shacl12-ui/index.html
```diff
@@ -534,8 +539,265 @@ <h2>Core Constraints</h2>
...
+                    The following example illustrates how predicate and inverse paths can be used in edit mode to modify
+                    a person’s name and manage their membership in departments.
+                </p>
+
+                <aside class="example" title="Predicate and inverse paths in edit mode">
+                    <div class="shapes-graph">
+                        <div class="turtle">
+ex:PersonShape
+    a sh:NodeShape ;
+    sh:targetClass ex:Person ;
```
```suggestion
```

bergos (changes requested):
I propose removing the targets from the examples. Only the section about target finding should contain them. The other sections should not need them, which makes the examples shorter and easier to grasp.

The rest looks good to me.

edmondchuc:
I’m wondering if we should mention the restrictions on alternative paths in edit mode as described in https://github.com/w3c/data-shapes/issues/534#issuecomment-3316912350, specifically on which predicate paths are allowed.

Should we also add a paragraph on the challenges of supporting complex paths in edit mode?

Otherwise, everything else looks good to me.

smessie:
@edmondchuc I’ve pushed a commit that addresses your comment. Is it what you had in mind, or are there specific things you’re missing, or is it too verbose? I’m open to suggestions.

bergos:
@smessie I think your last additions to alternative path and other complex paths are very useful.

One small thing: Can you add a shape to the *Other Complex Paths* example? I think it would be useful to show `sh:path` in that case.

smessie:
@bergos I wasn't sure yet if a sequence path (`sh:path ( ex:hasPart ex:hasPart ) ;`) would maybe make more sense than the zero-or-more-path I've used now, but if you think it does, I can still change it.

bergos (approved):

bergos:
@smessie not required to change anything. The path is fine like it is.

TallTed:
Screenshots may still be overly ambitious, but wireframes may be helpful to comprehension.

shacl12-ui/index.html
```diff
@@ -534,8 +539,338 @@ <h2>Core Constraints</h2>
...
-      <p>Content.</p>
+        <h2>Property Paths</h2>
+        <p>
+            <a href="../shacl12-core/#property-paths">SHACL Property Paths</a> can be used to render a SHACL shape as a
+            user interface.
+            Property paths define how data values are accessed or modified relative to a focus node.
+        </p>
+        <p>
+            The following subsections outline the scenarios in which SHACL UI implementations
+            <strong>MUST</strong>, <strong>SHOULD</strong>, or <strong>MAY</strong> support different kinds of property
```
edmondchuc: We might not need to state these since it's already mentioned in section 1.4.

smessie: I kind of liked this just as an introduction of what is coming. It is not really defining the semantic meaning of the keywords as in 1.4. In fact, it even relies on that definition to introduce the subsequent subsections.

But if this is a strong opinion, I could remove the paragraph.

edmondchuc: I agree that the paragraph is useful as an intro to the upcoming subsections. I was just wondering if mentioning the RFC2119 variations is necessary, since it feels a bit repetitive. No strong opinions either way! And I see the `<strong>` tags have been removed. I'll give it a tick now. Thanks!

edmondchuc (approved):

shacl12-ui/index.html
smessie: Aha, right, thanks for the clarification! I have removed the repetitive keywords now.

PR #601, opened 2025-10-09
Short IRIs

nicholascar:
* Update all docs to use the short IRIs, as voted on by the WG in September
* rename the shacl12-compact-sytax directory to shacl12-cs, to match short IRI

Note that the vast majority of "changes" in this PR are just file renaming due to the above folder name change.

I'm asking @jeswr and @VladimirAlexiev to review because the folder name change will affect them the most.


jeswr (approved):
Thanks for tagging @nicholascar - no issues with the change on my side!

PR #605, opened 2025-10-13
#276: Fixed prefix declarations in test cases and adjusted examples to use .com instead of .org

### LINKED ISSUE #276: Namespace of `ex:` in tests

HolgerKnublauch:
Closes #276

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1713,7 +1713,7 @@ ex:Person-fullName
 	sh:name "full name" ;
 	sh:path ex:fullName ;
 	sh:values <b>[
-		sh:prefixes &lt;http://example.org/ns&gt; ;
+		sh:prefixes &lt;http://example.com/ns&gt; ;
 		sh:select """
 			SELECT ?fullName
 			WHERE {
@@ -1725,11 +1725,11 @@ ex:Person-fullName
 	]</b> ;
 	sh:datatype xsd:string .
 
-&lt;http://example.org/ns&gt;
+&lt;http://example.com/ns&gt;
 	a owl:Ontology ;
 	sh:declare [
 		sh:prefix "ex" ;
-		sh:namespace "http://example.org/ns#"^^xsd:anyURI ;
+		sh:namespace "http://example.com/ns#"^^xsd:anyURI ;
 	] .
 						</div>
 						<div class="jsonld">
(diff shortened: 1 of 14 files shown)


shacl12-sparql/index.html
```diff
@@ -1713,7 +1713,7 @@ <h3>Select Expressions</h3>
 	sh:name "full name" ;
 	sh:path ex:fullName ;
 	sh:values <b>[
-		sh:prefixes &lt;http://example.org/ns&gt; ;
+		sh:prefixes &lt;http://example.com/ns&gt; ;
```
afs, line 1716: (Why change `.org` to `.com`?)

Even better change - use `http://example/`.

It is specifically for this purpose.

[RFC 2606 section 2](https://www.rfc-editor.org/rfc/rfc2606#section-2)

>      ".example" is recommended for use in documentation or as examples.

It is registered and fixed. It does not have a website at that address by design.

HolgerKnublauch: The choice between .org and .com was made before, not by me. I am merely aligning these test cases with all others.

HolgerKnublauch: @afs may I ping you on this old PR?

TallTed: The domains `example.org.` and `example.com.` (and `example.net.`) are just as acceptable as `example.` See [RFC 2606 section 3](https://www.rfc-editor.org/rfc/rfc2606#section-3).

Things do get more entertaining when trying to test (and not just document) with these domains, as discussed in [RFC 6761 section 6.5](https://www.rfc-editor.org/rfc/rfc6761), where it is said:
```
   The domains "example.", "example.com.", "example.net.",
   "example.org.", and any names falling within those domains, are
   special in the following ways:

   1.  Users SHOULD understand that example names are reserved for use
       in documentation.

   2.  Application software SHOULD NOT recognize example names as
       special and SHOULD use example names as they would other domain
       names.

   3.  Name resolution APIs and libraries SHOULD NOT recognize example
       names as special and SHOULD NOT treat them differently.  Name
       resolution APIs SHOULD send queries for example names to their
       configured caching DNS server(s).
```

TallTed: > The choice between `.org` and `.com` was made before, not by me. I am merely aligning these test cases with all others.

I think it would have helped @afs (as well as myself) if the initial comment had noted that "adjusted examples to use `.com` instead of `.org`" was just a "make all the examples use the same `.tld`" alignment, and not any kind of judgement about `.com` being better than `.org`. That said, I'm wondering whether there is more value in varying the `.tld` than in unifying them.

afs: > @afs may I ping you on this old PR?

To quote [RFC 2606 section 2](https://www.rfc-editor.org/rfc/rfc2606#section-2):  ".example" is recommended for use in documentation or as examples.

That's a suggestion, not a blocker.

afs (approved):

PR #628, opened 2025-10-29
move to local copies of CSS files

### LINKED ISSUE #626: Publication of Core and SPARQL fails due to common css file dependency
While the other build errors have disappeared, the publication process is now stuck because the shacl12-common/style.css is not in the same folder and/or cannot be found:

https://github.com/w3c/data-shapes/actions/runs/18834484351/job/53732261624

nicholascar:
Fixes #626

Copying CSS file from `common/` to each spec's folder to overcome a W3C tool processing issue.

I've also moved UI & CS to using the common CSS file, not inline CSS.


HolgerKnublauch (approved):
Awesome, let's hope this unblocks the build

edmondchuc (approved):
LGTM, thanks!

PR #655, opened 2025-11-17
#650: Added link to VarExpression to clarify what "variable" means here

### LINKED ISSUE #650: No definition of "variable"
There isn't a definition of "variable" nor does it say it is the same as a SPARQL variable.

Section: https://www.w3.org/TR/sparql12-query/#sparqlQueryVariables
Definition: https://www.w3.org/TR/sparql12-query/#defn_QueryVariable

HolgerKnublauch:
Closes #650

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-650/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -980,7 +980,7 @@ ex:Person-displayName
 							with the <a>function name</a> <code>shnex:EmptyExpression</code>.
 						</span>
 					</p>
-					<div class="def" id="VarExpression-evaluation">
+					<div class="def" id="EmptyExpression-evaluation">
 						<div class="def-header">EVALUATION OF EMPTY EXPRESSIONS</div>
 						<p>
 							An <a>empty expression</a> has the empty list <code>[]</code> as its <a>output nodes</a>.
@@ -2574,8 +2574,8 @@ ex:Person-fullName
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
-					Note that the scope in the evaluation of expression constraints contains the <a href="#VarExpression">variable</a> <code>value</code>
-					for the current <a>value node</a> while the variable <code>focusNode</code> is the <a>focus node</a>.
+					Note that the <code>scope</scope> in the evaluation of expression constraints the mapping
+					<code>value</code> to be the current <a>value node</a>.
 				</p>
 				<p>
 					The following example uses some SPARQL-based node expressions to declare
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -2256,7 +2256,7 @@ ex:Person-fullName
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
-					Note that the scope in the evaluation of expression constraints contains the variable <code>value</code>
+					Note that the scope in the evaluation of expression constraints contains the <a href="#VarExpression">variable</a> <code>value</code>
 					for the current <a>value node</a> while the variable <code>focusNode</code> is the <a>focus node</a>.
 				</p>
 				<p>
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -1039,9 +1039,9 @@ ex:Person-displayName
 						creating a self-reference.
 					</p>
 				
-						<aside class="example" title="A var expression stating that any Person loves him or herself by default.">
-							<div class="shapes-graph">
-								<div class="turtle">
+					<aside class="example" title="A var expression stating that any Person loves him or herself by default.">
+						<div class="shapes-graph">
+							<div class="turtle">
 ex:Person
 	a sh:ShapeClass ;
 	sh:property ex:Person-loves .
@@ -1050,9 +1050,9 @@ ex:Person-loves
 	a sh:PropertyShape ;
 	sh:path ex:loves ;
 	sh:defaultValue <b>[ shnex:var "focusNode" ] .</b>
-								</div>
 							</div>
-						</aside>
+						</div>
+					</aside>
 					
 				</section>
 
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2006,7 +2006,7 @@ ex:SomeClass-alternativePathExample
 					During evaluation, the engine can access <a>triples</a> related to <code>expr</code> in the <a>shapes graph</a>.</li>
 					<li><code>focusGraph</code> is a <a>graph</a>, called the <dfn>focus graph</dfn>. This is the default query graph for the evaluation of the node expression.</li>
 					<li><code>focusNode</code> is a <a>node</a>, called the <dfn>input focus node</dfn>. This variable may have no value.</li>
-					<li><code>scope</code> is a map from (key) <a>nodes</a> to individual (value) <a>nodes</a>.
+					<li><code>scope</code> is a map from (key) <a>terms</a> to individual (value) <a>terms</a>.
 					The empty map is written as <code>{}</code>.
 					</li>
 				</ul>


shacl12-node-expr/index.html
```diff
@@ -2256,7 +2256,7 @@ <h3>sh:expression</h3>
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
-					Note that the scope in the evaluation of expression constraints contains the variable <code>value</code>
+					Note that the scope in the evaluation of expression constraints contains the <a href="#VarExpression">variable</a> <code>value</code>
```
afs: At least:

```suggestion
					Note that the scope in the evaluation of expression constraints contains the <a href="#dfn-var-expression">variable</a> <code>value</code>
```
although this called a "var expression" not "a variable" so that seems to make it:

```suggestion
					Note that the scope in the evaluation of expression constraints contains the <a href="#VarExpression">var expression</a> <code>value</code>
```

The definition:

> A [blank node](https://www.w3.org/TR/rdf12-concepts/#dfn-blank-node) that is the [subject](https://www.w3.org/TR/rdf12-concepts/#dfn-subject) of the following properties is called a var expression with the [function name](https://www.w3.org/TR/shacl12-core/#dfn-function-name) shnex:VarExpression:

says the blank node _is_ the var expression and _is_ an RDF term.

If you want to keep it as the current design, please keep using "var expression" and then shacl12-rules can define `variable`. (There are only two other uses of "variable" in the node expressions doc at the moment.)

HolgerKnublauch: We don't even need to talk VarExpressions at all here. The key is the term "scope". And scope is defined in SHACL 1.2 core as

					<li><code>scope</code> is a map from (key) <a>nodes</a> to individual (value) <a>nodes</a>.
					The empty map is written as <code>{}</code>.

So I think we can avoid the term variable in that definition and just refer to mappings in the scope, where the key in the map is the value of shnex:var of the var expression.

I guess the proper fix is then to create a deep link to the scope definition and reference that from node expressions.

afs: The map is from a string (taken from `shnex:var`) to RDF term.

If it is a node (RDF Term), then

```
[ shnex:var "foo" ] .
[ shnex:var "foo" ] .
```
is two different scope entries.

If there is no varExpression, how does a node expression work

```
  [ sparql:plus ( [ shnex:var foo" ] 1 ) ]
```
Something has to define 
```
   exprEval([ shnex:var "foo" ])
```
and that to be an error if "foo" is not in the scope, which is returning an empty list from scope lookup.

HolgerKnublauch: I struggle to parse your response.

```
[ shnex:var "foo" ] .
[ shnex:var "foo" ] .
```

are two times the same expression. Why should those be "different scope entries"?

The sentence above the sparql:plus example seems unfinished. What do you mean?

We could change the handling of undefined variables but then how could we distinguish unbound (optional) parameters from those called by accident in a node expression? There is no "null" value here like in JavaScript.

Anyway, I made further adjustments to the PR and believe external documents can refer to the variable <code>scope</code> by name when they refer to evalExpr.

afs: > I struggle to parse your response.
> 
> ```
> [ shnex:var "foo" ] .
> [ shnex:var "foo" ] .
> ```
> 
> are two times the same expression.

And what _RDF term_ do they evaluate to? That's necessary because:

> scope is a map from (key) [nodes](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-node) to individual (value) [nodes](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-node).

(PS better to say "RDF term", not "nodes" which are the existing subject/objects of a graph)

HolgerKnublauch: If "foo" is not in the scope, then this rule applies

otherwise evalExpr(expr, focusGraph, focusNode, scope) -> []

i.e. the var node expression returns the empty list.

And OK I have switched the definition to "term". If we want to be consistent and serious about that, I believe many many other places would be affected though, e.g. node expressions would need to be renamed to term expressions. This would probably alienate users though, as the term "term" is barely used outside of W3C specs, while "node" is well-known.

ajnelson-nist: This is probably me still shaking cobwebs off from being AFK a while, but I need a reminder on one of the less intuitive ideas from RDF:

Is a literal a node?
Is a literal a term?

Update for RDF 1.2:

Is a triple term a node?
Is a triple term a term?  (...If it isn't, why?)

afs: RDF Terms : https://www.w3.org/TR/rdf12-concepts/#section-terms 
IRIs, Blank nodes, literals, triple terms.

Nodes: https://www.w3.org/TR/rdf12-concepts/#dfn-node

"Nodes of a graph" -- the set of Subjects and Objects in use in a graph (not if only in the predicate position, of if only inside triple terms). As in "nodes and arcs" for graphs.

Triple terms: https://www.w3.org/TR/rdf12-concepts/#section-triple-terms
It is a triple in the role of an RDF term.

> Is a literal a term?

Yes

> Is a literal a node?

If it is in the graph

> Is a triple term a node?

If it is in the graph as a `<<( )>>` thingy.

>Is a triple term a term? (...If it isn't, why?)

Yes

afs: > If "foo" is not in the scope, then this rule applies

So scope is mapping a string, not a variable.

HolgerKnublauch: Yes, it maps term to term, and I have taken the word "variable" out of the paragraph that started this thread.

Are you suggesting to rename shnex:var to something else?

afs: Why can't you say that the subject of `[ shnex:var "foo" ]` is (as in "describes", "represents" or "denotes", not in the RDF syntax where is a blank node) a <dfn>variable</dfn> with name "foo"?

After all:
```
shnex:var	sh:datatype xsd:string       The variable name, e.g. "focusNode".
            sh:minLength 1	        
```


Var expressions are expressions. They yield an RDF term when evaluated.
Here, they are the use of a variable in a role.

HolgerKnublauch: I wanted to avoid the (formal) term variable because it is overloaded and in #650 you complain that there is no formal definition of variable. At this stage I am really not sure what you want here. Would you like to produce a PR or specific suggestion yourself instead of me guessing?

afs: Indeed #650 is a request have a definition or at least an anchor for "variable".

How can rules say "treat `[ shnex:var "foo" ]` as a (SPARQL) variable in filter expressions?
How can node expressions say "The variable name" in defining `shnex:var` without the domain being a variable?

HolgerKnublauch: The anchor is for "var expression" https://w3c.github.io/data-shapes/shacl12-node-expr/#dfn-var-expression and you can then refer to the variable name as 

```
the <a>value</a> of <code>shnex:var</code> of the <a>var expression</a>.
```

this avoids me having to introduce a formal term for "variable" and other documents can freely redefine it.

If that's acceptable, please still approve this PR because it does some clean up.

afs: > this avoids me having to introduce a formal term for "variable" and other documents can freely redefine it.

Rules will now have to define evaluating node expressions because the input is a variable from earlier in the rule pattern.

Since we aren't defining a common evaluation model anymore, go ahead and merge this.
I'm not an editor of "Node Expression" so I will approve it this once.

afs (approved):
We aren't defining a common evaluation model anymore and I'm not an editor of "Node Expression".

shacl12-node-expr/index.html
```diff
@@ -2574,8 +2574,8 @@ <h3>sh:expression</h3>
 				</div>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
-					Note that the scope in the evaluation of expression constraints contains the variable <code>value</code>
-					for the current <a>value node</a> while the variable <code>focusNode</code> is the <a>focus node</a>.
+					Note that the <code>scope</scope> in the evaluation of expression constraints the mapping
```
TallTed, line 2577:
```suggestion
					Note that the <code>scope</code> in the evaluation of the expression constrains the mapping
```

Issue #643, opened 2025-11-07
Packaging rules, shapes, and data, with a defined execution order


liviorobaldo:
In the SHACL Inference Rules Task Force, we are defining the table of contents for the SHACL 1.2 Rules draft.

See [discussion 637](https://github.com/w3c/data-shapes/discussions/637). The provisional table of contents includes "7. Attaching Rules to Shapes", where we should explain how rules and shapes should work together. Below in the discussion, I propose moving this content to [2. Packaging SHACL](https://w3c.github.io/data-shapes/shacl12-profiling/#packaging). Perhaps in "7. Attaching Rules to Shapes" we could just include a brief pointer to [2. Packaging SHACL](https://w3c.github.io/data-shapes/shacl12-profiling/#packaging).

Concerning how rules and shapes should be executed together, in [discussion 603](https://github.com/w3c/data-shapes/discussions/603#discussioncomment-14727042), based on my experience with SHACL-SPARQL, I proposed to associate _sets_ of shapes with _sets_ of rules. This can be done with a new resource we might call "cluster" (but we should actually pick a better name, like "bundle", "package", "module", etc.), which would group data with shapes, with rules, or with both (three options in total). Something like:

```
:cluster-3
  rdf:type srl:Cluster;
  srl:data (
    ...
  );
  srl:ruleSet (
    ...
  );
  srl:shapeSet (
    ...
  ).
```

If both shapes and rules are present, we need to decide the execution order: should shapes run first, or the rules? As I argued in my recent paper (link provided at the beginning of the discussion), shapes should be executed at least once _after_ the rules. This is because what we want to validate is the _inferred knowledge graph_, consisting of the initially asserted triples plus all triples that can be logically inferred from them. This aligns with the reasonable assumption that if something logically inferred is invalid, then the originally asserted triples must also be considered invalid. Rules serve then to "modularize the effort" when it is too complex, if not impossible, to write a single shape that internally computes all inferred triples before validation.

However, nothing prevents executing the shapes multiple times. For example, if the shapes already invalidate the initially asserted triples, there is no need to compute the inferred knowledge graph. Shapes could also be re-executed after each "round" of rules execution (rules are iteratively re-executed until no new triple is inferred). However, this approach could be computationally expensive.

[...]

PR #659, opened 2025-11-18
Define rule set evaluation

afs:
Work to define rule set evaluation.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/def-eval/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -49,10 +49,15 @@
           }
         ],
 
+        //implementationReportURI: "http://w3c.github.io/data-shapes/shacl12-test-suite/tests/rules",
+        //testSuiteURI: "http://w3c.github.io/data-shapes/shacl12-test-suite/tests/rules",
+
+        // No previous REC
         //previousPublishDate: "2017-07-20",
         //previousMaturity: "REC",
         //prevRecShortname: "shacl",
         //prevRecURI: "https://www.w3.org/TR/2017/REC-shacl-20170720/",
+
 
         localBiblio: {
           "shacl12-shacl-shacl": {
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -288,7 +288,7 @@ RULE { ?a :ancestorOf ?b } WHERE { ?a :ancestorOf ?c . ?c :ancestorOf ?b }
 # Default value - calculate a name
 RULE { ?x :name ?FN } WHERE { 
     ?x rdf:type :Person 
-    NOT EXISTS { ?x :name ?FN }
+    NOT { ?x :name ?someName }
     ?x :givenName ?name1 ;
        :familyName ?name2 .
     BIND(concat(?name1, " ", ?name2) AS ?FN)
@@ -471,6 +471,12 @@ RULE { ?x :name ?FN } WHERE {
             assignment in a rule introduces a new variable,
             not used earlier in the rule body.
           </p>
+          <div class=ednote>
+            <p>
+              Add a rule for nested patterns e.g. negation.
+            </p>
+          </div>
+
           <p>
             A [=rule=] is a <dfn>well-formed rule</dfn> if all of the following
             conditions are met:
@@ -930,8 +936,9 @@ the result is GI
     <section id="shapes-rules-grammar">
       <h2>Shapes Rules Language Grammar</h2>
 
-        <div class="grammarTable">
-          <table><tbody>
+      <!-- GRAMMAR -->
+      <div class="grammarTable">
+        <table><tbody>
 
               <tr style="vertical-align: baseline">
                 <td><code>[1]&nbsp;&nbsp;</code></td>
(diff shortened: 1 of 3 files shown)


afs:
Very early work to show direction.

osavkovic:
> Very early work to show direction.

Hi Andy, could you say a bit more about what kind of input you expect here? For instance, should I provide comments and edits, and should I make them in the document or here?

Thanks, O.

shacl12-rules/index.html
```diff
@@ -459,14 +436,23 @@ <h2>Shape Rules Abstract Syntax</h2>
...
 
-          <dt><dfn data-lt="expression">node expression</dfn></dt>
+          <dt><dfn>expression</dfn></dt>
           <dd>
-            <em>Link to shacl12-node-expr#?</em> -- needs to be single value, list arguments.
-            Often called an <em>expression</em>.
+            <p>
+              An [=expression=] is a function, or functional form.
+              An expression is evaluated with respect to a [=solution mapping=] to give
+              an [=RDF term=] result.
```
```suggestion
              An expression is evaluated with respect to a [=solution mapping=] to
              result in an [=RDF term=].
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
         It does not prescribe the algorithm as the method of implementation.
-        An implementation can use any algorithm that generates the same outcome.
+        An implementation can use any algorithm that generates the same outcomes.
       </p>
-
-<pre class="algorithm">
+      <pre class="algorithm">
 Inputs: data graph G and a rule set RS.
 Output: an RDF graph GI of inferred triples
+      <p>The inferred triples does not include any triple present in the set of triples of G.</p>
```
```suggestion
      <p>The inferred triples do not include any triple present in the set of triples of G.</p>
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
+            is a function <code>subst(<var>μ</var>, [=triple pattern=])</code> that returns a triple pattern
+            where each occurrence in the triple pattern of a variable that is in the <code><var>dom</var>(<var>μ</var>)</code>
+            is replaced by the [=RDF Term=] given by the [=solution mapping=] for <var>var</var>.
+            If the triple pattern result has no variables, then it is an [=RDF Triple=].
+          </dd>
+          
+          <dt><dfn>Evaluation Graph</dfn></dt>
+          <dd>
+            A [=evaluation graph=] is an [=RDF Graph=] that combines the base graph 
+            and all triples produce by rule during the the evaluation of a rule set.
```
```suggestion
            and all triples produced by rules during the the evaluation of a rule set.
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
+
+          <dt><dfn data-lt="compatible">solution compatible</dfn></dt>
+          <dd>
+            Two solutions S1 and S2 are [=compatible=] if they agree on the variables in common.
+            <div class="def">
+              <p>
+              Let S1 and S2 be solutions.
+              </p>
+              <pre>compatible(<var>μ</var><sub>1</sub>, <var>μ</var><sub>2</sub>) = true
+                      if forall v in dom(μ1) intersection dom(μ2)
```
TallTed, line 761: "Forall" (with no space) is a programming-language-specific function. "For all" (with space) is English. These blocks appear to be an English narration, which calls for "for all". If these are blocks of some programming language, they all need to be reviewed with an eye to that vocabulary, so there must be a declaration of what that language is.
```suggestion
                      if for all v in dom(μ1) intersection dom(μ2)
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
+          <dd>
+            A [=solution sequence=] is a multi-set of solutions.
+            There is no defined order to the sequence.
+            It is equivalent to an unordered list and it can contain duplicates.
+          </dd>
+
+          <dt><dfn data-lt="merge">solution merge</dfn></dt>
+          <dd>
+            If two solutions are comptible, the merge of two solutions is the solution that maps variables of
+            each solution to the [=RDF term=] from one or other of the solutions.
```
```suggestion
            If two solutions are compatible, the merge of two solutions is the solution that maps
            variables of each solution to the [=RDF term=] from one or other of the solutions. 
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
+            A [=solution sequence=] is a multi-set of solutions.
+            There is no defined order to the sequence.
+            It is equivalent to an unordered list and it can contain duplicates.
+          </dd>
+
+          <dt><dfn data-lt="merge">solution merge</dfn></dt>
+          <dd>
+            If two solutions are comptible, the merge of two solutions is the solution that maps variables of
+            each solution to the [=RDF term=] from one or other of the solutions. 
+            (It is the union of solutions if defined as sets of pairs.)
```
```suggestion
            (It is the union of solutions if they are defined as sets of pairs.)
```

shacl12-rules/index.html
```diff
@@ -728,40 +714,135 @@ <h4>Compact Syntax Abbreviations</h4>
...
 <pre class="algorithm">
 
 Let F(arg1, arg2, ...) be an expression.
 where arg1, arg2 are RDF terms.
 
-Let [x/row] be
+Let [x/<var>μ</var>] be
     if x is an RDF term, the [x/row] is x
-    if x is a variable then [x/row] is the value of x in the row
+    if x is a variable then [x/row] is <var>μ</var>(x)
```
```suggestion
    if x is a variable, then [x/row] is <var>μ</var>(x)
```

afs:
> > Very early work to show direction.
> 
> Hi Andy, could you say a bit more about what kind of input you expect here? For instance, should I provide comments and edits, and should I make them in the document or here?
> 
> Thanks, O.

This is a (currently draft - i.e. unfinished) PR.

You can make comments and suggest changes here.

All changes come in via pull requests; we do not edit the document directly. 

PRs are checked for IPR acceptability automatically. Merging PRs has a policy; in this repo, the policy is to have at least one approval. That is, at a minimum, for governance; someone else checked the submitter has followed process and isn't doing anythign silly. But usually an approval is a general agreement on the content and implied direction. 

Changes are not fixed - a later PR can revise designs. This is common.

The default branch of this repo, unusually called `gh-pages`, not `main`, should be branch-protected, meaning no one can push directly to it.

liviorobaldo:
Dear Andy and all,

I tried to update the document. Apologies in advance if I made any mistakes or misunderstood something. It’s my first time working on a W3C working group repo. However, from what I understood from Andy, my update won’t go through until some of you approve it, so feel free to delete anything if I got it wrong 🙂

I thought we should clarify in “2. Outline” that rules can also create new RDF resources and new literals. I included an example of the former (an example of the latter is already in “3. Speculative features”, where a new literal is created and bound to `?FN` using `concat`). I also pointed out that this may lead to infinite loops if a user keeps re-applying the rules. However, as I understand it, we agreed [in this issue](https://github.com/w3c/data-shapes/issues/643#issuecomment-3517263369) that avoiding such loops is beyond the scope of SHACL rules; it is up to the programmer to iterate rules only under conditions that prevent infinite loops.

By the way, in “3. Speculative features”, I think we should discuss negation-as-failure and aggregates. The former is already there. However, I’m not clear on what the title “Speculative features” is meant to convey. From what I recall, we decided that we want negation-as-failure and aggregates, so they shouldn’t really be described as speculative.

Cheers
Livio

shacl12-rules/index.html
afs: It is "programming language" in a general sense as is used in formal descriptions, including academic papers.
Later, we can replace `forall` with `∀`. Or not. This draft PR is work-in-progress.

afs:
@liviorobaldo 

This PR is titled "Define rule set evaluation".
That is its focus and the sections of the document it changes.
The PR was "draft".

You have broken the work I have in-progress and made it hard to deal with @TallTed's suggestions.

Please make a separate PR.

liviorobaldo:
Dear Andy and all,

My apologies, I didn’t do this on purpose. As I mentioned in my previous message, I’m still learning how to use these tools and understand the collaboration policies for working with all of you.

I now understand that this PR is meant to focus only on Section "6. Shape Rules Evaluation" of the document, so we should restrict comments and edits to that section.

Later today, I will open a new PR titled "SHACL 1.2 Rules - Introduction & Outline", dedicated solely to the first three sections of the document. I will also make sure to explicitly note the scope and invite all of you to review it.

Could you kindly confirm that this is the correct approach? Is there anything else I should pay attention to when I'll create the PR? If I'll make some mistake I will of course delete the new PR.

Thanks again for your patience — I’m still learning and will do my best to avoid causing any inconvenience.

Livio

afs:
Please remove your commit from this PR.

liviorobaldo:
Dear Andy,

Ok, apologies again. I am trying to fix my mistake while learning how to use this platform and, of course, to avoid making further errors. I’ve installed GitHub Desktop and opened this branch. I located my commit and can see that “Revert changes in commit” is one of the available options:

image

Is this the option I should click on?

Thanks for your patience
Livio

afs:
I don't use GitHub Desktop. It's not available for my OS.

I would have thought you want "reset to commit" where the commit is the place you want to keep. "Define ..."

I hope changes will be made locally so you can check them before pushing them to github.

liviorobaldo:
Hi Andy,

"Reset to commit" is not selectable, as you can see in the screenshot above. Let me try “Revert changes in commit”. If it does not work or create other troubles, I will remove the content I've added by hand.

As soon as I've time, I will then open a new PR titled "SHACL 1.2 Rules - Introduction & Outline" as I indicated above.

Livio

liviorobaldo (approved):

liviorobaldo:
PS. I think I managed, can't see anymore the text I've added before in the [document rendered online here](https://raw.githack.com/w3c/data-shapes/def-eval/shacl12-rules/index.html).

shacl12-rules/index.html
TallTed: Hm. I would prefer `for all` (and `for each`), but I won't block on this, especially if the overall WG prefers `forall` (and `foreach`).

afs:
> "Reset to commit" is not selectable,

Says you can use right-click on the commit you wish to go back to.
https://docs.github.com/en/desktop/managing-commits/resetting-to-a-commit-in-github-desktop

Your screenshot shows the HEAD as focus, not the "Define ..."  commit.

“Revert changes in commit” applies your commit but in reverse. See all the lines starting `-`; the result is 2 commits.

I have reset the branch with a `git push -f`. A dangerous way to do it (I took a local copy for safety first.) Also - reset the draft flag and cleared the approved flag.

The HEAD is `[77aed92] andy@apache.org (3 days ago) - Define rule set evaluation`

liviorobaldo:
Andy, thank you very much for fixing the troubles I caused. Is there anything I am supposed to do? I think no, because I can now only see one commit, so I think the situation has been restored.

shacl12-rules/index.html

This is a draft at the moment.
There are significant changes in-progress.

shacl12-rules/index.html

shacl12-rules/index.html
afs: `for each` is different - it's `for` loop.

It could be `forall X in ...`.

This PR is someway from being "finished". I put it up to show the direction.

shacl12-rules/index.html
afs: Removed. The whole `( )` is wrong.

shacl12-rules/index.html
afs: Done - and the line above it.

shacl12-rules/index.html
afs: Done - though should be rewritten to use mapping composition. `S1 &compfn; S2`

liviorobaldo:
Dear Andy and all,

I saw a new commit "[Define stratification](https://github.com/w3c/data-shapes/pull/659/commits/2b2293985187dd1146f110845d02df1000bc82f5)" has been opened, so I assume we are now expected to comment on that specific part.

Since last time I caused mistakes you had to fix, I would like to double-check with you before making any changes to the file.

I would like to add a bit more of intro here: 

<pre>
<p> Stratification is the process of labelling <a>rules</a> based on their <a>dependency</a> relationships in a <a>rule set</a>. The set of rules with the same label are called a <a>layer</a>. </p>
</pre>

More specifically, I think we should explain why stratification is discussed in this document, i.e., that stratification is needed to avoid non-determinism when non-monotonic operators such as, in our case, negation-as-failure and aggregates, are involved, that it enforces an implicit execution order among rules ensuring consistent results, that all automated reasoners that implement the SHACL Rules specification are required to support stratification to ensure consistent behavior across systems, etc.

The intro shouldn't go into tech details, of course, but I think it is important to give newcomers a clear idea of the role of stratification and why the SHACL Rules specification requires it. I do not think we can take it for granted. For example, although I do not know how much I am representative of the standard SHACL users, as you may recall, when I joined this working group I did not know what stratification was, although I am been using SHACL for four years already 😅

Thus, I will proceed to edit index.html as described above unless you think this would create further issues that would need to be fixed. Please bear with me as I am still learning how to work with these tools.

From the web interface, I understood I have to press "Edit" here and then change the HTML code:

image

Will do asap
Livio

afs:
> we should explain why stratification is discussed in this document,

yes - but explanation goes elsewhere, not this PR.

Explanation is not definition.

>  I understood I have to press "Edit" here and then change the HTML code:

No. 
We use pull requests.

Your previous attempt was to section 2.
You can make a separate PR for any part of the document, relative to the state on branch `gh-pages`. Editing or PR'ing gains a another branch will be time consuming to keep in-step.

As the PR here does not change section 2, you can start a PR now.

Read up about git workflow. There is plenty of online material.

A _brief_ summary:

1. Clone the repo
2. Create a local branch (`git switch -c`)
3. Make local changes, commit locally with sensible log messages.
4. Push the branch ("git push -u")
5. Create a PR by going to the github interface. Step 3 gives you the URL.
6. You can keep making changes locally.
7. You can respond to suggestions in the github UI. NB This update your brach so you need to pull it into you rlcoal copy to stay in-step.

For a PR of any size, create a github issue first to record what it is able.

I will try to get this PR into a state where it can be put into the document to provide a baseline for the formal definition sections.
It is a _draft_ so you can see progress or lack thereof. 

As a draft I regular reset the commits in-place to keep the commit history clean (something you can't do on the `gh-pages` branch).

Please give me time to do some light tidying up (as I said in the telecon).

liviorobaldo:
Ok! Fortunately, this time I asked before taking action 😅
- I will study the Git workflow.
- I will wait a bit for you to do some light tidying up.
- I will create a completely new PR (on the same document, but for a different part, i.e., the introduction).

Thanks
Livio

shacl12-rules/index.html
```diff
@@ -1805,6 +1877,70 @@ <h2>Shapes Rules Language Grammar</h2>
...
+          <dd>"TEXT"</dd>
+          <dt>Person & email address to contact for further information:</dt>
+          <dd>@@</dd>
+          <dt>Intended usage:</dt>
+          <dd>COMMON</dd>
+          <dt>Restrictions on usage:</dt>
+          <dd>None</dd>
+          <dt>Author/Change controller:</dt>
+          <dd>The SHACL Rules 1.2 specification is a work product of the World Wide Web Consortium's
+            RDF-star Working Group. The W3C has change control over these specifications.</dd>
```
robert-david: Shouldn't this be the data shapes working group?

robert-david (changes requested):

shacl12-rules/index.html
afs: Done.

afs:
@robert-david -- you have directly changed this branch, rather than either making a reviewable suggestion via the GH interface, or by making a separate PR to this branch. 

The direct commit invalidated work I had on my copy and merge/rebase lost that content. I have taken you commit and made the changes manually, then restored the lost content, and made a step towards a cleaner commit to set the initial content for evaluation.

Later today or tomorrow, I will update the grammar with the negation feature.

Then I think it is adequate (barely) to be merged, mistakes and all.

Merge does not mean the content can not be changed; it means further PRs can made. 

I will ask the editors for approval to merge.

There is still a lot for work to do.

afs:
> the grammar with the negation feature.

Syntax for negation:
First take at a keyword is `NOT`, with an argument which is restricted form of "body" that allows only triple patterns and filters.

```
RULE { ?x :name ?FN } WHERE { 
    ?x rdf:type :Person 
    NOT { ?x :name ?fullName }
    ?x :givenName ?name1 ;
       :familyName ?name2 .
    BIND(concat(?name1, " ", ?name2) AS ?FN)
}
```

Editors - I think this is better as the first-public editors working draft (that is, a document at https://www.w3.org/TR/shacl12-rules) than the current document on the default branch.

Please review and if, it is adequate for a FPWD, please approve it.

ajnelson-nist:
> > the grammar with the negation feature.
> 
> Syntax for negation: First take at a keyword is `NOT`, with an argument which is restricted form of "body" that allows only triple patterns and filters.
> 
> ```
> RULE { ?x :name ?FN } WHERE { 
>     ?x rdf:type :Person 
>     NOT { ?x :name ?fullName }
>     ?x :givenName ?name1 ;
>        :familyName ?name2 .
>     BIND(concat(?name1, " ", ?name2) AS ?FN)
> }
> ```
> 
[snip]

A potential friction point: Current SPARQL muscle-memory may lead many users to type `NOT EXISTS` instead of `NOT`.

osavkovic:
A small observation: In logic programming, datalog, etc, rules typically assume safe negation meaning that every variable appearing under a negated (NOT) part must already be bound by a positive predicate in the rule. In this example, this condition is satisfied since concat provides the necessary bindings. However implementing such a syntactic check in the general case may not be trivial because it is not always clear what other kinds of variable dependencies it may exist in the rule.

Best
Ognjen

afs:
> A potential friction point: Current SPARQL muscle-memory may lead many users to type `NOT EXISTS` instead of `NOT`.

Actually, that's by design! I wanted to call out the differences for discussion. The name can change.

In SPARQL, `NOT EXISTS` is inside a filter, and maybe combined with other expressions including "not" in the form of `!`. The graph pattern is arbitrary SPARQL.

In Shape Rules, the `NOT` is a body element, not inside `FILTER` - this means it is not embedded inside an expression.
The pattern is triple patterns and filters only.

There is also @osavkovic 's point -- we may need other restrictions as well so really the path forward is keep separate until proven that the muscle memory is reliable.

Negation is important to discuss.  There's now issue #674 for negation discussion.

afs:
> A small observation: In logic programming, datalog, etc, rules typically assume safe negation meaning that every variable appearing under a negated (NOT) part must already be bound by a positive predicate in the rule. In this example, this condition is satisfied since concat provides the necessary bindings. However implementing such a syntactic check in the general case may not be trivial because it is not always clear what other kinds of variable dependencies it may exist in the rule.

Longer comment now on #674. 

I think this is possible because variables inside the `NOT` pattern are wildcards as far as the overall rule body is concerned and is conceptually another rule bidy.

The example is not great as it used `?FN` in two ways. (PR fix applied - rename the `NOT` usage of `?FN` as a variable not used anywhere else.)

(You can't write `NOT` inside `NOT { }`)

We do have well-formedness conditions. We need something for stratification and that's not yet included.

shacl12-rules/shacl-rules.ttl
```diff
@@ -0,0 +1,8 @@
+# SHACL Rules namespace file.
+
+PREFIX :       <http://example/>
+PREFIX rdf:    <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX sh:     <http://www.w3.org/ns/shacl#>
+PREFIX shrl:   <http://www.w3.org/ns/shacl-rules#>
+PREFIX shnex:  <http://www.w3.org/ns/shacl-rules#>
```
simonstey:
```suggestion
PREFIX shnex:  <http://www.w3.org/ns/shnex#>
```

shacl12-rules/index.html
```diff
@@ -196,163 +151,11 @@ <h3>Terminology</h3>
...
-          ,
-          <dfn data-lt="values">
-            <a href="https://www.w3.org/TR/shacl/#dfn-value">value</a>
-          </dfn>
-          ,
-          <dfn data-lt="value nodes">
-            <a href="https://www.w3.org/TR/shacl/#dfn-value-nodes">value node</a>
-          </dfn>
-          .
+          The following definitions from other specications are used in this document: @@
```
```suggestion
          The following definitions from other specifications are used in this document: @@
```

shacl12-rules/index.html
```diff
@@ -428,26 +243,58 @@ <h3>Document Conventions</h3>
     <section id="overview">
       <h2>Outline</h2>
 
-      <p>SHACL rules infer new triples. The input is a data graph and a shape
-      graph with rules, the output is a graph of inferred triples that do not
-      occur in the data graph.</p>
+      <p>
+        SHACL rules infer new triples. The input is a data graph
+        and a set of rules. The output is a graph of inferred triples that do not
```
simonstey, line 257: so what if a rule infers triples that are already in the data graph?

robert-david:
> @robert-david -- you have directly changed this branch, rather than either making a reviewable suggestion via the GH interface, or by making a separate PR to this branch.
> 
> The direct commit invalidated work I had on my copy and merge/rebase lost that content. I have taken you commit and made the changes manually, then restored the lost content, and made a step towards a cleaner commit to set the initial content for evaluation.

@afs apologies for not following the right procedure. It wasn't clear to me. I didn't find the option to suggest changes via review, but will check again and use this in the future.

shacl12-rules/index.html
```diff
@@ -428,26 +243,58 @@ <h3>Document Conventions</h3>
...
+RULE { ?a :ancestorOf ?b } WHERE { ?a :ancestorOf ?c . ?c :ancestorOf ?b }
+</pre>
+      </section>
+
+      <section id="outline-negation">
+        <h3>Negation</h3>
+<pre class="example-rules">
+# Default value - calculate a name
+RULE { ?x :name ?FN } WHERE { 
+    ?x rdf:type :Person
```
simonstey, line 299:
```suggestion
    ?x rdf:type :Person .
```

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
 
         <p>The Shape Rules Abstract Syntax</p>
 
-        <p class="ednote">?? Abstract Rules Syntax</p>
-
         <dl>
-          <dt><dfn data-lt="variable">variable</dfn></dt>
-          <dd><code>[ sh:var "name" ]</code>
+          <dt><dfn data-cite="SPARQL12-QUERY#defn_QueryVariable">variable</dfn></dt>
```
simonstey, line 316: this cites SPARQL12, intended?

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            Variables are also used in <a>expressions</a>.
           </dd>
 
-          <dt><dfn data-lt="expression">node expression</dfn></dt>
+          <dt><dfn>expression</dfn></dt>
           <dd>
-            <em>Link to shacl12-node-expr#?</em> -- needs to be single value, list arguments.
-            Often called an <em>expression</em>.
+            <p>
+              An [=expression=] is a function, or functional form.
```
simonstey, line 325: should we define what a function and functional form is?

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+              It's arguments are [=RDF terms=].
+              An expression is evaluated with respect to a [=solution mapping=] to give
+              an [=RDF term=] as the result.
```
simonstey, line 328: to give an RDF term as the result to what?

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+              <a>Expressions</a> are compatible with
+              <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
+                 >SHACL list parameter functions</a>
```
simonstey, line 331: only list ones?

robert-david:
> A small observation: In logic programming, datalog, etc, rules typically assume safe negation meaning that every variable appearing under a negated (NOT) part must already be bound by a positive predicate in the rule. In this example, this condition is satisfied since concat provides the necessary bindings. However implementing such a syntactic check in the general case may not be trivial because it is not always clear what other kinds of variable dependencies it may exist in the rule. Best Ognjen

@osavkovic why is this not trivial? I understand that there needs to be some 'positive' triple pattern in the body to bind it. So it is a sort of dependency tree, but it is limited to the body of a rule (only). Do i miss something here? Can you please give a non-trivial example.
Also, we need to check this in a similar way for the rule heads regarding the body.

robert-david:
> A potential friction point: Current SPARQL muscle-memory may lead many users to type `NOT EXISTS` instead of `NOT`.

NOT is common for logic programming, so i think it is a good choice for a name.
+1 for the differentiation to SPARQL's NOT EXIST.

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            A [=triple template=] is 3-tuple where each element is either
             a variable or an RDF term (which might be a triple term).
             [=Triple templates=] appear in the [=head=] of a [=rule=].
           </dd>
 
           <dt><dfn>triple pattern</dfn></dt>
           <dd>
-            A <em>triple pattern</em> is 3-tuple where each element is either a
+            A [=triple pattern=] is 3-tuple where each element is either a
             variable, or an RDF term  (which might be a triple term).
```
simonstey: so this could also be (tt,tt,tt) right?

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-            A <em>data block</em> is a set of triples.
+            A [=data block=] is a set of triples.
             These form extra facts that are included in the inference process.
           </dd>
 
           <dt><dfn>triple template</dfn></dt>
           <dd>
-            A <em>triple template</em> is 3-tuple where each element is either
+            A [=triple template=] is 3-tuple where each element is either
             a variable or an RDF term (which might be a triple term).
```
simonstey: should we define what a triple term is?

afs:
> I didn't find the option to suggest changes via review,

It's available on the "file changes" tab.

liviorobaldo:
> Also, we need to check this in a similar way for the rule heads regarding the body.

@robert-david, what do you mean here? According to the current version of the [3.1 Well-formedness Conditions](https://w3c.github.io/data-shapes/shacl12-rules/#wellformed), as far as I understand, all variables in the head must be bound in the body, possibly via an assignment. Now that NOT has been introduced, Section 3.1 should be updated to clarify that head variables may only be bound to variables occurring in positive triple patterns (or assignments) in the body. This is because variables appearing exclusively inside NOT are intended to quantify over individuals assumed _not_ to exist, and therefore cannot provide bindings for variables in the head.

shacl12-rules/shacl-rules.ttl
liviorobaldo: Can someone explain me what the prefix "shnex" refers to? Thanks in advance.

shacl12-rules/index.html
```diff
@@ -456,78 +303,106 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-            a [=rule body=] (often just "body").
+            An [=aggregation element=] is ..@@.. .
+             [=Aggregation elements=] appear in the [=body=] of a [=rule=].
           </dd>
 
-          <dt><dfn data-lt="head|rule head">rule head</dfn></dt>
+          <dt><dfn data-lt="rule body element|rule element|element">rule body element</dfn></dt>
           <dd>
-            A <em>rule head</em> is a sequence where each element
+            A [=rule body element=] (often just "rule element") is one of elements that
```
```suggestion
            A [=rule body element=] (often just "rule element") is one of the elements that 
```

shacl12-rules/index.html
```diff
@@ -619,59 +544,87 @@ <h2>Concrete Syntax forms for Shapes Rules</h2>
...
+        [ shrl:subject [ shrl:var "x" ] ; shrl:predicate :q ; shrl:object [ shrl:var "v2" ] ]
+        [ shrl:expr [ sparql:function-or (
+              [ sparql:equals ( [ shrl:var "v1" ] 0 ) ]
+              [ sparql:equals ( [ shrl:var "v2" ] 0 ) ]
             ) ]
         ]
       );
     ]
   ) .
 </pre>
```
simonstey, line 604: do we want to use `shnex-sparql` instead `sparql`?

shacl12-rules/index.html
```diff
@@ -619,59 +544,87 @@ <h2>Concrete Syntax forms for Shapes Rules</h2>
 RULE { ?x :oneIsZero true . }
 WHERE { ?x :p ?v1 ;  :q ?v2  FILTER ( ( ?v1 = 0 ) || ( ?v2 = 0 ) )  }
 </pre>
+      </div>
 
-<p>RDF:</p>
+      <p>RDF Rules syntax:</p>
 
 <pre class="example-rules">
 PREFIX :       &lt;http://example/&gt;
```
simonstey, line 563: above we define:

`ex: http://example.com/ns#`

should we align that?

robert-david:
> > Also, we need to check this in a similar way for the rule heads regarding the body.
> 
> @robert-david, what do you mean here? According to the current version of the [3.1 Well-formedness Conditions](https://w3c.github.io/data-shapes/shacl12-rules/#wellformed), as far as I understand, all variables in the head must be bound in the body, possibly via an assignment. Now that NOT has been introduced, Section 3.1 should be updated to clarify that head variables may only be bound to variables occurring in positive triple patterns (or assignments) in the body. This is because variables appearing exclusively inside NOT are intended to quantify over individuals assumed _not_ to exist, and therefore cannot provide bindings for variables in the head.

To make it more precise: the variables in the rule head triple patterns need to be bound in some positive (not negated) body element. If they are also bound (additionally) in a negated body element, this is still ok. If they are only bound in a negated body element, it is not ok, i.e., not a well-formed rule.

shacl12-rules/index.html
```diff
@@ -700,163 +653,292 @@ <h3>RDF Rules Syntax</h3>
...
+          <dd>
+            A [=binding=] is a pair 
+            (<a>variable</a>, [=RDF term=]), consistent with the 
+            <a data-cite="sparql12-query#defn_QueryVariable">definition</a> used in [[!sparql12-query]].
+          </dd>
+
+          <dt><dfn data-lt="solution|sparql12-query#defn_sparqlSolutionMapping">solution mapping</dfn></dt>
+          <dd> 
+            A <a data-lt="sparql12-query#defn_sparqlSolutionMapping">solution mapping</a>, 
+            <var>μ</var>, is a partial function <code><var>μ</var> : V → T</code>,
```
```suggestion
            <var>μ</var>, is a partial function <code><var>μ</var> : <var>V</var> → <var>T</var></code>, 
```

robert-david (approved):

shacl12-rules/index.html
afs: Link to RDF Concepts added.

shacl12-rules/index.html
afs: The predicate has to be URI (by syntax). It might be better to put that in the following paragraph about subject/prfidctae/object to avoid a contorted definition. "which might" supposed to be promising it.

Background:
SPARQL is "symmetric  RDF".
https://www.w3.org/TR/sparql12-query/#sparqlTriplePatterns
https://www.w3.org/TR/rdf12-concepts/#section-symmetric-rdf

Before anyone asks - yes, literals-as-subjects - these occur in RDFS (`"abc" rdf:type xsd:string`).
In RDF data they simply don't match.
Ditto triple terms.

shacl12-rules/index.html
afs: Changed to "is any element that can appear in a [=rule body=];"

shacl12-rules/index.html
afs: > do we want to use `shnex-sparql` instead `sparql`?

What would the difference/advantage be?

SPARQL goes further than node expressions, e.g. error handling, and node expressions can't be used for in triple patterns so thats the `shrl:var` (better as `shrl:varName`).

Also - [raw/data-shapes/shacl12-node-expr](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-node-expr/index.html#sparql-functions) uses `sparql:` :-)

shacl12-rules/index.html
afs: Aligned as `http://example.com/`.

shacl12-rules/index.html
afs: They do not appear in the output graph. This is for the sake of definition and also for provenance.
(alternatives like putting directly into the data graph are "optimizations".)

If a rule infers several triples, some new triples and some old triples, the new ones will be in the output.

(Loosely, multiple triples in the head behave like having instead one rule for each triple in the head and the same body.)

At each rule evaluation, the graph used for matching is "base+inferred triples from last round".

shacl12-rules/index.html
afs: Omitting the final DOT before a keyword is legal.

shacl12-rules/index.html
afs: "expression evaluation"

shacl12-rules/index.html
afs: At the moment, yes.  This includes all the SPARQL functions in the Rule Syntax.

We would have to define what returning a list means and what  list of some failures and some result means.

Many of the other forms access the graph - that would require a stability guarantee (the same as negation - must be dependent on only strictly lower stratification layers).

`List Operator Expressions` need a list which does not make sense in many places so composition of node expression needs refining for general use.

shacl12-rules/index.html
afs: Yes. SPARQL variables can appear in places other than expressions (triple patters, triple templates) - they not "var expressions" as currently defined.

shacl12-rules/index.html
afs: That is supposed to be done by referring to
`<a data-cite="sparql12-query#expressions">SPARQL expressions</a>`

so it can include error handling for `FILTER` and be written with `?x`.

shacl12-rules/index.html
afs: PS added wording to say the 2nd element of the tuple is an IRI or a variable.

shacl12-rules/index.html
simonstey: haha you are right.. <sup>I wonder who the editor was that worked on the sparql functions part</sup>

shacl12-rules/index.html
simonstey: I just found 

`SHACL rules infer new triples. ` 

to be a bit confusing without: 

`If a rule infers several triples, some new triples and some old triples, the new ones will be in the output.` 

can we add this sentence?

shacl12-rules/index.html
simonstey: nvm, I confused `List Parameter Functions` with `Named parameter functions` that have a list as their parameter:

> Note that some [named parameter functions](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-node-expr/index.html#dfn-named-parameter-function) — such as shnex:IntersectionExpression — also use a [SHACL list](https://www.w3.org/TR/shacl12-core/#dfn-shacl-list) as an object of the [key parameter](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-node-expr/index.html#dfn-key-parameter), similar to [list parameter functions](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-node-expr/index.html#dfn-list-parameter-function) which always have a [SHACL list](https://www.w3.org/TR/shacl12-core/#dfn-shacl-list) as the object of their [list parameter property](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-node-expr/index.html#dfn-list-parameter-property). However, these may produce more than one [output nodes](https://www.w3.org/TR/shacl12-core/#dfn-output-nodes), and also accept lists as input nodes.

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
     </section>
 
     <section id="shapes-rules-grammar">
       <h2>Shapes Rules Language Grammar</h2>
 
-        <div class="grammarTable">
-          <table><tbody>
+      <!-- GRAMMAR -->
+      <div class="grammarTable">
+        <table><tbody>
```
simonstey, line 955: @afs did you manually revise/update the grammar table or do you have a script or similar that created it for you?

shacl12-rules/index.html
```diff
@@ -867,942 +951,1014 @@ <h2>Shapes Rules Language Grammar</h2>
...
                 <td><code class="gRuleBody"><a href="#rPathPrimary">PathPrimary</a></code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[50]&nbsp;&nbsp;</code></td>
+                <td><code>[51]&nbsp;&nbsp;</code></td>
                 <td><code><span class="doc-ref" id="rPathPrimary">PathPrimary</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
                 <td><code class="gRuleBody"><a href="#riri">iri</a> | <span class="token">'a'</span> | <span class="token">'('</span> <a href="#rPath">Path</a> <span class="token">')'</span> </code></td>
               </tr>
```
simonstey, line 1312: this does only cover sequence `/` and inverse `^` directly.  but not others like `|, +, *, or !`. 

Although they could be expressed/dealt with differently:

Missing Construct | Logic Type | Replacement Mechanism in SHACL Rules
-- | -- | --
Alternative (\|) | Disjunction | Multiple Rules
One-or-more (+) | Recursion | Recursive Rules (Base case + Step)
Zero-or-more (*) | Recursion + Identity | Recursive Rules + BIND (Identity)
Negated Set (!) | Negation | Wildcard Predicate (?p) + FILTER

So was this restriction in the grammar deliberate to move complexity out of the property path string and **into the rule structure** itself?

shacl12-rules/index.html
simonstey, line 962: a file containing only prefixes would be invalid in this case, which is different from e.g. Turtle's Grammar (https://www.w3.org/TR/rdf12-turtle/#sec-grammar-grammar)

```
[1] turtleDoc  ::= statement*
[2] statement  ::= directive | (triples '.')
```

it's, however, aligned with [SPARQL](https://www.w3.org/TR/sparql12-query/#sparqlGrammar):

```
[1] QueryUnit  ::=  Query
[2] Query  ::=  Prologue( SelectQuery | ConstructQuery | DescribeQuery | AskQuery ) ValuesClause
```

shacl12-rules/index.html
```diff
@@ -867,942 +951,1014 @@ <h2>Shapes Rules Language Grammar</h2>
...
                 <td><code class="gRuleBody"><a href="#rNIL">NIL</a> | <span class="token">'('</span> <a href="#rExpression">Expression</a> ( <span class="token">','</span> <a href="#rExpression">Expression</a> )* <span class="token">')'</span> </code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[33]&nbsp;&nbsp;</code></td>
+                <td><code>[34]&nbsp;&nbsp;</code></td>
                 <td><code><span class="doc-ref" id="rTriplesSameSubject">TriplesSameSubject</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
-                <td><code class="gRuleBody"><a href="#rVarOrTerm">VarOrTerm</a> <a href="#rPropertyListNotEmpty">PropertyListNotEmpty</a> | <a href="#rTriplesNode">TriplesNode</a> <a href="#rPropertyList">PropertyList</a> | <a href="#rReifiedTripleBlock">ReifiedTripleBlock</a></code></td>
+                <td><code class="gRuleBody"><a href="#rVarOrTerm">VarOrTerm</a> <a href="#rPropertyListNotEmpty">PropertyListNotEmpty</a> |	<a href="#rTriplesNode">TriplesNode</a> <a href="#rPropertyList">PropertyList</a> |	<a href="#rReifiedTripleBlock">ReifiedTripleBlock</a></code></td>
```
simonstey, line 1192: So I guess with the `RDF Rules` serialization of `SHACL Rules`, we would take care of producing standard (not generalized!) RDF, right?

shacl12-rules/index.html
```diff
@@ -619,59 +546,87 @@ <h2>Concrete Syntax forms for Shapes Rules</h2>
...
+        <p>The <a href="#shapes-rules-grammar">grammar</a> is given below.</p>
+        <p class="ednote">Mapping the AST to the abstract syntax.</p>
+
+        <section id="shape-rules-abbrev">
+          <h4>Shape Rules Language Abbreviations</h4>
+
+          <p>Additional helpers (short-hand abbreviations):</p>
+
+          <p class=note>
+            These allow for well-known rul epatterns and also specialised implementations in basic engines
```
```suggestion
            These allow for well-known rule patterns and also specialised implementations in basic engines
```

shacl12-rules/index.html
simonstey, line 247:
```suggestion
        <p class="ednote">RFC 2119 language should automatically be inserted here.</p>
```

shacl12-rules/index.html
```diff
@@ -428,26 +243,58 @@ <h3>Document Conventions</h3>
...
+        <div class="rules">
 <pre>
 RULE { ?x :childOf ?y } WHERE { ?y :fatherOf ?x }
 RULE { ?x :childOf ?y } WHERE { ?y :motherOf ?x }
 
 RULE { ?x :descendedFrom ?y } WHERE { ?x :childOf ?y }
 RULE { ?x :descendedFrom ?y } WHERE { ?x :childOf ?z . ?z :childOf ?y }
 </pre>
+        </div>
+        <p>The above rules, applied ot the data, will conclude
```
```suggestion
        <p>The above rules, applied to the data, will conclude 
```

shacl12-rules/index.html
```diff
@@ -456,78 +303,108 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            An [=aggregation element=] is ..@@.. .
+             [=Aggregation elements=] appear in the [=body=] of a [=rule=].
           </dd>
 
-          <dt><dfn data-lt="head|rule head">rule head</dfn></dt>
+          <dt><dfn data-lt="rule body element|rule element|element">rule body element</dfn></dt>
           <dd>
-            A <em>rule head</em> is a sequence where each element
+            A [=rule body element=] (often just "rule element") is any element that
+            can appear in a [=rule body=];</dd>
```
```suggestion
            can appear in a [=rule body=];
```

shacl12-rules/index.html
```diff
@@ -456,78 +303,108 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-            Often called an <em>expression</em>.
+            <p>
+              An [=expression=] is a function, or functional form.
+              It's arguments are [=RDF terms=].
+              An expression is evaluated with respect to a [=solution mapping=] to give
+              an [=RDF term=] as the result.
+              <a>Expressions</a> are compatible with
+              <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
+                 >SHACL list parameter functions</a>
+              and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.
```
```suggestion
              and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
```

shacl12-rules/index.html
```diff
@@ -539,6 +416,52 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            Stratification is the process of labelling <a>rules</a> based on their
+            <a>dependency</a> relationships in a <a>rule set</a>.
+            The set of rules with the same label are called a <a>layer</a>.
+          </p>
+          
+          <dl>
+            <dt><dfn data-lt="stratification layer|layer">stratification layer</dfn></dt>
+            <dd>
+              A stratification layer, usually, just "layer", is the set of rules 
+              with same straification labelling.
```
```suggestion
              with same stratification labelling.
```

shacl12-rules/index.html
```diff
@@ -456,78 +303,108 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            A [=rule head=] is a sequence where each element
             of the sequence is a [=triple template=].
           </dd>
 
           <dt>
-            <dfn data-lt="body|rule body">rule body</dfn>
+            <dfn data-lt="rule body|body">rule body</dfn>
           </dt>
+          <dd> 
+            A [=rule body=] is a sequence of [=rule body elements=].</dfn>
```
```suggestion
            A [=rule body=] is a sequence of [=rule body elements=].
```

shacl12-rules/index.html
```diff
@@ -593,22 +522,20 @@ <h3>Well-formedness Conditions</h3>
 
     <section id="concrete-syntax">
       <h2>Concrete Syntax forms for Shapes Rules</h2>
+      <P>There are two concrete syntaxes.</p>
```
```suggestion
      <p>There are two concrete syntaxes.</p>
```

shacl12-rules/index.html
```diff
@@ -619,59 +546,87 @@ <h2>Concrete Syntax forms for Shapes Rules</h2>
...
-  sh:ruleSet (
+  shrl:ruleSet (
     [
-      rdf:type sh:Rule;
-      sh:head (
-        [ sh:subject [ sh:var "x" ] ; sh:predicate :bothPositive; sh:object true ]
+      rdf:type shrl:Rule;
+      shrl:head (
+        [ shrl:subject [ shrl:var "x" ] ; shrl:predicate :bothPositive; shrl:object true ]
       )
```
```suggestion
      ) ;
```

shacl12-rules/index.html
```diff
@@ -619,59 +546,87 @@ <h2>Concrete Syntax forms for Shapes Rules</h2>
...
       );
     ]
     [
-      rdf:type sh:Rule;
-      sh:head (
-        [ sh:subject [ sh:var "x" ] ; sh:predicate :oneIsZero ; sh:object true ]
+      rdf:type shrl:Rule;
+      shrl:head (
+        [ shrl:subject [ shrl:var "x" ] ; shrl:predicate :oneIsZero ; shrl:object true ]
       )
```
```suggestion
      ) ;
```

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
-    T = T1
-    endif
+for each rule element rElt in B
+    let S1 = { }
+
+    if rElt is a triple pattern TP:
+        X = graphMatch(G, TP)
+        SEQ1 = { <var>μ</var><sub>0</sub> }
+        for each <var>μ</var><sub>1</sub> in X:
+            for each <var>μ</var><sub>2</sub> in T:
```
simonstey: `T` was removed, this should be `SEQ` I guess?

```suggestion
            for each <var>μ</var><sub>2</sub> in SEQ:
```

T was removed

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+   
     endfor
 
 # Evaluate rule head
 let H = empty set
-for each R in T:
-    Let S = set of triples obtained by replacing variables
-            in the triple templates of the head with values from R
-    define [head(R)/row]
+for each <var>μ</var> in T:
```
simonstey: same here, T was removed

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+        Let triple = subst(<var>μ</var>, TT)
+        Add triple to S
     H = H union S
     endfor
 
 result eval(R, G) is H
 
 </pre>
-        <p>Note that `H` may contain triples that are also in the data grapoh.</p>
+      <p>Note that `H` may contain triples that are also in the data grapoh.</p>
```
```suggestion
      <p>Note that `H` may contain triples that are also in the data graph.</p>
```

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+              <pre>
+merge(S1, S2) = { <var>μ</var> |
+                    <var>μ</var><sub>1</sub> in S1, <var>μ</var><sub>2</sub> in S2
+                    and compatible(<var>μ</var><sub>1</sub>, <var>μ</var><sub>2</sub>)
+                        μ(v) = μ1(v) if v in dom(μ1)
+                        μ(v) = μ2(v) if v in dom(μ2) }</pre>
+            </div>
+            <div class="ednote">
+              <p>
+              Say the domain is `dom(S1) &cups; dom(S2)`.
```
simonstey, line 796: `&cups;` is technically not correct, it should be `&cup;` https://toolstud.io/web/charmap.php?encoded=cup

(works anyway though)

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+for each rule element rElt in B
+    let S1 = { }
+
+    if rElt is a triple pattern TP:
+        X = graphMatch(G, TP)
+        SEQ1 = { <var>μ</var><sub>0</sub> }
+        for each <var>μ</var><sub>1</sub> in X:
+            for each <var>μ</var><sub>2</sub> in T:
+              if compatible(<var>μ</var><sub>1</sub>, <var>μ</var><sub>2</sub>)
+                <var>μ</var><sub>3</sub> = merge(<var>μ</var>1, <var>μ</var><sub>2</sub>)
```
simonstey, line 869: Merge function uses merge(μ1, μ2) but earlier defined merge(S1, S2) for solution sequences:

```
solution merge
    If two solutions are compatible, the merge of two solutions is the solution that maps variables of each solution to the [RDF term](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-term) from one or other of the solutions.
    Let S1 and S2 be solution sequences.

    merge(S1, S2) = { μ |
                        μ1 in S1, μ2 in S2
                        and compatible(μ1, μ2)
                            μ(v) = μ1(v) if v in dom(μ1)
                            μ(v) = μ2(v) if v in dom(μ2) }
```

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+Let L be the sequence of layers after stratification
+
+for each layer in L
+    Let finished = false
+    while !finished:
+        finished = true
+        foreach rule in RS
+            let GI = G0 &cup; GI
+            let X = eval(rule, GI)
+            let Y = those triples in X that are not in G0
```
```suggestion
            let Y = those triples in X that are not in GI
```

this should be GI, otherwise this never terminates

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
-
-result is GI
+Apply stratification to RS
+Let L be the sequence of layers after stratification
+
+for each layer in L
+    Let finished = false
+    while !finished:
+        finished = true
+        foreach rule in RS
```
simonstey: The inner loop iterates over the entire Rule Set.. shouldnt It strictly iterate over rules belonging to the current layer only?

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
-          add row1 to T1
-          endfor
-    T = T1
-    endif
+for each rule element rElt in B
+    let S1 = { }
+
+    if rElt is a triple pattern TP:
+        X = graphMatch(G, TP)
+        SEQ1 = { <var>μ</var><sub>0</sub> }
```
simonstey: isn't SEQ1 the destination bucket for the results of the join between graph match and SEQ? 

what happens if the loops fail to find any matches? wouldnt the algo then pass the identity mapping forward instead of the empty set?

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
 
     if rElt is an assignment(V, expr)
-        T1 = empty list
-        for each row in T:
-            let x = eval(expr, row)
-            let R1 = R union (V, x)
-            add R1 to T1
+        SEQ1 = {}
+        for each solution S in SEQ:
+            let x = eval(expr, <var>μ</var>)
```
simonstey: `μ` is undefined here, shouldnt it be S?
```suggestion
            let x = eval(expr, S)
```

shacl12-rules/index.html
simonstey: (it was also GI previously)

shacl12-rules/index.html
simonstey:
```suggestion
        foreach rule in layer
```

shacl12-rules/index.html
```diff
@@ -700,163 +655,292 @@ <h3>RDF Rules Syntax</h3>
...
+            let X = eval(rule, GI)
+            let Y = those triples in X that are not in G0
+            if Y is not empty:
+                finished = false
+                endif
+            let GI = Y &cup; GI
+            endfor
+        endwhile
+    endfor
+the result is GI
```
```suggestion
for each layer in L:
    let finished = false
    while !finished:
        finished = true
        for each rule in layer: # only iterate rules in the current layer  
            let X = eval(rule, GI) # evaluate against current accumul.
            let Y = { t | t in X AND t not in GI } # check against GI not G0 (base)       
            if Y is not empty:
                finished = false
                GI = Y &cup; GI
           
return GI
```

shacl12-rules/index.html
afs: ?? I see that text already in the PR document.

shacl12-rules/index.html
afs: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references

`&cup;` is HTML4, `&cups;` is HTML5.

shacl12-rules/index.html
afs: > The inner loop iterates over the entire Rule Set.. shouldn't It strictly iterate over rules belonging to the current layer only?

Yes. 
Stratification work has only just started.

This PR gets definitions in - it isn't the final state (of anything). I want to rewrite the algorithms, especially the rule one, to pull out the different elements and define eval of each element separately, then use those definitions in a more concise rule evaluation.

Working long term on a PR is undesirable. It's already stressing the github UI!

https://github.com/w3c/data-shapes/pull/659#issuecomment-3575287869
> Later today or tomorrow, I will update the grammar with the negation feature.
> 
> Then I think it is adequate (barely) to be merged, mistakes and all.
> 
> Merge does not mean the content can not be changed; it means further PRs can made.

shacl12-rules/index.html
afs: > can we add this sentence?

Not while there are review changes outstanding because it causes merge conflicts!

Made into an issue #678 - although it may well become evident as the document is completed.

I really did mean:
> > Very early work to show direction.

shacl12-rules/index.html
afs: The grammar, between the two `<!-- GRAMMAR -->`, is generated by a script (open source)
The grammar is too big to manage manually. Keeping the links right would be impossible.

The primary version is a javacc file (open source), which is then use to make HTML. This means it is checked e.g. no dangling productions. The grammar is LL(1) and LALR(1) - i.e. simple technology, many parser generators.

There will be a plain text BNF file as well for implementers to use as a starting point for their own parsers, and not have to work from the HTML.

The HTML and BNF are the normative forms.

The javacc also generates a java parser, which means test cases for syntax can be run against the input to making HTML and BNF.

(This has been the process for all versions of the SPARQL grammar. Even v1.0 was impractical to manage without tooling.)

shacl12-rules/index.html
afs: It's deliberate. Sequence and inverse are syntactic sugar for a triple pattern.

The others introduce path operators. I could not see that they will be analysable for rule dependency calculations, and even if they are, whether they need well-formness constraints.

Negated set in particular looks in conflict with any datalog negation because the set of predicated is not bounded.

As you note, there are rules that have similar effects.

So - an area to consider later.

Note also: `TRANSITIVE` - a short form of writing transitive properties. Clarity for the rule users; easier for an engine to provide a custom implementation. If simple engines may not implement full rules, it gives one case of positive recursion without needing full support. Also `SYMMETRIC` and `INVERSE`.

shacl12-rules/index.html
afs: Right, in both the RDF syntax is not symmetric RDF (or generalized RDF) and the output inferred graph is a an RDF graph, so RDF triples only.

The RDF syntax does not have triple patterns in it, only written out as `[ :subject S ; :predciate P ; :object O ]`.

For inferring, same as `CONSTRUCT` - non-triples get dropped.

A consequence is that writing a rule with `"123" rdf:type xsd:integer` does not match the base or the infer graph. The latter may be a problem.

shacl12-rules/index.html
afs: Captured as issue #679.

It is legal in SPARQL-update

It can't be in SPARQL Query because a query needs an output, one of the 4 kinds of query.

afs:
> > Very early work to show direction.
> 
> Hi Andy, could you say a bit more about what kind of input you expect here? For instance, should I provide comments and edits, and should I make them in the document or here?
> 
> Thanks, O.

We are now at - or exceeding - the usability of the github review UI . Moving around the PR is slow; the new UI can't copy with the large file; accepting suggestions resets the display, hiding some input.

I think it is better for reviewers and for processing review comments to merge this, and then create PRs and raise issues on the document on `gh-pages`.

shacl12-rules/index.html
afs: For now, I've added a definition for `merge(μ1, μ2)`.

shacl12-rules/index.html

shacl12-rules/index.html
simonstey: cool! just wanted to know, whether I should check it for any typos or similar (which shouldnt be an issue when generated by tooling :) )

shacl12-rules/index.html
simonstey: later on we use &cup; though.. should we change those to cups then?

simonstey: `&cup;`

afs: Done

shacl12-rules/index.html
afs: Fix: `SEQ1 = { }` (It's SEQ that is μ0 initially)

`RULE { :s :p :o } WHERE {}` generates `:s :p :o`.
`DATA` is syntactic sugar.

shacl12-rules/index.html
afs: I've taken this suggestion, and others in this area, and reworked the block.

We'll be back to this algorithm later.

shacl12-rules/index.html
afs: A human chose the production names :-)

shacl12-rules/index.html
```diff
@@ -867,942 +954,1014 @@ <h2>Shapes Rules Language Grammar</h2>
...
               </tr>
 
               <tr style="vertical-align: baseline">
                 <td><code>[20]&nbsp;&nbsp;</code></td>
-                <td><code><span class="doc-ref" id="rTriplesBlock">TriplesBlock</span></code></td>
+                <td><code><span class="doc-ref" id="rBodyBasic">BodyBasic</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
-                <td><code class="gRuleBody"><a href="#rTriplesSameSubjectPath">TriplesSameSubjectPath</a> ( <span class="token">'.'</span> <a href="#rTriplesBlock">TriplesBlock</a>? )?</code></td>
+                <td><code class="gRuleBody"><a href="#rBodyTriplesBlock">BodyTriplesBlock</a>? ( <a href="#rFilter">Filter</a> <a href="#rBodyTriplesBlock">BodyTriplesBlock</a>? )*</code></td>
               </tr>
```
simonstey, line 1095:
```
[19] Negation ::= 'NOT' '{' BodyBasic '}' 
[20] BodyBasic ::= BodyTriplesBlock? ( Filter BodyTriplesBlock? )*
```

means we do not allow for BIND (or nested NOTs) . bug or feature?

liviorobaldo: Good catch @simonstey! Without nested `NOT`s, we cannot represent rules such as:

```∀x​[ (A(x)^∀x​[C(x)⇒D(x)]) ⇒ B(x)]```

i.e., rules containing universals in the antecedent. For example:

``IF { ?p a :Program . NOT{ ?l :in ?p . NOT{ ?l a :correct }} } THEN { ?p a :correct }``

This means: "every program `?p` whose lines are all correct is correct". Note that this also entails that programs with _no lines_ (i.e., empty programs) are correct, which makes sense (an empty program is, of course, correct).

The nested condition 

```NOT { ?l :in ?p . NOT { ?l a :correct } }```

reads as: "there is no line `?l` in program `?p` that is _not_ correct". This is how we represent the nested universal:

```∀x​[C(x)⇒D(x)]```

using negation when the `∀` symbol is not available.

I have worked on some use cases where this type of condition is actually necessary, for instance in the Time Ontology (see this paper, which defines its original semantics: https://www.isi.edu/people-hobbs/wp-content/uploads/sites/58/2023/11/hobbs-pan-TALIP04-1.pdf).

Therefore, I suggest we allow nested `NOT`s.

On the other hand, I think we should not allow `BIND`s nested inside `NOT`s. I cannot see any practical use for them. But if anyone can provide a good example, I’m happy to reconsider.

afs: Feature. 

There is an issue specifically about negation #674 so we can merge this PR as a partial draft and keep developing it.

Not having features that are outside "datalog" nested inside `NOT` simplifies the design (AKA simpler getting it right, simpler to get to a complete system we can add features to).

Adding features needs to be proven with use cases to show general applicability, and also a detailed proposal as to how they would work so we can investigate the consequences.

`BIND`::
The RHS of pattern does not change the body solution. `NOT` is a specific yes/no condition with no new variable binding [^1].  So even if allowed, the assignment would not contribute to the overall body solution which I think is confusing to allow it - it is a mild lost of convenience not being able to calculate a temporary value.

If (that's a big _if_) we have named tuples as evaluation workspace, the `NOT` body can be a tuple and then all features are available. That's a more useful feature.

`NOT` in `NOT`::
Double negation! What would that mean in this context? 

I can imagine (=intuition, =guess) if the inner one was dependent on layers beneath the layer-1 dependency of the outer `NOT`making it layer-2 then something might work but that's complicated.  Maybe it's better to say "run two rulesets, one after the other". And again - "named tuples".

[^1]: See Ognjen's comment above https://github.com/w3c/data-shapes/pull/659#issuecomment-3581922057

liviorobaldo: I did some online research. The standard Datalog solution to represent nested `NOT`s, such as the rule I showed above:

```IF { ?p a :Program . NOT { ?l :in ?p . NOT { ?l a :correct } } } THEN { ?p a :correct }```

is to introduce an extra predicate and write two rules:

```IF { ?l :in ?p . NOT { ?l a :correct } }  THEN at_least_one_line_not_correct(?p)```
```IF { ?p a :Program . NOT { at_least_one_line_not_correct(?p) } }  THEN { ?p a :correct }```

The semantics is the same.

I agree with Andy that this is simpler from the reasoner's point of view:

> Not having features that are outside "datalog" nested inside `NOT` simplifies the design (AKA simpler getting it right, simpler to get to a complete system we can add features to).

The drawback is that we need to introduce extra rules with auxiliary predicates, which are used only to implement the universal nested in the antecedent. For example, the predicate `at_least_one_line_not_correct` above is used only in the second rule. If a use case requires many universals in the antecedent, this could lead to verbose rule sets.

> Adding features needs to be supported by use cases to show general applicability, and also a detailed proposal for how they would work so we can investigate the consequences.

We would need to empirically measure how frequently rules with universals nested in the antecedent occur in existing use cases. If they occur often, we might need to introduce some kind of "shortcut", i.e., a compact way to represent them (but this should be an "additional feature").

simonstey (approved):

afs:
Added:
* rebase to `gh-pages`
* add new `shnex:` namespace
* placeholder for tests and implementation report

PR #668, opened 2025-11-25
#649: Fixed definitions to singular, added uniqueness

### LINKED ISSUE #649: Definitions and linking  of `binding` and `solution`
The definition of binding is  `id="#dfn_bindings"` and the singular does not link. This is needed for reference from other documents.
It should be `<dfn>binding</dfn>` and reSpec will also allow the plural within the document.

"solutions" has the same linking issue.  

The definition of "solution" is incomplete and not compatible with SPARQL.

It is defined as a set of pairs but does not say that the the variables of the pairs are unique.The current definition allows `(?x, 1)`, `(?x , 2)`.

[SPARQL section](w3.org/TR/sparql12-query/#defn_sparqlSolutionMapping) may help:
http://w3.org/TR/sparql12-query/#defn_sparqlSolutionMapping:

> A solution mapping, μ, is a partial function μ : V → T, where V is the set of all [variables](https://www.w3.org/TR/sparql12-query/#defn_QueryVariable) and T is the set of all [RDF terms](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-terms).

SHACL says:

"Variables are not required to be bound in a solution." so `(?x , _nothing_)` is legal.

HolgerKnublauch:
Closes #649

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-649/shacl12-sparql/index.html)

@afs here is a PR that hopefully addresses the uniqueness constraints (thanks for pointing that out). I am not sure if the language around unbound mappings is good enough now, and I would appreciate if you could add a suggestion in GitHub with a better wording.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -297,15 +297,7 @@
 					</div>
 				</div>
 				<div class="def">
-					<div class="term-def-header">Binding, Solution</div>
-					<div>
-						A <dfn data-lt="bindings">binding</dfn> is a pair (<a data-cite="sparql12-query/#defn_QueryVariable">variable</a>, <a>RDF term</a>), consistent with the term's use in [[!sparql12-query]].
-					    A <dfn data-lt="solutions">solution</dfn> is a set of bindings, informally often understood as one row in the body of the result table of a SPARQL query.
-					    Variables are not required to be bound in a solution.
-					</div>
-				</div>
-				<div class="def">
-				<div class="term-def-header">SHACL Subclass, SHACL superclass</div>
+					<div class="term-def-header">SHACL Subclass, SHACL superclass</div>
 					<div>
 						A <a>node</a> <code>Sub</code> in an <a>RDF graph</a> is a <dfn data-lt="SHACL subclass|subclass|subclasses">SHACL subclass</dfn> of another <a>node</a> <code>Super</code>
 						in the <a>graph</a> if there is a sequence of <a>triples</a> in the <a>graph</a> each with predicate <code>rdfs:subClassOf</code> such that the <a>subject</a> of the first <a>triple</a> is <code>Sub</code>,
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -247,7 +247,6 @@
 						A <dfn>binding</dfn> is a pair (<a data-cite="sparql12-query/#defn_QueryVariable">variable</a>, <a>RDF term</a>), consistent with the term's use in [[!sparql12-query]].
 					    A <dfn>solution</dfn> is a set of bindings where each variable must be unique.
 						Informally, a solution is often understood as one row in the body of the result table of a SPARQL query.
-					    Variables are not required to be bound in a solution.
 					</div>
 				</div>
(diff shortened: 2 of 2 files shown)


shacl12-sparql/index.html
```diff
@@ -244,8 +244,9 @@ <h3>Terminology</h3>
 				<div class="def">
 					<div class="term-def-header">Binding, Solution</div>
 					<div>
-						A <dfn data-lt="bindings">binding</dfn> is a pair (<a data-cite="sparql12-query/#defn_QueryVariable">variable</a>, <a>RDF term</a>), consistent with the term's use in [[!sparql12-query]].
-					    A <dfn data-lt="solutions">solution</dfn> is a set of bindings, informally often understood as one row in the body of the result table of a SPARQL query.
+						A <dfn>binding</dfn> is a pair (<a data-cite="sparql12-query/#defn_QueryVariable">variable</a>, <a>RDF term</a>), consistent with the term's use in [[!sparql12-query]].
+					    A <dfn>solution</dfn> is a set of bindings where each variable must be unique.
+						Informally, a solution is often understood as one row in the body of the result table of a SPARQL query.
 					    Variables are not required to be bound in a solution.
```
afs: Saying "not bound" is ambiguous - could be either the variable is absent from all pairs, or there is some `(var, ???)`.

If they are not bound they should be absent, otherwise it is not a set of pairs of (variable, term).


```suggestion
```

afs (approved):

afs:
"solution mapping" is used in core and should be a link to SPARQL12.

HolgerKnublauch:
> "solution mapping" is used in core and should be a link to SPARQL12.

Yes, new PR created for that https://github.com/w3c/data-shapes/pull/695

shacl12-core/index.html
```diff
@@ -297,15 +297,7 @@ <h3>Terminology</h3>
...
-					<div class="term-def-header">Binding, Solution</div>
-					<div>
-						A <dfn data-lt="bindings">binding</dfn> is a pair (<a data-cite="sparql12-query/#defn_QueryVariable">variable</a>, <a>RDF term</a>), consistent with the term's use in [[!sparql12-query]].
-					    A <dfn data-lt="solutions">solution</dfn> is a set of bindings, informally often understood as one row in the body of the result table of a SPARQL query.
-					    Variables are not required to be bound in a solution.
-					</div>
-				</div>
-				<div class="def">
-				<div class="term-def-header">SHACL Subclass, SHACL superclass</div>
+					<div class="term-def-header">SHACL Subclass, SHACL superclass</div>
```
TallTed, line 300: Why uppercase `S` on `Subclass`, and lowercase `s` on `superclass`? It seems like these should be the same, but I'm not sure which... I'm *guessing* lowercase, based on line 302.
```suggestion
					<div class="term-def-header">SHACL subclass, SHACL superclass</div>
```

HolgerKnublauch: Will be addressed in this PR because this one here is already closed. https://github.com/w3c/data-shapes/pull/705

Issue #496, opened 2025-08-11
Decide on all the SHACL Namespaces


nicholascar:
We have more than one SHACL namespace, but how many do we have and what are they? Please list all the proposed Namespaces here and discuss!

nicholascar:
Name | Prefix | Namespace
--- | --- | ---
SHACL | `sh` | `http://www.w3.org/ns/shacl#`
SHACL-SHACL | `sh-sh` | `http://www.w3.org/ns/shacl-shacl#`
SHACL Tests | `sht` | `http://www.w3.org/ns/shacl-test#`

afs:
Name | Prefix | Namespace
--- | --- | ---
SPARQL-RL | `srl` | `http://www.w3.org/ns/sparql-rl#`
SPARQL-RL Tests | `srt` | `http://www.w3.org/ns/sparql-rl#`

There will be a MIME type registration for the rules language, `applciation/sparql-rl`, with suggested file extension of `srl`.

The tests namespace would be small - test type to use within the existing http://www.w3.org/2001/sw/DataAccess/tests/test-manifest# vocabulary, A rules evaluation test three parts - rules in various syntax forms, input data, and a result graph which needs to be separate from the inputs.

HolgerKnublauch:
Name | Prefix | Namespace
--- | --- | ---
SHACL Node Expressions | `shnex` | `http://www.w3.org/ns/shnex#`

This is currently implemented in #490

robert-david:
> Name	Prefix	Namespace
> SHACL Rules	`srl`	`http://www.w3.org/ns/shacl-rules#`
> There will be a MIME type registration for the rules language, with suggested file extension `srl`.
> 
> (though I'm not so convinced about involving the `C`/"constraint" here.)

Shouldn't we follow the 'sh' scheme, i.e. 'shrl' for rules as well?

afs:
It is picking the file extension as prefix, which are often 3 characters (8.3 lives!).

"shrl" might be pronounced "shrill" :smile:

edmondchuc:
| Name                  | Prefix | Namespace                  |
|-----------------------|--------|----------------------------|
| SHACL User-Interfaces | shui   | http://www.w3.org/ns/shui# |

Will add this to the UI subgroup meeting agenda for this week to discuss.

afs:
There are two different URI styles.

/ns/shacl#
/ns/shacl-shacl#
/ns/shacl-test#
/ns/shacl-rules#
[...]

PR #681, opened 2025-12-01
#496: Replaced shnex and shui namespaces

HolgerKnublauch:
As discussed in https://github.com/w3c/data-shapes/issues/496#issuecomment-3586054793

This includes changes to shacl-ui, which I usually don't edit, thus the review requests.


edmondchuc (approved):
Looks good. No issues with the namespace change from my side.

afs (approved):

PR #692, opened 2025-12-06
Some SRL tests

afs:
This PR is no more than an illustrative starting point for the SHACL Rules tests for `syntax` and `eval`.

The names and organisation of tests is subject to change.

There are likely to be other categories of test.

Eventually, syntax tests will have two representations, SRL and RDF.

### DIFF
--- a/shacl12-test-suite/tests/rules/eval/data-01.ttl
+++ b/shacl12-test-suite/tests/rules/eval/data-01.ttl
@@ -1,0 +1,3 @@
+PREFIX : <http://example/>
+
+:s :p :o .
--- a/shacl12-test-suite/tests/rules/eval/data-02.ttl
+++ b/shacl12-test-suite/tests/rules/eval/data-02.ttl
@@ -1,0 +1,4 @@
+PREFIX : <http://example/>
+
+:s :p :z .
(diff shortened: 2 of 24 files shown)


robert-david (approved):

simonstey:
fwiw, I created a bunch of SRL rule examples here as well https://github.com/simonstey/py-srl/tree/05149ab3d97147a40a26fca8352e77966d8f5cd2/examples

simonstey (approved):
looks good!

recalcitrantsupplant:
Looks good - I've added them in two libraries here: https://shacl-rules.app.dev.kurrawong.ai/

Would it make sense to have the data graph input in the compact syntax to go with the compact srl format? - I realise they are equivalent.

recalcitrantsupplant (approved):

afs:
>> Eventually, syntax tests will have two representations, SRL and RDF.

> Would it make sense to have the data graph input in the compact syntax to go with the compact srl format? - I realise they are equivalent.

One step at a time!

It is more than just having two forms. Ideally, there should be tests of 3 kinds: "SRL syntax", "SRL to RDF", "RDF to SRL".

PR #694, opened 2025-12-08
Correct link to "solution"

afs:
Correct link to "solution", remove "binding".

(SHACL Core uses "binding" for prefixes)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -313,7 +313,7 @@ RULE { ?x :name ?FN } WHERE {
         <p>The Shape Rules Abstract Syntax</p>
 
         <dl>
-          <dt><dfn data-cite="SPARQL12-QUERY#defn_QueryVariable">variable</dfn></dt>
+          <dt><dfn data-cite="sparql12-query#defn_QueryVariable">variable</dfn></dt>
           <dd>
             A [=variable=] represents a possible [=RDF term=] in a triple pattern.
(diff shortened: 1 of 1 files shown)


robert-david (approved):
Not sure if i understood this correctly, but the anchors
https://www.w3.org/TR/shacl12-core/#dfn-binding
https://www.w3.org/TR/shacl12-core/#dfn-solution
still don't work.

HolgerKnublauch (approved):

HolgerKnublauch (changes requested):
Actually I noticed that https://github.com/w3c/data-shapes/pull/668 is not yet approved and has the same fix but in shacl-sparql. I suggest to approve that branch instead and change the links to shacl12-sparql as my PR deletes them from shacl-core.

afs:
> Not sure if i understood this correctly, but the anchors https://www.w3.org/TR/shacl12-core/#dfn-binding https://www.w3.org/TR/shacl12-core/#dfn-solution still don't work.

The TR version is the state of the default to branch (`gh-pages`) and gets updated-by-bot soon after a merge.

afs:
A link for solution now goes to SPARQL Query. 

There is nothing gained by involving SHALCL SPARQL.

afs:
@HolgerKnublauch - this needs your review - the previous block remains active.

HolgerKnublauch (approved):

PR #708, opened 2026-01-05
#702: Changed definition of sh:deactivated to mean "ignored" or skipped.

### LINKED ISSUE #702: Clarification of `sh:deactivated` semantics
In the SHACL RDF specification, the definition of `sh:deactivated` states that "If set to true then all nodes conform to this."

While this is true, it is not immediately clear that it is not a typo. It might be more immediately intuitive if it read something like **"If set to true then all nodes will be treated as conforming to the shape, effectively 'deactivating' it."**

The documentation (Section 2.1.6) states very similarly "All RDF terms conform to a deactivated shape." which might be more immediately intuitive if it read instead: **"All RDF terms are treated as conforming to a deactivated shape."**

HolgerKnublauch:
Closes #702

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-702/shacl12-core/index.html)


YoucTagh (approved):

afs:
Another reason to change the definition of `sh:deactivated` way from "always conforms" is that if `sh:deactivated` is in a negation, it is causes non-conformance.

jeswr (approved):

PR #711, opened 2026-01-05
#676: Rewrote security section and used MUST for now

### LINKED ISSUE #676: SHACL-SPARQL dataset definition
It isn't defined in SHACL-SPARQL as to the graphs available int the dataset for the query. This leads to security risks of unauthorized access to local data.

A SHACL engine may not be executing with the same permissions as the user making the request.

SPARQL's `FROM` and `FROM NAMED` describe the dataset to be queried by giving some URIs.

There is no fixed required way to use the URIs to identify graphs. One common way is to choose named graphs from the dataset.
However, downloading graphs over HTTP is also a valid way to obtain graphs. The latter is away to leak information because the SHACL engine may not be executing with different permissions to the user making the request. SHACL engines do not typically have multi-tenant isolation.

Implicitly, the dataset for SPARQL queries is the one with the only graph being the default graph; no named graphs.
But the text does not forbid, or warn about, the dataset having other graphs accessible via `GRAPH`.

* `FROM` and `FROM NAMED` should be forbidden.
* The dataset for SPARQL execution should be defined as a single graph dataset.

Alternatively, the security section needs to have warnings. Delegating to the SPARQL security section is not enough because SHACL-SPARQL makes a restricted use of SPARQL.

HolgerKnublauch:
Closes #676

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-676/shacl12-sparql/index.html)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1993,7 +1993,7 @@ WHERE {
 		<section id="security" class="appendix">
 			<h2>Security and Privacy Considerations</h2>
 			<p>
-				Note that SPARQL directives such as <code>GRAPH</code> and <code>FROM</code> may provide access to
+				Note that SPARQL key words such as <code>GRAPH</code> and <code>FROM</code> may provide access to
 				other graphs than the active data graph in the dataset.
 				A SHACL-SPARQL engine MUST ensure that the SPARQL engine only provides access to the same
 				named graphs as the user who has triggered the validation.
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1990,20 +1990,17 @@ WHERE {
 			</section>
 		</section>
 
-		<section id="security" class="appendix informative">
+		<section id="security" class="appendix">
 			<h2>Security and Privacy Considerations</h2>
 			<p>
-				Like most RDF-based technologies, SHACL processors may operate on graphs that are combined
-				from various sources.  Some applications may have an open "linked data" architecture and dynamically
-				assemble RDF triples from sources that are outside of an organization's network of trust.
-				Since RDF allows anyone to add statements about any resource, triples may modify the originally
-				intended semantics of shape definitions or nodes in a data graph and thus lead to misleading results.
-				Protection against this (and the following) scenario can be achieved by only using trusted
-				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
-				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
+				Note that SPARQL directives such as <code>GRAPH</code> and <code>FROM</code> may provide access to
+				other graphs than the active data graph in the dataset.
+				A SHACL-SPARQL engine MUST ensure that the SPARQL engine only provides access to the same
+				named graphs as the user who has triggered the validation.
 			</p>
 			<p>
-				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>.
+				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>
+				and the <a data-cite="shacl12-core/#security">security issues of SHACL Core</a>.
 			</p>
 		</section>
 
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1990,13 +1990,13 @@ WHERE {
 			</section>
 		</section>
 
-		<section id="security" class="appendix">
+		<section id="security" class="appendix informative">
 			<h2>Security and Privacy Considerations</h2>
 			<p>
 				Note that SPARQL key words such as <code>GRAPH</code> and <code>FROM</code> may provide access to
-				other graphs than the active data graph in the dataset.
-				A SHACL-SPARQL engine MUST ensure that the SPARQL engine only provides access to the same
-				named graphs as the user who has triggered the validation.
+				graphs other than the active data graph in the dataset.
+				A SHACL-SPARQL engine should ensure that the SPARQL engine does not provide access to
+				named graphs which the user who has triggered the validation is not permitted to access.
 			</p>
 			<p>
 				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1999,8 +1999,10 @@ WHERE {
 				named graphs which the user who has triggered the validation is not permitted to access.
 			</p>
 			<p>
-				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>
-				and the <a data-cite="shacl12-core/#security">security issues of SHACL Core</a>.
+				Security considerations of SHACL-SPARQL include all the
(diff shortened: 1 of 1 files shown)


shacl12-sparql/index.html
```diff
@@ -1990,20 +1990,17 @@ <h3>sh:lessThanOrEquals</h3>
...
-				assemble RDF triples from sources that are outside of an organization's network of trust.
-				Since RDF allows anyone to add statements about any resource, triples may modify the originally
-				intended semantics of shape definitions or nodes in a data graph and thus lead to misleading results.
-				Protection against this (and the following) scenario can be achieved by only using trusted
-				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
-				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
+				Note that SPARQL directives such as <code>GRAPH</code> and <code>FROM</code> may provide access to
+				other graphs than the active data graph in the dataset.
+				A SHACL-SPARQL engine MUST ensure that the SPARQL engine only provides access to the same
+				named graphs as the user who has triggered the validation.
```
```suggestion
				Note that SPARQL directives such as <code>GRAPH</code> and <code>FROM</code> may include
				graphs other than the active data graph in the dataset.
				A SHACL-SPARQL engine MUST ensure that the SPARQL engine does not provide access to
				named graphs which the user who has triggered the validation is not permitted to access.
```

shacl12-sparql/index.html
```diff
@@ -1990,20 +1990,17 @@ <h3>sh:lessThanOrEquals</h3>
...
 			<p>
-				Like most RDF-based technologies, SHACL processors may operate on graphs that are combined
-				from various sources.  Some applications may have an open "linked data" architecture and dynamically
-				assemble RDF triples from sources that are outside of an organization's network of trust.
-				Since RDF allows anyone to add statements about any resource, triples may modify the originally
-				intended semantics of shape definitions or nodes in a data graph and thus lead to misleading results.
-				Protection against this (and the following) scenario can be achieved by only using trusted
-				and verified RDF sources and eliminating the possibility that graphs are dynamically added via
-				<code>owl:imports</code> and <code>sh:shapesGraph</code>.
+				Note that SPARQL directives such as <code>GRAPH</code> and <code>FROM</code> may provide access to
```
```suggestion
				Note that SPARQL key words such as <code>GRAPH</code> and <code>FROM</code> may provide access to
```

"directives" are things like `VERSION`, `BASE` and `PREFIX`.
https://www.w3.org/TR/rdf12-turtle/#grammar-production-directive

shacl12-sparql/index.html
```diff
@@ -1990,20 +1990,17 @@ <h3>sh:lessThanOrEquals</h3>
...
+				other graphs than the active data graph in the dataset.
+				A SHACL-SPARQL engine MUST ensure that the SPARQL engine only provides access to the same
```
afs: I don't see how this can be "MUST". Why not "SHOULD"? In a enclosed environment, it does not matter so why would that make it non-compliant?

More generally, if this is normative-MUST then the section should cover all, or at least most of, the possible security issues.

MUST needs tests.

shacl12-sparql/index.html
```diff
@@ -1990,20 +1990,17 @@ <h3>sh:lessThanOrEquals</h3>
...
+				named graphs as the user who has triggered the validation.
 			</p>
 			<p>
-				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>.
+				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>
+				and the <a data-cite="shacl12-core/#security">security issues of SHACL Core</a>.
```
afs: SHACL Core has a "Privacy and Security" section and it is non-normative.

shacl12-sparql/index.html
TallTed: `SHOULD` SHOULD also be tested. That said, I have no objection to changing this `MUST` to a `SHOULD`.

shacl12-sparql/index.html
TallTed: I think the "Privacy and Security" section should be split. I would also tweak the text, a la --
```suggestion
				Security considerations of SHACL-SPARQL include all the
				<a data-cite="sparql12-query/#security">security considerations
				of SPARQL</a> and the <a data-cite="shacl12-core/#security">security
				considerations of SHACL Core</a>.
```

shacl12-sparql/index.html
ajnelson-nist: On the first line's revision, `s/provide access to/include`:
I'm not sure "include" captures the right potential of a security action.  "Integrate"?  "Induce access"?  "Attempt access"?

shacl12-sparql/index.html
ajnelson-nist: I think "MUST" would induce a certain review level for implementations and access control policies and mechanisms.

In an enclosed environment like @afs describes, the access policies could be described as "All connections/accesses are trusted."  Then the SHACL-SPARQL engine would be conformant with policy.

If "MUST" here would induce that the SHACL test suite needs to add an XFAIL pertaining to blocked access...I appreciate that might be difficult with the current test scaffolding, if SPARQL hasn't done similar already.  @TallTed , I agree it would be good for "SHOULD" to be tested, but do we have scaffolding, or the resources to design scaffolding, for access control tests too?  What would the additions to `sh:ValidationResult`s look like?
If worth considering, please feel free to quote this in a new Issue.

Following on to @afs 's later remark on SHACL Core's "Privacy and Security" section being non-normative, I checked SPARQL 1.2's "Security Considerations" section and noted there's no usage of "MUST".

Since the highlighted "MUST" here is due to mechanisms available to SPARQL, it seems overly assertive to strengthen "should"s [sic., lowercase] of SPARQL to "MUST"s downstream.

So, we should probably drop this back to "SHOULD" [sic.].

shacl12-sparql/index.html
HolgerKnublauch: I also prefer to keep "provide access" but have accepted the other changes.

shacl12-sparql/index.html
HolgerKnublauch: I only switched to normative because I thought we wanted MUST here. But I have now switched to "should" and made it informative again.

shacl12-sparql/index.html
```diff
@@ -1993,17 +1993,16 @@ <h3>sh:lessThanOrEquals</h3>
...
+				graphs other than the active data graph in the dataset.
+				A SHACL-SPARQL engine should ensure that the SPARQL engine does not provide access to
+				named graphs which the user who has triggered the validation is not permitted to access.
 			</p>
 			<p>
-				SHACL-SPARQL includes all the <a data-cite="sparql12-query/#security">security issues of SPARQL</a>.
+				Security considerations of SHACL-SPARQL include all the
+				<a data-cite="sparql12-query/#security">security considerations
+				of SPARQL</a> and the <a data-cite="shacl12-core/#security">security
+				considerations of SHACL Core</a>.
```
TallTed: This might be better --
```suggestion
				Security considerations of SHACL-SPARQL include all the
				security considerations of both
				<a data-cite="sparql12-query/#security">SPARQL</a> and
				<a data-cite="shacl12-core/#security">SHACL Core</a>.
```

ajnelson-nist (approved):

PR #713, opened 2026-01-05
#712: Switched to sh:targetClass, added note on how to handle this test

### LINKED ISSUE #712: Errors in test for sh:conformanceDisallows and some prefix declarations
See upcoming PR, the test uses sh:targetNode instead of sh:targetClass and also should clarify how the test harness needs to run it.

https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-test-suite/tests/core/validation-reports/conformance-disallows-001.ttl

HolgerKnublauch:
Also removed the sh:resultMessage as that's implementation-specific.

Closes #712

### DIFF
--- a/shacl12-test-suite/tests/core/validation-reports/conformance-disallows-001.ttl
+++ b/shacl12-test-suite/tests/core/validation-reports/conformance-disallows-001.ttl
@@ -10,7 +10,7 @@
 ex:PersonShape
   rdf:type sh:NodeShape ;
   sh:property ex:PersonShape-age ;
-  sh:targetNode ex:Person ;
+  sh:targetClass ex:Person ;
 .
 ex:PersonShape-age
   rdf:type sh:PropertyShape ;
@@ -31,7 +31,11 @@ ex:Bob
 <conformance-disallows-001>
   rdf:type sht:Validate ;
   rdfs:label "Test of path sh:conformanceDisallows 001" ;
-  rdfs:comment "conformanceDisallows presence lets sh:conforms be true despite presence of an sh:Warning-severity result.  Test contents and behavior otherwise match conformance-disallows-002."@en ;
+  rdfs:comment """
+      Note that in order to pass this test, the test framework needs to use the values of sh:conformanceDisallows
+      from the mf:result.
+      Here, the presence of sh:conformanceDisallows lets sh:conforms be true despite presence of an sh:Warning-severity result.
+      Test contents and behavior otherwise match conformance-disallows-002."""@en ;
   rdfs:seeAlso <conformance-disallows-002> ;
   mf:action [
       sht:dataGraph <> ;
@@ -47,7 +51,6 @@ ex:Bob
           sh:focusNode ex:Bob ;
           sh:resultPath ex:age ;
           sh:value "twenty two" ;
-          sh:resultMessage "ex:age expects a literal of datatype xsd:integer." ;
           sh:sourceConstraintComponent sh:DatatypeConstraintComponent ;
           sh:sourceShape ex:PersonShape-age ;
       ]
--- a/shacl12-test-suite/tests/sparql/node/sparql-003.ttl
+++ b/shacl12-test-suite/tests/sparql/node/sparql-003.ttl
@@ -11,7 +11,7 @@
 <http://example.com/ns#>
   sh:declare [
       rdf:type sh:PrefixDeclaration ;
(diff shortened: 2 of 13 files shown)


shacl12-test-suite/tests/sparql/node/sparql-003.ttl
```diff
@@ -24,12 +24,12 @@ ex:LanguageExampleShape
   rdf:type sh:SPARQLConstraint ;
   sh:message "Values are literals with German language tag." ;
   sh:prefixes <http://example.com/ns#> ;
-  sh:select """
-			SELECT $this (ex:germanLabel AS ?path)
-			WHERE {
-				$this ex:germanLabel ?value .
-				FILTER (!isLiteral(?value) || !langMatches(lang(?value), \"de\"))
-			}
+  sh:select """
```
ajnelson-nist, line 27: There's some text-editor noise in here.  The line endings got affected by patching line 14.

Unfortunately, a lot of this noise got introduced in patch 2 ( https://github.com/w3c/data-shapes/pull/713/commits/b84d507cb9d8f7770dab6e4d299a549db04a80c7 ).

shacl12-test-suite/tests/sparql/property/property-select-001.ttl
```diff
@@ -27,23 +27,18 @@ ex:Person-fullName
...
 ex:JohnMuir
   a ex:Person ;
   ex:firstName "John" ;
   ex:lastName "Muir" .
 
-ex:JohnWayne
-  a ex:Person ;
-  ex:firstName "John" ;
-  ex:lastName "Wayne" .
-
```
ajnelson-nist, line 46: This is a non-editorial test revision.  Including the `ex:JohnWayne` node shows the test's demonstrated folly-point of writing a too-broad targeter---should be `sh:targetNode` instead of `sh:targetClass`---but I'm not sure if there's a huge benefit to retaining the `ex:JohnWayne` node and splitting the property shape into two, one a validator targeting `ex:JohnMuir`, one a property value constructor targeting the class.

ajnelson-nist: By "non-editorial," I mean I thought the change to this file was just going to be tweaking the namespace.  The test design seems to still be in the spirit of #712 .

ajnelson-nist (approved):
Most of this is fine.  I won't stand in the way of `property-select-001.ttl`'s change going forward, but I think the change should be confirmed by anyone else.  I'm fine with @HolgerKnublauch or @bergos affirming the change should proceed.

My preference would be the second patch be revised by force-push to remove the whitespace text editor noise, because someone else's editor is going to come in and just oscillate that change back.

shacl12-test-suite/tests/sparql/node/sparql-003.ttl
HolgerKnublauch: I don't even see what's different. Is that Windows vs Unix line endings?

In any case, I hope we can live with that. I edited this on a Mac, and I somehow assume that's the default platform for most people here.

shacl12-test-suite/tests/sparql/property/property-select-001.ttl
HolgerKnublauch: Yes it is more than editorial, but the test wasn't using JohnWayne so I thought I'd clean it up.

HolgerKnublauch:
With @bergos on holidays for another week, I grant myself permission to proceed here. Test cases will continue to evolve and this here isn't the final word.

shacl12-test-suite/tests/core/validation-reports/conformance-disallows-001.ttl
```diff
@@ -31,7 +31,11 @@ ex:Bob
 <conformance-disallows-001>
   rdf:type sht:Validate ;
   rdfs:label "Test of path sh:conformanceDisallows 001" ;
-  rdfs:comment "conformanceDisallows presence lets sh:conforms be true despite presence of an sh:Warning-severity result.  Test contents and behavior otherwise match conformance-disallows-002."@en ;
+  rdfs:comment """
+      Note that in order to pass this test, the test framework needs to use the values of sh:conformanceDisallows
+      from the mf:result.
+      Here, the presence of sh:conformanceDisallows lets sh:conforms be true despite presence of an sh:Warning-severity result.
```
TallTed, line 37:
```suggestion
      Here, the presence of sh:conformanceDisallows lets sh:conforms be true despite the presence of an sh:Warning-severity result.
```

PR #716, opened 2026-01-06
Issue 706: Rewritten Abstract

### LINKED ISSUE #706: It's SHACL, and not RVL
The abstract says "SHACL is a language for validating RDF graphs against a set of conditions"
This is indeed an important use case and one that boosted this standard. But it is called SHACL and not RVL (RDF Validation Language), because its purpose is more general. The SHACL standard should stimulate re-use (including innovative re-purposing) of SHACL and not reinforce a single use-case narrative. 

Suggestion: "SHACL is a language for expressing constraints and rules on RDF graphs"

HolgerKnublauch:
Closes #706

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-706/shacl12-core/index.html#abstract)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -182,14 +182,18 @@
 		<section id="abstract">
 			<p>
 				This document defines the Core of the SHACL Shapes Constraint Language.
-				SHACL is a language for validating RDF graphs against a set of conditions.
-				These conditions are provided as shapes and other constructs expressed in the form of an RDF graph.
-				RDF graphs that are used in this manner are called "shapes graphs" in SHACL and
-				the RDF graphs that are validated against a shapes graph are called "data graphs".
+				SHACL is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of
+				these classes can have.
+				More general than classes and instances, SHACL introduces the notion of shapes
+				that can describe constraints on RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF in so-called "shapes graphs".
+				The RDF graphs that are validated against a shapes graph are called "data graphs".
 				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
 				they can also be viewed as a description of the data graphs that do satisfy these conditions.
 				Such descriptions may be used for a variety of purposes beside validation, including
-				user interface building, code generation and data integration.
+				domain modeling, ontologies to inform AI agents, user interface building,
+				code generation and data integration.
 			</p>
 		</section>
 
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -185,7 +185,7 @@
             </p>
             <p>
 				SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
-				SHACL can be used to define classes and the properties that the instances of these classes can have.
+				SHACL can be used to define classes and the properties that instances of these classes can have.
 				More general than classes and instances, SHACL introduces the notion of shapes
 				that can formally specify constraints on the structure of RDF nodes and edges.
 				SHACL shapes are themselves represented in RDF graphs called shapes graphs.
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -181,8 +181,10 @@
 	<body>
 		<section id="abstract">
 			<p>
-				This document defines the Core of SHACL, the Shapes Constraint Language.
-				SHACL is a language for describing the structure of RDF graphs.
+				This document defines the Core of SHACL.
+            </p>
+            <p>
+			 SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
 				SHACL can be used to define classes and the properties that the instances of
 				these classes can have.
 				More general than classes and instances, SHACL introduces the notion of shapes
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -186,13 +186,13 @@
 				SHACL can be used to define classes and the properties that the instances of
 				these classes can have.
 				More general than classes and instances, SHACL introduces the notion of shapes
-				that can describe constraints on the structure of RDF nodes and edges.
+				that can formally describe constraints on the structure of RDF nodes and edges.
 				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
 				The RDF graphs that are validated against a shapes graph are called data graphs.
 				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions,
-				they can be viewed as a description of the data graphs that do satisfy these conditions.
+				they can be viewed as a description of the structure of data graphs that do satisfy these conditions.
 				Such descriptions may be used for a variety of purposes beside validation, including
-				domain modeling, ontologies to inform other software systems, user interface building,
+				domain modeling, ontologies to inform other agents, user interface building,
 				code generation and data integration.
 			</p>
 		</section>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -181,18 +181,18 @@
 	<body>
 		<section id="abstract">
 			<p>
-				This document defines the Core of the SHACL Shapes Constraint Language.
+				This document defines the Core of SHACL, the Shapes Constraint Language.
 				SHACL is a language for describing the structure of RDF graphs.
 				SHACL can be used to define classes and the properties that the instances of
 				these classes can have.
 				More general than classes and instances, SHACL introduces the notion of shapes
-				that can describe constraints on RDF nodes and edges.
-				SHACL shapes themselves are represented in RDF in so-called "shapes graphs".
-				The RDF graphs that are validated against a shapes graph are called "data graphs".
-				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
-				they can also be viewed as a description of the data graphs that do satisfy these conditions.
+				that can describe constraints on the structure of RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
+				The RDF graphs that are validated against a shapes graph are called data graphs.
+				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions,
+				they can be viewed as a description of the data graphs that do satisfy these conditions.
 				Such descriptions may be used for a variety of purposes beside validation, including
-				domain modeling, ontologies to inform AI agents, user interface building,
+				domain modeling, ontologies to inform other software systems, user interface building,
 				code generation and data integration.
 			</p>
 		</section>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -184,16 +184,15 @@
 				This document defines the Core of SHACL.
             </p>
             <p>
-			 SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
-				SHACL can be used to define classes and the properties that the instances of
-				these classes can have.
+				SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of these classes can have.
 				More general than classes and instances, SHACL introduces the notion of shapes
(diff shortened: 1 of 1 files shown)


shacl12-core/index.html
```diff
@@ -181,15 +181,19 @@
...
-				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
-				they can also be viewed as a description of the data graphs that do satisfy these conditions.
+				This document defines the Core of SHACL, the Shapes Constraint Language.
+				SHACL is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of
+				these classes can have.
+				More general than classes and instances, SHACL introduces the notion of shapes
+				that can formally describe constraints on the structure of RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
+				The RDF graphs that are validated against a shapes graph are called data graphs.
```
liviorobaldo:
```suggestion
RDF graphs that satisfy the constraints expressed by a set of SHACL shapes are said to be <i>valid</i> with respect to those shapes and are called <i>data graphs</i>.
```

The previous version used the wording "…that are validated against a shapes graph…". This was the first occurrence of the verb "validate" in the document. However, I think we should define what “validate” means before using the verb, given the central role of validation in SHACL.

shacl12-core/index.html
```diff
@@ -181,15 +181,19 @@
...
-				the RDF graphs that are validated against a shapes graph are called "data graphs".
-				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
-				they can also be viewed as a description of the data graphs that do satisfy these conditions.
+				This document defines the Core of SHACL, the Shapes Constraint Language.
+				SHACL is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of
+				these classes can have.
+				More general than classes and instances, SHACL introduces the notion of shapes
+				that can formally describe constraints on the structure of RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
```
liviorobaldo:
```suggestion
SHACL shapes are themselves represented in RDF graphs called shapes graphs; they group together different constraints on the same focus nodes.
```

I have incorporated Robert's suggestion here, which I agree with.

shacl12-core/index.html
```diff
@@ -181,15 +181,19 @@
...
 			<p>
-				This document defines the Core of the SHACL Shapes Constraint Language.
-				SHACL is a language for validating RDF graphs against a set of conditions.
-				These conditions are provided as shapes and other constructs expressed in the form of an RDF graph.
-				RDF graphs that are used in this manner are called "shapes graphs" in SHACL and
-				the RDF graphs that are validated against a shapes graph are called "data graphs".
-				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
-				they can also be viewed as a description of the data graphs that do satisfy these conditions.
+				This document defines the Core of SHACL, the Shapes Constraint Language.
+				SHACL is a language for describing the structure of RDF graphs.
```
```suggestion
				This document defines the Core of SHACL.
            </p>
            <p>
			 SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
```

With this suggestion the text is SHACL generally might be reused in other documents, e.g. any "SHACL New" note and possibly shared via shacl12-common.

shacl12-core/index.html
TallTed: Tweaking @liviorobaldo's change... The reason for my tweak is that I think it's easier (especially for people new to the world of SHACL and/or RDF) to conceive of _descriptions_ being put to multiple uses than it is to conceive of _constraints_ being put to those same multiple uses.
```suggestion
SHACL shapes are themselves represented in RDF graphs called _shapes graphs_,
multiples of which may be collected into one or more RDF graphs or documents.
```

shacl12-core/index.html
TallTed: Tweak similar to my suggestion of a few minutes ago.
```suggestion
RDF graphs that contain data that is to be tested for conformance with the
descriptions expressed by a set of SHACL shapes are called <i>data graphs</i>.
If they are found to conform with those descriptions, they are considered to
be <i>valid</i> with respect to those shapes.
```

liviorobaldo: Ok, but @afs was suggesting removing “valid”, “validation”, “validating”, etc. from the abstract. I tend to agree with that suggestion, so I propose mentioning validation only in the final sentence, together with the other tasks that SHACL can be used for.

shacl12-core/index.html
```diff
@@ -181,15 +181,19 @@
...
+				More general than classes and instances, SHACL introduces the notion of shapes
+				that can formally describe constraints on the structure of RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
+				The RDF graphs that are validated against a shapes graph are called data graphs.
+				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions,
+				they can be viewed as a description of the structure of data graphs that do satisfy these conditions.
 				Such descriptions may be used for a variety of purposes beside validation, including
-				user interface building, code generation and data integration.
+				domain modeling, ontologies to inform other agents, user interface building,
+				code generation and data integration.
```
liviorobaldo:
```suggestion
Such descriptions may be used for a variety of purposes: validating, inferencing, modeling domains, generating ontologies to inform other agents, building user interfaces, generating code, and integrating data.
```

I tried to incorporate @TallTed’s suggestion from [Issue #706](https://github.com/w3c/data-shapes/issues/706#issuecomment-3716364538), but I don’t think I did it correctly. The suggestion above should span lines 194–195–196, and I’m not sure how to replace all three lines with a single suggestion. Can someone kindly explain me how to do that (if that's possible)?

shacl12-core/index.html
afs: The grouping mentioned earlier is about having many shapes in one graph/file to be applied to the same data.
It is not about multiple about a focus nodes. At this level of description, the abstract should not mention focus nodes.

shacl12-core/index.html
afs: > <i>valid</i> with respect to those shapes and are called <i>data graphs</i>.

A data graph is a graph used for the data, as opposed to the shapes graph. It is an input to the validation process.

A data graph becomes valid or invalid _after_ the shapes are used. It does not change its status of being a data graph.

liviorobaldo: True. So maybe we could write above "SHACL shapes themselves are represented in RDF graphs called <i>shapes graphs</i>, which are then applied to other RDF graphs called <i>data graph</i>."

And here we could then write:

"Data graphs that conform to the descriptions expressed by a set of SHACL shapes are said to be <i>valid</i> with respect to those shapes"

But again, I tend to agree with your suggestion to avoid emphasizing validation too much in the abstract; I would only mention it in the final sentence, along with the other tasks.

TallTed: Suggestion revised.

shacl12-core/index.html
TallTed: Revised suggestion.

shacl12-core/index.html
TallTed: At the moment, suggesting one change for multiple lines requires use of the old "changed files" interface, not the new one. GitHub is not making our work easy! :-(

Also, GitHub does not accept suggestions that start before a changed line, and end after it. This means that such suggestions must be broken up. I'm making a pair of suggestions now, both of which will need to be merged into the PR.

shacl12-core/index.html
```diff
@@ -181,15 +181,19 @@
...
+				SHACL is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of
+				these classes can have.
+				More general than classes and instances, SHACL introduces the notion of shapes
+				that can formally describe constraints on the structure of RDF nodes and edges.
+				SHACL shapes themselves are represented in RDF graphs called shapes graphs.
+				The RDF graphs that are validated against a shapes graph are called data graphs.
+				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions,
+				they can be viewed as a description of the structure of data graphs that do satisfy these conditions.
 				Such descriptions may be used for a variety of purposes beside validation, including
```
TallTed: Suggestion 1 of 2
```suggestion
```

shacl12-core/index.html
TallTed: Suggestion 2 of 2
```suggestion
Such descriptions may be used for a variety of purposes: validating, inferencing, modeling domains, generating ontologies to inform other agents, building user interfaces, generating code, and integrating data.
```

scotthenninger: > Suggestion 2 of 2

Is it "modeling domains" or "domain modeling"? I think the term domain modeling should be used here as it is more familiar as a modeling topic.

TallTed: I'm not dead set on "modeling domains". I used this to keep all the list items in the same shape. More opinions?

HolgerKnublauch:
I made some further simplifications. Given that we focus on the term "describes" I found that the extra justification that we added to explain that SHACL is more than a constraint language was no longer needed. Please revisit if anyone disagrees with those simplifications, but as an editor someone needs to keep the various discussions together and move the document forward somehow...

shacl12-core/index.html
HolgerKnublauch: BTW the term "valid" is not used in the main spec so shouldn't go into the abstract either. We use "conforms" everywhere.

shacl12-core/index.html
liviorobaldo: I agree with @scotthenninger

liviorobaldo: Thanks. I suspected it wasn't possible, just wanted to double-check. It would be a useful feature indeed.

liviorobaldo:
Thanks @HolgerKnublauch, this looks fine to me. I’m not particularly fond of the verb “declare” in “...that can formally _declare_ constraints on the structure of RDF nodes and edges”; “specify” sounds slightly better to my ear (i.e., “...that can formally _specify_ constraints on the structure of RDF nodes and edges”). But this is purely a matter of taste, please feel free to ignore it.

YoucTagh (approved):

shacl12-core/index.html
```diff
@@ -181,19 +181,20 @@
...
 		<section id="abstract">
 			<p>
-				This document defines the Core of SHACL, the Shapes Constraint Language.
-				SHACL is a language for describing the structure of RDF graphs.
-				SHACL can be used to define classes and the properties that the instances of
-				these classes can have.
+				This document defines the Core of SHACL.
+            </p>
+            <p>
+				SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
```
TallTed, line 187:
```suggestion
				This document defines the Core of SHACL, the Shapes Constraint Language.
            </p>
            <p>
				SHACL is a language for describing the structure of RDF graphs.
```

HolgerKnublauch: I prefer the wording that Andy had suggested because otherwise it can be interpreted as

"Core is the Shapes Constraint Language"

shacl12-core/index.html
```diff
@@ -181,15 +181,20 @@
...
-				the RDF graphs that are validated against a shapes graph are called "data graphs".
-				As SHACL shape graphs are used to validate that data graphs satisfy a set of conditions
-				they can also be viewed as a description of the data graphs that do satisfy these conditions.
-				Such descriptions may be used for a variety of purposes beside validation, including
-				user interface building, code generation and data integration.
+				This document defines the Core of SHACL.
+            </p>
+            <p>
+				SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs.
+				SHACL can be used to define classes and the properties that the instances of these classes can have.
```
afs: Minor:

```suggestion
				SHACL can be used to define classes and the properties that instances of these classes can have.
```

afs (approved):

bergos (approved):

shacl12-core/index.html
TallTed: Hm, yes, I can see that.

PR #717, opened 2026-01-07
#709: Added sh:unit property

### LINKED ISSUE #709: Support for units and currencies
Many properties implicitly assume that values are stored in a certain unit of measurement. For example, some ontologies define properties such as priceUSD or weightKG. It would be better to have this lifted into the ontology itself and made explicit. For example, we could have

```
ex:Fridge-height
    sh:path ex:height ;
    sh:datatype xsd:decimal ;
    sh:unit "cm" ;    # Or shui:unit   or sh:unitCode
.
```

This is particularly useful for UI builders, which can render values as "180 cm" instead of just "180". The UI (and server) can also use this info to automatically translate between units, for example to show feet instead of metric units for US visitors.

There are various units ontologies esp QUDT that define proper URIs for these units. However, I think it would be less commitment and more flexible to just use them as strings, because those strings can be mapped to the units as they are standardized already by International System of Units (SI) and ISO. So people can still use QUDT with a simple join in SPARQL.

And we can expect people to add currencies too, such as USD, EUR and AUD and those are more fluent as countries and their financial systems keep evolving.

As the primary use case is for UI, we could elect to move such a property into SHACL UI namespace. However, I think units are so central that they should be promoted to the main sh: namespace because they may be picked up by other algorithms. For example, with temperatures in Kelvin, I think 0 is the implicit sh:minInclusive value. Therefore I suggest we add them to the section that already contains sh:name, sh:codeIdentifier and sh:order.

HolgerKnublauch:
Closes #709

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-709/shacl12-core/index.html#unit)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5372,20 +5372,39 @@ ex:RainbowPony ex:color ex:Pink .
 				<p>
 					Shapes may have <a>values</a> for the property <code>sh:unit</code> to indicate a unit of
 					measure, currency or similar information about the <a>value nodes</a>.
-					The values of <code>sh:unit</code> must be <a>literals</a> with <a>datatype</a> `xsd:string`
-					or `rdf:langString` or `rdf:dirLangString`.
+					The values of <code>sh:unit</code> can be either <a>literals</a> with <a>datatype</a> `xsd:string`
+					or `rdf:langString` or `rdf:dirLangString`, or <a>IRIs</a>.
+					SHACL 1.2 does not prescribe specific values for <code>sh:unit</code> because, at the time of writing (2026),
+					there was no official W3C standard to represent them uniformly, and multiple standards exist outside of the W3C.
 				</p>
 				<p>
-					For units of measure the recommended values for <code>sh:unit</code> are <code>xsd:string</code>
-					literals that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>
-					as used by the `qudt:ucumCode` property of the <a href="https://qudt.org/">QUDT</a> units vocabulary.
-					In addition to these canonical values, there may be additional language-tagged literals
-					with abbreviations or display names that are commonly used in the selected languages
-					(e.g. <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches).
+					For SHACL 1.2, the following recommendations are given:
 				</p>
+				<ul>
+					<li>
+						For units of measure the values for <code>sh:unit</code> may be <code>xsd:string</code>
+						<a>literals</a> that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>,
+						for example <code>sh:unit "cm"</code>.
+					</li>
+					<li>
+						For currencies the values for <code>sh:unit</code> may be <code>xsd:string</code> <a>literals</a>
+						that exactly match the codes defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>,
+						for example <code>sh:unit "AUD"</code>.
+					</li>
+					<li>
+						The values for <code>sh:unit</code> may be the <a>IRIs</a> defined by the
+						<a href="https://qudt.org/">QUDT</a> units vocabulary,
+						for example <code>sh:unit qudt:CentiM</code>.
+					</li>
+					<li>
+						In addition to these canonical values, there may be additional language-tagged literals
+						with abbreviations or display names that are commonly used in the selected languages,
+						for example <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches.
+					</li>
+				</ul>
 				<p>
-					For currencies the recommended values for <code>sh:unit</code> are <code>xsd:string</code> literals
-					that exactly match the codes defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>.
+					While these syntax rules are intentionally left flexible, specific applications may enforce more specific rules,
+					for example by declaring a <code>sh:class qudt:Unit</code> constraint on the <code>sh:unit</code> property.
 				</p>
 				<p>
 					User interface tools based on SHACL can use the values of <code>sh:unit</code> to render values.
@@ -5751,8 +5770,8 @@ ex:NameGroup
 				<li>Generalized <a href="#order"></a> to also allow xsd:integers; see <a href="https://github.com/w3c/data-shapes/issues/479">Issue 479</a></li>
 				<li>Added new <a href="#conformanceDisallows"><code>sh:conformanceDisallows</code> property</a> to the validation report; see <a href="https://github.com/w3c/data-shapes/issues/453">Issue 453</a></li>
 				<li>Added annotation property <a href="#codeIdentifier"></a>; see <a href="https://github.com/w3c/data-shapes/issues/559">Issue 559</a></li>
-				<li>Added the target types <a href="#targetWhere"><code>sh:targetWhere</code></a> and <a href="#explicit-shape-target"><code>sh:shape</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/517">Issue 517</a></li>
-				<li>Added support for units and currencies <a href="#unit"><code>sh:unit</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/709">Issue 709</a></li>
+				<li>Added the target types <a href="#targetWhere"></a> and <a href="#explicit-shape-target"></a>, see <a href="https://github.com/w3c/data-shapes/issues/517">Issue 517</a></li>
+				<li>Added support for units and currencies <a href="#unit"></a>, see <a href="https://github.com/w3c/data-shapes/issues/709">Issue 709</a></li>
 			</ul>
 		</section>
   </body>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5372,7 +5372,8 @@ ex:RainbowPony ex:color ex:Pink .
 				</p>
 				<p>
 					For units of measure the recommended values for <code>sh:unit</code> are <code>xsd:string</code>
-					literals that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>.
+					literals that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>
+					as used by the `qudt:ucumCode` property of the <a href="https://qudt.org/">QUDT</a> units vocabulary.
 					In addition to these canonical values, there may be additional language-tagged literals
 					with abbreviations or display names that are commonly used in the selected languages
 					(e.g. <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches).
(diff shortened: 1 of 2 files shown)


shacl12-core/index.html
```diff
@@ -5362,6 +5362,77 @@ <h3>sh:codeIdentifier</h3>
 					Similar requirements exist for API generators, for example in Java, JavaScript or Python.
 				</p>
 			</section>
+			<section id="unit">
+				<h3>sh:unit</h3>
+				<p>
+					Property shapes may have <a>values</a> for the property <code>sh:unit</code> to indicate a unit of
+					measure, currency or similar information about the values of a property.
+					The values of this property must be <a>literals</a> with <a>datatype</a> <code>xsd:string</code>
+					or <code>rdf:langString</code>.
```
```suggestion
					The values of this property must be <a>literals</a> with <a>datatype</a> `xsd:string`
					or `rdf:langString` or `rdf:dirLangString`.
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -1749,6 +1749,14 @@ sh:order
 	# range: xsd:decimal or xsd:integer ;
 	rdfs:isDefinedBy sh: .
 
+sh:unit
+	a rdf:Property ;
+	rdfs:label "unit"@en ;
+	rdfs:comment "The code of a unit that the (literal) values of a property should be interpreted with. Recommended values are plain strings from the UCUM and ISO 4217 standards."@en ;
```
YoucTagh:
```suggestion
	rdfs:comment "The code of a unit that the (literal) values of a property should be interpreted with. Recommended values are plain strings from the UCUM and ISO 4217 standards for units of measure and currencies respectively."@en ;
```

YoucTagh (approved):

shacl12-core/index.html
```diff
@@ -5653,6 +5747,7 @@ <h2>Changes between the original SHACL Core and SHACL 1.2 Core</h2>
 				<li>Added new <a href="#conformanceDisallows"><code>sh:conformanceDisallows</code> property</a> to the validation report; see <a href="https://github.com/w3c/data-shapes/issues/453">Issue 453</a></li>
 				<li>Added annotation property <a href="#codeIdentifier"></a>; see <a href="https://github.com/w3c/data-shapes/issues/559">Issue 559</a></li>
 				<li>Added the target types <a href="#targetWhere"><code>sh:targetWhere</code></a> and <a href="#explicit-shape-target"><code>sh:shape</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/517">Issue 517</a></li>
+				<li>Added support for units and currencies <a href="#unit"><code>sh:unit</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/709">Issue 709</a></li>
```
afs: (for all new properties and featues)

If you put in `<a href="#unit></a>`, respec will put in a section link as if it were `<a href="#unit>8.3 sh:unit</a>`.

shacl12-core/index.html
```diff
@@ -5372,20 +5372,39 @@ <h3>sh:unit</h3>
...
+						<a>literals</a> that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>,
+						for example <code>sh:unit "cm"</code>.
+					</li>
+					<li>
+						For currencies the values for <code>sh:unit</code> may be <code>xsd:string</code> <a>literals</a>
+						that exactly match the codes defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>,
+						for example <code>sh:unit "AUD"</code>.
+					</li>
+					<li>
+						The values for <code>sh:unit</code> may be the <a>IRIs</a> defined by the
```
JervenBolleman: I read this to suggest that only QUDT maybe used, I don't think that is the intent.

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
 			<section id="unit">
 				<h3>sh:unit</h3>
 				<p>
-					Property shapes may have <a>values</a> for the property <code>sh:unit</code> to indicate a unit of
-					measure, currency or similar information about the values of a property.
-					The values of this property must be <a>literals</a> with <a>datatype</a> `xsd:string`
-					or `rdf:langString` or `rdf:dirLangString`.
+					Shapes may have <a>values</a> for the property <code>sh:unit</code> to indicate a unit of
+					measure, currency or similar information about the <a>value nodes</a>.
```
```suggestion
					measure, currency, or similar information about the <a>value nodes</a>.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
+					The values of <code>sh:unit</code> can be either <a>literals</a> with <a>datatype</a> `xsd:string`
+					or `rdf:langString` or `rdf:dirLangString`, or <a>IRIs</a>.
```
```suggestion
					A value of <code>sh:unit</code> can be either of the following: (1) a <a>literal</a> with <a>datatype</a> `xsd:string`,
					`rdf:langString`, or `rdf:dirLangString`; or (2) an <a>IRI</a>.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
-					In addition to these canonical values, there may be additional <code>rdf:langString</code> literals
-					with abbreviations or display names that are commonly used in the selected languages
-					(e.g. <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches).
+					For SHACL 1.2, the following recommendations are given:
 				</p>
+				<ul>
+					<li>
+						For units of measure the values for <code>sh:unit</code> may be <code>xsd:string</code>
+						<a>literals</a> that exactly match the codes defined in <a href="http://ucum.org">UCUM</a>,
+						for example <code>sh:unit "cm"</code>.
```
TallTed, line 198:
```suggestion
						For units of measure a value of <code>sh:unit</code> may be an <code>xsd:string</code>
						<a>literal</a> that exactly matches a code defined in <a href="http://ucum.org">UCUM</a>;
						for example, <code>sh:unit "cm"</code>.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
+					</li>
+					<li>
+						For currencies the values for <code>sh:unit</code> may be <code>xsd:string</code> <a>literals</a>
+						that exactly match the codes defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>,
+						for example <code>sh:unit "AUD"</code>.
```
TallTed, line 5371:
```suggestion
						For currencies, a value of <code>sh:unit</code> may be an <code>xsd:string</code> <a>literal</a>
						that exactly matches a code defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>;
						for example, <code>sh:unit "AUD"</code>.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
+					</li>
+					<li>
+						The values for <code>sh:unit</code> may be the <a>IRIs</a> defined by the
+						<a href="https://qudt.org/">QUDT</a> units vocabulary,
+						for example <code>sh:unit qudt:CentiM</code>.
```
TallTed: Has the bonus of solving @JervenBolleman's [concern](https://github.com/w3c/data-shapes/pull/717/files/248dbcb7fdb87d2d8fa29bcee901c74f259a759a..ac6756cd764074dacc447a173b8269b0701ebfc0#r2687138546).
```suggestion
						A value of <code>sh:unit</code> may be an <a>IRI</a> defined by the
						<a href="https://qudt.org/">QUDT</a> or similar units vocabulary;
						for example, <code>sh:unit qudt:CentiM</code>.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
+					</li>
+					<li>
+						In addition to these canonical values, there may be additional language-tagged literals
+						with abbreviations or display names that are commonly used in the selected languages,
+						for example <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches.
```
TallTed, line 5377:
```suggestion
						with abbreviations or display names that are commonly used in the selected language(s);
						for example, <code>"Zoll"@de</code> in addition to <code>"in"</code> for inches.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
+					</li>
+				</ul>
 				<p>
-					For currencies the recommended values for <code>sh:unit</code> are <code>xsd:string</code> literals
-					that exactly match the codes defined in <a href="https://www.iso.org/iso-4217-currency-codes.html">ISO 4217</a>.
+					While these syntax rules are intentionally left flexible, specific applications may enforce more specific rules,
+					for example by declaring a <code>sh:class qudt:Unit</code> constraint on the <code>sh:unit</code> property.
```
```suggestion
					While these syntax rules are intentionally left flexible, specific applications may enforce more specific rules;
					for example, by declaring a <code>sh:class qudt:Unit</code> constraint on the <code>sh:unit</code> property.
```

shacl12-core/index.html
```diff
@@ -5370,28 +5370,48 @@ <h3>sh:codeIdentifier</h3>
...
 				</p>
 				<p>
-					User interface tools based on SHACL can use the values of <code>sh:unit</code> to render values of
-					a property.  Depending on a user's locale such tools may even automatically convert between units,
+					User interface tools based on SHACL can use the values of <code>sh:unit</code> to render values.
+					Depending on a user's locale such tools may even automatically convert between units,
 					or allow data entry in different units than those that the actual values will be stored in.
```
```suggestion
					Depending on a user's locale, such tools may automatically convert between units,
					or even allow data entry in units different than those that the actual values will be stored as.
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -1752,9 +1752,8 @@ sh:order
 sh:unit
 	a rdf:Property ;
 	rdfs:label "unit"@en ;
-	rdfs:comment "The code of a unit that the (literal) values of a property should be interpreted with. Recommended values are plain strings from the UCUM and ISO 4217 standards for units of measure and currencies respectively."@en ;
-	rdfs:domain sh:PropertyShape ;
-	# range: xsd:string or rdf:langString ;
+	rdfs:comment "The code of a unit that the (literal) value nodes should be interpreted with. Recommended values are plain strings from the UCUM and ISO 4217 standards for units of measure and currencies respectively, or QUDT unit URIs."@en ;
```
```suggestion
	rdfs:comment "The code of a unit that the (literal) value nodes should be interpreted with. Recommended values are plain strings from the UCUM and ISO 4217 standards for units of measure and currencies respectively, or unit URIs from QUDT or a similar ontology."@en ;
```

shacl12-core/index.html
HolgerKnublauch: I applied all your feedback except this. I don't think adding inline enumerations with (1) and (2) improve readability, and the intent was IMHO already clear enough because literals and IRIs are disjoint.

PR #719, opened 2026-01-09
#491: Added node expression test cases and clarified spec in two places

### LINKED ISSUE #491: Create test cases for node expressions
While most of the node expression functions are now documented, they lack systematic test cases.

HolgerKnublauch:
Closes #491

This is meant to be a START only and there certainly can and should be more tests in the future.


shacl12-test-suite/tests/node-expr/constraints/expression-001.ttl
```diff
@@ -1,48 +1,48 @@
-@prefix dash: <http://datashapes.org/dash#> .
```
ajnelson-nist, line 1: Non-blocking remark:
Here's the line ending Git noise again.

ajnelson-nist: ...Hm. Closer-to-blocking remark: The noise is pervasive enough it's hard to tell what's changed in the tests, at least through the GitHub review interface.

shacl12-test-suite/tests/node-expr/shnex/instancesOf.ttl
```diff
@@ -0,0 +1,59 @@
...
+<>
+  rdf:type mf:Manifest ;
+  mf:entries (
+      <instancesOf-base-class>
+      <instancesOf-sub-class>
+    ) ;
+.
+
+<instancesOf-base-class>
+  rdf:type sht:EvalNodeExpr ;
```
ajnelson-nist, line 20: Apologies, it's likely I've missed some conversations in the last few weeks.  Is there a reason this is `sht:EvalNodeExpr`, rather than `shnex:EvalNodeExpr`?

ajnelson-nist: Ok, not finding `EvalNodeExpr` in the node expressions document, I assume there's a test suite specification update I, or this PR, has missed.

Is [this document](https://w3c.github.io/data-shapes/data-shapes-test-suite/) going to get an entry for `EvalNodeExpr`?

HolgerKnublauch: Yes correct, there was no description yet of the new format. I have just added a readme.

We cannot modify the old test suite report, but I guess it will be better if we put the descriptions of the format into the test suite itself anyway.

And it's sht:nodeExpr because this vocabulary is part of the test suite only, not shnex.

shacl12-test-suite/tests/node-expr/constraints/expression-001.ttl
HolgerKnublauch: Thanks, I managed to roll those back. Not sure where this is coming from suddenly.

shacl12-test-suite/tests/node-expr/shnex/instancesOf.ttl
ajnelson-nist: Thank you for documenting the vocabulary.  Is the README in the source tree the better spot, though, vs. the test-suite document?

shacl12-test-suite/tests/node-expr/shnex/concat.ttl
```diff
@@ -0,0 +1,60 @@
...
+        ( 2 1 )
+      )
+    ] ;
+  ] ;
+  mf:result ( 5 4 3 2 1 ) ;
+  mf:status sht:approved ;
+.
+<concat-three-constants>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of a concat expression that concatenates three constants" ;
```
ajnelson-nist: Did this mean three *distinct* constants?  The list has four constants, three distinct.

shacl12-test-suite/tests/node-expr/shnex/distinct.ttl
```diff
@@ -0,0 +1,55 @@
...
+  mf:entries (
+      <distinct-focusNode>
+      <distinct-list>
+      <distinct-termEquality>
+    ) ;
+.
+
+<distinct-focusNode>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of an distinct expression on just the focus node" ;
```
ajnelson-nist:
```suggestion
  rdfs:label "Test of a distinct expression on just the focus node" ;
```

shacl12-test-suite/tests/node-expr/shnex/distinct.ttl
```diff
@@ -0,0 +1,55 @@
...
+      ] ;
+    ] ;
+    sht:focusNode ex:TestNode ;
+  ] ;
+  mf:result ( ex:TestNode ) ;
+  mf:status sht:approved ;
+.
+<distinct-list>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of an distinct expression on a list with 3 members and one duplicate" ;
```
ajnelson-nist:
```suggestion
  rdfs:label "Test of a distinct expression on a list with three members and one duplicate" ;
```

ajnelson-nist: Hm, on second thought, this should just be grammatically consistent with the label of line 47.  I'm fine with word-spelling or using a number.

shacl12-test-suite/tests/node-expr/shnex/distinct.ttl
```diff
@@ -0,0 +1,55 @@
...
+  mf:action [
+    sht:nodeExpr [
+      shnex:distinct ( 4 2 4 ) ;
+    ] ;
+  ] ;
+  mf:result ( 4 2 ) ;
```
ajnelson-nist, line 42: Where does `sht:ignoreOrder` go in this `EvalNodeExpr` node?  I imagine `( 2 4 )` should also be an acceptable test result.

shacl12-test-suite/tests/node-expr/shnex/distinct.ttl
```diff
@@ -0,0 +1,55 @@
...
+  mf:status sht:approved ;
+.
+<distinct-termEquality>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of an distinct expression on a list with 3 members and no term-equal duplicate" ;
```
ajnelson-nist, line 47:
```suggestion
  rdfs:label "Test of an distinct expression on a list with 3 members and no term-equal duplicate" ;
  rdfs:seeAlso <https://www.w3.org/TR/rdf12-concepts/#dfn-literal-term-equality> ;
```

It took me a few link-traversals to get to the part of the RDF 1.2 spec to refresh on literal term-equality.  This `seeAlso` may be a helpful time saver for test implementers.

shacl12-test-suite/tests/node-expr/shnex/exists.ttl
```diff
@@ -0,0 +1,40 @@
...
+      <exists-empty>
+    ) ;
+.
+
+<exists-constant>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of an exists expression with a constant expression as input" ;
+  mf:action [
+    sht:nodeExpr [
+      shnex:exists ex:SomeNode ;
```
ajnelson-nist, line 24: You should NACK these suggestions, but suppose this example were spelled like so:

```suggestion
      shnex:exists ex:UndefinedNode ;
```

or:

```suggestion
      shnex:exists [ rdfs:comment "Unlinked bnode" ] ;
```

It just looks a little odd.  I don't object, but just wanted to confirm whether this is consistent with the intended design of `shnex:exists`.

ajnelson-nist: ...Actually, looking at line 35, I'm now confused about how a blank node could be interpreted as a constant node or an empty node expression.

shacl12-test-suite/tests/node-expr/shnex/flatMap.ttl
```diff
@@ -0,0 +1,77 @@
...
+      ] ;
+    ] ;
+  ] ;
+  mf:result ( 1 2 3 ) ;
+  mf:status sht:approved ;
+.
+
+<flatMap-allSkills>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of a flatMap expression, similar to the one in the spec example" ;
```
ajnelson-nist, line 36: Non-blocking remark: I'd personally prefer "the one in the spec example" be replaced with a stable link to the spec. section or example-block.

shacl12-test-suite/tests/node-expr/shnex/nodesMatching.ttl
```diff
@@ -0,0 +1,63 @@
...
+    ] ;
+    sht:ignoreOrder true ;
+  ] ;
+  mf:result (
+    ex:Org1
+    ex:Company1
+    ex:Company2
+  ) ;
+  mf:status sht:approved ;
+.
```
ajnelson-nist, line 36: I've forgotten which thread this was happening in, but we've had some discussion of how to match nodes where types are directly asserted, pre-inferencing; vs. nodes that follow the ~"SHACL class-instance, respecting subclass hierarchy".  So, shouldn't this test's `mf:result` list differ from the one in `instancesOf.ttl`?  I thought `instancesOf.ttl` would have the result being just `( ex:Org1 )`, while this test would have the result `( ex:Org1 ex:Company1 ex:Company2 )`.

shacl12-test-suite/tests/node-expr/shnex/offset.ttl
```diff
@@ -0,0 +1,57 @@
...
+  mf:action [
+    sht:nodeExpr [
+      shnex:nodes [
+        shnex:nodes ( 1 2 3 4 ) ;
+        shnex:limit 2 ;
+      ] ;
+      shnex:offset 1 ;
+    ] ;
+  ] ;
+  mf:result ( 2 ) ;
```
ajnelson-nist: Suggestion offered because I may have been looking at too many tests in one fell swoop, and I found myself wanting clearer disambiguation from `shnex:max` and `sh:maxInclusive` [sic.].  No hard feelings if NACK'd.

```suggestion
  mf:action [
    sht:nodeExpr [
      shnex:nodes [
        shnex:nodes ( 5 6 7 8 ) ;
        shnex:limit 2 ;
      ] ;
      shnex:offset 1 ;
    ] ;
  ] ;
  mf:result ( 6 ) ;
```

shacl12-test-suite/tests/node-expr/shnex/sum.ttl
```diff
@@ -0,0 +1,54 @@
...
+  mf:entries (
+      <sum-empty>
+      <sum-list-3-integers>
+      <sum-list-3-mixed>
+    ) ;
+.
+
+<sum-empty>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of a sum expression on an empty list" ;
```
ajnelson-nist, line 23:
```suggestion
  rdfs:label "Test of a sum expression on an empty list" ;
  rdfs:seeAlso <https://www.w3.org/TR/sparql12-query/#aggSum> ;
```

I had to check SPARQL 1.2 to confirm that the sum of a cardinality-0 list is defined as (integer) 0.  Linking that spec. section might save future questions.

shacl12-test-suite/tests/node-expr/shnex/var.ttl
```diff
@@ -0,0 +1,53 @@
...
+    ) ;
+.
+<var-bound>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of a var expression querying a bound variable" ;
+  mf:action [
+    sht:nodeExpr [
+      shnex:var "bound" ;
+    ] ;
+    sht:scope-bound 4.2 ;
```
ajnelson-nist, line 26: `sht:scope-bound` needs its own documentation.

ajnelson-nist (commented):
I think there are conversation threads in `exists.ttl`, `instancesOf.ttl`/`nodesMatching.ttl`, and `var.ttl` that should be resolved before this is merged.

For some logistics reasons, please consider addressing those threads as my "Request changes", but I won't push "Request changes" on GitHub in case I end up unavailable to unblock in a timely manner.

shacl12-test-suite/tests/node-expr/shnex/instancesOf.ttl
HolgerKnublauch: We do not have a test-suite document right now for 1.2, and I am not sure if we actually still need one. The test and implementation report are more like living documents that are best kept on GitHub and follow GitHub best-practice.

The rules folder in shacl12-test-suite also already has a README.md and this would allow its editors to define a test suite for their own needs.

shacl12-test-suite/tests/node-expr/shnex/distinct.ttl
HolgerKnublauch: shnex:distinct is defined to preserve the order, so I think the test is correct as-is.

shacl12-test-suite/tests/node-expr/shnex/exists.ttl
HolgerKnublauch: The blank node with the rdfs:comment would be a malformed shapes graph, because it is none of the supported node expression types. Constant nodes can only be literals and IRIs, and then there is the empty node expression [] that is not allowed to have any triples at all. Using ex:UndefinedNode would just be another constant node expression, which is valid for exists.

shacl12-test-suite/tests/node-expr/shnex/nodesMatching.ttl
HolgerKnublauch: Both sh:instancesOf and sh:class have the same interpretation of rdfs:subClassOf, so their output is the same here. See the definition of https://www.w3.org/TR/shacl12-node-expr/#InstancesOfExpression which uses "SHACL Instances"

HolgerKnublauch:
Thanks Alex, for your review and comments. I hope I have addressed them. I went ahead and marked the threads as resolved but please feel free to reopen if you disagree or have further questions.

shacl12-node-expr/index.html
```diff
@@ -1500,6 +1500,7 @@ <h3>Intersection Expressions</h3>
 							The <a>output nodes</a> of the <a>intersection expression</a> are the <a>nodes</a> that form the intersection of the <a>output nodes</a>
 							produced by each <a>node expression</a> <code>NE</code> in <code>members</code>, using <code>evalExpr(NE, focusGraph, focusNode, scope)</code>.
 							Nodes must be equal using <a>term equality</a>, i.e. <code>"01"^^xsd:integer</code> is distinct from <code>"1"^^xsd:integer</code>.
```
robert-david:
```suggestion
							Nodes must be equal using <a>term equality</a>, e.g., <code>"01"^^xsd:integer</code> is distinct from <code>"1"^^xsd:integer</code>.
```

shacl12-node-expr/index.html
```diff
@@ -1999,7 +2003,64 @@ <h3>Offset Expressions</h3>
...
+						</span>
+					</p>
+					<div class="def" id="OrderByExpression-evaluation">
+						<div class="def-header">EVALUATION OF ORDER BY EXPRESSIONS</div>
+						<p>
+							Let <code>orderBy</code> be the <a>value</a> of <code>shnex:orderBy</code>,
+							<code>nodes</code> be the <a>value</a> of <code>shnex:nodes</code> and
+							<code>desc</code> be the <a>value</a> of <code>shnex:desc</code> in the <a>order by expression</a>.
+							Let <code>N</code> be the <a>output nodes</a> of <code>evalExpr(nodes, focusGraph, focusNode, scope)</code>.
+							Let <code>c(n)</code> be the first output node of <code>evalExpr(orderBy, focusGraph, n, scope)</code>
```
robert-david, line 2055: I don't understand what c(n) returns. Why is it a single (first) node? I guess this should be an ordering function or returning and index, like described in '...sorted by c(n)...'. Also, shouldn't the value of shnex:desc be included in the evalExpr?

shacl12-test-suite/tests/node-expr/shnex/count.ttl
```diff
@@ -0,0 +1,41 @@
...
+
+<count-empty>
+  rdf:type sht:EvalNodeExpr ;
+  rdfs:label "Test of a count expression on an empty list" ;
+  mf:action [
+    sht:nodeExpr [
+      shnex:count [] ;
+    ] ;
+  ] ;
+  mf:result ( 0 ) ;
```
robert-david, line 27: Is a result (list) of 0 correct for a bnode? Shouldn't shnex:count define a return value of ( 0 ) in this case, i.e., 
The [output nodes](https://www.w3.org/TR/shacl12-core/#dfn-output-nodes) of the [count expression](https://raw.githack.com/w3c/data-shapes/issue-491/shacl12-node-expr/index.html#dfn-count-expression) is the list consisting of exactly one xsd:integer [literal](https://www.w3.org/TR/rdf12-concepts/#dfn-literal) that is computed as the length of N, if N is a list; otherwise, the output nodes are ( 0 ).

shacl12-test-suite/tests/node-expr/shnex/offset.ttl
```diff
@@ -0,0 +1,57 @@
...
+  mf:action [
+    sht:nodeExpr [
+      shnex:nodes [
+        shnex:nodes ( 5 6 7 8 ) ;
+        shnex:limit 2 ;
+      ] ;
+      shnex:offset 1 ;
+    ] ;
+  ] ;
+  mf:result ( 6 ) ;
```
robert-david, line 55:
```suggestion
  mf:result ( 6 7 ) ;
```

shacl12-test-suite/tests/node-expr/shnex/var.ttl
robert-david: I agree. I cannot verify if this test case is correct. Does "bound" refer to any bindings in scope?

robert-david (approved):
I approve to not block it further, but please look at my comments. Thanks

shacl12-node-expr/index.html
HolgerKnublauch: Basically c(n) returns one value for each node from the input list. Then these (derived) values are sorted. For example, assume we want to sort IRIs by strlen

```
    sh:values [
        shnex:nodes [ <urn:Medium> <urn:Short> <urn:VeryLong> ] ;
        shnex:orderBy [ sparql:strlen ( shnex:var "focusNode" ) ] ;
    ]
```

then c(n) produces a list like [ 10 9 12 ] where the first item represent urn:Medium etc.

Can anyone suggest how to make this clearer in the definition?

Desc cannot be included in evalExpr because that has a fixed set of parameters. However, desc is used as a post-processing step here, on the resulting list.

shacl12-test-suite/tests/node-expr/shnex/count.ttl
HolgerKnublauch: The value of shnex:count is a node expression. Here, this is the empty node expression, which is represented by a blank node that has no other values. And the count of the empty list is 0.

shacl12-test-suite/tests/node-expr/shnex/offset.ttl
HolgerKnublauch: You need to read these from the inside out:

1) First it will evaluate the shnex:limit 2 on ( 5 6 7 8 ) which returns ( 5 6 )
2) Second it will evaluate the offset of ( 5 6 ) which is ( 6 )

So the test is correct.

shacl12-node-expr/index.html
```diff
@@ -1499,7 +1499,8 @@ <h3>Intersection Expressions</h3>
 							Let <code>members</code> be the <a>members</a> of the <a>value</a> of <code>shnex:intersection</code> in the <a>intersection expression</a>.
 							The <a>output nodes</a> of the <a>intersection expression</a> are the <a>nodes</a> that form the intersection of the <a>output nodes</a>
 							produced by each <a>node expression</a> <code>NE</code> in <code>members</code>, using <code>evalExpr(NE, focusGraph, focusNode, scope)</code>.
-							Nodes must be equal using <a>term equality</a>, i.e. <code>"01"^^xsd:integer</code> is distinct from <code>"1"^^xsd:integer</code>.
+							Nodes must be equal using <a>term equality</a>, e.g., <code>"01"^^xsd:integer</code> is distinct from <code>"1"^^xsd:integer</code>.
+							The intersection does not include duplicates and the order is undefined.
```
TallTed, line 1503:
```suggestion
							The intersection does not include duplicates, and the order is undefined.
```

HolgerKnublauch: @TallTed the PR was already merged and closed. I suggest we do a final read-through of these documents before they are considered frozen and moved to the next stage, to fix commas etc. Creating another PR just for one line is just not scalable.

PR #720, opened 2026-01-13
Update rule grammar

afs:
This PR updates the SRL grammar:

1. Allow a prologue anywhere between rules or data blocks (c.f. SPARQL Update)
   A prologue is `PREFIX`/`BASE`/`VERSION`/`IMPORTS`
3. Remove `BOUND` as a function.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/afs/rules-grammar/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -245,7 +245,7 @@
 
         <p>TODO</p>
 
-        <p class="ednote">RFC 2119 language should automatically be inserted here.</p>
+        <p class="ednote">RFC 2119 language should be automatically inserted here.</p>
 
       </section>
     </section>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -245,7 +245,7 @@
 
         <p>TODO</p>
 
-        <p class="ednote">RFC 2119 language should autmatically be inserted here.</p>
+        <p class="ednote">RFC 2119 language should automatically be inserted here.</p>
 
       </section>
     </section>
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -245,7 +245,7 @@ <h3>Document Conventions</h3>
 
         <p>TODO</p>
 
-        <p class="ednote">RFC 2119 language should autmatically be inserted here.</p>
+        <p class="ednote">RFC 2119 language should automatically be inserted here.</p>
```
```suggestion
        <p class="ednote">RFC 2119 language should be automatically inserted here.</p>
```

recalcitrantsupplant (approved):
updates match comments - looks good to me

simonstey (approved):
looks good to me!

shacl12-rules/index.html
```diff
@@ -952,707 +952,721 @@ <h2>Shapes Rules Language Grammar</h2>
...
+                <td><code>[12]&nbsp;&nbsp;</code></td>
                 <td><code><span class="doc-ref" id="rRule2">Rule2</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
                 <td><code class="gRuleBody"><span class="token">'IF'</span> <a href="#rBodyPattern">BodyPattern</a> <span class="token">'THEN'</span> <a href="#rHeadTemplate">HeadTemplate</a></code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[11]&nbsp;&nbsp;</code></td>
+                <td><code>[13]&nbsp;&nbsp;</code></td>
                 <td><code><span class="doc-ref" id="rRule3">Rule3</span></code></td>
```
liviorobaldo, line 1037: We decided at the TF meeting to drop Rule3, correct? I'm not sure how to propose deletions spanning multiple lines. assuming this is even possible via the GitHub web interface (my understanding from TallTed is that it isn't).

afs: That will be the next PR. We only decided today.

The grammar table is generated and then put in the document between the two `<!-- GRAMMAR -->` markers.

Removing Rule3 will renumber everything following.

shacl12-rules/index.html
afs: FYI @tallted 

This is an `ednote` and will go away before the document is finished. Ditto `@@` markers.

`class="note"` are the "NOTE" blocks for publication.

afs: FYI @Tallted -- this is an "ednote" , not a "note".

`ednote` are editors notes and will go away. (see also `@@`).
`class="note"` are `NOTE` blocks in the final document.

shacl12-rules/index.html
TallTed: > propose deletions spanning multiple lines. assuming this is even possible via the GitHub web interface (my understanding from TallTed is that it isn't).

It is, but you have to switch back to the old `files` view. Click to the "Files changed" tab, and look to the upper right. You will either see "change back" which goes to the old presentation, or "try the new" which goes to the new presentation.

*Both* of these will warn that "some comments can't be seen" when this is so, so you can stay with the old view for most purposes, and only click to the new when it says there are some comments you cannot see, and click back to the old once you've seen the ones you can in the new -- which may not be any that you could't see in the old, because there are bizarre "max" limits on the new.

liviorobaldo: @TallTed, Thanks! ❤️ I tried "Try the new experience" on another GitHub account that I use for testing, and I was able to propose suggestions spanning multiple lines 🙂

TallTed: The "new experience" is a moving target. They've got 2500+ comments on a feedback post, and they're gradually fixing things that commenters have flagged, but there's no clarity of timing, no roadmap, no nothing for people who aren't GitHub-internal.

PR #722, opened 2026-01-14
Rewrite the SHACL Rules introduction, incorporating all the topics we have discussed so far and that we want the spec to cover

liviorobaldo:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/SHACL-rules-introduction/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -142,18 +142,17 @@
         These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
         such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
         to express and evaluate rules that generate new data.</p>
-      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'initial knowledge graph', producing
-        a new graph called the 'inferred knowledge graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF 
-        resources and literals that were not present in the initial knowledge graph. While most rule operators are monotonic, the grammar also includes
-        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether 
-        certain triples are absent from the initial knowledge graph, whereas aggregates generate new literals that can be subsequently added to the 
-        inferred knowledge graph.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
+        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
+        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators: 
+        <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
+        from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
         non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
           be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
         forward chaining, so that for all rules whose bodies match the initial knowledge graph, the content in their heads is added to the inferred 
         knowledge graph, and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
-        the initial knowledge graph using the rules.</p>
+        the base graph using the rules.</p>
       <section id="terminolgy">
         <h3>Terminology</h3>
 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -138,20 +138,22 @@
     <section id="introduction">
       <h2>Introduction</h2>
 
-      <p>
-        This document introduces the concept of inference rules for SHACL 1.2, a mechanism for
-        deriving new RDF triples from existing data using declarative rules defined in shapes
-        graphs. This extends SHACL’s capabilities beyond validation, enabling reasoning and data
-        enrichment.
-      </p>
-
-      <p>
-        This document complements other SHACL 1.2 specifications, such as SHACL Core, by defining
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
+        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'initial knowledge graph', producing
+        a new graph called the 'inferred knowledge graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF 
+        resources and literals that were not present in the initial knowledge graph. While most rule operators are monotonic, the grammar also includes
+        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates (e.g., SUM, COUNT)</i>. Negation as failure allows testing whether 
+        certain triples are absent from the initial knowledge graph, whereas aggregates generate new literals that can be subsequently added to the 
+        inferred knowledge graph.</p>
+      <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
+        non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
+          be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
+        forward chaining, so that for all rules whose bodies match the original knowledge graph, the content in their heads is added to the inferred 
+        knowledge graph, and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
+        the initial knowledge graph using the rules.</p>
       <section id="terminolgy">
         <h3>Terminology</h3>
 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -138,23 +138,16 @@
     <section id="introduction">
       <h2>Introduction</h2>
 
-      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
-        These rules extend SHACL's capabilities by enabling inference. The document complements other SHACL 1.2 specifications, 
-        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
-        to express and evaluate rules that generate new data.</p>
-      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
-        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
-        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators: 
-        <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
-        from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
-      <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
-        non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
-          be applied within implementations.
-          </p>
-          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies the rules so that, for every rule whose body matches the base graph, 
-            the content of its head is added to the inferred graph, and <i>query()</i>, which determines whether a given goal pattern can be inferred from the base graph using 
-            the rules.</p>
-      <section id="terminolgy">
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules. 
+        The document defines the syntax and semantics of rule-based inference.</p>
+      <p>Implementations of SHACL Rules provide two operations. The <i>infer()</i> operation applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
+        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <i>query()</i> operation determines
+        whether a given goal pattern can be derived from the base graph using the rules.</p>
+      <p>The inference graph may contain RDF triples with new blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
+        of rules that generate new blank nodes or literals do not result in infinite loops.</p>
+      <p>SHACL Rules also support constructs, such as negation as failure, that could lead to different inferred graphs depending on the order in which rules are executed. To avoid this, 
+        rules are evaluated using the technique of <i>stratification</i>, which establishes a single, implicit ordering among rules, thus ensuring that the same inference graph is always produced.</p>
+      <section id="terminology">
         <h3>Terminology</h3>
 
         <p class="ednote">Connect to definitions in RDF 1.2 Concepts.</p>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -148,7 +148,7 @@
         <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
         from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
-        non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
+        non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
           be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
         forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
         and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -149,7 +149,9 @@
         from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
         non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
-          be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
+          be applied within implementations.
+          </p>
+          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
         forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
         and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
         the base graph using the rules.</p>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -141,7 +141,7 @@
       <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules. 
         The document defines the syntax and semantics of rule-based inference.</p>
       <p>Implementations of SHACL Rules provide two operations. The <em>infer</em> operation that applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
-        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <i>query()</i> operation determines
+        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <em>query</em> operation determines
         whether a given goal pattern can be derived from the base graph using the rules.</p>
       <p>The inference graph may contain RDF triples with new blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
         of rules that generate new blank nodes or literals do not result in infinite loops.</p>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -151,7 +151,7 @@
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
         non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
           be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
-        forward chaining, so that for all rules whose bodies match the original knowledge graph, the content in their heads is added to the inferred 
+        forward chaining, so that for all rules whose bodies match the initial knowledge graph, the content in their heads is added to the inferred 
         knowledge graph, and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
         the initial knowledge graph using the rules.</p>
       <section id="terminolgy">
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -145,7 +145,7 @@
       <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'initial knowledge graph', producing
         a new graph called the 'inferred knowledge graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF 
         resources and literals that were not present in the initial knowledge graph. While most rule operators are monotonic, the grammar also includes
-        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates (e.g., `SUM`, `COUNT`)</i>. Negation as failure allows testing whether 
+        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether 
         certain triples are absent from the initial knowledge graph, whereas aggregates generate new literals that can be subsequently added to the 
         inferred knowledge graph.</p>
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -139,7 +139,7 @@
       <h2>Introduction</h2>
 
       <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
-        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
+        These rules extend SHACL's capabilities by enabling inference. The document complements other SHACL 1.2 specifications, 
         such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
         to express and evaluate rules that generate new data.</p>
       <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -145,7 +145,7 @@
       <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'initial knowledge graph', producing
         a new graph called the 'inferred knowledge graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF 
         resources and literals that were not present in the initial knowledge graph. While most rule operators are monotonic, the grammar also includes
-        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates (e.g., SUM, COUNT)</i>. Negation as failure allows testing whether 
+        two non-monotonic operators: <i>negation as failure</i> and <i>aggregates (e.g., `SUM`, `COUNT`)</i>. Negation as failure allows testing whether 
         certain triples are absent from the initial knowledge graph, whereas aggregates generate new literals that can be subsequently added to the 
         inferred knowledge graph.</p>
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -150,8 +150,8 @@
       <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
         non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
           be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
-        forward chaining, so that for all rules whose bodies match the initial knowledge graph, the content in their heads is added to the inferred 
-        knowledge graph, and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
+        forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
+        and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
         the base graph using the rules.</p>
       <section id="terminolgy">
         <h3>Terminology</h3>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -143,7 +143,7 @@
       <p>Implementations of SHACL Rules provide two operations. The <em>infer</em> operation that applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
         containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <em>query</em> operation determines
         whether a given goal pattern can be derived from the base graph using the rules.</p>
-      <p>The inference graph may contain RDF triples with new blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
+      <p>The inference graph may contain RDF triples with IRIs, blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
         of rules that generate new blank nodes or literals do not result in infinite loops.</p>
       <p>SHACL Rules also support constructs, such as negation as failure, that could lead to different inferred graphs depending on the order in which rules are executed. To avoid this, 
         rules are evaluated using the technique of <i>stratification</i>, which establishes a single, implicit ordering among rules, thus ensuring that the same inference graph is always produced.</p>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -163,7 +163,7 @@
         that could lead to different inferred graphs depending on the order in which
         rules are executed. To avoid this, rules are evaluated using the technique
         of <em>stratification</em>, which establishes a single, implicit ordering
-        among rules, thus ensuring that the same inference graph is always produced.
+        among rules, ensuring that the same inference graph is always produced.
       </p>
       <section id="terminology">
         <h3>Terminology</h3>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -138,15 +138,33 @@
     <section id="introduction">
       <h2>Introduction</h2>
 
-      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules. 
-        The document defines the syntax and semantics of rule-based inference.</p>
-      <p>Implementations of SHACL Rules provide two operations. The <em>infer</em> operation that applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
-        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <em>query</em> operation determines
-        whether a given goal pattern can be derived from the base graph using the rules.</p>
-      <p>The inference graph may contain RDF triples with IRIs, blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
-        of rules that generate new blank nodes or literals do not result in infinite loops.</p>
-      <p>SHACL Rules also support constructs, such as negation as failure, that could lead to different inferred graphs depending on the order in which rules are executed. To avoid this, 
-        rules are evaluated using the technique of <i>stratification</i>, which establishes a single, implicit ordering among rules, thus ensuring that the same inference graph is always produced.</p>
+      <p>
+        This document introduces inference rules for SHACL 1.2, a mechanism for
+        deriving new RDF triples from existing RDF data through declarative rules. 
+        The document defines the syntax and semantics of rule-based inference.
+      </p>
+      <p>
+        Implementations of SHACL Rules provide two operations.
+        The <em>infer</em> operation that applies the rules to a given
+        <i>base graph</i> and produces an <i>inference graph</i> containing the
+        RDF triples derived by rule execution.
+        Combining the inference graph with the base graph is optional and
+        left to users. The <em>query</em> operation determines whether a
+        given goal pattern can be derived from the base graph using the rules
+      </p>
+      <p>
+        The inference graph may contain RDF triples with IRIs, blank nodes or
+        literals that were not present in the base graph. Users are responsible to
+        ensure that iterative applications of rules that generate new blank nodes or
+        literals do not result in infinite loops.
+      </p>
+      <p>
+        SHACL Rules also support constructs, such as negation as failure, 
+        that could lead to different inferred graphs depending on the order in which
+        rules are executed. To avoid this, rules are evaluated using the technique
+        of <em>stratification</em>, which establishes a single, implicit ordering
+        among rules, thus ensuring that the same inference graph is always produced.
+      </p>
       <section id="terminology">
         <h3>Terminology</h3>
 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -151,10 +151,9 @@
         non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
           be applied within implementations.
           </p>
-          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
-        forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
-        and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
-        the base graph using the rules.</p>
+          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies the rules so that, for every rule whose body matches the base graph, 
+            the content of its head is added to the inferred graph, and <i>query()</i>, which determines whether a given goal pattern can be inferred from the base graph using 
+            the rules.</p>
       <section id="terminolgy">
         <h3>Terminology</h3>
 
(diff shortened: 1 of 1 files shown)


liviorobaldo:
My first PR ☺️

Some comments:

- I wrote “for deriving new content” rather than “for deriving new RDF triples” because rules can also add new RDF resources and RDF literals.

- I removed “beyond validation” because, as discussed in the last PR for SHACL Core, we should avoid over-emphasising validation and instead present it as one task among others. The same applies to the phrase “While SHACL Core focuses on constraint validation”.

- Regarding “SHACL Rules only <i>add</i> content to the knowledge graph”, rules do not <i>remove</i> content. See the [discussion here](https://github.com/w3c/data-shapes/discussions/603#discussioncomment-14798786): allowing rules to <i>remove</i> content, although potentially useful, could introduce several issues, particularly with respect to guaranteeing termination of rule execution. For this reason, I think we should avoid this.

- For the discussion on <i>infer()</i> and <i>query()</i>, see [discussion here](https://github.com/w3c/data-shapes/discussions/637).

- Concerning the statement “Well-formedness conditions in SHACL Rules prevent infinite loops”, see [discussion here ](https://github.com/w3c/data-shapes/issues/643#issuecomment-3560383490). However, I am not sure whether stratification also plays a role in preventing infinite loops; if so, we could consider stating this explicitly in the abstract as well.

afs:
> “Well-formedness conditions in SHACL Rules prevent infinite loops"

The well-formedness conditions are not about loops. They are "set variable before use" where use is needing the variable setting - a filter or a head - or for stopping re-assignment.
https://raw.githack.com/w3c/data-shapes/SHACL-rules-introduction/shacl12-rules/index.html#wellformed

Infinite loops: There is an open issue here related to term introduction where stratification *may* be part of the solution.

liviorobaldo:
Hi Andy, okay. My understanding is that this should still be implemented as part of the Rule Set Evaluation (§5). Perhaps the abstract should not go into these details, but I think it is important to at least state that correct implementations of SHACL Rules prevent infinite loops and non-deterministic behaviour.

By the way, I have just noticed that what I called the "initial knowledge graph" in the current version of the abstract is referred to as the "base graph" in §5. I am going to update index.html in this branch to keep the terminology consistent with §5.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
+        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph
```
afs: Rules produce an inference graph - they don't add to the input graph. You say this later but this text contradicts that.

A user may choose to add the inferences but that is not a requirement of SHACL Rules.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
-
-      <p>
-        This document complements other SHACL 1.2 specifications, such as SHACL Core, by defining
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
+        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications,
```
afs:
```suggestion
        These rules extend SHACL's capabilities by enabling inference. The document complements other SHACL 1.2 specifications, 
```
or

```suggestion
        These rules extend SHACL's capabilities by enabling data enrichment. The document complements other SHACL 1.2 specifications, 
```
"reasoning" is part of data enrichment.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
-        evaluate rules that generate new data.
-      </p>
-
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
+        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
+        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
+        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators:
```
afs: I don't think an introduction needs to say "monotonic" which has a very precise meaning.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
+        These rules extend SHACL's capabilities by enabling reasoning and data enrichment. The document complements other SHACL 1.2 specifications, 
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
+        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
+        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators: 
+        <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
+        from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
```
afs: People are reading the working draft. This is good!

There is still significant work time needed just to get the specification to one that covers negation, which IMO is more important. 

While it is unfortunate not to go further, a worse outcome is to rush and put something in which turns out to be wrong or not the bast design. There is opportunity to add features later; removing features is very painful for the spec and for users.

So I'm uncomfortable with putting anything about aggregation in the introduction at this time.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
+      <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
+        non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
+          be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using
```
afs:
```suggestion
          be applied within implementations.
          </p>
          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
```

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
+        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
+        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators: 
+        <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
+        from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
+      <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while
```
afs: This first part of this paragraph is getting into more detail than an introduction needs to.

To me, an introduction sets the scene - the common use. Error conditions etc are details for later.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph 
+        called the 'inferred graph'. In particular, rules may add new RDF triples but they may also <i>create</i> and add new RDF resources and literals 
+        that were not present in the base graph. While most rule operators are monotonic, the grammar also includes two non-monotonic operators: 
+        <i>negation as failure</i> and <i>aggregates</i> (e.g., `SUM`, `COUNT`). Negation as failure allows testing whether certain triples are absent 
+        from the base graph, whereas aggregates generate new literals that can be subsequently added to the inferred graph.</p>
+      <p>Well-formedness conditions in SHACL Rules prevent infinite loops that could arise from the creation of new RDF resources and literals, while 
+        non-deterministic behavior introduced by non-monotonic operators is avoided through the well-known technique of <i>stratification</i>, which must
```
afs:
```suggestion
        non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
```

shacl12-rules/index.html
afs:
```suggestion
          be applied within implementations.
          </p>
          <p>Implementations of the rules provide two possible operations: <i>infer()</i>, which applies rules using 
```
The "Implementations .. provide" text is important for an introduction - move it up.

Avoid, where possible, may/should/must language because, even if written as plain text, they suggest RFC 2119 meaning. Sometimes it is natural to use the words (lowercase) but if possible, reword.

shacl12-rules/index.html
```diff
@@ -138,20 +138,21 @@
...
+          be applied within implementations. Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
+        forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
+        and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from
```
afs: I don't think the introduction needs to mention "backward-chaining"/"forward-chaining". They are implementation strategies.

Text later can get into the details and say either can be used - and the spec defines outcomes using one possible forward-chaining approach.

afs:
> Perhaps the abstract should not go into these details, but I think it is important to at least state that correct implementations of SHACL Rules prevent infinite loops and non-deterministic behaviour.

Introduction - the abstract is a more general and shorter section in the standard W3C spec template.

I think the introduction is a scene-setting general outline. It is for the first-time reader. It does not have to have every detail.

The rest of section 2, the description/tutorial, can cover items at a deeper level. Mentioning deep details in the introduction does not do the details justice nor is the first-time reader yet in a position to pick up on these details.

afs (changes requested):

liviorobaldo:
Yes, sorry, here we are dealing with the "introduction", not the "abstract". Every time I wrote "abstract" above I meant "introduction". Totally agree about "I think the introduction is a scene-setting general outline. It is for the first-time reader". I will have a look at all your revisions asap.

shacl12-rules/index.html
liviorobaldo: I think it is important to make clear that the inferred graph includes all triples from the base graph, i.e., that these triples are preserved and not removed (see [issue #728](https://github.com/w3c/data-shapes/issues/728)).

Other readers might wonder whether SHACL Rules only add triples to the base graph or whether they can also remove them.

shacl12-rules/index.html
liviorobaldo: I think we should instead mention monotonic vs non-monotonic operators in the Introduction. The intended audience for SHACL Rules consists of people with a basic background in logic, for whom the term "monotone" is already part of the standard vocabulary. In other words, this term is not "too technical" for them. 

Readers approaching this document will naturally start wondering how SHACL Rules differ from OWL, SWRL, N3, RuleML, and similar formalisms. In light of this, in my view readers should be informed from the outset that SHACL Rules allow for NAF, aggregates, and the creation of new RDF resources and literals, while controlling their effects, e.g., through stratification.

shacl12-rules/index.html
liviorobaldo: Although I agree that NAF is _more important_ than aggregates (in the sense that there are many more use cases requiring NAF than those requiring aggregates) I still think that both are _necessary_. 

I was planning to open a new PR on aggregates, including them in the grammar and describing them in the relevant section (they are currently mentioned in §3). Let’s discuss negation and aggregates further in the forthcoming TF meetings.

shacl12-rules/index.html
liviorobaldo: Fair enough. Instead of explicitly mentioning well-formedness conditions, stratification, and similar technical details, we could just state more generally that: "implementations are expected to prevent infinite loops arising from the creation of new RDF resources and literals, as well as non-deterministic behaviour introduced by non-monotonic operators (see §3)."

However, as I mentioned above, I believe it is important to make clear from the outset that non-monotonic operators are part of the design and that their potential "collateral effects" have been considered.

shacl12-rules/index.html
liviorobaldo: I agree. Unless anyone else disagrees (let’s wait a couple of days for comments), I’ll update the text accordingly.

shacl12-rules/index.html
afs: It can not remove them.

shacl12-rules/index.html
afs: What do you mean by "monotonic" in this context?

shacl12-rules/index.html
afs: > I still think that both are necessary.

Why? One specualtive use case is not enough to justify choosing an approach.

We need to scope the possibilities (there are several) and survey the state of the art. Until that ground work is done, any changes are at significant risk of being wrong and have to be removed.

For a standard, a document that should be stable for years, removing or materially changing a feature is a massive problem.

New features, as they are proven, can come along later if they are also new syntax as this would be.

Do the core feature well - and that is a lot more work than just paragraphs in a document.

shacl12-rules/index.html
afs: > Readers approaching this document will naturally start wondering how SHACL Rules differ from OWL, SWRL, N3, RuleML, and similar formalisms.

Would you care to draft a comparison?

This does not need go in the spec itself (they become out of date or incomplete). 

It is good material for the "What's new in SHACL 1.2" document.

shacl12-rules/index.html
afs: > implementations are expected to prevent infinite loops arising

If you leave it to implementations, there will be inconsistency on the web.

shacl12-rules/index.html
liviorobaldo: > What do you mean by "monotonic" in this context?

By (non-)monotonic, I mean whether previously derived conclusions can be retracted. What is retracted may depend on the execution order of the rules. This corresponds to the standard and well-known definition. A canonical example is:

Bird(x) → CanFly(x)
Penguin(x) → ¬CanFly(x)
Bird(x) → Penguin(x)

If Tweety is a bird, under one execution order it may first be inferred that Tweety can fly and only later that it cannot (so the first conclusion is retracted); under another execution order, the opposite occurs.

That said, this notion is quite standard and should be familiar to anyone with a background in logic... including you. So perhaps I have misunderstood your question?

My point is that, since the term "monotonic" is well known to the potential readership of SHACL rules, I do not consider it overly specific or too technical for the introduction. This was in response to your initial claim:

> I don't think an introduction needs to say "monotonic", which has a very precise meaning.

To answer your question:

> Would you care to draft a comparison?

Yes, I think comparisons with existing solutions are of clear interest. We discussed this ourselves during the TF meeting, for example comparisons with N3. Indeed, the very first question a colleague of mine asked was what is new in SHACL rules compared to RuleML.

I fully agree that these discussions should be incorporated into the "What’s New in SHACL 1.2" document. However, I still think it is useful to mention (non-)monotonicity and stratification in the introduction, as these refer to general categories and can help clarify the direction of our work from the outset. The "What’s New in SHACL 1.2" document can then provide the detailed comparisons.

shacl12-rules/index.html
liviorobaldo: Ok, I totally agree that we need to scope the possibilities. Let’s discuss this at the TF meetings. My understanding from [the current version of the spec](https://w3c.github.io/data-shapes/shacl12-rules/) was that NAF, aggregates, and the creation of new RDF resources and literals were included in the scope.

shacl12-rules/index.html
liviorobaldo: Ok, that's also a good point. Let’s then avoid to say “are expected”. What do you (and everyone else) think of the following instead:

"The evaluation of the rules prevents infinite loops arising from the creation of new RDF resources and literals, as well as non-deterministic behaviour introduced by non-monotonic operators (see §5)."

recalcitrantsupplant:
As a general comment I also think there is too much detail for an introduction, but the content could go elsewhere. Paraphrasing the addition around base/inference graphs we could add something like "SHACL Rules are evaluated over a base graph to produce an inference graph."

shacl12-rules/index.html
```diff
@@ -151,10 +151,9 @@ <h2>Introduction</h2>
         non-deterministic behavior introduced by non-monotonic operators is avoided through the technique of <i>stratification</i>, which must
           be applied within implementations.
           </p>
-          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies rules using 
-        forward chaining, so that for all rules whose bodies match the base graph, the content in their heads is added to the inferred graph, 
-        and <i>query()</i>, which applies rules using backward chaining, determining whether a given goal pattern can be inferred from 
-        the base graph using the rules.</p>
+          <p>Implementations of the rules must also support two operations: <i>infer()</i>, which applies the rules so that, for every rule whose body matches the base graph, 
+            the content of its head is added to the inferred graph, and <i>query()</i>, which determines whether a given goal pattern can be inferred from the base graph using 
+            the rules.</p>
```
TallTed: It's helpful to maintain line lengths similar to those already in the source. RFC2119 terminology SHOULD be avoided, even in lowercase, unless intended to be interpreted as in RFC2119. If this text is non-normative, then the MUST below SHOULD be deleted (changing `rules MUST also` to `rules also`)
```suggestion
          <p>Implementations of the rules MUST also support two operations: `infer()`, which applies the rules so that, for every rule
            whose body matches the base graph, the content of its head is added to the inferred graph; and `query()`, which
            tests whether a given goal pattern can be inferred by applying the rules to the base graph.</p>
```

shacl12-rules/index.html
TallTed: > The intended audience for SHACL Rules consists of people with a basic background in logic, for whom the term "monotone" is already part of the standard vocabulary.

I imagine myself to be among the intended audience, yet I have no "basic background in logic", nor is "the term 'monotone' ... already part of [my] standard vocabulary".

At minimum, we should provide a basic definition of such terms, especially but not only where we expect such basic definition to influence the reader's understanding of the text that follows.

afs: > Bird(x) → CanFly(x)
> Penguin(x) → ¬CanFly(x)
> Bird(x) → Penguin(x)
> 
> If Tweety is a bird, under one execution order it may first be inferred that Tweety can fly and only later that it cannot (so the first conclusion is retracted); under another execution order, the opposite occurs.
> 
> That said, this notion is quite standard and should be familiar to anyone with a background in logic... including you.

What would they look like with data in RDF and with SHACL Rules? (specifically, how to model "can fly")

liviorobaldo: You need to model _strong_ negation in RDF/SPARQL, which is not natively supported. SPARQL only provides negation-as-failure, namely `NOT EXISTS`.

In [this paper of mine](https://academic.oup.com/logcom/article/35/8/exaf054/8320660), I propose using RDF reification for this purpose. Until now, RDF reification has mainly been used to timestamp triples or to attribute authorship. In that paper, I pioneer the use of RDF reification to assign truth values to triples (to the best of my knowledge, this had not been done before). For example, `¬CanFly(Tweety)` is represented as:

```<<:Tweety a :CanFly>> a :false.```

One can also state:

```<<:Tweety a :CanFly>> a :true.```

which corresponds to `CanFly(Tweety)`. However, this is redundant, since truth is already expressible in RDF via the presence of the triple itself, i.e., `:Tweety a :CanFly`.

Of course, a rule is also required stating that if a triple is false, it cannot also be true; if both occur in the graph, an inconsistency must be derived.

---------

On a separate note, it is clear to me that the majority considers the Introduction proposed in this PR to be too specific. Fair enough, I bow to the majority 😅. I will re-read all the comments and rewrite the Introduction accordingly. I plan to do this after the next TF meeting, so if anyone has further comments in the meantime, I will be happy to take them into account as well.

Comment (afs) [reply shacl12-rules/index.html]: > using RDF reification for this purpose.

You can publish statements about claims as a modelling layer on top of core RDF (RDF 1.2, and original RDF reification).

Comment (liviorobaldo) [reply shacl12-rules/index.html]: Yes, that's exacly what I'm doing, to represent falsity but also disjunction (v), possibility/necessity, etc. But maybe I didn't understand you comment?

Comment (afs) [reply shacl12-rules/index.html]: > You need to model strong negation in RDF/SPARQL

> But maybe I didn't understand you comment?

You have only just mentioned reification. The introduction to SHACL Rules isn't concerned with domain specialization.

---

You can not add strong negation to the semantics of RDF. You can publish data that, in the domain model, expresses that.

```
<<:Tweety a :CanFly>> a :false.
```
isn't logically false in RDF semantics. RDF only has "asserted" and "not asserted". `:false` is just a class name - a sequence of letters that denotes a resource.

https://www.w3.org/Talks/2025/iswc-tutorial-rdfsparql-12/slides/2-foundations.pdf

(and the RDF&SPARQL WG is likely to write one or two guidance documents (time permitting).)

(I'm only talking about RDF 1.2 - as RDF-star, either the form in Olaf Hartig's paper or the CG  - there is an additional problem identified and addressed by RDF 1.2.)

If you write the SHACL rules, you'll get something unexpected. Either stratification violation, or you get a domain model contradiction where the RDF is happy but says something you didn't want in the domain model.

Try writing the RDF and the rules.

Comment (liviorobaldo) [reply shacl12-rules/index.html]: > You cannot add strong negation to the semantics of RDF.

I did not mean to do that. I am _modelling_ strong negation at the meta-level in my research paper. The same applies to the following remark:

> The introduction to SHACL Rules isn't concerned with domain specialization.

I have no intention of mentioning any content of my paper in the introduction.

> Try writing the RDF and the rules.

The paper I linked above is associated with [this GitHub repository](https://github.com/liviorobaldo/conflict-tolerantDeonticTraditionalScheme), where the RDF resources and the rules are available. However, this is a research work that proposes a way to model strong negation in RDF, since, as you correctly point out, this is not natively supported by RDF semantics. I referred to it only to answer your earlier question about how I represent `¬CanFly(x)`.

Perhaps I should have chosen a different example without the ¬ symbol to illustrate non-monotonicity (see the beginning of this discussion), for instance:

```
Bird(x) → Yellow(x)
Penguin(x) → Red(x)
Bird(x) → Penguin(x)
```

where `Yellow` and `Red` are disjoint.

I think we are mixing too many different topics together; perhaps we should just drop this discussion 😅

Comment (afs) [reply shacl12-rules/index.html]: I had looked at the repo - it uses Apache Jena 4.10.0 -- 2023-10-24.
Apache Jena 4.10.0 has limited support for RDF-star (original version) whereas Jena 6.0 has RDF 1.2 support.

Comment (liviorobaldo) [reply shacl12-rules/index.html]: Yes, I know, I started that work in 2023. Now I'm working on new papers and I'm using the latest version of Apache Jena... and I'm looking forward for the versions that will support SHACL 1.2 ☺️

File: shacl12-rules/index.html [hunk 10]
```diff
@@ -138,20 +138,22 @@
...
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new content from existing data through declarative rules. 
+        These rules extend SHACL's capabilities by enabling inference. The document complements other SHACL 1.2 specifications, 
+        such as SHACL Core, by defining the syntax and semantics of rule-based inference. The SHACL Rules specification provides a standardized way 
+        to express and evaluate rules that generate new data.</p>
+      <p>SHACL Rules only <i>add</i> content to the knowledge graph to which they are applied, referred to as the 'base graph', producing a new graph
```
robert-david: 'content' sounds weird for me like some unstructured text.
I think RDF triples fits well, because this is precisely what rules do. Maybe you could expand it with:
... RDF triples, with potential fresh nodes, ...

shacl12-rules/index.html
robert-david: You would understand the meaning of monotonic if you have a background in logic, which users of SHACL not necessarily have. I cannot think of a better term, though, but i wonder if we really need to explain monotonic and non-monotonic. I think it becomes clear in the definition later how the rules work.
What might be worth mentioning is that the reasoning will return one inference graph only.

shacl12-rules/index.html
robert-david: The evaluation always terminates with (exactly) one inference graph?

robert-david (changes requested):

shacl12-rules/index.html
liviorobaldo: That's indeed a good suggestion. Instead of "monotonic" we can simply say that "the rules and evaluation algorithm are designed so that a single inference graph is always returned, regardless of the order in which the rules are executed" or similar.

As I told Andy, I'll prepare a new version of the introduction after the next TF meeting.

TallTed: @liviorobaldo — Note that you appear to be [misusing `a`](https://github.com/w3c/data-shapes/pull/722#discussion_r2758594055), which is built into Turtle and defined as syntactic sugar for `rdf:type`. Your sample statements (which I presume to be in Turtle) thus expand to the following —
```turtle
<< :Tweety rdf:type :CanFly >> rdf:type :false .

<< :Tweety rdf:type :CanFly >> rdf:type :true .
```
— which might be considered useful in some universe, but I think are not what you intended them to say.

liviorobaldo: Uh? I thought that `rdf:type` and `a` were interchangeable. Are they not?

TallTed: > I thought that `rdf:type` and `a` were interchangeable

Indeed, *in Turtle*, they are. `a` is not a defined shortcut in any other serialization.

It did not appear to me that you intended to imply 
```
:false  a rdfs:Class . 
:true   a rdfs:Class . 
:CanFly a rdfs:Class .
```
I now gather, that was your intention. It doesn't seem terribly useful to me, but perhaps there is a use case for it.

liviorobaldo: > It doesn't seem terribly useful to me, but perhaps there is a use case for it.

`:true a rdfs:Class.` is redundant, since every asserted triple is true. However, it can still make sense to talk about the truth of a triple even when the triple itself is not asserted. This is the case, for example, when we need to represent a (negated) disjunction, such as:

`¬(A(x) ∨ ¬B(x))`

In my paper, I represent this as follows:

```
<< <<<<:x a :A>> a :true>> :disjunction <<<<:x a :B>> a :false>> >> a :false.
```


In this example, we do not know whether the triple `:x a :A` actually holds in the knowledge graph. The reification of the triple `<<:x a :A>> a :true` appears inside a disjunction, i.e., as the first argument of the predicate `:disjunction`.

This approach can be useful whenever one wants to represent strong negation or disjunction, which are not natively available in RDF. Whether this is _terribly_ useful in practice or mostly of interest to true nerds… that, I don’t know 😂

shacl12-rules/index.html
TallTed: > it is important to make clear that the inferred graph includes all triples from the base graph

Isn't that inaccurate? The inferred graph includes only the triples that are inferred from the base graph, and the base graph remains as it was to start.

The user (or a rule they set) can combine the inferred graph with the base graph, by using graph marge, or named graphs, or other methods, but this is entirely optional!

liviorobaldo: You are right. According to the new definitions, the union of the base graph and the inferred knowledge graph is called the evaluation graph. I don't recall whether the comment above was made before we agreed on these definitions, though.

shacl12-rules/index.html
liviorobaldo: @robert-david: Yes, that is also what you observed above:

> What might be worth mentioning is that the reasoning will return only one inference graph.

Stratification establishes a single, implicit ordering among rules, thereby ensuring that the same inference graph is always returned.

liviorobaldo:
Dear SHACL rules, 

I have created a new version of the introduction trying to incorporate all comments above.

Please have a look
Cheers
Livio

shacl12-rules/index.html
```diff
@@ -138,21 +138,16 @@
...
-
-      <p>
-        This document complements other SHACL 1.2 specifications, such as SHACL Core, by defining
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
-      <section id="terminolgy">
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules.
```
afs: Could we have short-ish lines please, like the rest of the document?  Say, upto 85-ish characters.

It makes editing easier, depending on the editing tools used. Github reviewing in particular is better with no line wrap and change suggestions can be more focused.

liviorobaldo: Sure, I'll try to remember for the next times.

shacl12-rules/index.html
```diff
@@ -138,21 +138,16 @@
...
-        This document complements other SHACL 1.2 specifications, such as SHACL Core, by defining
-        the syntax and semantics of rule-based inference. While SHACL Core focuses on constraint
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
-      <section id="terminolgy">
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules. 
+        The document defines the syntax and semantics of rule-based inference.</p>
+      <p>Implementations of SHACL Rules provide two operations. The <i>infer()</i> operation applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i>
```
afs:
```suggestion
      <p>Implementations of SHACL Rules provide two operations. The <em>infer</em> operation that applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
```

shacl12-rules/index.html
```diff
@@ -138,21 +138,16 @@
...
-        validation, the SHACL Rules specification provides a standardized way to express and
-        evaluate rules that generate new data.
-      </p>
-
-      <section id="terminolgy">
+      <p>This document introduces inference rules for SHACL 1.2, a mechanism for deriving new RDF triples from existing RDF data through declarative rules. 
+        The document defines the syntax and semantics of rule-based inference.</p>
+      <p>Implementations of SHACL Rules provide two operations. The <i>infer()</i> operation applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
+        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <i>query()</i> operation determines
+        whether a given goal pattern can be derived from the base graph using the rules.</p>
```
afs:
```suggestion
        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <em>query</em> operation determines
        whether a given goal pattern can be derived from the base graph using the rules.</p>
```

shacl12-rules/index.html
```diff
@@ -138,21 +138,16 @@
...
+      <p>The inference graph may contain RDF triples with new blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications
```
afs:
```suggestion
      <p>The inference graph may contain RDF triples with IRIs, blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
```

afs (changes requested):

afs:
Other than the editorial request changes, this is good to go.

We can come back to it when we have a whole document. Some of the detail may be better at the start of "outline" when we have that.

liviorobaldo:
Ok, should I click “Merge pull request”? I cannot do so, however, because it says “2 reviews requesting changes by reviewers with write access.”

image

I believe the reviewers are @afs and @robert-david, so I have sent them another request for review.

Sorry if I do not know what to do at this stage... as I mentioned above, this was my first PR 😅

Thanks for guidance

shacl12-rules/index.html
```diff
@@ -138,21 +138,16 @@
...
+      <p>Implementations of SHACL Rules provide two operations. The <em>infer</em> operation that applies the rules to a given <i>base graph</i> and produces an <i>inference graph</i> 
+        containing the RDF triples derived by rule execution. Combining the inference graph with the base graph is optional and left to users. The <em>query</em> operation determines
+        whether a given goal pattern can be derived from the base graph using the rules.</p>
+      <p>The inference graph may contain RDF triples with IRIs, blank nodes or literals that were not present in the base graph. Users are responsible to ensure that iterative applications 
+        of rules that generate new blank nodes or literals do not result in infinite loops.</p>
+      <p>SHACL Rules also support constructs, such as negation as failure, that could lead to different inferred graphs depending on the order in which rules are executed. To avoid this, 
+        rules are evaluated using the technique of <i>stratification</i>, which establishes a single, implicit ordering among rules, thus ensuring that the same inference graph is always produced.</p>
```
afs:
```suggestion
      <p>
        This document introduces inference rules for SHACL 1.2, a mechanism for
        deriving new RDF triples from existing RDF data through declarative rules. 
        The document defines the syntax and semantics of rule-based inference.
      </p>
      <p>
        Implementations of SHACL Rules provide two operations.
        The <em>infer</em> operation that applies the rules to a given
        <i>base graph</i> and produces an <i>inference graph</i> containing the
        RDF triples derived by rule execution.
        Combining the inference graph with the base graph is optional and
        left to users. The <em>query</em> operation determines whether a
        given goal pattern can be derived from the base graph using the rules
      </p>
      <p>
        The inference graph may contain RDF triples with IRIs, blank nodes or
        literals that were not present in the base graph. Users are responsible to
        ensure that iterative applications of rules that generate new blank nodes or
        literals do not result in infinite loops.
      </p>
      <p>
        SHACL Rules also support constructs, such as negation as failure, 
        that could lead to different inferred graphs depending on the order in which
        rules are executed. To avoid this, rules are evaluated using the technique
        of <em>stratification</em>, which establishes a single, implicit ordering
        among rules, thus ensuring that the same inference graph is always produced.
      </p>
```
Use shorter lines for the HTML.

afs:
Requesting review was the right thing to do.

While this WG does not (AFAIK) have formal policy other than 1+ approvals, it is usually editors who do the merge; they are responsible for process.

I just put in a change request that shortens the long lines (which better for later).
Please handle that and then I will approve the PR.

There will be 15 on this commits, most of which are "Update index.html" which isn't very helpful.
I'll sort out the commits by squashing the PR down to one commit, then merge it.

Merge to the default branch is permanent. Documents can be edited but commits are never removed - it would cause chaos, and, potentially lost work, for everyone especially on a multiple document repo like `data-shapes`.

liviorobaldo:
Hi Andy, I've accepted your suggestion. I can see that "Squash and merge" is one of the available options, but the button is still disabled, presumably because, as you mentioned, I'm not an editor. If you're able to do this on my behalf, we can close the PR. Thanks.

afs (approved):

afs:
> but the button is still disabled, 

because there is an outstanding request for change. 

image

shacl12-rules/index.html
```diff
@@ -139,20 +139,33 @@
...
+        literals that were not present in the base graph. Users are responsible to
+        ensure that iterative applications of rules that generate new blank nodes or
+        literals do not result in infinite loops.
+      </p>
+      <p>
+        SHACL Rules also support constructs, such as negation as failure, 
+        that could lead to different inferred graphs depending on the order in which
+        rules are executed. To avoid this, rules are evaluated using the technique
+        of <em>stratification</em>, which establishes a single, implicit ordering
+        among rules, thus ensuring that the same inference graph is always produced.
```
```suggestion
        among rules, ensuring that the same inference graph is always produced.
```

recalcitrantsupplant (approved):
Looks good - just one suggestion I don't think the word "thus" is needed here

shacl12-rules/index.html
```diff
@@ -163,7 +163,7 @@ <h2>Introduction</h2>
         that could lead to different inferred graphs depending on the order in which
         rules are executed. To avoid this, rules are evaluated using the technique
         of <em>stratification</em>, which establishes a single, implicit ordering
```
TallTed, line 165: It just occurred to me that this stratification is analogous to the `PEMDAS` (Parentheses, Exponents, Multiplication/Division, Addition/Subtraction) order of mathematical operations, and this might be helpful in making this explanation clearer for readers.

However, mathematics also includes a number of Properties (including commutative (`a+b=b+a`), associative (`a+(b+c)=(a+b)+c`), distributive (`a(b+c)=ab+ac`), identity (`a+0=a,a·1=a`), and inverse) which do not seem to flow cleanly nor fully into SHACL nor RDF. I hope I'm just having a slow brain day, and someone can quickly and easily explain the error of my ways!

All that said... I still do not see how a "single, implicit ordering" is established, and I thus don't see how "the same inference graph is always produced." Please help!

liviorobaldo: Hi @TallTed,

sorry, I can't see the parallel between stratification and mathematical operations 🙁.

I’m happy to help, but I’m not sure the issue is entirely clear, specifically the non-determinism caused by negation-as-failure, aggregates, and other non-monotonic operators. This has already been discussed above. When these operators are used in if–then rules, different outcomes may arise depending on the order in which the rules are executed.

Stratification is simply a technique that ensures rules containing non-monotonic operators (currently only negation-as-failure in SHACL rules, at least for now) are executed _after_ (some of) the rules without such operators, the ones  on which the former depend. This is why an algorithm is being defined to build the dependency graph.

In short, stratification establishes a single execution order among the rules, thereby preventing non-determinism: the same unique result is always produced.

As mentioned, I still do not see the parallel with mathematical operators.

I hope this helps

robert-david (approved):

afs:
(Long comment because we can seed section 2.3 with the discussion.)

Datalog evaluation is only adding inferences, never taking them away. Datalog does not have "asserted negation" (a fact known to be false), only "negation by fact not found".

To test for absence, the pattern in the `NOT` is evaluated and then it matches or it doesn't. We need know that the inputs to the matching will not change otherwise "`NOT` true" (i.e. pattern not found) might become "`NOT` false" (pattern is now found and hence inferred triples are not generated).

The NAF requirement of `NOT` pattern elements is that they only depend on triples generated in lower layers or from the base graph.

The stratification we give is that level 0 is rules that have no negation or have a `NOT` that depends only on the base graph (the EDB in the datalog jargon) - they can be executed first.

Then level 1 is rules that have a `NOT` and the `NOT` pattern depends on level 0 (and maybe also the base graph).

Level 0 is executed completely, then level 1 can be, because `NOT`element won't go true -> false.

Level 2 is rules that have a `NOT` that depends on level 1 generated triples.

The _stratification condition_ forbids a `NOT` path to depend on itself - that is, a cycle in the dependency graph with a `NOT` in the cycle. Rules in such a cycle can't be partitioned.

Stratification puts the rules into layers. Evaluation has two parts. An outer loop which executes each layer, and an inner loop to execute the subset of rules in the layer as a conjunction, possibly with recursion but only with positive dependencies in the recursion) from bottom (level 0) to top. There is no specific order of execution in the inner loop; it's execute until no change.

There is a special case "semi-positive datalog" where rules with `NOT` only depend on the base data (the EDB in the datalog jargon).  "semi-positive datalog" does not need stratification or - equivalently, the stratification is one level, level 0, and it has all the rules in it.

So stratification is a _partition_ of the rules (every rule is assigned to exactly one stratum) and can be thought of as a partial ordering of the rules based on stratum number. It is a "strict partial ordering" because two distinct rules can be same stratum and there isn't an an ordering condition between the pair of rules. See https://en.wikipedia.org/wiki/Partially_ordered_set.

The spec gives an algorithm for _a_ stratification. There are other possible stratifications that meet the NAF requirement. The general technique of stratification (dividing evaluation to a number of subsets of rules) can be used for purposes other than providing negation.

afs:
Feel free to continue the general discussion here.

This PR has 16 commits with bland log messages and woudl benefit from tidying up before merging into the default branch.

Rather than make changes over the top of this PR, PR #801 takes this PR, commits squashed down to one commit with a more long-term log message. You can compare the diffs. They have the same content checksums `d80f395..27edf2f`.

I'll assume that the approvals from this PR carry over.

If @liviorobaldo agrees the PRs are the same by approving #801, then I will merge that one.

afs:
> If @liviorobaldo agrees the PRs are the same by approving #801, then I will merge that one.

Done.

> My first PR ☺️

:tada:

liviorobaldo:
Ahahahah, thanks! I'll try to minimize the number of commits in future PRs 👍

afs:
It is the final product that matters, not way it got there.

If the first commit had been significant not "Update index.html" it would have been easier (did you use the GH UI to write the PR?)

liviorobaldo:
Ok, as I mentioned in my email, I’ve now understood that I should avoid accepting each suggestion individually, since GitHub creates a new “Update index.html” commit every time. Instead, I should:
- edit the file manually (without committing each suggestion);
- apply all changes first and then create a single meaningful commit;
- use Squash and merge when the PR is ready.
This is what ChatGPT suggested... I hope he's right 😅

afs:
> I hope he's right

He isn't. 

1. Start with a meaningful commit.
2. Accept the changes either via UI or edit locally (often other changes follow from one review suggestion)
3. Then tidy up.
4. Repeat 2-3 as necessary as discussions happen.

liviorobaldo:
Noted. Will try my best.
Thanks

shacl12-rules/index.html
TallTed: @liviorobaldo —

The parallel with mathematical operators is that when you have a formula made up of those operators, you execute them in a particular order _according to the operators_ rather than according to the order they're written, e.g., a Parenthetical expression may follow any number of other operators as written, but it must be executed *before* those other operators.

That is, PEMDAS provides an initial execution order among these operators, parallel to _"stratification establishes a single execution order among the rules"_. 

But again, though you've said several times, "stratification establishes a single execution order among the rules," and insisted that this means "the same unique result is always produced," you've also conceded that, "When these operators are used in if–then rules, different outcomes may arise depending on the order in which the rules are executed", and this leads me once again to say, I _still_ do not see how a "single, implicit ordering" is established, and I thus don't see how "the same inference graph is always produced."

I have a feeling that part of this is due to differences between our primary languages. Perhaps someone else will be able to understand one or both of us better, and offer some translation assistance.

liviorobaldo: > But again, though you've said several times, "stratification establishes a single execution order among the rules," and insisted that this means "the same unique result is always produced," you've also conceded that, "When these operators are used in if–then rules, different outcomes may arise depending on the order in which the rules are executed

I’m sorry, I think there may have been a misunderstanding. What I meant was:

1. _Without stratification_: when non-monotonic operators are used in if–then rules, different outcomes may arise depending on the order in which the rules are executed.

2. _With stratification_: the problem described in 1. is resolved; thus a single, well-defined result is always produced.

About the problem in 1., perhaps you may have a look at [my recent paper on the Semantic Web journal](https://www.semantic-web-journal.net/content/interplay-between-validation-and-inference-shacl-investigation-time-ontology-1) (section 8.2). As for stratification, there are many resources available online. For example, [these slides](https://www.doc.ic.ac.uk/~mjs/teaching/KnowledgeRep491/Stratified_491-2x1.pdf) provide a clear explanation of why "every stratified logic program/database has a unique model".

Hope this helps 😊

afs: > a clear explanation of why "every stratified logic program/database has a unique model".

and seeAlso https://en.wikipedia.org/wiki/Datalog#Semantics

afs: Text in the abstract:

> which establishes a single, implicit ordering among rules,

This clause could be dropped if it is causing confusion.

At this point in the text, what matters to the reader is the outcome -- "ensuring that the same inference graph is always produced."

"single" isn't quite right (stratification is a partial ordering of rules, https://github.com/w3c/data-shapes/pull/722#issuecomment-3966747277, and execution within a stratum isn't order sensitive - [Datalog evaluation](https://en.wikipedia.org/wiki/Datalog#Evaluation)).

liviorobaldo: Makes sense. So the final sentence should be: "To avoid this, rules are evaluated using the technique of stratification, ensuring that the same inference graph is always produced."

How do we proceed? Do we create a new PR?

TallTed: I am looking for some instruction for the implementer, that tells them how to stratify the rules, such that the problem described in 1. is resolved, thereby always producing a single, well-defined result.

What I'm seeing now is, "upon intonation of the magic word, _stratification_, a single, well-defined result is magically delivered."

I can instruct a teenage student of mathematics in the use of PEMDAS, and they can follow that order of operations to always calculate the same result.

I'm hoping that we can provide some similar instructions such that a rather more sophisticated student of data science can start with a few visibly different sets of rules and a dataset, and, through application of the art of stratification, transform those different sets of rules into things that will produce the same inference graph.

Thus far, I don't see it.

afs: > I am looking for some instruction for the implementer, that tells them how to stratify the rules, such that the problem described in 1. is resolved, thereby always producing a single, well-defined result.

Ted - that text is in the document _introduction_ - 4 short paragraphs.

> What I'm seeing now is, "upon intonation of the magic word, stratification, a single, well-defined result is magically delivered."

Have you read https://www.w3.org/TR/shacl12-rules/#stratification-algorithm and https://w3c.github.io/data-shapes/shacl12-rules/#rule-set-evaluation ?

There are many slide sets and lecture notes online for free about Datalog.

The "Alice" book is also available. http://webdam.inria.fr/Alice/ - chapters 12,13,14,15

Issue #215, opened 2025-01-20
Shall the work on Inferencing start earlier?


HolgerKnublauch:
The plan outlined in the first WG meeting was to first bring the Core and SPARQL specs to conclusion before starting the other documents.

However, arguably there is no strong reason to not at least start the work on Inferencing, perhaps through a dedicated Task force, assuming this does not block or hinder the deliverables from the first round.

First, there seems to be a LOT of interest in this topic, so I assume there will be volunteers.

But more importantly, I believe there is a huge potential in expanding the expressiveness of SHACL Core if the Inferencing work leads to the standardization of SHACL-AF Node Expressions: https://w3c.github.io/shacl/shacl-af/#node-expressions

In a nutshell, Node Expressions start at a focus node and then compute a stream of value nodes. There are various kinds of such node expressions but it's essentially an extension point of its own.

IF Node Expressions would become a deliverable alongside the Core Spec (maybe even as a chapter in core) the following new use cases would become possible:

1. Dynamically computed targets, see also https://github.com/w3c/data-shapes/issues/213 (split from https://github.com/w3c/data-shapes/issues/199#issuecomment-2602401903)

```
ex:SomeShape
    a sh:NodeShape ;
    sh:targetExpression [
        sh:if [ ... ] ;
        sh:then [ ... ] ;
        sh:else [ ... ] ;
    ]
```

(This would also remove the need to add SPARQL-based targets to SHACL-SPARQL because sh:select is one of the standard node expression types).

2. Dynamically computed constraint parameters, see https://datashapes.org/dynamic.html#example-state

3. Generalizing property path expressions as a subset of node expressions. This is possible because path expressions also just start at a focus node and then produce a stream of value nodes. This would mean that the sh:path of a property shape may make much richer computations.

So depending on the output of the inferencing work, SHACL Core and SHACL SPARQL may look quite different.

VladimirAlexiev:
Hi @HolgerKnublauch! If you give a more prominent status to **node expressions**, that would mean less emphasis on SPARQL, right?
Maybe even the reference implementation of Constraint Components would be based more on expressions and less on SPARQL?

To my mind, the more machinery is based on declarative RDF notations, the better.
- So far all implementations have taken SPARQL as opaque: parameterize it and run it, but don't try to figure out what's going on inside.
- Whereas with declarative RDF, implementations can more easily pick it apart and implement it in a more optimized way.

HolgerKnublauch:
[...]

PR #729, opened 2026-01-20
Revise SHACL 1.2 Rules abstract

afs:
This PR takes text from the revised SHACL 12. Core abstract as related to SHACL Rules.

This is no more than a slightly better abstract. The final abstract can be written when the document is complete.

* [See the revised abstract online here](https://raw.githack.com/w3c/data-shapes/afs/rules-abstract/shacl12-rules/index.html#abstract)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -123,11 +123,21 @@
   </head>
   <body>
     <section id="abstract">
-      <p>This document describes Shapes Constraint Language (SHACL) Rules.</p>
+      <p>This document defines <em>SHACL Rules</em>.</p>
+      <p>
+        SHACL, the Shapes Constraint Language, is a language for describing the structure
+        of RDF graphs.
+        SHACL may be used for a variety of purposes such as validating, inferencing,
+        modeling domains, generating ontologies to inform other agents, building user
+        interfaces, generating code, and integrating data.
+      </p>
+      <p>
+        SHACL Rules provides inferencing with the generation of new RDF data from a
+        combination of a set of rules and a base data graph. Rules can be expressed as RDF
+        or in the SHACL Rules Language (SRL).
       <p>
         This specification is published by the
-        <a href="https://www.w3.org/groups/wg/data-shapes/">Data Shapes Working Group</a>
-        .
+        <a href="https://www.w3.org/groups/wg/data-shapes/">Data Shapes Working Group</a>.
       </p>
     </section>
 


liviorobaldo:
I think the abstract of SHACL Rules should not mention shapes. Accordingly, I would shorten the abstract as follows (since it is short, I prefer to include it here rather than proposing the changes directly in "Files changed"):

This document defines SHACL Rules.

SHACL, the Shapes Constraint Language, is a language for describing the structure of RDF graphs. SHACL may be used for a variety of purposes such as validating, inferencing, modeling domains, generating ontologies to inform other agents, building user interfaces, generating code, and integrating data.

SHACL Rules provides inferencing with the generation of new RDF data from a combination of a set of rules and a base data graph. Rules can be expressed as RDF or in the SHACL Rules Language (SRL).

shacl12-rules/index.html
```diff
@@ -123,11 +123,24 @@
   </head>
   <body>
     <section id="abstract">
-      <p>This document describes Shapes Constraint Language (SHACL) Rules.</p>
+      <p>This document defines SHACL Rules.</p>
+      <p>
+        SHACL, the Shapes Constraint Language, is a language for describing the structure
+        of RDF graphs.  SHACL introduces the notion of shapes that can specify constraints
+        on the structure of RDF nodes and edges.  graphs.
```
```suggestion
        on the structure of RDF graphs.
```

robert-david (changes requested):

shacl12-rules/index.html
afs: @liviorobaldo's change proposal removes the sentence.

Do you think that the rules abstract should/should not mention "shapes" and "constraints"?

I'm neutral on this ATM - the abstract is likely to change later anyway.

afs:
> I prefer to include it here rather than proposing the changes directly in "Files changed"

That makes it harder to see the changes - the text layout does not correspond to the HTML layout. I have made the change but it would be helpful to have the comparison that the GH UI provides - it highlights the changes in a suggestion.

The text still has "shapes" because of writing out "SHACL".

liviorobaldo:
Ok, thanks for making the changes. I'll make them myself next time. For me the abstract is fine.

liviorobaldo (approved):

shacl12-rules/index.html
```diff
@@ -123,11 +123,21 @@
   </head>
   <body>
     <section id="abstract">
-      <p>This document describes Shapes Constraint Language (SHACL) Rules.</p>
+      <p>This document defines SHACL Rules.</p>
+      <p>
+        SHACL, the Shapes Constraint Language, is a language for describing the structure
+        of RDF graphs.
+        SHACL may be used for a variety of purposes such as validating, inferencing,
```
TallTed, line 130: `inferencing` has long bothered me, and I am hoping it bothers others here, and maybe we can stop verbing the noun (`inference`) that was already a nouning of a verb (`infer`). Maybe we can just use `inferring`.
```suggestion
        SHACL may be used for a variety of purposes such as validating, inferring,
```

afs: "Inferencing" is a technical term that is used in several fields, such as AI, and is used in the semantic web area. 

e.g. 
https://blog.okfn.org/2010/08/02/about-inferencing/
https://github.com/w3c/data-shapes/issues/215

Note: the wording for the rules abstract is based on:
https://www.w3.org/TR/shacl12-core/#abstract

robert-david (approved):

shacl12-rules/index.html
simonstey: do we then actually need to copy SHACL Cores's abstract 1:1?

shacl12-rules/index.html
```diff
@@ -123,11 +123,21 @@
...
+        modeling domains, generating ontologies to inform other agents, building user
```
simonstey, line 131: >  generating ontologies to inform other agents,

I know this is coming from SHACL-core, but: what other agents?!

shacl12-rules/index.html
```diff
@@ -123,11 +123,21 @@
...
+        interfaces, generating code, and integrating data.
+      </p>
+      <p>
+        SHACL Rules provides inferencing with the generation of new RDF data from a
+        combination of a set of rules and a base data graph. Rules can be expressed as RDF
```
simonstey, line 136: is a `set of rules` equivalent to `rule set`?

<dt><dfn>rule set</dfn></dt>
    <dd>
      A [=rule set=] is a collection of zero or more [=rules=]
      and a collection of zero or more [=data blocks=].
    </dd>

----

one could argue that `rule set` is `set of rules \cups set of data blocks`

shacl12-rules/index.html
HolgerKnublauch: I picked that term for either humans or software agents such as LLMs. But since these software agents are currently changing on a weekly basis, I wanted to remain vague and not talk about AI agents.

shacl12-rules/index.html
```diff
@@ -123,11 +123,21 @@
...
+      <p>This document defines SHACL Rules.</p>
+      <p>
+        SHACL, the Shapes Constraint Language, is a language for describing the structure
+        of RDF graphs.
+        SHACL may be used for a variety of purposes such as validating, inferencing,
+        modeling domains, generating ontologies to inform other agents, building user
+        interfaces, generating code, and integrating data.
+      </p>
+      <p>
+        SHACL Rules provides inferencing with the generation of new RDF data from a
```
simonstey, line 135: >SHACL Rules provides 

here _SHACL Rules_ is used as a name rather than the plural of rule right? should we then somehow indicate this by formatting it differently?

shacl12-rules/index.html
afs: @simonstey -- I'm not sure which way to read your comment:

Are you saying we should take verbatim the SHACL Core?
Or that we should take none of it?

As [pointed out](https://github.com/w3c/data-shapes/pull/729#issuecomment-3773040951) the SHACL Core text emphases validation. It also puts that before the general text.

We don't have, yet, WG text that works for core, sparql, profiles, ui, rules, it is better to take what we can and adapt it for now. For example, it would be restructured to go put the general SHACL description in the main paragraph, then have a final short section to cover the specific document. 

That will take awhile. This PR for a slightly better abstract, not the final one, given we know people are reading the live drafts.

shacl12-rules/index.html
afs: I think "agents" will do for now.
What other suggestions are there? 

It is a WG task to craft some abstract text that can be used across all the documents. The core abstract is a good start. https://github.com/w3c/data-shapes/pull/729#discussion_r2735489206.

I've created #747 for that.

shacl12-rules/index.html
afs: This is intended to be a general description. 

I don't think the abstract needs forward pointers and the text should work if lifted out of the document.

Technical details like "RuleSet", "Data blocks" haven't been introduced. The key point is a set of rules (data blocks can be considered to be, technically, axioms - rules with a body of `{ }`. "set" / "collection" / "group of"

shacl12-rules/index.html
afs: We could do that at first mention: "This document defines _SHACL Rules_". Change made.

It ought to be done systematically for all SHACL documents -- #747.

recalcitrantsupplant (approved):

simonstey (approved):

shacl12-rules/index.html
```diff
@@ -123,11 +123,21 @@
   </head>
   <body>
     <section id="abstract">
-      <p>This document describes Shapes Constraint Language (SHACL) Rules.</p>
+      <p>This document defines <em>SHACL Rules</em>.</p>
+      <p>
+        SHACL, the Shapes Constraint Language, is a language for describing the structure
+        of RDF graphs.
+        SHACL may be used for a variety of purposes such as validating, inferencing,
+        modeling domains, generating ontologies to inform other agents, building user
```
TallTed, line 131: I wonder if the only "purpose" note shouldn't be removed, here, i.e.,
```suggestion
        modeling domains, generating ontologies, building user
```

afs: The discussion on the general abstract for SHACL is on #747.

Issue #677, opened 2025-11-27
[Rules] Tests


afs:
We need rule syntax tests - a test that takes a file, parses it as a rule set and passes (positive test) or expects a faulure (negative test).
That includes well-formedness and stratification test coverage.

We need evaluation tests that take a rule set, a data graph and an expected output graph. The test run by evaluation to give an actual graph which is compared with the expected graph. Graph compare is "blank node isomorphic". Evaluation tests don't have a lot of data so "blank node isomorphic" is cheap. 

```
:test_1 rdf:type   srt:RulesPositiveSyntaxTest ;
   mf:name    "A test of good syntax" ;
   mf:action  <syntax-rule-01.srl> ;
   .

:test_2 rdf:type   srt:RulesNegativeSyntaxTest ;
   mf:name    "A test that syntax is rejected" ;
   mf:action  <syntax-rule-02.srl> ;
   .
```

```
:eval-01 rdf:type srt:RulesEvalTest ;
    mf:name "Evaluation test basic-1";
    mf:action
         [ srt:ruleset  <eval-basic-01.shr> ;
           srt:data   <data-01.ttl> ] ;
    mf:result  <eval-results-basic-01.ttl> ;
    .
```

See http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#.
There is a reporting format using [W3C EARL](https://www.w3.org/TR/EARL10-Guide/).

SHACL validation tests can use a single file because its checking the validation report which is easy to identify within a graph.
General testing for "expected graph" as part of a single graph is not so easy.

https://w3c.github.io/rdf-tests/

afs:
We have tests!
More tests is better!

PR #735, opened 2026-01-24
More SHACL Rules tests

afs:
More syntax tests.
Add wellformedness and stratification test areas.

The file names are not stable - as more tests are added they may fit better along side existing tests which alters the numbering.

Part of #677

At the moment, the syntax tests and wellformedness are [generated from a script](https://github.com/afs/jena-rules/tree/main/src/test/files). This only really matters for syntax tests because there are a lot of them and new tests are likely to affect naming/numbering. To make a PR for new tests, create a new directory and we can combine them when the test suite settles down. When we have a first complete set, scripting shouldn't be needed.

The evaluation tests are hand-crafted so it is possible to make PRs for corrections and new tests now.


shacl12-test-suite/tests/rules/stratification/stratification-05.srl
```diff
@@ -0,0 +1,2 @@
+PREFIX : <http://example>
+RULE {} WHERE {}
```
robert-david: This is a copy of stratification-05.srl. Is this on purpose?

shacl12-test-suite/tests/rules/syntax/syntax-data-13.srl
```diff
@@ -0,0 +1,2 @@
+PREFIX : <http://example/>
+DATA { <<:s :p :o >> }
```
robert-david, line 2: OIs a triple term by itself allowed?

shacl12-test-suite/tests/rules/syntax/syntax-reification-03.srl
```diff
@@ -0,0 +1,4 @@
+PREFIX :    <http://example/>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+
+RULE { } WHERE { ?s :p << :a :b :c ~:r >> }
```
robert-david, line 4: Maybe this is the wrong place to discuss this, but i don't understand the meaning of ~ in this context.

shacl12-test-suite/tests/rules/syntax/syntax-reification-08.srl
```diff
@@ -0,0 +1,4 @@
+PREFIX :    <http://example/>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+
+RULE { } WHERE { :s :p :o {| :q1 :r1 ; :q2 :r2 |} . }
```
robert-david, line 4: Duplicate of syntax-reification-07.srl.

shacl12-test-suite/tests/rules/syntax/syntax-reification-05.srl
```diff
@@ -0,0 +1,4 @@
+PREFIX :    <http://example/>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+
+RULE { } WHERE { :s :p :o {| :q :r |} . }
```
robert-david, line 4: I don't have an intuition for the {| |} and it is not explained in the spec. Again, probably wrong place to discuss, but still noting it to pick up later.

robert-david (approved):

shacl12-test-suite/tests/rules/syntax/syntax-reification-03.srl
ajnelson-nist: The only place the character-pair `~:` appears in RDF 1.2 Turtle is [here](https://www.w3.org/TR/rdf12-turtle/#reifying-triples).  If I'm reading that document right, AND if this example used individuals using this prefix ...

```turtle
PREFIX ex: <http://example.com/ns#>
```

... then this rule looks like it could read:

```turtle
RULE { } WHERE { ?s ex:p << ex:a ex:b ex:c ~ex:r >> }
```

As for what it's accomplishing - does this rule just produce nothing, matching statements where some subject uses predicate `:p` and a reified triple `:a :b :c` that has a triple-term identifier `:r`?

shacl12-test-suite/tests/rules/syntax/syntax-data-13.srl
afs: A triple term is `<<( ... )>>`.

`<<:s :p :o >>` is a "reifiedTriple" (Turtle grammar rule name) that expands to:
```
   _:b rdf:reifies  <<( :s :p :o )>> .
```

shacl12-test-suite/tests/rules/syntax/syntax-reification-03.srl
afs: It matches the data 

```
:ABC ex:p << ex:a ex:b ex:c ~ex:r >>
```
which is legal Turtle. The SRL grammar (and SPARQL FWIW) lifts all the Turtle convenience syntax, allowing variables.

```
:ABC ex:p ex:r .
ex:r rdf:reifies <<( ex:a ex:b ex:c )>> .
```

All these convenience syntax forms are expanded to be RDF triples.

afs: > Maybe this is the wrong place to discuss this, but i don't understand the meaning of ~ in this context.

It's a reifier.

It allows for giving an explicit URI or blank node for the expansion that can be referred to elsewhere in the file, or, in the case of a URI, from outside. If none is given, a fresh blank node is a allocated. c.f. lists `(1 2 3)` and `[ ]`.

Turtle:

```
<<:s :p :o >> 
=>
_:b rdf:reifies  <<( :s :p :o )>>
```
but there is no way to get the blank node in the file.

```
<<:s :p :o ~:r>> 
:r :saidBy "Alice" .
=>
:r rdf:reifies  <<( :s :p :o )>> .
:r :saidBy "Alice" .
```

shacl12-test-suite/tests/rules/syntax/syntax-reification-05.srl
afs: https://www.w3.org/TR/rdf12-turtle/#annotation-syntax
https://www.w3.org/TR/rdf12-primer/#section-turtle-reifier-representation

and may be a longer NOTE on reification from the RDF WG (time-permitting).

You can try these in Turtle and SPARQL at http://sparql.org/ .

shacl12-test-suite/tests/rules/stratification/stratification-05.srl
afs: fixed

shacl12-test-suite/tests/rules/syntax/syntax-reification-08.srl
afs: Changed 07 to use `,`.

shacl12-test-suite/tests/rules/README.md
```diff
@@ -1 +1,44 @@
...
+All the test are syntactically legal, i.e. conform to the SHACL Rules grammar and any
+addition parsing rules.
+
+To pass a test, the parser must accept or reject a rule set as stated by the test
+stype. A negative well-formedness is a ruleset that violates the well-formed
+conditions of the SHACL abstract rule syntax.
+
+### Illegal stratification
+
+These test rule sets to detect where the startification condition is violated.
```
simonstey:
```suggestion
These test rule sets to detect where the stratification condition is violated.
```

shacl12-test-suite/tests/rules/eval/eval-basic-02-results.ttl
simonstey, line 1:
```ttl
PREFIX : <http://example/>

:x :q :o
```

this is missing a `.`

shacl12-test-suite/tests/rules/eval/rdfs.srl
```diff
@@ -0,0 +1,18 @@
+## RDFS (without no axoimatic triples)
```
simonstey, line 1:
```suggestion
## RDFS (without no axiomatic triples)
```

shacl12-test-suite/tests/rules/syntax/syntax-ruleset-structure-02.srl
simonstey, line 1: should be renamed to `syntax-ruleset-structure-02.srl`

shacl12-test-suite/tests/rules/syntax/syntax-ruleset-structure-03.srl
simonstey, line 1: should be renamed to `syntax-ruleset-structure-03.srl`

simonstey:
this should eventually also add an entry for rules to: https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-test-suite/tests/manifest.ttl right?

afs:
> this should eventually also add an entry for rules to: https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-test-suite/tests/manifest.ttl right?

Unclear.

First - it assumes a system is a complete implementation of all of SHACL (UI? Profiles?). That does not reflect the client/server split.

Second - while at the highest level, they use the same vocabulary, the SHACL 1.0 test framework diverged from the original DataAccess work in the details.

The barrier for rules is that the result graph needs to be compared to a reference graph. Having the input and expected output in one graph does not work. `NOT` would also makes it "interesting".

afs:
@recalcitrantsupplant @simonstey --

Would you mind if this PR is merged? Any issues with the tests can be dealt with as focused issues, which is more helpful than refining here on a large, long-lived PR.

recalcitrantsupplant:
Yes agree, happy to merge - will add these to my implementation and raise issues/PRs based on that.

recalcitrantsupplant (approved):

PR #736, opened 2026-01-25
GH-728: Inference graph

### LINKED ISSUE #728: Does output include base graph triples
The section [Rule Set Evaluation](https://w3c.github.io/data-shapes/shacl12-rules/#rule-set-evaluation) indicates that:

"The inferred triples do not include triples present in the set of triples of the [base graph](https://w3c.github.io/data-shapes/shacl12-rules/#dfn-base-graph)."

However, the definition given for an _evaluation graph_ is:

"A [evaluation graph](https://w3c.github.io/data-shapes/shacl12-rules/#dfn-evaluation-graph) is an [RDF Graph](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-graph) that combines the [base graph](https://w3c.github.io/data-shapes/shacl12-rules/#dfn-base-graph) and all triples produced during the evaluation of a rule set."

* Q1: So I conclude that an _evaluation graph_ is not the output but an interim thing. Is this correct?

If I want to use a SHACL Rules engine to remove triples from a graph, I could do so by generating (selecting) the triples to be removed and then removing them from the base graph if the output was inferred triples only but not if it included the base graph as then I would get no result.

* Q2: Is removal by inferring triples to be removed and then removing them from the base graph an expected use pattern?

afs:
Closes #728

From https://github.com/w3c/data-shapes/issues/728#issuecomment-3775598896

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/inference-graph/shacl12-rules/index.html)
  - [End of section 3](https://raw.githack.com/w3c/data-shapes/inference-graph/shacl12-rules/index.html#rules-abstract-syntax)
  - [Algorithm in 5.4](https://raw.githack.com/w3c/data-shapes/inference-graph/shacl12-rules/index.html#eval-rule-set)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -342,7 +342,7 @@ RULE { ?x :name ?FN } WHERE {
           <dt><dfn>triple template</dfn></dt>
           <dd>
             A [=triple template=] is 3-tuple where each element is either
-            a [=variable=] or an [=RDF term] (which might be a [=triple term=]).
+            a [=variable=] or an [=RDF term=] (which might be a [=triple term=]).
             The second element of the tuple must be an [=IRI=] or a [=variable=].
             [=Triple templates=] appear in the [=head=] of a [=rule=].
           </dd>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -417,6 +417,20 @@ RULE { ?x :name ?FN } WHERE {
             A [=rule set=] is a collection of zero or more [=rules=]
             and a collection of zero or more [=data blocks=].
           </dd>
+
+          <dt><dfn>base graph</dfn></dt>
+          <dd>
+            The [=base graph=] is the [=RDF Graph=] given as input to the evaluation
+            process.
+          </dd>
+
+          <dt><dfn>inference graph</dfn></dt>
+          <dd>
+            The [=inference graph=] is an [=RDF Graph=] that is the outcome of rule set evaluation.
+            It contains all triples not found in the [=base graph=] that are inferred from
+            evaluation of the [=rule set=] on the [=base graph=].
+          </dd>
+
         </dl>
 
         <p>
@@ -503,7 +517,7 @@ RULE { ?x :name ?FN } WHERE {
             <li>
               For every [=variable=] in an [=expression=] at position <em>i</em>
               of the [=body=], there is a corresponding [=variable=] of the same
-              name occuring in a [=triple pattern=] at a position <em>j</em>,
+              name occurring in a [=triple pattern=] at a position <em>j</em>,
               or occurring as an [=assignment variable=] at a position <em>j</em>,
               where <em>i &gt; j</em>.
             </li>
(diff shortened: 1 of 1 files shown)


liviorobaldo:
Looks correct to me. The only question is whether we want to return the pair ⟨GI, GE⟩ rather than only GI, but that's not really necessary.

robert-david (approved):

shacl12-rules/index.html
```diff
@@ -916,23 +925,32 @@ <h4>Evaluation of a Rule Set</h4>
         <p class=ednote>Sketch</p>
 
 <pre class="algorithm">
-let G0 be the input RDF graph.
+let G0 be input base graph
```
```suggestion
let G0 be the input base graph
```

recalcitrantsupplant (approved):
should still have "the" here I think

shacl12-rules/index.html
```suggestion
            a [=variable=] or an [=RDF term=] (which might be a [=triple term=]).
```

not directly added by this PR, but a minor thing to fix

shacl12-rules/index.html
```diff
@@ -417,6 +417,20 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          <dt><dfn>base graph</dfn></dt>
+          <dd>
+            The [=base graph=] is the [=RDF Graph=] given as input to the evaluation
+            process.
+          </dd>
+
+          <dt><dfn>inference graph</dfn></dt>
+          <dd>
+            The [=inference graph=] is an [=RDF Graph=] that is the outcome of rule set evaluation.
+            It contains all triples not found in the [=base graph=] that are inferred from
```
simonstey, line 430: so what's the connection between the base graph triples and the ones provided in the form of data blocks? I assumed the latter are basically just a more concise way of providing triples to apply rules to. In which case, this should read something like this: 

```suggestion
            It contains all triples not found in the [=base graph=] or any [=data block=] that are inferred from
```

however later on we say:

```
let G0 be the input base graph
let RS be the rule set
let D be the graph of all DATA triples in RS

Apply stratification to RS

let L be the sequence of layers after stratification

# Inference graph
let GI = D                                 # <--- so now D triples are already in the inference graph

# Evaluation graph.
let GE = G0 &cups; D              # <--- but also in the evaluation graph together with the base graph

for each layer in L:
    let finished = false
    while !finished:
        finished = true
        for each rule in layer:
            let X = eval(rule, GE)
            let Y = { t &isin; X | t &notin; in GE }
            if Y is not empty:
                finished = false
                GI = Y &cups; GI
                GE = Y &cups; GE
                endif
            endfor
        endwhile
    endfor
the result is GI
```

so any triples provided via DATA blocks are therefore automatically returned, regardless of whether they were also in the base graph. 

bug or feature?

shacl12-rules/index.html
simonstey, line 957: see https://github.com/w3c/data-shapes/pull/736/changes#r2735006556

tldr: any triples coming from data blocks are automatically returned as part of the inference graph, regardless of whether they are also in the base graph or not.

simonstey (commented):
see comments

shacl12-rules/index.html
```suggestion
              name occurring in a [=triple pattern=] at a position <em>j</em>,
```

typo fix

shacl12-rules/index.html
afs: DATA are inferred. They are axioms - constant rules.

```
DATA { :s :p :o }
```
is the same effect as:
```
RULE { :s :p :o } WHERE {}
```

The inference graph is defined to be the triples not found in the base graph. (See tests)

So the fix needed is `let GI = { t &isin;D  | t &notin; in G0 }`

(I'm avoiding using set difference notation -- `\` or sometimes `-`)

shacl12-rules/index.html
afs: Comment above.

shacl12-rules/index.html
simonstey: > So the fix needed is `let GI = { t &isin;D  | t &notin; in G0 }` 

👍

simonstey (approved):

PR #738, opened 2026-01-26
Updates permitted datatypes of `sh:message` and `sh:description`, and other properties previously permitting `rdf:langString` (Issues 723 & 724)

### LINKED ISSUE #723: Allow `xsd:string`, `rdf:langString`, and `rdf:dirLangString` as values for `sh:message` and `sh:labelTemplate`.
2.2 Syntax of SPARQL-based Constraints
https://w3c.github.io/data-shapes/shacl12-sparql/#sparql-constraints-syntax

> [SPARQL-based constraints](https://w3c.github.io/data-shapes/shacl12-sparql/#dfn-sparql-based-constraint) may have values for the property sh:message and these are either xsd:string literals or literals with a language tag.

3.2.2 Label Templates (sh:labelTemplate)
https://w3c.github.io/data-shapes/shacl12-sparql/#labelTemplate

> The property sh:labelTemplate can be used at any [constraint component](https://www.w3.org/TR/shacl12-core/#dfn-constraint-component) to suggest how [constraints](https://www.w3.org/TR/shacl12-core/#dfn-constraint) could be rendered to humans. The values of sh:labelTemplate are strings (possibly with language tag) and are called label templates.

The I18n WG recommends that the `sh:message` and `sh:labelTemplate` properties accept `xsd:string`, `rdf:langString` and `rdf:dirLangString` as values, with `rdf:dirLangString` being the most recommended. This preserves backwards compatibility while encouraging implementers to specify both the language and the base direction.

### LINKED ISSUE #724: Allow `xsd:string`, `rdf:langString`, and `rdf:dirLangString` as values for `sh:message`
3.1.5 Declaring Messages for a Shape or Constraint
https://w3c.github.io/data-shapes/shacl12-core/#message

> Shapes can have values for the property sh:message. The values of sh:message are either xsd:string literals or literals with a language tag. A subject should not have more than one value for sh:message with the same language tag.

The I18n WG recommends that the `sh:message` property accept `xsd:string`, `rdf:langString` and `rdf:dirLangString` as values, with `rdf:dirLangString` being the most recommended. This preserves backwards compatibility while encouraging implementers to specify both the language and the base direction. Additionally, we wonder if using a [languageMap](https://www.w3.org/TR/string-meta/#language-maps) might be a better choice here.

HolgerKnublauch:
Also allowing rdf:dirLangString where rdf:langString was allowed.

Also allowing rdf:HTML at sh:message and sh:description.

Closes #723
Closes #724

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-723-724/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -881,7 +881,7 @@ ex:Person
 						</div>
 						<div class="diagram-class-properties-section">
 							<div><a href="#deactivated">sh:deactivated</a> : xsd:boolean</div>
-							<div><a href="#message">sh:message</a> : xsd:string or rdf:langString</div>
+							<div><a href="#message">sh:message</a> : text</div>
 							<div><a href="#severity">sh:severity</a> : sh:Severity</div>
 						</div>
 					</div>
@@ -913,8 +913,8 @@ ex:Person
 							<div><a href="#NodeConstraintComponent">sh:node</a> : sh:NodeShape</div>
 						</div>
 						<div class="diagram-class-properties-section">
-							<div><a href="#name">sh:name</a> : xsd:string or rdf:langString</div>
-							<div><a href="#description">sh:description</a> : xsd:string or rdf:langString</div>
+							<div><a href="#name">sh:name</a> : xsd:string or rdf:dirLangString or rdf:langString</div>
+							<div><a href="#description">sh:description</a> : text</div>
 							<div><a href="#syntax-rule-path-defaultValue">sh:defaultValue</a> : any</div>
 							<div><a href="#syntax-rule-path-defaultValue">sh:values</a> : any</div>
 							<div><a href="#group">sh:group</a> : sh:PropertyGroup</div>
@@ -1606,7 +1606,8 @@ ex:MyShape-myProperty2
 					<h4>Declaring Messages for a Shape or Constraint</h4>
 					<p class="syntax">
 						Shapes can have values for the property <code>sh:message</code>.
-						<span data-syntax-rule="message-datatype">The values of <code>sh:message</code> are either <code>xsd:string</code> literals or literals with a language tag.</span>
+						<span data-syntax-rule="message-datatype">The values of <code>sh:message</code> are <a>literals</a> with <a>datatype</a>
+						<code>xsd:string</code>, <code>rdf:dirLangString</code>, <code>rdf:langString</code>, or <code>rdf:HTML</code>.
 						A subject should not have more than one value for <code>sh:message</code> with the same language tag.
 					</p>
 					<p>
@@ -2774,7 +2775,6 @@ ex:Alice ex:age "23"^^xsd:integer .
 					<p>
 						The following example illustrates the list-based syntax, meaning that all values of
 						<code>rdfs:label</code> must be either <code>xsd:string</code> or <code>rdf:langString</code>.
-						Note that using <code>rdf:langString</code> as value of <code>sh:datatype</code> can be used to test if value nodes have a language tag.
 					</p>
 					<aside class="example" title="Shape with a list of IRIs as sh:datatype property constraint">
 						<div class="shapes-graph">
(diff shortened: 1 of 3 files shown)


HolgerKnublauch:
The PR also includes a minor clean up to the SPARQL abstract, reflecting our recent generalization of the scope of SHACL beyond just validation.

Note that in the diagram I switched to "text" instead of the (too long) xsd:string or rdf:langString or rdf:dirLangString or rdf:HTML.

YoucTagh (approved):

shacl12-core/index.html
```diff
@@ -2774,7 +2775,6 @@ <h4>sh:datatype</h4>
 					<p>
 						The following example illustrates the list-based syntax, meaning that all values of
 						<code>rdfs:label</code> must be either <code>xsd:string</code> or <code>rdf:langString</code>.
```
TallTed, line 2777: It seems very strange to not include `rdf:dirLangString` as a possible type of `rdfs:label` values. It seems less so but still strange to not include `rdf:HTML`.

shacl12-core/index.html
```diff
@@ -3461,7 +3461,7 @@ <h4>sh:singleLine</h4>
 ex:SingleLineExampleShape-comment
     a sh:PropertyShape ;
     sh:path rdfs:comment ;
-    sh:datatype ( xsd:string rdf:langString ) ;
+    sh:datatype ( xsd:string rdf:dirLangString rdf:langString ) ;
```
TallTed, line 3464: Again, why not include `rdf:HTML`?

shacl12-sparql/index.html
```diff
@@ -147,12 +147,11 @@
 		<section id="abstract">
 			<p>
 				This document defines SPARQL-related extensions of the SHACL Shapes Constraint Language.
-				SHACL is a language for validating RDF graphs against a set of conditions.
-				These conditions are provided as shapes and other constructs expressed in the form of an RDF graph.
 				While the Core part of SHACL defines the basic syntax of shapes and the most common constraint components
 				supported by SHACL, the SPARQL-related extensions cover features that extend the expressiveness of Core
 				by means of SPARQL.
-				In particular, this document defines how constraints and constraint components can be defined using SPARQL.
+				In particular, this document defines how constraints and constraint components can be defined using SPARQL,
```
TallTed, line 153:
```suggestion
				In particular, this document defines how constraints and constraint components can be defined using SPARQL
```

TallTed:
> Note that in the diagram I switched to "text" instead of the (too long) xsd:string or rdf:langString or rdf:dirLangString or rdf:HTML.

Perhaps we should define `rdf:text` as a supertype, having subtypes of `xsd:string`, `rdf:langString`, `rdf:dirLangString`, & `rdf:HTML` (and others?)...?

HolgerKnublauch:
Introducing a supertype would require a new type of inference, and is certainly out of scope for the SHACL WG, esp if it would go into the rdf: namespace.

TallTed:
I should have suggested `sh:text`, but I see your point. Still, that suggests that the "too long" string of optional type values should still be used, instead of "text".

shacl12-sparql/index.html
HolgerKnublauch: No, the Oxford comma was added here intentionally - SPARQL-based node expressions are not used to define constraints and constraint components.

shacl12-core/index.html
HolgerKnublauch: This is just an example, not normative. And it doesn't make sense to have rdf:HTML in the context of singleLine (which only checks for line breaks in the syntax, which doesn't mean the same in HTML rendering).

shacl12-core/index.html
HolgerKnublauch: Ok I'll fix that, but no, rdf:HTML in labels wouldn't be a good practice. Labels are for example often rendered in tree structures, and those shouldn't contain markup such as line breaks or formatting that the tree viewer will override anyway.

HolgerKnublauch:
@TallTed I have opened a follow-up PR - please approve if that addresses the remaining issues https://github.com/w3c/data-shapes/pull/756

PR #748, opened 2026-01-28
GH-743: Remove some SRL functions

### LINKED ISSUE #743: Remove some SRL functions
After today's (2026-01-26) WG meeting, I took a look at the list of functions to see what can be removed:

| Function | Reason |
| -- | -- |
| `COALESCE` | Failure to find an valid argument expression causes an error |
| `BOUND`  | Well-formedness means this is does not happen |
| `MD5`, `SHA*` | Nowadays, there are other important hash functions |
| `RAND` | Not a well-behaved function (calling twice gives different values) |
| `BNODE()` | Zero-args form - not a well-behaved function |

afs:
Removes COALESCE, RAND, SHA*, MD5. (BOUND was removed by an earlier commit)

Closes #743

* [See the grammar section rendered online here](https://raw.githack.com/w3c/data-shapes/gh-743-remove-srl-functions/shacl12-rules/index.html#shapes-rules-grammar)


robert-david (approved):

recalcitrantsupplant:
I know we discussed NOW() but can't see it in the issue or here - what was the status on it? also UUID/STRUUID.

recalcitrantsupplant (approved):

afs:
> NOW()

https://www.w3.org/TR/sparql12-query/#func-now
which is modelled after
https://www.w3.org/TR/xpath-functions-31/#func-current-dateTime

It returns the same value xsd:dateTime through an evaluation. It is a well-behaved function.

> UUID/STRUUID.

I would group the decision with BNODE -- have both UUID and BNODE, or neither.

PR #756, opened 2026-02-02
issue-723/724: Further cosmetic edits related to dirLangString

HolgerKnublauch:
Also turned the value range of sh:closed into an enum

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-723-edits/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -882,7 +882,7 @@ ex:Person
 						</div>
 						<div class="diagram-class-properties-section">
 							<div><a href="#deactivated">sh:deactivated</a> : xsd:boolean</div>
-							<div><a href="#message">sh:message</a> : text</div>
+							<div><a href="#message">sh:message</a> : xsd:string, rdf:dirLangString/langString or rdf:HTML</div>
 							<div><a href="#severity">sh:severity</a> : sh:Severity</div>
 						</div>
 					</div>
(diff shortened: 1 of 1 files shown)


shacl12-core/index.html
```diff
@@ -882,7 +882,7 @@ <h2>Shapes and Constraints</h2>
 						</div>
 						<div class="diagram-class-properties-section">
 							<div><a href="#deactivated">sh:deactivated</a> : xsd:boolean</div>
-							<div><a href="#message">sh:message</a> : text</div>
+							<div><a href="#message">sh:message</a> : xsd:string, rdf:dirLangString/langString or rdf:HTML</div>
```
TallTed, line 885:
```suggestion
							<div><a href="#message">sh:message</a> : xsd:string, rdf:dirLangString, rdf:langString, or rdf:HTML</div>
```

shacl12-core/index.html
```diff
@@ -896,26 +896,26 @@ <h2>Shapes and Constraints</h2>
...
 							<div><a href="#MinLengthConstraintComponent">sh:minLength</a>, <a href="#MaxLengthConstraintComponent">sh:maxLength</a> : xsd:integer</div>
-							<div><a href="#ClassConstraintComponent">sh:class</a>, <a href="#DatatypeConstraintComponent">sh:datatype</a> : rdfs:Resource or rdf:List of rdfs:Resources</div>
+							<div><a href="#ClassConstraintComponent">sh:class</a>, <a href="#DatatypeConstraintComponent">sh:datatype</a> : rdfs:Resource or list of rdfs:Resource</div>
 							<div><a href="#NodeConstraintComponent">sh:node</a> : sh:NodeShape</div>
 						</div>
 						<div class="diagram-class-properties-section">
-							<div><a href="#name">sh:name</a> : xsd:string or rdf:dirLangString or rdf:langString</div>
-							<div><a href="#description">sh:description</a> : text</div>
+							<div><a href="#name">sh:name</a> : xsd:string or rdf:dirLangString/langString</div>
+							<div><a href="#description">sh:description</a> : xsd:string, rdf:dirLangString/langString or rdf:HTML</div>
```
TallTed, line 918:
```suggestion
							<div><a href="#description">sh:description</a> : xsd:string, rdf:dirLangString, rdf:langString, or rdf:HTML</div>
```

TallTed (changes requested):
Small punctuation changes, and a couple of `/` replaced with `, rdf:`.

shacl12-core/index.html
HolgerKnublauch: I did that intentionally because otherwise it was too wide for the box. That's exactly the problem here, and why I went with "text" earlier.

shacl12-core/index.html
HolgerKnublauch: Maybe we should just delete the "UML" diagram altogether and move it to some other overview document.

shacl12-core/index.html
TallTed: `too wide for the box` depends on the active viewport, the available viewport, the user-set base font size, and other factors. Just let it flow to a second (or third, or more) line.

TallTed: Here's what I'm now seeing in [the PR preview](https://raw.githack.com/w3c/data-shapes/issue-723-edits/shacl12-core/index.html), which is displaying in dark mode for unknown reasons --

Screen Shot 2026-02-04 at 01 12 12 PM



Note that the `sh:Shape` has a `sh:message` that wraps, even with your `/`.

If *exact* presentation matters, the diagram should be an immutable image ... but even then, various user- or device-based settings, preferences, etc., may change what the user actually sees.

Better we be precise with our language, naming, etc., and figure out presentation challenges later — optimally with decisions that optimize presentation throughout our document(s).

HolgerKnublauch: Here is how it looks for me 
image
and you see the sh:description barely fits in the property shape box, and that was after I have added a few pixels.

Yes, all this is very fragile and perhaps of questionable value because it's just a small subset of features anyway. There are various unresolved/proposed additions related to bringing in different tools but they are appear stalled. Unless someone has sufficient time to truly fix this diagram, I would tend to just discard it completely.

afs:
From the discussions between RDF WG and i18n WG, 

https://github.com/w3c/rdf-concepts/pull/265#issuecomment-3820365443
and specifically:
https://github.com/w3c/rdf-concepts/pull/265#issuecomment-3842904195

it may not be necessary to include rdf:HTML (unless SHACL wants to).

HolgerKnublauch:
> From the discussions between RDF WG and i18n WG,
> 
> [w3c/rdf-concepts#265 (comment)](https://github.com/w3c/rdf-concepts/pull/265#issuecomment-3820365443) and specifically: [w3c/rdf-concepts#265 (comment)](https://github.com/w3c/rdf-concepts/pull/265#issuecomment-3842904195)
> 
> it may not be necessary to include rdf:HTML (unless SHACL wants to).

Thanks. This would have to be applied to a separate PR than this here, which is entirely cosmetic.

But I think allowing rdf:HTML in sh:message makes sense from a user perspective, as it gives more freedom in rendering hyperlinks, bullet lists etc.

PR #757, opened 2026-02-02
Shacl shacl location update

nicholascar:
Changes the reference to SHACL-SHACL being in a stand-alone Note to being in an Appendix of the Overview.


nicholascar:
BAD PR, will re-issue

PR #758, opened 2026-02-03
SHACL-SHACL-ref

nicholascar:
Update the reference to SHACL-SHACL from the old Note to the Overview's Appendix


nicholascar:
Second bad PR, closing, will reissue

Issue #767, opened 2026-02-08
Provide an algorithm to calculate a stratification


afs:
SHACL Rules ought to provide an algorithm for stratification.

While we could just state the requirements for consistent evaluation, it would be helpful to give an algorithm that achieves this.

Ideally, this algorithm would be "minimal" in the sense that it produces a least number of strata to make evaluation works. Other stratification might be used by an implementation for other evaluation implementation purposes but they must enforce the stratification requirement in order have the same output to be the same for all implementations.

This is a global stratification - it only depends on the rule set not the data. 

My sense is that a _minimal_ stratification is one that separates the negation elements, and any other stratification will further refine the set of rules at each layer of the minimal stratification. A proof of this would be nice to have 🍰 .

PR #768, opened 2026-02-08
Definitions for dependency and stratification

### LINKED ISSUE #674: [Rules] Negation
The initial proposal for the Shape Rule Language includes a form of negation: 

```
RULE { ?x :name "unset" } WHERE { 
    ?x rdf:type :Person 
    NOT { ?x :name ?someName }
}
```

The name is an initial proposal.

The inner pattern is rule-body like only allowing triple patterns and filters only.

Variable only mentioned in the inner body do not cause solution bindings for the rule.

`NOT` is a rule body element. It is not identical to SPARQL's `NOT EXISTS`.

In SPARQL, `NOT EXISTS` is an expression inside a SPARQL `FILTER`.
* it can be combined with other expressions including the boolean negation in the form of `!`.
* the graph pattern is arbitrary SPARQL.
* there is a companion `EXISTS` expression.

### LINKED ISSUE #687: [Rules] Stratification
Stratification is used to gives conditions on the order rules are evaluated so that the output is predicable. Negation depends on order of execution.
https://en.wikipedia.org/wiki/Datalog#Negation

The [spec](https://w3c.github.io/data-shapes/shacl12-rules) should define [stratification](https://w3c.github.io/data-shapes/shacl12-rules/#stratification) and provide an algorithm to calculate the layer numbering.

This can be done for [rule](https://w3c.github.io/data-shapes/shacl12-rules/#dfn-rule) but if applied to [rule elements](https://w3c.github.io/data-shapes/shacl12-rules/#dfn-rule-body-element) we may be able to define evaluation for slighter more generally.

For example. `NOT` can defined to require conditions dependencies on triple patterns with variables that are used in the `NOT` pattern, and not for unrelated variables.

Recursion of unrelated triple patterns could then be allowed. This is equivalent to rewrite, replacing the rule with several rules. On the other hand, it might not be a good idea to over-complicate the conditions for something that can be written, albeit in a different form.

afs:
Closes #674
Closes #687 - this is a general issue. Further work has been recorded in #767.

This PR puts in sub-sections for dependency and stratification into the abstract syntax section.
The earlier general definitions have some indentation and formatting fixes.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/afs/dependency-stratification/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -13,13 +13,13 @@
         group: "wg/data-shapes",
         github: "w3c/data-shapes",
         //preProcess:     [  ],
-        //@@
         edDraftURI: "https://w3c.github.io/data-shapes/shacl12-rules/",
         shortName: "shacl12-rules",
         copyrightStart: "2025",
         // Can remove SHACL12-*?
         xref: ["RDF12-CONCEPTS", "SPARQL12-QUERY", "SHACL12-CORE", "SHACL12-SPARQL", "SHACL12-NODE-EXPR" ] ,
         lint: { "no-unused-dfns": false },
+        maxTocLevel: 2,
 
         editors: [
           {
@@ -58,8 +58,6 @@
         //previousMaturity: "REC",
         //prevRecShortname: "shacl",
         //prevRecURI: "https://www.w3.org/TR/2017/REC-shacl-20170720/",
-
-
         localBiblio: {
           "shacl12-shacl-shacl": {
             title: "SHACL 1.2 SHACL-SHACL",
@@ -67,12 +65,12 @@
             status: "ED",
             publisher: "W3C",
           }
-        }
+        },
         // Remove in final version
-        , lint: {
+        lint: {
           "informative-dfn": false,
           "no-unused-dfns": false
-        }
+        },
       };
     </script>
     <link rel="stylesheet" href="style.css"/>
@@ -317,19 +315,21 @@ RULE { ?x :name ?FN } WHERE {
     </section>
 
     <section id="rules-abstract-syntax">
-        <!-- Export -->
-        <h2>Shape Rules Abstract Syntax</h2>
+      <h2>Shape Rules Abstract Syntax</h2>
 
-        <p>The Shape Rules Abstract Syntax</p>
+      <p>The Shape Rules Abstract Syntax</p>
 
+      <section id="rules-defns">
+        <h3>Elements of the Abstract Syntax</h3>
         <dl>
-          <dt><dfn data-cite="sparql12-query#defn_QueryVariable">variable</dfn></dt>
+
+          <dt><dfn data-cite="sparql12-query#defn_QueryVariable">Variable</dfn></dt>
           <dd>
             A [=variable=] represents a possible [=RDF term=] in a triple pattern.
             Variables are also used in <a>expressions</a>.
           </dd>
 
-          <dt><dfn>expression</dfn></dt>
+          <dt><dfn>Expression</dfn></dt>
           <dd>
             <p>
               An [=expression=] is a function, or functional form.
@@ -342,13 +342,13 @@ RULE { ?x :name ?FN } WHERE {
               and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
           </dd>
 
-          <dt><dfn data-lt="data-block">data block</dfn></dt>
+          <dt><dfn data-lt="data-block">Data block</dfn></dt>
           <dd>
             A [=data block=] is a set of triples.
             These form extra facts that are included in the inference process.
           </dd>
 
-          <dt><dfn>triple template</dfn></dt>
+          <dt><dfn>Triple template</dfn></dt>
           <dd>
             A [=triple template=] is 3-tuple where each element is either
             a [=variable=] or an [=RDF term=] (which might be a [=triple term=]).
@@ -356,28 +356,36 @@ RULE { ?x :name ?FN } WHERE {
             [=Triple templates=] appear in the [=head=] of a [=rule=].
           </dd>
 
-          <dt><dfn>triple pattern</dfn></dt>
+          <dt><dfn>Triple pattern</dfn></dt>
           <dd>
             A [=triple pattern=] is 3-tuple where each element is either a
-            [=variable=], or an [=RDF term=] (which might be a triple term).
-            The second element of the tuple must be an [=IRI=].
-            [=Triple patterns=] appear in the [=body=] of a [=rule=].
+            [=variable=] or an [=RDF term=] (which might be a triple term).
+            The second element of the tuple must be an [=IRI=] or a [=variable=].
+            [=Triple patterns=] can appear as [=triple pattern elements=]
+            as well as inside [=negation elements=].
           </dd>
 
-          <dt><dfn>condition expression</dfn></dt>
+          <dt><dfn>Condition expression</dfn></dt>
           <dd>
             A [=condition expression=] is a function, or functional form,
             that evaluates to true or false.
             [=Condition expressions=] appear in the [=body=] of a [=rule=].
           </dd>
 
-          <dt><dfn>negation element</dfn></dt>
+          <dt><dfn>Triple pattern element</dfn></dt>
           <dd>
-            A [=negation element=] is ..@@.. .
-            [=Negation elements=] appear in the [=body=] of a [=rule=].
+            A [=triple pattern element=] is a [=triple pattern=]
+            used as a [=rule body element=].
           </dd>
 
-          <dt><dfn>assignment</dfn></dt>
+          <dt><dfn>Negation element</dfn></dt>
+          <dd>
+            A [=negation element=] is a [=rule body element=].
+            It takes a sequence of [=triple patterns=] and
+            [=condition expressions=].
+          </dd>
+
+          <dt><dfn>Assignment</dfn></dt>
           <dd>
             An [=assignment=] is a pair of a <a>variable</a>, called the
             <dfn>assignment variable</dfn>, and an <a>expression</a>, called the
@@ -385,61 +393,63 @@ RULE { ?x :name ?FN } WHERE {
             [=Assignments=] appear in the [=body=] of a [=rule=].
           </dd>
 
+          <!-- Pending
           <dt><dfn>aggregation element</dfn></dt>
           <dd>
             An [=aggregation element=] is ..@@.. .
              [=Aggregation elements=] appear in the [=body=] of a [=rule=].
           </dd>
+          -->
 
-          <dt><dfn data-lt="rule body element|rule element|element">rule body element</dfn></dt>
+          <dt><dfn 
+                data-lt="rule body element|rule element"
+                >Rule body element</dfn></dt>
           <dd>
             A [=rule body element=] (often just "rule element") is any element that
-            can appear in a [=rule body=];
+            can appear in a [=rule body=],
             a [=triple pattern=],
             a [=condition expression=],
             a [=negation element=],
-            an [=assignment=],
+            or an [=assignment=].
+            <!-- Pending
             or an [=aggregation element=].
+            -->
           </dd>
 
-          <dt><dfn data-lt="rule head|head">rule head</dfn></dt>
+          <dt><dfn data-lt="rule head|head">Rule head</dfn></dt>
           <dd>
-            A [=rule head=] is a sequence where each element
-            of the sequence is a [=triple template=].
+            A [=rule head=] is a sequence of [=triple templates=].
           </dd>
 
-          <dt>
-            <dfn data-lt="rule body|body">rule body</dfn>
-          </dt>
+          <dt><dfn data-lt="rule body|body">Rule body</dfn></dt>
           <dd> 
             A [=rule body=] is a sequence of [=rule body elements=].
           </dd>
 
-          <dt><dfn>rule</dfn></dt>
+          <dt><dfn>Rule</dfn></dt>
           <dd>
             A [=rule=] is a pair of a [=rule head=] (often just "head") and
             a [=rule body=] (often just "body").
           </dd>
 
-          <dt><dfn>rule set</dfn></dt>
+          <dt><dfn>Rule set</dfn></dt>
           <dd>
             A [=rule set=] is a collection of zero or more [=rules=]
             and a collection of zero or more [=data blocks=].
           </dd>
 
-          <dt><dfn>base graph</dfn></dt>
+          <dt><dfn>Base graph</dfn></dt>
           <dd>
             The [=base graph=] is the [=RDF Graph=] given as input to the evaluation
             process.
           </dd>
 
-          <dt><dfn>inference graph</dfn></dt>
+          <dt><dfn>Inference graph</dfn></dt>
           <dd>
             The [=inference graph=] is an [=RDF Graph=] that is the outcome of rule set evaluation.
             It contains all triples not found in the [=base graph=] that are inferred from
             evaluation of the [=rule set=] on the [=base graph=].
           </dd>
-
         </dl>
 
         <p>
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            required to implement the algrothm as presented.
+          </p>
+        </div>
+
+        <dl>
+          <dt><dfn>Stratification</dfn></dt>
+          <dd>
+            A [=stratification=] of a [=rule set=] is a sequence of sets of
+            rules.  Each rule in a [=rule set=] appears in exactly one set of
+            rules.
```
```suggestion
            rules.  Each rule in a [=rule set=] appears in exactly one [=stratification layer=].```

File: shacl12-rules/index.html [hunk 2]
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            Each set of rules in a [=stratification=] is called a 
+            [=stratification layer=], also called a "stratum".
+          </dd>
+        </dl>
+
+        <section id="stratification-condition">
+          <h4>Stratification Condition</h4>
+          <p>
+            Stratification is only defined when the following condition is
+            satisfied. If a [=rule set=] does not meet vthis condition, then this
```
```suggestion
            satisfied. If a [=rule set=] does not meet this condition, then this
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          </pre>
+        </div>
+      </section>
+      
+      <section id="stratification">
+        <h3>Stratification</h3>
+        <p>
+          Stratification is the process of assigning a label to each [=rule=] in a
+          [=rule set=].  The set of rules with the same label is called a
+          [=stratification layer=] or [=stratum=].
```
recalcitrantsupplant: Stratification is the process of partitioning a [=rule set=] into an ordered sequence of [=stratification layer=]s (strata), forming a [=stratification=].

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        </p>
+        <p>
+          A [=stratification=] is used to order the evaluation of rules.
```
```suggestion
          Stratification is the process of partitioning a [=rule set=] into an
          ordered sequence of [=stratification layer=]s (strata), forming a 
          [=stratification=]. Rules in lower strata are evaluated before rules
          in higher strata.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        </p>
+        <p>
+          Stractification introduces requirements on the relationships between
+          rules to ensure that 'NOT' rule elements depend in such a way that
+          there is a single, well-defined, output from the evaluation of a [=rule set=]
+          on a given graph.
```
```suggestion
          Stratification imposes constraints on dependencies between [=rule=]s
          to ensure that [=negation element=]s depend only on results computed
          in earlier strata, guaranteeing a single, well-defined outcome from
          the evaluation of a [=rule set=] over a given [=base graph=].
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            specification does not define the outcome of [=rule set=] evaluation.
+          </p>
+          <dl>
+            <dt><dfn>Stratification Condition</dfn></dt>
             <dd>
-              A rule `R1` depends on rule `R2` if any rule element of `R1` depends on `R2`.
-              <br/>@@formalize
+              In the [=dependency graph=] for a [=rule set=]
+              there MUST NOT be a [=transitive dependency=] involving a
+              <em>negative</em> dependency.
```
recalcitrantsupplant: should the criteria here be around recursion through negation, not transitivity?

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-            </li>
-          </ul>
-          <p>
-            A [=rule set=] is "well-formed" if and only if all of the [=rules=] of the
-            rule set are "well-formed".
+            An outcome of the stratification requirement is that when a rule
+            containing a [=negation element=]
+            is <a href="#eval-rule">evaluated</a> the data, whether in the
+            [=base graph=] or the [=inference graph=] that determines
+            the outcome of the [=negation element=] will not change.
```
```suggestion
            A consequence of the stratification requirement is that when a rule
            containing a [=negation element=] is
            <a href="#eval-rule">evaluated</a>, the data used to determine the
            outcome of that [=negation element=], whether in the
            [=base graph=] or the [=inference graph=], is fixed and will not
            change during evaluation.
```

recalcitrantsupplant (commented):
Please see my comment re. Stratification Condition - should this be defined w.r.t recursion not transitivity?

Rest of comments are rewordings.

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+                data-lt="rule body element|rule element|element"
+                >Rule body element</dfn></dt>
           <dd>
             A [=rule body element=] (often just "rule element") is any element that
             can appear in a [=rule body=];
             a [=triple pattern=],
             a [=condition expression=],
             a [=negation element=],
-            an [=assignment=],
+            or an [=assignment=]
```
TallTed: When the `Pending` line comes out of the HTML comment (and this `or an` changes to `an`), this `.` should change to a `,`.
```suggestion
            or an [=assignment=].
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          expression has a value at the point of evaluation; and that each
+          assignment in a rule introduces a new variable,
+          not used earlier in the rule body.
+        </p>
+        <div class=ednote>
           <p>
-            Stratification is the process of labelling <a>rules</a> based on their
-            <a>dependency</a> relationships in a <a>rule set</a>.
-            The set of rules with the same label are called a <a>layer</a>.
+            Add a rule for nested patterns e.g. negation.
```
```suggestion
            Add a rule for nested patterns, e.g., negation.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        </div>
+
+        <p>
+          A [=rule=] is a <dfn>well-formed rule</dfn> if all of the following
+          conditions are met:
+        </p>
+        <ul>
+          <li>For every [=variable=] appearing in a [=triple template=]
+            of the [=head=] of the [=rule=],
+            there is one or more occurrences of a [=variable=]
```
TallTed, line 483:
```suggestion
            there is at least one occurrence of a [=variable=]
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            if the [=triple pattern=] is a [=rule body element=]
+            and the dependency is <em>negative</em> if the
+            [=triple pattern=] appears in a [=negation element=].
+          </dd>
+
+          <dt><dfn>Rule dependency</dfn></dt>
+          <dd>
+            A rule `R1` depends on rule `R2` if any rule element of `R1`
+            depends on `R2`.
+            The dependency is <em>negative</em> any of the rule elements
```
```suggestion
            The dependency is <em>negative</em> if any of the rule elements
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            have negative dependency on `R2`; otherwise the dependency is
+            <em>positive</em>.
+          </dd>
+
+          <dt><dfn>Dependency graph</dfn></dt>
+          <dd>
+            A [=dependency graph=] of a [=rule set=] is a graph where the
+            vertices are rules and an edge exists where one rule R1 depends on
+            rule R2. Edge are labeled <em>positive</em> or <em>negative</em>
```
```suggestion
            rule R2. Edges are labeled <em>positive</em> or <em>negative</em>
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        <p>Notes:</p>
+        <ul>
+          <li>
+            <p>
+              A [=triple template=] with components `ts`, `tp`, `to`
+              can possibly generate a triple with component RDF terms
+              `s`, `p`, `o` if 
+              `ts` is a variable or `ts` is the
+              <a data-cite="RDF12-CONCEPTS##dfn-rdf-term-equality"
+                 >same RDF term</a> as `s`.
```
```suggestion
              A [=triple template=] with components `ts`, `tp`, `to`
              can possibly generate a triple with component RDF terms
              `s`, `p`, `o`, if 
              `ts` is a variable or `ts` is the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
                 >same RDF term</a> as `s`.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+              <br/>
+              @@ same for `p` and `o`.
+            </p>
+            <p>
+              In addition, if any of `ts`, `tp` and `to` are the same variable,
+              then the corresponding  `s`, `p` or `o` must be the same.
```
```suggestion
              In addition, if any pair of `ts`, `tp`, and `to` are the same variable,
              then the corresponding pair of `s`, `p`, and `o` must be the same.
```

shacl12-rules/index.html
TallTed: Tweaking @recalcitrantsupplant's suggestion...
```suggestion
          [=Stratification=] is the process of partitioning a [=rule set=]
          into an ordered sequence of [=stratification layers=] (also known
          as [=strata=], singular [=stratum=]), forming a [=stratification=].
          Rules in lower [=strata=] are evaluated before rules in higher
          [=strata=].
```

shacl12-rules/index.html
TallTed: Tweaking @recalcitrantsupplant's suggestion
```suggestion
          [=Stratification=] imposes constraints on dependencies between [=rules=]
          to ensure that [=negation elements=] depend only on results computed
          in earlier [=strata=], guaranteeing a single, well-defined outcome
          from the evaluation of a [=rule set=] over a given [=base graph=].
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          on a given graph.
+        </p>
+        <div class="ednote">
+          <p>
+            A stratification process may also be used to make other evaluation
+            decisions. This document describes the necessary conditions for
+            consistent evaluation and gives one possible way to form a
+            stratification.  Implementations need to meet the conditions
+            described here in order to get compatible behavior but are not
+            required to implement the algrothm as presented.
```
```suggestion
            A [=stratification=] process may also be used to make other
            evaluation decisions. This document describes the necessary
            conditions for consistent evaluation and gives one possible way to
            form a [=stratification=]. Implementations need to meet the
            conditions described here to get compatible behavior, but are not
            required to implement the algorithm as presented.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-          </ul>
-          
-          <p>Conditions</p>
-          <ul>
-            <li>For a rule, all its (rule elements) negation/aggregation dependencies must 
-              have a label that is less than the layer label of this rule.
-            </li> 
-          </ul>
+          <p>
+            In order words, there is no `NOT` used in any rule that
```
```suggestion
            In other words, there is no `NOT` used in any rule that
```

shacl12-rules/index.html
TallTed:
```suggestion
            rules.  Each rule in a [=rule set=] appears in exactly one
            [=stratum=] ([=stratification layer=]).
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        <dl>
+          <dt><dfn>Stratification</dfn></dt>
+          <dd>
+            A [=stratification=] of a [=rule set=] is a sequence of sets of
+            rules.  Each rule in a [=rule set=] appears in exactly one set of
+            rules.
+          </dd>
+          <dt><dfn 
+                data-lt="stratification layer|layer|stratum"
+                >Stratification layer</dfn></dt>
```
TallTed, line 645: I believe ReSpec will handle the "s" plurals (judging by other examples throughout this document), but perhaps not the `strata`/`stratum`. Possibly even the leading `stratification layer|` could be removed, as I *think* ReSpec will get _that_ from the content of the `<dfn>`.
```suggestion
                data-lt="stratification layer|layer|stratum|strata"
                >Stratification layer</dfn></dt>
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          <dd>
+            Each set of rules in a [=stratification=] is called a 
+            [=stratification layer=], also called a "stratum".
```
TallTed, line 648:
```suggestion
            [=stratification layer=], also called a [=stratum=]
            (plural, [=strata=]).
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          </dd>
+        </dl>
+
+        <section id="stratification-condition">
+          <h4>Stratification Condition</h4>
+          <p>
+            Stratification is only defined when the following condition is
```
```suggestion
            [=Stratification=] is only defined when the following condition is
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            satisfied. If a [=rule set=] does not meet vthis condition, then this
+            specification does not define the outcome of [=rule set=] evaluation.
+          </p>
+          <dl>
+            <dt><dfn>Stratification Condition</dfn></dt>
             <dd>
-              A rule `R1` depends on rule `R2` if any rule element of `R1` depends on `R2`.
-              <br/>@@formalize
+              In the [=dependency graph=] for a [=rule set=]
```
TallTed: But this sentence does not appear to define `Stratification Condition`‽ 
```suggestion
              In the [=dependency graph=] for a [=rule set=],
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          <h4>Stratification Algorithm</h4>
           <p>
-            Well-formedness is a set of conditions on the abstract syntax of
-            shapes rules.  Together, these rules ensure that a [=variable=] in the
-            [=head=] of a rule has a value defined in the [=body=] of the rule;
-            that each variable in an condition expression or assignment
-            expression has a value at the point of evaluation; and that each
-            assignment in a rule introduces a new variable,
-            not used earlier in the rule body.
+            The following algorithm gives one possible stratification based solely
```
TallTed, line 677:
```suggestion
            The following algorithm gives one possible [=stratification=], based solely
```

shacl12-rules/index.html
TallTed:
```suggestion
            One consequence of the [=stratification=] requirement is that when
            a rule containing a [=negation element=] is
            <a href="#eval-rule">evaluated</a>, the data used to determine the
            outcome of that [=negation element=], whether in the
            [=base graph=] or the [=inference graph=], is fixed and will not
            change further during evaluation.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        </p>
+        <p>
+          Stractification introduces requirements on the relationships between
+          rules to ensure that 'NOT' rule elements depend in such a way that
+          there is a single, well-defined, output from the evaluation of a [=rule set=]
+          on a given graph.
+        </p>
+        <div class="ednote">
+          <p>
+            A stratification process may also be used to make other evaluation
```
liviorobaldo, line 628: Such as what? Does this refer to pending work (e.g., aggregates)? Even then, I find this sentence vague. I think we should define very clearly what stratification does and what it does not do. That is why we define an explicit algorithm.

I agree with all the other text and revisions, both in this Editor's note and in the rest of the document.

afs: Stratification has all sorts of uses for evaluation and optimization. It is not possible to list them all (and there seems to be some active research in this area). We should not imply that any particular possibility other than its use in making negation stable is required.

See #767.

liviorobaldo: Ok. In that case, I would suggest phrasing this more explicitly, for example by stating that the evaluation described below is minimal, as you explained in #767, and that other extended or optimized stratification approaches that produce the same results are also acceptable, etc. 

If this starts to become too complex or overly detailed, IMHO we should remove the sentence “A stratification process may also be used to make other evaluation decisions” as I find that sentence on its own to be too vague.

shacl12-rules/index.html
TallTed: Might be better...
```suggestion
              A [=triple template=] with components `ts`, `tp`, `to`
              can possibly generate a triple with component RDF terms
              `s`, `p`, `o`, if 
              `ts` is either a variable or the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
                 >same RDF term</a> as `s`.
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
             a [=variable=] or an [=RDF term=] (which might be a [=triple term=]).
             The second element of the tuple must be an [=IRI=] or a [=variable=].
             [=Triple templates=] appear in the [=head=] of a [=rule=].
           </dd>
 
-          <dt><dfn>triple pattern</dfn></dt>
+          <dt><dfn>Triple pattern</dfn></dt>
           <dd>
             A [=triple pattern=] is 3-tuple where each element is either a
             [=variable=], or an [=RDF term=] (which might be a triple term).
```
```suggestion
            [=variable=] or an [=RDF term=] (which might be a [=triple term=]).
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-            The second element of the tuple must be an [=IRI=].
-            [=Triple patterns=] appear in the [=body=] of a [=rule=].
+            The second element of the tuple must be an [=IRI=] or a [=variable=].
+            [=Triple patterns=] appear as [=triple pattern elements=]
+            and also inside [=negation elements=].
```
```suggestion
            [=Triple patterns=] can appear as [=triple pattern elements=]
            as well as inside [=negation elements=].
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
           <dd>
             A [=condition expression=] is a function, or functional form,
             that evaluates to true or false.
             [=Condition expressions=] appear in the [=body=] of a [=rule=].
           </dd>
 
-          <dt><dfn>negation element</dfn></dt>
+          <dt><dfn>Triple pattern element</dfn></dt>
+          <dd>
+            A [=triple pattern element=] is a  [=triple pattern=]
```
```suggestion
            A [=triple pattern element=] is a [=triple pattern=]
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            used as a [=rule body element=].
+          </dd>
+
+          <dt><dfn>Negation element</dfn></dt>
           <dd>
-            A [=negation element=] is ..@@.. .
-            [=Negation elements=] appear in the [=body=] of a [=rule=].
+            A [=negation element=] is [=rule body element=].
```
```suggestion
            A [=negation element=] is a [=rule body element=].
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
              [=Aggregation elements=] appear in the [=body=] of a [=rule=].
           </dd>
+          -->
 
-          <dt><dfn data-lt="rule body element|rule element|element">rule body element</dfn></dt>
+          <dt><dfn 
+                data-lt="rule body element|rule element|element"
+                >Rule body element</dfn></dt>
           <dd>
             A [=rule body element=] (often just "rule element") is any element that
```
TallTed, line 408:
```suggestion
            A [=rule body element=] (often just [=rule element=]) is any [=element=] that
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
             can appear in a [=rule body=];
```
```suggestion
            can appear in a [=rule body=],
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          <dt><dfn data-lt="rule body|body">Rule body</dfn></dt>
           <dd> 
             A [=rule body=] is a sequence of [=rule body elements=].
           </dd>
 
-          <dt><dfn>rule</dfn></dt>
+          <dt><dfn>Rule</dfn></dt>
           <dd>
             A [=rule=] is a pair of a [=rule head=] (often just "head") and
             a [=rule body=] (often just "body").
```
TallTed, line 432:
```suggestion
            A [=rule=] is a pair of a [=rule head=] (often just [=head=]) and
            a [=rule body=] (often just [=body=]).
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+          <dt><dfn>Rule set</dfn></dt>
           <dd>
             A [=rule set=] is a collection of zero or more [=rules=]
             and a collection of zero or more [=data blocks=].
           </dd>
 
-          <dt><dfn>base graph</dfn></dt>
+          <dt><dfn>Base graph</dfn></dt>
           <dd>
             The [=base graph=] is the [=RDF Graph=] given as input to the evaluation
```
TallTed, line 443:
```suggestion
            A [=base graph=] is an [=RDF Graph=] given as input to a [=rule set=] evaluation
```

shacl12-rules/index.html
```diff
@@ -332,104 +332,114 @@ <h2>Shape Rules Abstract Syntax</h2>
...
             process.
           </dd>
 
-          <dt><dfn>inference graph</dfn></dt>
+          <dt><dfn>Inference graph</dfn></dt>
           <dd>
             The [=inference graph=] is an [=RDF Graph=] that is the outcome of rule set evaluation.
```
TallTed, line 449:
```suggestion
            An [=inference graph=] is an [=RDF Graph=] that is the outcome of a [=rule set=] evaluation process.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+      <section id="wellformed">
+        <h3>Well-formedness Conditions</h3>
+        <p>
+          Well-formedness is a set of conditions on the abstract syntax of
+          shapes rules.  Together, these rules ensure that a [=variable=] in the
+          [=head=] of a rule has a value defined in the [=body=] of the rule;
+          that each variable in an condition expression or assignment
+          expression has a value at the point of evaluation; and that each
+          assignment in a rule introduces a new variable,
+          not used earlier in the rule body.
```
```suggestion
          Well-formedness is a set of conditions on the abstract syntax of
          shape rules.  Together, these rules ensure that a [=variable=] in the
          [=head=] of a rule has a value defined in the [=body=] of the rule;
          that each variable in a [=condition expression=] or
          [=assignment expression=] has a value at the point of evaluation; and
          that each assignment in a rule introduces a new variable, i.e., one
          not used earlier in the rule body.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,238 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+        <p>
+          A [=rule=] is a <dfn>well-formed rule</dfn> if all of the following
+          conditions are met:
+        </p>
+        <ul>
+          <li>For every [=variable=] appearing in a [=triple template=]
+            of the [=head=] of the [=rule=],
+            there is one or more occurrences of a [=variable=]
+            of the same name in the [=triple patterns=] in the [=body=],
+            or in an [=assignment=] in the body, or both.
```
```suggestion
            in an [=assignment=] in the body, or both.
```

shacl12-rules/index.html
afs: Yes, they should. Thanks for catching this.

shacl12-rules/index.html
afs: The link should be [=stratification condition=]. 

Change made.

shacl12-rules/index.html
afs: I prefer the "one of more" form.

shacl12-rules/index.html
afs: This is an "ednote" and all ednotes will be removed before publication.

shacl12-rules/index.html
afs: Changed to "An edge is labeled either <em>positive</em> or <em>negative</em> according to the rule dependency."

shacl12-rules/index.html
afs: Change made along with another correction.

shacl12-rules/index.html
```diff
@@ -439,108 +449,237 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            Each [=assignment variable=] is used in only one [=assignment=] in
+            the [=body=] of the [=rule=].
+          </li>
+          <li>
+            An [=assignment variable=] at position <em>i</em> of a [=rule body=] does not
+            occur in any [=triple pattern=] at position <em>j</em> where <em>i &gt; j</em>.
+          </li>
+          <li>
+            For every [=variable=] in an [=assignment expression=] at position <em>i</em>,
+            there is a corresponding variable in a triple pattern at position
```
```suggestion
            there is a corresponding variable in a [=triple pattern=] at position
```

shacl12-rules/index.html
TallTed: The problem with the "one or more" form is that it is neither (while both) singular nor (and) plural, which makes agreement with its surroundings more difficult to keep consistent. Neither "there is one or more occurrences" nor "there are one or more occurrence" read well to me, and "there are one or more occurrences" isn't much better; "there is one or more occurrence" isn't terrible, but there's always someone who comes later and wants "more occurrences" instead of "more occurrence", but I'll still suggest this
```suggestion
            there is one or more occurrence of a [=variable=]
```

shacl12-rules/index.html
```diff
@@ -705,25 +844,33 @@ <h3>Rule Set Evaluation</h3>
...
           </dd>
 
           <dt><dfn data-lt="substitution">substitution function</dfn></dt>
           <dd>
             A <a>substitution function</a>, or just <i>substitution</i>,
-            is a function <code>subst(<var>μ</var>, [=triple pattern=])</code> that returns a triple pattern
+            is a function 
+            <code>subst(<var>μ</var>, [=triple pattern=])</code>
+            that returns a triple pattern
             where each occurrence in the triple pattern of a variable that is in the
```
```suggestion
            that returns a [=triple pattern=]
            where each occurrence in the [=triple pattern=] of a variable that is in the 
```

shacl12-rules/index.html
```diff
@@ -705,25 +844,33 @@ <h3>Rule Set Evaluation</h3>
...
-            is a function <code>subst(<var>μ</var>, [=triple pattern=])</code> that returns a triple pattern
+            is a function 
+            <code>subst(<var>μ</var>, [=triple pattern=])</code>
+            that returns a triple pattern
             where each occurrence in the triple pattern of a variable that is in the 
             <code><var>dom</var>(<var>μ</var>)</code>
-            is replaced by the [=RDF term=] given by the [=solution mapping=] for <var>var</var>.
+            is replaced by the [=RDF term=] given by the 
+            [=solution mapping=] for <var>var</var>.
             If the triple pattern result has no variables, then it is an [=RDF Triple=].
```
TallTed, line 888:
```suggestion
            If the [=triple pattern=] result has no variables, then it is an [=RDF Triple=].
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,237 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+
+      <section id="rule-dependency">
+        <h3>Rule Dependency</h3>
+
+        <dl>
+          <dt><dfn>Triple pattern dependency</dfn></dt>
+          <dd>
+            A [=triple pattern=] in the rule body of rule `R1` depends on a
+            rule `R2` if the head of `R2` can possibly generate triples that
+            match the triple pattern in `R1`.
```
```suggestion
            match the [=triple pattern=] in `R1`.
```

shacl12-rules/index.html
```diff
@@ -439,108 +449,237 @@ <h2>Shape Rules Abstract Syntax</h2>
...
+            have a negative dependency on `R2`; otherwise the dependency is
+            <em>positive</em>.
+          </dd>
+
+          <dt><dfn>Dependency graph</dfn></dt>
+          <dd>
+            A [=dependency graph=] of a [=rule set=] is a graph where the
+            vertices are rules and an edge exists where one rule R1 depends on
+            rule R2. An edge is labeled either <em>positive</em> 
+            or <em>negative</em> according to the rule dependency.
```
```suggestion
            or <em>negative</em> according to the [=rule dependency=].
```

shacl12-rules/index.html
afs: > Possibly even the leading `stratification layer|` could be removed, as I _think_ ReSpec will get _that_ from the content of the `<dfn>`.

Empirically, if `data-lt` is give it hides the `<dfn>` text, which makes sense other wise there is no way to provide a link definition when the text may be quite extensive.

afs: Not 'strata`. We will just have to be careful.

This `um`/`a` stuff is why I prefer `layer`.

shacl12-rules/index.html
afs: No really necessary as it was just linked.

shacl12-rules/index.html
```diff
@@ -439,108 +449,237 @@ <h2>Shape Rules Abstract Syntax</h2>
...
-          </ul>
-          <p>
-            A [=rule set=] is "well-formed" if and only if all of the [=rules=] of the
-            rule set are "well-formed".
+            A consequence of the [=stratification condition=] is that when 
+            a rule containing a [=negation element=] is
+            <a href="#eval-rule">evaluated</a>, the data used to determine the
+            outcome of that [=negation element=], whether in the
+            [=base graph=] or the [=inference graph=], is fixed and will not
+            change during evaluation.
```
TallTed, line 692:
```suggestion
            One consequence of the [=stratification condition=] is that when 
            a rule containing a [=negation element=] is
            <a href="#eval-rule">evaluated</a>, the data used to determine the
            outcome of that [=negation element=], whether in the
            [=base graph=] or the [=inference graph=], is fixed and will not
            change during further evaluation.
```

shacl12-rules/index.html
afs: Splits it away from "results" which is the important part.

Left as it.

shacl12-rules/index.html
afs: "further" might suggest after this point whereas stability is at this point.

shacl12-rules/index.html
afs: It is a common pattern in formal definitions.

shacl12-rules/index.html
```diff
@@ -705,25 +844,33 @@ <h3>Rule Set Evaluation</h3>
...
             A <a data-lt="sparql12-query#defn_sparqlSolutionMapping">solution mapping</a>, 
-            <var>μ</var>, is a partial function <code><var>μ</var> : <var>V</var> → <var>T</var></code>, 
-            where <var>V</var> is the set of all variables and <var>T</var> is the set of all RDF terms.
-            The domain of <var>μ</var> is denoted by <code><var>dom</var>(<var>μ</var>)</code>,
-            and it is the subset of <var>V</var> for which <var>μ</var> is defined.
-            We use the term [=solution=] where it is clear that a [=solution mapping=] is meant.
+            <var>μ</var>, is a partial function 
+            <code><var>μ</var> : <var>V</var> → <var>T</var></code>,
+            where <var>V</var> is the set of all variables 
+            and <var>T</var> is the set of all RDF terms.
```
```suggestion
            and <var>T</var> is the set of all [=RDF terms=].
```

shacl12-rules/index.html
afs: The whole section needs writing 😞  to introduce term mapping, then use that for triple patterns.

ednote added.

shacl12-rules/index.html
afs: Only one base graph can be given. The suggestion can be read to cover multiple graphs.

shacl12-rules/index.html
afs: There is only one output.

afs: Adding the [==] has the downside that it is cut off from "evaluation".

shacl12-rules/index.html
afs: "..." is quoting the word (token), not using the concept.
See another working group. 😄

shacl12-rules/index.html
afs: This is an "ednote". It does not make it into the final document.

shacl12-rules/index.html

shacl12-rules/index.html
afs: It's making it a token.

afs: Removed _element_ being a definition alternative. Too broad.

shacl12-rules/index.html
afs: This is an ednote.

shacl12-rules/index.html
afs: This is already said from an earlier suggestion you made.

shacl12-rules/index.html
afs: Reworded.

shacl12-rules/index.html

afs: Definition of "triple term" is not SHACL rules.
See note above about importing terminology when the doc is stable.

shacl12-rules/index.html

shacl12-rules/index.html

shacl12-rules/index.html
afs: Done locally + other corrections.

afs:
> Please see my comment re. Stratification Condition - should this be defined w.r.t recursion not transitivity?
> 
> Rest of comments are rewordings.

Yes - "recursion". Correction made.

recalcitrantsupplant (approved):

shacl12-rules/index.html
TallTed: OK, my misread.

shacl12-rules/index.html
TallTed: OK, but that changes only one of [three suggestions in the line above](https://github.com/w3c/data-shapes/pull/768/files#r2787557981). Adopting your correction yields:
```suggestion
            A [=base graph=] is the [=RDF Graph=] given as input to a [=rule set=] evaluation
```

shacl12-rules/index.html
TallTed: > There is only one output.

Indefinite articlces are appropriate in many places. including the two highlighted below. `The inference graph` suggests there's only one, ever; likewise, `the rule set evaluation`. I have corrected the other, `an RDF Graph`, to a definite article, also highlighted.

> Adding the [==] has the downside that it is cut off from "evaluation".

This strongly suggests to me that there should be another `<dfn>` (or alias) for `rule set evaluation`, to which the below corrections would link. This senteence might be better without the trailing `process`.

```suggestion
            An [=inference graph=] is the [=RDF Graph=] that is the outcome of a [=rule set evaluation=] process.
```

shacl12-rules/index.html
TallTed: But you said --

> Not 'strata`. We will just have to be careful.

shacl12-rules/index.html
TallTed: Why not the `[= =]`?

shacl12-rules/index.html
TallTed: If I understand this correctly, it seems that while the data may be stable ("stability is at this point"), the evaluation may not be complete ("will not change during evaluation"). On that basis, the following suggestion --
```suggestion
            One consequence of the [=stratification condition=] is that when 
            a rule containing a [=negation element=] is
            <a href="#eval-rule">evaluated</a>, the data used to determine the
            outcome of that [=negation element=], whether in the
            [=base graph=] or the [=inference graph=], is fixed upon that
            determination, and will not change further during evaluation.
```

shacl12-rules/index.html
TallTed: see below

shacl12-rules/index.html
TallTed: My grammar instructors would have fits. But I'll try to live with it.

shacl12-rules/index.html
TallTed:
```suggestion
            If the resulting [=triple pattern=] has no variables, then it is an [=RDF Triple=].
```

PR #787, opened 2026-02-14
Stratification algorithm

### LINKED ISSUE #767: Provide an algorithm to calculate a stratification

afs:
Closes #767

Stratification algorithm - enough for SNAF.

* [See this section rendered online here](https://raw.githack.com/w3c/data-shapes/afs/stratification-algorithm/shacl12-rules/index.html#stratification-algorithm)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -19,7 +19,7 @@
         // Can remove SHACL12-*?
         xref: ["RDF12-CONCEPTS", "SPARQL12-QUERY", "SHACL12-CORE", "SHACL12-SPARQL", "SHACL12-NODE-EXPR" ] ,
         lint: { "no-unused-dfns": false },
-        maxTocLevel: 2,
+//        maxTocLevel: 2,
 
         editors: [
           {
@@ -766,7 +766,7 @@ RULE { ?x :name ?FN } WHERE {
           <dl>
             <dt><dfn>Stratification Condition</dfn></dt>
             <dd>
-              The [=stratification Condition=] requires that
+              The [=stratification condition=] requires that
               there is no [=recursive dependency=] involving a
               <em>negative</em> dependency
               in the [=dependency graph=] for a [=rule set=].
@@ -786,11 +786,73 @@ RULE { ?x :name ?FN } WHERE {
           </p>
           <div class="algorithm">
             <pre>
-              @@ one possible stratification algorithm
+
+## output -- Map: Integer -> Set of rules.
+
+define stratification(ruleSet):
+
+    let DP = Dependency graph for the rule set.
+    let stratumMap be a map from rule to integer
+
+    ## The dependency graph should satisfy the stratification condition.
+    ## Check unbounded work due to a violation of the stratification condition.
+    let limit = num rules  + 1
+    let maxStratum = 0
+
+    ## initialize stratumMap
+    foreach rule in ruleSet:
+        stratumMap.set(rule, 0)
+        endfor
+
+    boolean changed = true;
+    while changed:
+        changed = false;
+        foreach edge E in DP:
+            ## Edge from pRule to qRule with a label
+            let pRule = source of edge
+            let qRule = destination of the edge
+            let label = edge label
+
+            if label == "positive" :
+                if stratumMap.get(pRule) < stratumMap.get(qRule) :
+                    stratumMap.set(pRule, stratumMap.get(qRule))
+                    changed = true;
+                    endif
+                endif
+            if label == "negative" :
+                if stratumMap.get(pRule) <= stratumMap.get(qRule) :
+                    let xStratum = 1 + stratumMap.get(qRule)
+                    if ( xStratum > limit )
+                        ## Stratification requirement violated
+                        error "Stratification error"
+                        endif
+                    stratumMap.put(pRule, xStratum)
+                    maxStratum = max(maxStratum, xStratum)
+                    changed = true;
+                    endif
+                endif
+            endfor
+        endwhile
+
+    ## Initialize the result map.
+    let stratumLevels be a map from integer to a set of rules
+    for i = 0 to maxStratum
+        stratumLevels.set(i, {})
+        endfor
+
+    ## Gather rules in stratumMap with the same level number
+    for rule R in map stratumMap:
+        let stratumNum = stratumMap.get(R)
+        let S = {R} &cups; stratumLevel.get(stratumNum)
+        stratumLevel.set(stratumNum, S)
+        endfor
+
+    the result is stratumLevels
+    endfunction
             </pre>
           </div>
 
-          <p>
+          <p class="note">
             A consequence of the [=stratification condition=] is that when 
             a rule containing a [=negation element=] is
             <a href="#eval-rule">evaluated</a>, the data used to determine the
@@ -959,10 +1021,12 @@ PREFIX sparql: &lt;http://www.w3.org/ns/sparql#&gt;
         It does not prescribe the algorithm as the method of implementation.
         An implementation can use any algorithm that generates the same outcome.
       </p>
-      <pre class="algorithm">
+      <div class="algorithm">
+<pre>
 Inputs: data graph G, called the base graph, and a rule set RS.
 Output: an RDF graph GI of inferred triples
 </pre>
+      </div>
       <p>
         The inferred triples do not include triples present in the set 
         of triples of the [=base graph=].
@@ -1084,8 +1148,8 @@ merge(S1, S2) = { <var>μ</var> |
         <h4>Evaluation of an Expression</h4>        
 
         <p class=ednote>Sketch</p>
-
-<pre class="algorithm">
+        <div class="algorithm">
+<pre>
 @@ Reference SPARQL expression evaluation <a data-cite="SPARQL12-QUERY/#expression-evaluation">Expression Evaluation</a>
 @@ Reference SPARQL EBV <a data-cite="SPARQL12-QUERY/#ebv">Effective Boolean Value (EBV)</a>
 
@@ -1097,12 +1161,13 @@ define evalFunction(F, <var>μ</var>):
     eval(F(expr1, expr2 ...), row) = F(eval(expr1, row), eval(expr2, row), ...)
     eval(FF(expr1, expr2) , row) = ... things that are not functions like `IF`
 </pre>
+        </div>
       </section>
 
       <section id="eval-rule">
         <h4>Evaluation of a Rule</h4>
-
-<pre class="algorithm">
+        <div  class="algorithm">
+<pre>
 let R be a well-formed rule.
 
 let rule R = (H, B) where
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant:
Should the spec also cover an algorithm to generate the dependency graph or is this better left to implementations?

I can draft one based on the implementation I have

afs:
The PR on which this based on #768 (recently merged) that defines [dependency graph](https://www.w3.org/TR/shacl12-rules/#dfn-dependency-graph) and [rule dependency](https://www.w3.org/TR/shacl12-rules/#dfn-rule-dependency).

The dependency graph is defined by +ve/-ve immediate rule dependency, not path following. (Sometimes dependency graph includes reachablility). So between the definitions, I hope it's clear.

If you think an algorithm would be helpful, then, yes,  please do make a PR.

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+    let DP = Dependency graph for the rule set.
+    let stratumMap be a map from rule to integer
+
+    ## Safety condition for violation of the stratification condition
+    let limit = num rules  + 1
+    let maxStratum = 0
+
+    ## initialize stratumMap
+    foreach rule in ruleSet:
+        stratumMap[rule] = 0
```
liviorobaldo:
```suggestion
stratumMap.put(rule, 0)
```

For consistency with the format used below. stratumMap is a map, not an array.

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+                    changed = true;
+                    endif
+                endif
+            endfor
+        endwhile
+
+    ## Initialize the result map.
+    let stratumLevels be a map from integer to a set of rules
+    for i = 0 to maxStratum
+        stratumLevels[i]= {}
```
liviorobaldo:
```suggestion
        stratumLevels.put(i, {})
```

As above. Also stratumLevels is a map.

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+        endfor
+
+    ## Gather rules in startumMap with the same level number
+    for rule R in map stratumMap:
+        let stratumNum = startumMap[R]
```
liviorobaldo:
```suggestion
        let stratumNum = stratumMap.get(R)
```

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+        add R to the set stratumLevels[stratumNum]
```
liviorobaldo:
```suggestion
        add R to the set stratumLevels.get(stratumNum)
```

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+            endfor
+        endwhile
+
+    ## Initialize the result map.
+    let stratumLevels be a map from integer to a set of rules
+    for i = 0 to maxStratum
+        stratumLevels[i]= {}
+        endfor
+
+    ## Gather rules in startumMap with the same level number
```
liviorobaldo:
```suggestion
    ## Gather rules in stratumMap with the same level number
```

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+            if label == "positive" :
+                if stratumMap.get(pRule) < stratumMap.get(qRule) :
+                    stratumMap.put(pRule, stratumMap.get(qRule))
+                    changed = true;
+                    endif
+                endif
+            if label == "negative" :
+                if stratumMap.get(pRule) <= stratumMap.get(qRule) :
+                    let xStratum = 1 + stratumMap.get(qRule)
+                    if ( xStratum > limit )
```
liviorobaldo, line 825: I believe this condition can be true, and thus the exception raised, only if the input dependency graph DP is ill-formed. If DP is well formed, it should never be the case that maxStratum exceeds the limit. Therefore, we could explicitly state above that DP is assumed to be a well-formed dependency graph, constructed using the algorithm that @recalcitrantsupplant also proposed to define.

shacl12-rules/index.html
```diff
@@ -679,11 +679,71 @@ <h4>Stratification Algorithm</h4>
...
+        changed = false;
+        foreach edge E in DP:
+            ## Edge from pRule to qRule with a label
+            let pRule = source of edge
+            let qRule = destination of the edge
+            let label = edge label
+
+            if label == "positive" :
+                if stratumMap.get(pRule) < stratumMap.get(qRule) :
+                    stratumMap.put(pRule, stratumMap.get(qRule))
```
liviorobaldo: I do not understand this. If qRule produces triples that are matched in the body of pRule by positive elements (which I understand is the definitio of positive dependency), then qRule should be executed _before_ pRule. If both rules are placed in the same stratum, their execution order is unconstrained.

Although, in the presence of only positive dependencies, it would in principle be sufficient to re-execute all rules until saturation, my understanding is that this stratification algorithm not only prevents nondeterminism caused by negative dependencies, but also avoids iterative re-execution. In particular, it ensures that, even for positive dependencies, rules that produce triples are executed _before_ the rules that consume them.

Therefore, also in this case, we should set:

`stratumMap.put(pRule, stratumMap.get(qRule)+1)`

or am I missing something?

afs: > If qRule produces triples that are matched in the body of pRule by positive elements (which I understand is the definitio of positive dependency), then qRule should be executed _before_ pRule. If both rules are placed in the same stratum, their execution order is unconstrained.

Within a stratum, order does not matter - rules are not executed exactly once, they are executed until a fixed point is reached.

So the condition to update pRule if qRule itself requires a high level that pRule (i.e. place pRule into the stratum for qRule).

Consider where pRule and qRule are mutually dependent i.e. there is a cyclic dependency. There is no ordering where all of pRule can be done before qRule.

> my understanding is that this stratification algorithm not only prevents nondeterminism caused by negative dependencies, but also avoids iterative re-execution.

The stratification algorithm given is only to separate into strata for negation.

("nondeterminism" isn't accurate here - that means an operation _with the same arguments and same data_ returns different answers e.g. `BNODE()`; here, the _data_ would be changing as could be the argument from earlier in the rule body. What it is stable, or idempotent - later work will not invalidate a negation element in the sense that positive becomes negative. Even rules with a negation element may be executed more than once.

The standard [naive algorithm](https://en.wikipedia.org/wiki/Datalog#Na%C3%AFve_evaluation) 
(which, with early propagation, is what [5.4 RuleSet evaluation](https://www.w3.org/TR/shacl12-rules/#eval-rule-set)) evaluates a stratum until fixed point; rules are executed until there is no change.

It is possible to do better in many ways, but that's for implementation optimization.

> or am I missing something?

I would find it easier to discuss if you'd give a counterexample.

shacl12-rules/index.html
afs: > If DP is well formed, it should never be the case that maxStratum exceeds the limit.

Correct.
The next line says:
> `                        ## Stratification requirement violated`

> Therefore, we could explicitly state

Done.

By the way: well-formed is something else. A dependency graph has no conditions. it is straification that makes a requirement on the dependency graph.

A "well-formed dependency graph"
-> 
A "dependency graph that satisfies the stratification condition"

shacl12-rules/index.html
afs: A consistent notation should be used but there are choices.

Python has `[ ]` (and `get` but `[ ]` is more idiomatic?) 

As long as `{ }` is for sets.

liviorobaldo: Python uses `[...]` to access elements in both lists and dictionaries: in lists, the brackets contain a numeric index, while in dictionaries (or maps) they contain a key, typically a string. In many other programming languages, `[...]` is used only for arrays. For this reason, I find get less confusing.

Whatever we choose, we agree we've to consistently use that.

shacl12-rules/index.html
liviorobaldo: Ah, ok, I missed that within a stratum the rules are executed until saturation. The algorithm now makes sense to me.

> Consider where pRule and qRule are mutually dependent i.e. 
> there is a cyclic dependency.

In a well-formed dependency graph (that is, a dependency graph that satisfies the stratification condition), there should be no cyclic dependencies, correct? In other words, the dependency graph should be a DAG. Is that right?

afs: > In a well-formed dependency graph (that is, a dependency graph that satisfies the stratification condition), there should be no cyclic dependencies, correct?

No. Valid recursion is a cycle of one or more positive edges in the dependency graph.

Minimally:
```
RULE { ?x rdfs:subClassOf ?z } WHERE { ?x rdfs:subClassOf ?y . ?y rdfs:subClassOf ?z }
```
the dependency graph has a cycle of length 1 - there is a positive edge in the dependency graph with both ends being the rule.

The stratification condition only restricts cycles with one or more edge labels being negative.

shacl12-rules/index.html
```diff
@@ -845,10 +906,12 @@ <h3>Rule Set Evaluation</h3>
...
-      <pre class="algorithm">
+      <div class="algorithm">
+<pre>
 Inputs: data graph G, called the base graph, and a rule set RS.
 Output: an RDF graph GI of inferred triples
 </pre>
+      </div>
       <p>
         The inferred triples do not include triples present in the set 
         of triples of the [=base graph=].
```
TallTed, line 1032:
```suggestion
        of triples in the [=base graph=].
```

shacl12-rules/index.html
afs: Done, with `.set` to pair with `.get`

shacl12-rules/index.html
afs: Done.

shacl12-rules/index.html
```diff
@@ -786,11 +786,73 @@ <h4>Stratification Algorithm</h4>
...
           <div class="algorithm">
             <pre>
-              @@ one possible stratification algorithm
+
+## output -- Map: Integer -> Set of rules.
+
+define stratification(ruleSet):
+
+    let DP = Dependency graph for the rule set.
+    let stratumMap be a map from rule to integer
```
TallTed, line 795: Why sometimes `=` and sometimes `be`?

afs: The first is assignment to a specific value.
The second is a statement about the data structure.

shacl12-rules/index.html
```diff
@@ -786,11 +786,73 @@ <h4>Stratification Algorithm</h4>
...
+    ## Initialize the result map.
+    let stratumLevels be a map from integer to a set of rules
+    for i = 0 to maxStratum
+        stratumLevels.set(i, {})
+        endfor
+
+    ## Gather rules in stratumMap with the same level number
+    for rule R in map stratumMap:
+        let stratumNum = startumMap.get(R)
+        let S = {R} &cups; startumLevel.get(stratumNum)
```
```suggestion
        let stratumNum = stratumMap.get(R)
        let S = {R} &cups; stratumLevel.get(stratumNum)
```

recalcitrantsupplant (approved):

shacl12-rules/index.html

PR #754, opened 2026-02-02
Profiling data

nicholascar:
This PR starts the section "Profiling Data" but does not yet complete it.

The section needs an outcome or final salient word of some sort, otherwise it's a "so what??".

A new very small section on [Indicating Validity](https://raw.githack.com/w3c/data-shapes/profiling-data/shacl12-profiling/index.html#indicating-validity) is also added. What do you think?

* See the doc here: https://raw.githack.com/w3c/data-shapes/profiling-data/shacl12-profiling/index.html
  * in particular, the section [Profiling Data](https://raw.githack.com/w3c/data-shapes/profiling-data/shacl12-profiling/index.html#profiling-data)

PR #790, opened 2026-02-16
Profiling of Standards section

nicholascar:
Initial content for the Profiling of Standards (now renamed to Profiling of Specifications) section.

RDF code styling will be picked up by the merger of PR #754

* See the doc here: https://raw.githack.com/w3c/data-shapes/profiling-standards/shacl12-profiling/index.html#
    * in particular, the section [Profiling Specifications](https://raw.githack.com/w3c/data-shapes/profiling-standards/shacl12-profiling/index.html#profiling-specifications)


shacl12-profiling/img/profiling-specifications.svg
YoucTagh, line 1: Why does the blank node have three `role:validation`?

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      <p>
+        However, <code>SG1</code> is also playing the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a>,
+        so a more complete [[dx-prof]] characterisation of the situation would be:
+      </p>
+      <div class="turtle">
+  &lt;SG1>
+    a prof:Profile ;
+    prof:isProfileOf &lt;S1> ;
+    prof:hasResource [
+      prof:hasArtifact &lt;S1> ;
```
mgberg: Shouldn't the artifact (here and in subsequent examples) be the profile instead of the specification being profiled, i.e.
```suggestion
      prof:hasArtifact &lt;SG1> ;
```

nicholascar:
Use language akin to that in Core, e.g. [the definition of `sh:shape`](https://w3c.github.io/data-shapes/shacl12-core/#explicit-shape-target)

shacl12-profiling/img/key.svg
ajnelson-nist, line 1: Style request - can the inferred predicate line be green instead of red?  Using green draws from the `diff` visual scheme, where green indicates something added, and red something removed.  Inference of predicates is addition of statements.

ajnelson-nist, line 1: (Comment shifted to new thread - sorry, pushed wrong button earlier.)

An accessibility point: The colors for the "SHACL Class" and `owl:NamedIndividual` nodes are difficult to distinguish, at least to my eye on the "dark mode" background.

shacl12-profiling/img/profiling-specifications.svg
ajnelson-nist, line 1: One node-class point: I wasn't aware `dcterms:Standard` had an OWL casting.  I thought Dublin Core Terms just used RDFS.  I'm having trouble finding an OWL file from them containing `Standard`; I can see dcterms' RDFS-but-not-OWL schema [here](https://www.dublincore.org/schemas/rdfs/), and BIBO containing `bibo:Standard` [here](https://www.dublincore.org/specifications/bibo/).  Is there a `dcterms:Standard` OWL class posted by Dublin Core?

Saving on distraction energy: The figure-key could leave this matter aside by just defining the white box as "OWL or RDFS Class".

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+    <p class="note">
+      As per [[[dx-prof]]], we take the term <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>
+      to be synonymous with the term <em>Standard</em>.
+    </p>
+    <figure>
+      <a href="img/profiling-specifications.svg">
+        <img src="img/profiling-specifications.svg" alt="Figure of profiling of specifications example [[dx-prof]] characterization" />
+      </a>
+      <figcaption>
+        Figure of profiling of specifications example [[dx-prof]] characterization
```
ajnelson-nist: This caption was pretty choppy to me.  I offer this alternative wording, but it might need another tweak.

```suggestion
        Figure of profiling of specifications, characterizing an example per the style guidance of [[dx-prof]]
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      </figcaption>
+    </figure>
+    <section>
+      <h4>Characterizing the profiling situation</h4>
+      <p>
+        SHACL is often used to test that data matches a known <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>.
+        If the SHACL used for such testing implements no rules other than those in the specification, the SHACL
+        <a data-cite="shacl12-core/#dfn-shapes-graph">shapes graph</a> creates a <a>null profile</a> of the specification.
```
ajnelson-nist, line 868: This term "null profile" - is this part of dx-prof's vocabulary, or is it a term SHACL Profiling is devising?

shacl12-profiling/index.html
ajnelson-nist: I concur with @mgberg .

ajnelson-nist: `S1` is not in the role of validating `S1`.  `SG1` is.

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      <div class="turtle">
+  &lt;SG1>
+    a prof:Profile ;
+    prof:isProfileOf &lt;S1> ;
+    prof:hasResource [
+      prof:hasArtifact &lt;S1> ;
+      prof:hasRole
+        role:constraints ,
+        role:specification ,
+        role:validation ,
```
ajnelson-nist: Syntax fix.

```suggestion
        role:validation ;
```

shacl12-profiling/index.html
ajnelson-nist: And, it looks OK...I think...that `SG1` is the artifact-part of the resource that `SG1` has.

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      </div>
+      <p>
+        However, <code>SG1</code> is also playing the role of <a href="https://www.w3.org/TR/dx-prof/#Role:validation">Validation</a>,
+        so a more complete [[dx-prof]] characterisation of the situation would be:
+      </p>
+      <div class="turtle">
+  &lt;SG1>
+    a prof:Profile ;
+    prof:isProfileOf &lt;S1> ;
+    prof:hasResource [
```
ajnelson-nist, line 892: @nicholascar , I'd appreciate one other clarification - these blank nodes that are the objects of `prof:hasResource` - could they please include the most specific `rdf:type` values applicable in each of these example boxes?

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      <h4>Implications for data</h4>
+      <p>
+        If a <a data-cite="shacl12-core/#data-graph">data graph</a>, <code>DG1</code> is valid according to a <a data-cite="shacl12-core/#shapes-graph">shapes graph</a>,
+        <code>SG1</code>, then using the general [[DCTERMS]] definition of <code>dcterms:conformsTo</code>, as used by [[dx-prof]], we can then say:
+      </p>
+      <div class="turtle">
+&lt;DG1> dcterms:conformsTo &lt;SG1> .
+      </div>
+      <p>
+        Following the [[dx-prof]] axiom of <code>dct:conformsTo owl:propertyChainAxiom ( prof:isProfileOf dct:conformsTo )</code> then, if given
```
ajnelson-nist: OWL point: This property chain axiom only holds if `dct:conformsTo` is an `owl:ObjectProperty`, because OWL syntax requires everything in the list that is the object of `owl:propertyChainAxiom` must be an object property.  (Syntax source: See all the 1st and 5th occurrence of `propertyChainAxiom` on [this page](https://www.w3.org/TR/owl2-mapping-to-rdf/).)
I'm not aware of Dublin Core Terms having put this typing-statement out.

ajnelson-nist: Consistency point: `dcterms:` is the prefix used in the following example lines, but `dct:` is the prefix used here.

ajnelson-nist: Another OWL point: It seems the property chain axiom's list should be reversed.

I checked this example against pattern `prp-spo2` from [OWL2-Profiles Reasoning rules](https://www.w3.org/TR/owl2-profiles/#Reasoning_in_OWL_2_RL_and_RDF_Graphs_using_Rules):

```
If ...

T(?p, owl:propertyChainAxiom, ?x)
LIST[?x, ?p1, ..., ?pn]
T(?u1, ?p1, ?u2)
T(?u2, ?p2, ?u3)
...
T(?un, ?pn, ?un+1)

... Then:

T(?u1, ?p, ?un+1)
```

With the given axiom `dct:conformsTo owl:propertyChainAxiom ( prof:isProfileOf dct:conformsTo )`, here are the bindings:

| Variable | Value |
| --- | --- |
| `?p` | `dcterms:conformsTo` |
| `?x` | `_:x` |
| `n` | `2` |
| `?p1` | `prof:isProfileOf` |
| `?p2` | `dcterms:conformsTo` |
| `?u1` | `<DG1>` |
| `?u2` | `<SG1>` |
| `?u3` | `<S1>` |

And here is the rule expansion:

```
T(dcterms:conformsTo, owl:propertyChainAxiom, _:x)
LIST[_:x, prof:isProfileOf, dcterms:conformsTo]
T(<DG1>, prof:isProfileOf, <SG1>)
T(<SG1>, dcterms:conformsTo, <S1>)

->

T(<DG1>, dcterms:conformsTo, <S1>)
```

In English: "If the data graph is a profile of the shapes graph, and the shapes graph conforms to the specification, then the data graph conforms to the specification."

So, it seems the property chain axiom is backwards.

ajnelson-nist (changes requested):
The longer OWL point I noted on line 1033 needs review.

shacl12-profiling/img/profiling-specifications.svg
mgberg: I'm not aware of any OWL definitions posted by Dublin Core, I've tried looking for that before. I agree with the comment about the key though.

shacl12-profiling/index.html
mgberg: I tried to find a clear source of a definition for that term but didn't have a lot of luck. I did find a few places that used it though:
- A note from the DXWG (despite the fact it doesn't show up in the current version of DX-PROF): https://www.w3.org/2017/dxwg/wiki/ProfileContext
- https://www.w3.org/TR/2021/CR-ttml2-20210309/#profile-vocabulary-profile
- https://geoscienceaustralia.github.io/fsdf-supermodel/supermodel#null-profile-defn

shacl12-profiling/index.html
mgberg: I thought that property chain looked fishy but I didn't double check it because DX-PROF has the same one.

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      prof:hasArtifact &lt;S1> ;
+      prof:hasRole role:validation ;
+    ] ;
+  .
+      </div>
+      <p>
+        This is unexpected use of [[dx-prof]] where the profile is also playing a role usually performed by a part of,
+        or one of, the resources of the profile rather than a single profile resource playing all roles, but it is legitimate
+        and will still achieve the PROF goal of allowing a conceptual profile to indicate a resource that performs a common
+        profiling resource role - validation.
```
TallTed:
```suggestion
        This is an unexpected use of [[dx-prof]], where the profile is also playing a role usually performed by a part of,
        or one of, the resources of the profile, rather than having a single profile resource playing all roles, but it is legitimate,
        and will still achieve the PROF goal of allowing a conceptual profile to indicate a resource that performs a common
        profiling resource role — validation.
```

shacl12-profiling/index.html
```diff
@@ -934,7 +934,7 @@ <h3>Profiling Specifications</h3>
         <img src="img/profiling-specifications.svg" alt="Figure of profiling of specifications example [[dx-prof]] characterization" />
       </a>
       <figcaption>
-        Figure of profiling of specifications example [[dx-prof]] characterization
+        Figure of profiling of specifications, characterizing an example per the style guidance of [[dx-prof]]
```
TallTed:
```suggestion
        Figure of profiling of specifications, characterizing an example following the style guidance of [[dx-prof]]
```

shacl12-profiling/img/key.svg
TallTed: Also note that a18y guidance includes not using color alone to distinguish meaning in illustrations. While typography is often too thin to use patterns or shading, the background color of the text items *often* can be employed, using shading and/or patterns to distinguish their meanings. Sometimes it's better to put that meaning into visible words, even though that may substantially enlarge some elements of the illustration.

In addition, contrast evaluation should be applied to all text + background colors, to increase readability. `PROF Class` here should probably have white text; likewise with the three dark lozenges in `shacl12-profiling/img/profiling-specifications.svg`

shacl12-profiling/index.html
```diff
@@ -507,16 +518,24 @@ <h3>Document Conventions</h3>
...
       Grey boxes such as this include syntax rules that apply to the <a data-cite="shacl12-core/#dfn-shapes-graph">shapes graph</a>.
     </p>
-
     <p>
       <code>true</code> denotes the RDF term <code>"true"^^xsd:boolean</code>.
       <code>false</code> denotes the RDF term <code>"false"^^xsd:boolean</code>.
     </p>
-
+    <p>
+      Conventions within figures are as per this key:
```
TallTed:
```suggestion
      Conventions within figures are described by this key:
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
   <p>
     This section describes how to create profiles of different kinds of things with SHACL.
   </p>
-  <section id="profiling-standards">
-    <h3>Profiling Standards</h3>
-    <p class="todo">Incomplete</p>
+  <section id="profiling-specifications">
+    <h3>Profiling Specifications</h3>
+    <p class="note">
+      As per [[[dx-prof]]], we take the term <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>
```
TallTed:
```suggestion
      Conforming with [[[dx-prof]]], we take the term <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+        <img src="img/profiling-specifications.svg" alt="Figure of profiling of specifications example [[dx-prof]] characterization" />
+      </a>
+      <figcaption>
+        Figure of profiling of specifications, characterizing an example following the style guidance of [[dx-prof]]
+      </figcaption>
+    </figure>
+    <section>
+      <h4>Characterizing the profiling situation</h4>
+      <p>
+        SHACL is often used to test that data matches a known <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>.
```
TallTed:
```suggestion
        SHACL is often used to test whether data matches a known <a href="https://www.w3.org/TR/dx-prof/#Role:specification">Specification</a>.
```

shacl12-profiling/index.html
TallTed: "null profile" seems to be a term drawn from some discipline, which might be defined in an educational or business context, largely with an "if you know, you know" mentality. If it *has* a good definition somewhere, we absolutely should link to it. I think [this is the closest to "good"](https://linked.data.gov.au/def/fsdf-supermodel#fsdf-defn) that I've seen. If anyone has suggestions of improvement, maybe we can help others by applying such and generating our own definition.

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+  .
+      </div>
+      <p>
+        This is an unexpected use of [[dx-prof]], where the profile is also playing a role usually performed by a part of,
+        or one of, the resources of the profile, rather than having a single profile resource playing all roles, but it is legitimate,
+        and will still achieve the PROF goal of allowing a conceptual profile to indicate a resource that performs a common
+        profiling resource role — validation.
+      </p>
+      <p>
+        To indicate that profile resource is using SHACL (Core) and available in Turtle, the following can also be added to the characterisation:
```
TallTed: Are our docs (and this one in particular) using US or UK spelling?
```suggestion
        To indicate that a profile resource is using SHACL (Core) and available in Turtle, the following can also be added to
        the characterisation:
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+    prof:hasResource [
+      prof:hasArtifact &lt;S1> ;
+      prof:hasRole role:validation ;
+      dcterms:conformsTo &lt;https://www.w3.org/TR/shacl12-core/> ;
+      dcterms:format "text/turtle" ;
+    ] ;
+  .
+      </div>
+      <p>
+        If <code>SG1</code> is also considered the normative expression of the profile's requirements, as opposed to perhaps a
```
TallTed, line 921:
```suggestion
        If <code>SG1</code> is also considered to be the normative expression of the profile's requirements, as opposed to perhaps a
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+        separate human language document containing them, then the <code>prof:hasRole</code> value of <code>role:constraints</code>
+        may also be applied to the resource.
+      </p>
+      <p>
+        If there is no separate human language document about the profile, then the <code>prof:hasRole</code> value of
+        <code>role:specification</code> may also be applied to the resource, but this is not optimal, as the definition of
+        <code>role:specification</code> is that it "[defines] the profile in human-readable form" and while SHACL is
+        human-readable, natural language prose is expected.
```
TallTed:
```suggestion
        <code>role:specification</code> may also be applied to the resource. This is not optimal, however, as
        <code>role:specification</code> is specified to "[define] the profile in human-readable form" and while SHACL is
        human-readable, natural language prose is expected.
```

shacl12-profiling/index.html
```diff
@@ -904,9 +923,131 @@ <h2>Profiling with SHACL</h2>
...
+      dcterms:format "text/turtle" ;
+    ] ;
+  .
+      </div>
+    </section>
+    <section>
+      <h4>Implications for data</h4>
+      <p>
+        If a <a data-cite="shacl12-core/#data-graph">data graph</a>, <code>DG1</code> is valid according to a <a data-cite="shacl12-core/#shapes-graph">shapes graph</a>,
+        <code>SG1</code>, then using the general [[DCTERMS]] definition of <code>dcterms:conformsTo</code>, as used by [[dx-prof]], we can then say:
```
TallTed:
```suggestion
        If a <a data-cite="shacl12-core/#data-graph">data graph</a>, <code>DG1</code>, is valid according to a
        <a data-cite="shacl12-core/#shapes-graph">shapes graph</a>, <code>SG1</code>, then using the general
        [[DCTERMS]] definition of <code>dcterms:conformsTo</code>, as used by [[dx-prof]], we can say
        the following:
```

shacl12-profiling/img/key.svg
nicholascar: Certainly, done with Commit https://github.com/w3c/data-shapes/pull/790/commits/1c66af822851d968fe2b23074c069abc6037665b

shacl12-profiling/img/profiling-specifications.svg
nicholascar: Ooops, fixed now with Commit https://github.com/w3c/data-shapes/pull/790/commits/ae4117c02097e42b9429d0fc0c4caed50a75f24b

shacl12-profiling/img/key.svg
nicholascar: Addressed by using a pattern for SHACL Class and white text in Commit https://github.com/w3c/data-shapes/pull/790/commits/fc7566a5461404169f9d9ef4ef88f66ff238bdd6

shacl12-profiling/img/profiling-specifications.svg
nicholascar: You're right: only RDFS.

Great suggestion to save effort with the "OWL or RDFS Class". Implemented in Commit https://github.com/w3c/data-shapes/pull/790/commits/b5d18c39f35836aa7574e9ffa1088142f9d014ca

shacl12-profiling/index.html
nicholascar: I've improved the definition of "null profile" in this doc and I've added Issues to DX-PROF to adopt this definition and to provide ontology elements to indicate a null profile - `prof:isNullProfileOf` - in Commit https://github.com/w3c/data-shapes/pull/790/commits/f2de378cd6fae9d9b45f20672e5cf333099e6174.

Please raise new issues if the problem isn't resolved yet.

shacl12-profiling/index.html
nicholascar: They are all just `prof:ResourceDescriptor`, as per https://www.w3.org/TR/dx-prof/#Class:ResourceDescriptor.

So they are just containers to describe a resource with - artifact, format, role etc. - and don't have interesting classes of their own. Like a `prov:qualifiedAttribution` which links an `Entity` to an `Agent` with a role but the `Attribution` object itself isn't really interesting.

shacl12-profiling/index.html
nicholascar: This is true. Let's leave the improvement of the axiom to PROF! Either we can state that it only works in the all-OP case, or we subProperty `dcterms:conformsTo` with something like `prof:conformsTo` to ensure it does.

nicholascar: The axiom here and in PROF is correct, it's was my English that was mangled! Ted's fixed it now so this is resolved.

shacl12-profiling/index.html
ajnelson-nist: Cross-referencing, no further action needed:
https://github.com/w3c/dx-prof/issues/56

Thanks, @nicholascar !

shacl12-profiling/index.html
ajnelson-nist: Ok, I see that from the illustrated figure (2) now.

shacl12-profiling/index.html
ajnelson-nist: @nicholascar : I disagree.  The property chain axiom's sequence is backwards.

I looked over the usage of the property chain axiom in the PROF vocabulary, and I saw that in one place, it is _verbalized_ (/semi-encoded) correctly.  The header of [example 3](https://www.w3.org/TR/dx-prof/#eg-conformance-axiom) reads "Demonstrating conformance via the vocabulary's dct:conformsTo/prof:isProfileOf property chain.".  However, everywhere else in the document, including in the code body of that example, the axiom is spelled as you quoted: `dcterms:conformsTo owl:propertyChainAxiom ( prof:isProfileOf dcterms:conformsTo )`.  I worked through the OWL expansion above to fully bind all the variables and demonstrate the errant reversal, but maybe we need another example.

[The OWL Primer Section 6.2](https://www.w3.org/TR/owl2-primer/#Property_Chains) shows a property chain axiom for `hasGrandparent` (visible when pushing the "Show Turtle Syntax" button at the end of Section 1.2):

```turtle
:hasGrandparent  owl:propertyChainAxiom  ( :hasParent  :hasParent ) .
```

I'm going to adapt that to not use a repeated property:

```turtle
:hasAunt  owl:propertyChainAxiom  ( :hasParent  :hasSister ) .
```

With that property chain axiom, we can follow this if--then inference chain, using indentation to illustrate what the chain's shortcutting:

```turtle
# Given:
:Alice :hasParent :Bob .
                  :Bob :hasSister :Carol .

# Infer:
:Alice          :hasAunt          :Carol .
```

The property chain does not work in the other order:

```turtle
:hasAuntButWrong  owl:propertyChainAxiom  ( :hasSister :hasParent ) .

# Given:
:Alice :hasSister :Darcy .
                  :Darcy :hasParent :Bob .

# Infer:
:Alice       :hasAuntButWrong       :Bob .
```

Going back to what is in the document under revision in this PR, this is the illustration of the chain as of the current state ([`2276470`](https://github.com/w3c/data-shapes/pull/790/changes/22764705581b4ffaaade1bf2d5d585693df08576)):

```turtle
<DG1> dcterms:conformsTo <SG1> .
                         <SG1> prof:isProfileOf <S1> .

# To claim <DG1> dcterms:conformsTo <S1>, we need:
dcterms:conformsTo owl:propertyChainAxiom (
      dcterms:conformsTo       prof:isProfileOf
) .
```

How to resolve this: The backwards chain matter stems from PROF, where I see the property chain axiom is marked "at risk."  I believe it would be low risk to the PROF spec to correct the sequence's order, and improve its current state.  If you don't object, I'll forward this thread to their tracker.

ajnelson-nist (changes requested):
I'm only blocking on the property chain axiom thread.  The rest looks fine to me.

shacl12-profiling/index.html
mgberg: If the property chain axiom is marked as risk, would it be better to express this here as a property path instead?

mgberg (changes requested):
Looks good to me once the property chain thread is resolved.

shacl12-profiling/img/key.svg
TallTed, line 1: Much better. A somewhat larger white block behind `SHACL Class` would improve readability there. I suggest a similar white block behind `owl:NamedIndividual`. You could do the same with *all* the text elements, atop whatever shape or arrow. The yellowish tinge of `inferred predicate` is odd. As this stands, I would stick with black for all text, layered with the white block just behind, and then the shapes and arrows. Sorry I didn't make all these suggestions before.

shacl12-profiling/index.html
```diff
@@ -1109,7 +1040,8 @@ <h3>Creating Null Profiles</h3>
       The main purpose of creating a null profile of a specification using SHACL is to enable the testing of conformance
       to that specification by SHACL validation. This purpose exists because many RDF data models exist that have a
       model specification, perhaps an OWL model or only a natural-language document of a model, but do not provide a
-      mechanism for data validation, such as a SHACL validator.
+      mechanism for data validation, such as a SHACL validator. Examples of such models are: [[[vocab-dcat-3]]], [[[schema-org]]],
+      [[[vocab-data-cube]]] and many W3C standards.
```
TallTed, line 620:
```suggestion
      [[[vocab-data-cube]]], and many W3C standards.
```

shacl12-profiling/index.html
```diff
@@ -1126,6 +1058,11 @@ <h3>Creating Null Profiles</h3>
         SPARQL queries, allowing for RDF data validation.
       </p>
     </div>
+    <p>
+      To create a null profile of a specification, no universal and precise methodology can be given for, by the task's
+      very nature, the specification's rule may be imprecisely defined and thus the method used to characterise them in
+      SHACL, which is precise, will vary just as the imprecise rule definition does.
```
TallTed, line 658:
```suggestion
      No universal and precise methodology can be given for creation of a null profile of a specification,
      because, by the task's very nature, the specification's rule(s) may be imprecisely defined, and thus
      the method used to characterize them in SHACL, which is precise, will vary just as the imprecise rule definition does.
```

shacl12-profiling/index.html
```diff
@@ -1047,11 +973,18 @@ <h4>Implications for data</h4>
       <div class="turtle">
 &lt;DG1> dcterms:conformsTo &lt;S1> .
       </div>
+      <p>
+        This reasoning allows for static conformance claims of data to profiles of specifications and specifications
+        themselves to be made which will have utility for data whose conformance to these things must be known but which
+        cannot be continuously validated.
```
TallTed, line 980: This definitely needs some punctuation and rephrasing, but, lacking those, it's complicated enough that I am unsure of the intended meaning, so I cannot suggest revision yet.

shacl12-profiling/img/key.svg
nicholascar: I've expanded the white block in Commit https://github.com/w3c/data-shapes/pull/790/commits/e6842ba11b937f9277217245e08d43671f96ee26

nicholascar: I've altered the boldness of the diagrams and changes the green used for the inferred predicate. I've used black text for the inferred predicate: Commit https://github.com/w3c/data-shapes/pull/790/commits/6a0b24cd711231bc2a3d5a902511a82fa65e473a

shacl12-profiling/index.html
nicholascar: OK, got it now! Yes, I agree that the correct form is 

```turtle
dcterms:conformsTo 
    owl:propertyChainAxiom ( 
        dcterms:conformsTo 
        prof:isProfileOf 
) .
```

When I was arguing with you above, I was reading a local copy of PROF which is set that way but, on review, yes all the instances in PROF are incorrect. I reread the OWL spec, the aunty example, and indeed couldn't understand the problem raised as I thought it was stated correctly...

I triple checked by recreating the axiom in protege:

image

image

...and indeed created that way, it comes out as expected too.

I've fixed this in this spec here:  Commit https://github.com/w3c/data-shapes/pull/790/commits/fc8f857cea1ed8fd86cb272d495914b2f59ebeab

I've raised a ticket for fixing PROF here: Issue https://github.com/w3c/dx-prof/issues/58

While I'm at it, I've raised a ticket to replace the use of `dcterms:conformsTo`: Issue https://github.com/w3c/dx-prof/issues/58


Thanks for sticking with the correct call here @ajnelson-nist and @mgberg

ajnelson-nist: (Looks like the second issue you meant to link was [dx-prof 59](https://github.com/w3c/dx-prof/issues/59)?)

I appreciate the chance to discuss.  Thank you, @nicholascar and @mgberg .

ajnelson-nist (approved):
The githack page appears to have a stale cache, but looking at the updated source, the property chain axiom matter is resolved.

YoucTagh (approved):

mgberg (approved):

PR #796, opened 2026-02-23
Add sh:rootClass property

### LINKED ISSUE #792: Add `sh:rootClass` (similar to `sh:class` but not an instance)
While we have `sh:class` which is an equivalent of `ASK { $value rdf:type/rdfs:subClassOf* $class . }`, I believe it is interesting to also support a similar property `sh:rootClass` which expresses that all value nodes must be subclasses of the specified root class (and not an instance of any of these classes).
This would be the equivalent of `ASK { $value rdfs:subClassOf* $class . }`.

This was originally defined already in DASH: https://datashapes.org/constraints.html#RootClassConstraintComponent

This already has a use case in the SHACL UI for the SubClassEditor, and I believe it could also enhance the completeness of SHACL Core, considering that we already have `sh:class`.

Should SHACL Core include this feature, or should it be scoped to the SHACL UI namespace? Does it add value to include it in SHACL Core?
I'm also open to discuss if this property can already be expressed using different properties.

smessie:
Closes #792 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-792-rootclass-property/shacl12-core/index.html)


HolgerKnublauch:
This would require at least one test case to close this.

YoucTagh:
If you can also update the [Changes between the original SHACL Core and SHACL 1.2 Core](https://raw.githack.com/w3c/data-shapes/issue-792-rootclass-property/shacl12-core/index.html#changes-12) section.

HolgerKnublauch:
I think I would also rather like to see this under Other Constraints instead of under Value Types. Many people will not really understand the difference with sh:class, so it's IMHO better to have it near sh:in instead.

smessie:
I've addressed the above comments. I've also added the sh:rootClass constraint to the shacl.ttl vocabulary file.

HolgerKnublauch:
Yes this is getting much better now. Could you kindly also add an example as an informative subsection?

And does this feature make sense for node shapes at all? I think it will be commonly used with property shapes, and the test case should reflect that IMHO.

smessie:
I think it makes as much sense to have this feature for node shapes as it makes sense for sh:class to have it, so I think it definitely makes sense, but I've changed the test case to reflect your comment.
Also added an example focussing on property shapes.

ajnelson-nist:
Since biological taxons have come up in discussion, perhaps the informative subsection could include an example?

Here's a data graph, which I started by looking for flamingos, and found an interesting effect of the eBird name-compaction procedure and ticking the URL counter up and down.

```turtle
PREFIX ex: <http://example.com/ns#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

################################################################
# Higher-order classes, instantiated by Class individuals.

ex:TaxonomicType
	a owl:Class ;
	.

################################################################
# Metaclasses, instantiated by Class individuals.

ex:Order
	a owl:Class , ex:TaxonomicType ;
	owl:disjointWith ex:Family ;
	.

ex:Family
	a owl:Class , ex:TaxonomicType ;
	owl:disjointWith ex:Specie ;
	.

ex:Specie
	a owl:Class , ex:TaxonomicType ;
	.

################################################################
# Classes, instantiated by non-Class individuals.

ex:Phoenicopteriformes
	a owl:Class , ex:Order ;
	rdfs:label "Phoenicopteriformes"@la ;
	rdfs:seeAlso <https://en.wikipedia.org/wiki/Flamingo> ;
	.

ex:Phoenicopteridae
	a owl:Class , ex:Family ;
	rdfs:label "Phoenicopteridae"@la ;
	rdfs:subClassOf ex:Phoenicopteriformes ;
	.

ex:Picidae
	a owl:Class , ex:Family ;
	rdfs:label "Picidae"@la ;
	rdfs:subClassOf ex:Piciformes ;
	.

ex:Piciformes
	a owl:Class , ex:Order ;
	rdfs:label "Piciformes"@la ;
	.

ex:grefla1
	a owl:Class , ex:Specie ;
	rdfs:label "Chrysocolaptes guttacristatus"@la ;
	rdfs:label "Greater Flameback"@en ;
	rdfs:subClassOf ex:Picidae ;
	rdfs:seeAlso <https://ebird.org/species/grefla1> ;
	.

ex:grefla2
	a owl:Class , ex:Specie ;
	rdfs:label "American Flamingo"@en ;
	rdfs:label "Phoenicopterus ruber"@la ;
	rdfs:subClassOf ex:Phoenicopteridae ;
	rdfs:seeAlso <https://ebird.org/species/grefla2> ;
	.

ex:grefla3
	a owl:Class , ex:Specie ;
	rdfs:label "Greater Flamingo"@en ;
	rdfs:label "Phoenicopterus roseus"@la ;
	rdfs:subClassOf ex:Phoenicopteridae ;
	rdfs:seeAlso <https://ebird.org/species/grefla3> ;
	.

ex:grefla4
	a owl:Class , ex:Specie ;
	rdfs:label "Chrysocolaptes socialis"@la ;
	rdfs:label "Malabar Flameback"@en ;
	rdfs:subClassOf ex:Picidae ;
	rdfs:seeAlso <https://ebird.org/species/grefla4> ;
	.
```

Writing the example data graph up, I see some potential wording issues:

* "Root" seems to be expanding to cover *instantiation* (`rdf:type`) and *specialization* (`rdfs:subClassOf`), but the current text ([6c2c5a0](https://github.com/w3c/data-shapes/pull/796/commits/6c2c5a094f615e79a04b003d141788c5222f9221)) doesn't cover that.
* Should a higher-order class---i.e., two or more "hops" along `rdf:type`---be usable as a "root" class?

shacl12-core/index.html
```diff
@@ -5607,7 +5690,7 @@ <h3>sh:unit</h3>
 					</p>
 					<code>180.5 cm</code>
 					<p>
-						or, for US-based visitors, as 
+						or, for US-based visitors, as
 					</p>
 					<code>180.5 cm (71 in)</code>
```
```suggestion
						or, for US-based visitors, as
					</p>
					<code>71 in (180.5 cm)</code>
```

shacl12-core/index.html
```diff
@@ -5607,7 +5690,7 @@ <h3>sh:unit</h3>
 					</p>
 					<code>180.5 cm</code>
```
```suggestion
					<code>180.5 cm (71 in)</code>
```

shacl12-core/index.html
```diff
@@ -5376,6 +5376,89 @@ <h4>sh:in</h4>
...
+						<tr>
+							<th>Property</th>
+							<th>Summary and Syntax Rules</th>
+						</tr>
+						</thead>
+						<tbody>
+						<tr>
+							<td><code>sh:rootClass</code></td>
+							<td>
+								The root of the class hierarchy that all value nodes must belong to.
```
HolgerKnublauch:
```suggestion
								The root(s) of the class hierarchy that all value nodes must belong to.
```

shacl12-core/index.html
```diff
@@ -5376,6 +5376,89 @@ <h4>sh:in</h4>
...
+					<div class="def def-text">
+						<div class="def-header">TEXTUAL DEFINITION</div>
+						<div class="def-text-body" data-validator="RootClass">
+							Let <code>$rootClass</code> be a <a>parameter value</a> for <code>sh:rootClass</code>.
+							Let <code>classes</code> be a set of <a>IRIs</a> so that
+							when <code>$rootClass</code> is an <a>IRI</a> then the set only consists of exactly that IRI,
+							and when <code>$rootClass</code> is a <a>blank node</a> <a>SHACL list</a> then the set consists of
+							exactly the members of the list.<br><br>
+							For each <a>value node</a>
+							that is either a <a>literal</a>, or a non-literal for which there exists no <code>class</code> in <code>classes</code>
```
HolgerKnublauch:
```suggestion
							that is either not an <a>IRI</a>, or an <a>IRI</a> for which there exists no <code>class</code> in <code>classes</code>
```

This also needs to exclude blank nodes and triple terms.

shacl12-core/index.html
```diff
@@ -5376,6 +5376,89 @@ <h4>sh:in</h4>
...
+							such that the <a>data graph</a> entails
+							<code>valueNode rdfs:subClassOf* class</code>,
+							there is a <a>validation result</a> with the <a>value node</a> as <code>sh:value</code>.
+						</div>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						The <code>sh:rootClass</code> constraint is typically used with property shapes
+						to restrict the allowed subclass hierarchy of value nodes.
```
HolgerKnublauch, line 5498:
```suggestion
						The <code>sh:rootClass</code> constraint is typically used with property shapes
						that can have classes as their values (e.g., via <code>sh:class rdfs:Class</code>)
						to restrict the allowed subclass hierarchy of value nodes.
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -1257,6 +1257,27 @@ sh:reificationRequired
...
+	a sh:Parameter ;
+	sh:path sh:rootClass ;
+	sh:nodeKind sh:BlankNodeOrIRI ;
+	rdfs:isDefinedBy sh: .
+
+sh:rootClass
+	a rdf:Property ;
+	rdfs:label "root class"@en ;
+	rdfs:comment "Specifies a class so that each value node must be that class or a (transitive) subclass of that class."@en ;
+	rdfs:range rdfs:Class ;
```
HolgerKnublauch:
```suggestion
```

Range would also need to allow lists.

HolgerKnublauch (approved):
Approved assuming the comments get applied.

shacl12-core/index.html
smessie: Would you perhaps mean `e.g., via sh:rootClass rdfs:Class`?

HolgerKnublauch: No, it was intentionally sh:class. It means "the values of the property must be instances of rdfs:Class, and must be subclasses of the root class".

HolgerKnublauch: Maybe the subsentence that I am suggesting here isn't ideal, feel free to insert something else. I just think it's important to make this clear and understandable for everyone as this is a bit meta.

smessie: Right, my bad. Going over it again, it makes more sense now.

shacl12-core/index.html
smessie: This suggestion (and the one below) does not apply to the `sh:rootClass` property being added in this PR, but rather to the `sh:unit` property added before in #717 by @HolgerKnublauch, but I can definitely accept this suggestion here.

shacl12-vocabularies/shacl.ttl
smessie: Right. Am I correct that this should also be the case for `sh:class` then? I see that's currently not the case.
https://github.com/w3c/data-shapes/blob/e3d0238daf51412f6330eda735f3a07cdbe01f5e/shacl12-vocabularies/shacl.ttl#L546-L551

HolgerKnublauch: Correct, good catch. Also sh:datatype maybe? Would be great if you can fix them here in one go.

PR #798, opened 2026-02-24
Clarify where data block triples go

### LINKED ISSUE #791: Data block definition / inclusion in data graph
1. The naming of Data Blocks may lead to ambiguity about which graph they contribute to. Because the term "data" is also used in "data graph", users may incorrectly assume that Data Blocks contribute to the data graph rather than the inference graph. I think this can be clarified by a definition update, I'll open a PR with this:

    Current: A [data block](https://raw.githack.com/w3c/data-shapes/outline/shacl12-rules/index.html#dfn-data-block) is a set of triples. These form extra facts that are included in the inference process.

    Proposed: A [data block](https://raw.githack.com/w3c/data-shapes/outline/shacl12-rules/index.html#dfn-data-block) is a set of triples that are added to the inference graph as inferred triples.

2. Related to the above, implementations could optionally treat Data Blocks as contributing to the data graph instead of the inference graph, as an alternate execution mode. This would provide a lightweight way to define inline reference data, such as constants for currency conversions or unit definitions. The benefit would be they are first class objects, so are explicitly scoped with rules (in rulesets) for readability/clarity, versus needing to handle/ensure reference data is included in the input data graph (which would be the logical alternative). I don't think it would be good to cover such an alternate execution mode in the spec as there would be a few places to add if/else logic and it could confuse readers. A drawback of offering such functionality could be that users write their rules/datablocks/data with an intended execution mode in mind, and could get different results if using the other mode.

recalcitrantsupplant:
This closes #791

Contains (updated) rewording suggested in #791

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/dhabgood/issue791/shacl12-rules/index.html#dfn-data-block)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -437,7 +437,8 @@ RULE { ?x :name ?FN } WHERE {
           <dt><dfn data-lt="data-block">Data block</dfn></dt>
           <dd>
             A [=data block=] is a set of triples.
-            These form extra facts that are included in the inference process.
+            These triples are added to the inference graph as additional facts and are
+            included in the inference process.
           </dd>
 
           <dt><dfn>Triple template</dfn></dt>


afs (approved):

robert-david (approved):

afs:
I just used this PR to flush the publication of /TR/shacl12-rules.

It was stuck on the first merge of the day from Feb 26th. It's a W3C website caching problem.

/TR/shacl12-rules is now today's merge.

If anyone has problems, the GH page version https://w3c.github.io/data-shapes/shacl12-rules/ is up-to-date.

PR #799, opened 2026-02-24
Add dependency graph algorithm example

recalcitrantsupplant:
Following from discussion here: https://github.com/w3c/data-shapes/pull/787#issuecomment-3908169987
Provides an example dependency graph calculation algorithm. 
I think the existing definitions are sufficient on their own, and this example can help clarify.
As such, no strong opinion on its inclusion.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/dhabgood/dep-graph-algo/shacl12-rules/index.html#dependency-graph-construction-algorithm)

Branch was based on afs/stratification-algorithm

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -689,9 +689,8 @@ RULE { ?x :name ?FN } WHERE {
           <p>
             This section is non-normative. The following algorithm gives one
             possible method for constructing the [=dependency graph=] from a
-            [=rule set=]. Implementations MAY use any equivalent algorithm.
-            Conformance depends on the dependency definitions in this section,
-            not on use of this procedure.
+            [=rule set=]. Conformance depends on the dependency definitions 
+			in this section, not on use of this procedure.
           </p>
           <div class="algorithm">
             <pre>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -709,7 +709,7 @@ define buildDependencyGraph(ruleSet):
         else:
             return "positive"
         endif
-        endfunction
+        enddefine
 
     foreach rule R1 in ruleSet:
 
@@ -722,7 +722,7 @@ define buildDependencyGraph(ruleSet):
             else:
                 item.depLabel="positive"
             add item to bodyDependencies
-                    endfor
+        endfor
 
         foreach pair (triplePattern, depLabel) in bodyDependencies:
 
@@ -753,7 +753,7 @@ define buildDependencyGraph(ruleSet):
     endfor
 
     the result is DP
-    endfunction
+    enddefine
             </pre>
           </div>
         </section>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -684,10 +684,10 @@ RULE { ?x :name ?FN } WHERE {
           </dd>
         </dl>
 
-        <section id="dependency-graph-construction-algorithm">
+        <section id="dependency-graph-construction-algorithm" class="informative">
           <h4>Dependency Graph Algorithm</h4>
           <p>
-            This section is non-normative. The following algorithm gives one
+            The following algorithm gives one
             possible method for constructing the [=dependency graph=] from a
             [=rule set=]. Conformance depends on the dependency definitions 
 			in this section, not on use of this procedure.
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -683,8 +683,78 @@ RULE { ?x :name ?FN } WHERE {
             in the [=dependency graph=] involving `R`.
           </dd>
         </dl>
-          
-        
+
+        <section id="dependency-graph-construction-algorithm">
+          <h4>Dependency Graph Algorithm</h4>
+          <p>
+            This section is non-normative. The following algorithm gives one
+            possible method for constructing the [=dependency graph=] from a
+            [=rule set=]. Implementations MAY use any equivalent algorithm.
+            Conformance depends on the dependency definitions in this section,
+            not on use of this procedure.
+          </p>
+          <div class="algorithm">
+            <pre>
+
+## output -- Dependency graph with rule vertices and labeled edges.
+
+define buildDependencyGraph(ruleSet):
+
+    ## edgeLabelMap maps (R1, R2) to "positive" or "negative"
+    let edgeLabelMap be a map from (rule, rule) to label
+
+    define mergeLabel(oldLabel, newLabel):
+        ## Negative dependency dominates positive dependency.
+        if oldLabel == "negative" or newLabel == "negative":
+            return "negative"
+        else:
+            return "positive"
+        endif
+    endfunction
+
+    foreach rule R1 in ruleSet:
+
+        ## Each item is (triplePattern, depLabel) where depLabel is
+        ## "positive" for a rule body element and "negative" for a
+        ## negation element.
+        let bodyDependencies = body triple patterns of R1 with dependency labels
+
+        foreach item in bodyDependencies:
+            let TP = item.triplePattern
+            let depLabel = item.depLabel
+
+            foreach rule R2 in ruleSet:
+                foreach triple template TT in head of R2:
+
+                    ## "possibly generate" / matching is defined in
+                    ## 3.3 Rule Dependency.
+                    if TT can possibly generate a triple matching TP:
+                        let key = (R1, R2)
+                        if edgeLabelMap contains key:
+                            let oldLabel = edgeLabelMap.get(key)
+                            let merged = mergeLabel(oldLabel, depLabel)
+                            edgeLabelMap.set(key, merged)
+                        else:
+                            edgeLabelMap.set(key, depLabel)
+                        endif
+                    endif
+
+                endfor
+            endfor
+        endfor
+    endfor
+
+    let DP be a graph with vertices = ruleSet
+    foreach entry ((R1, R2), label) in edgeLabelMap:
+        add edge (R1 -&gt; R2) to DP labeled label
+    endfor
+
+    the result is DP
+    endfunction
+            </pre>
+          </div>
+        </section>
+
         <p>Notes:</p>
         <ul>
           <li>
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -713,16 +713,16 @@ define buildDependencyGraph(ruleSet):
 
     foreach rule R1 in ruleSet:
 
-let bodyDependencies = {}
-foreach triplePatternElement TP in the body of R1:
-    let item be a new pair (triplePattern, depLabel)
-    item.triplePattern = TP
-	if TP is a negation element:
-		item.depLabel="negative"
-	else:
-		item.depLabel="positive"
-	add item to bodyDependencies
-            endfor
+        let bodyDependencies = {}
+        foreach triplePatternElement TP in the body of R1:
+            let item be a new pair (triplePattern, depLabel)
+            item.triplePattern = TP
+            if TP is a negation element:
+                item.depLabel="negative"
+            else:
+                item.depLabel="positive"
+            add item to bodyDependencies
+                    endfor
 
         foreach pair (triplePattern, depLabel) in bodyDependencies:
 
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -709,12 +709,12 @@ define buildDependencyGraph(ruleSet):
         else:
             return "positive"
         endif
-    endfunction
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -786,11 +856,73 @@ <h4>Stratification Algorithm</h4>
           </p>
           <div class="algorithm">
             <pre>
-              @@ one possible stratification algorithm
+
```
afs, line 878: #787 has been merged (with your corrections).

This text and some of the other fixes in this PR should then already be in `gh-pages`.

liviorobaldo:
I am having some difficulty understanding this algorithm, in particular the phrase "body triple patterns of R1 with dependency labels".

My first understanding was that, at the beginning, no dependencies are defined. If that is the case, is this set initially empty? That does not seem likely, as the algorithm would then terminate immediately.

Reading the comment more carefully:

```
## Each item is (triplePattern, depLabel) where depLabel is
## "positive" for a rule body element and "negative" for a
## negation element.
```

Does this mean that _all_ triple patterns in the body of R1 are included, and that each of them is necessarily associated with a dependency label (either positive or negative)? If so, it is unclear why the phrase "with dependency labels" is needed above, since there would be no body triple patterns of R1 without such labels. Or, I may be misunderstanding which body triple patterns are considered to be _without_ dependency labels.

In any case, it seems that some explicit initialization of bodyDependencies should be added. In my view, this initialization should rely on the definitions of [Rule body element](https://raw.githack.com/w3c/data-shapes/dhabgood/dep-graph-algo/shacl12-rules/index.html#dfn-rule-body-element) and [triplePattern](https://raw.githack.com/w3c/data-shapes/dhabgood/dep-graph-algo/shacl12-rules/index.html#dfn-triple-pattern).

Something like:

```
let bodyDependencies = {}
foreach "triplePattern" TP in "Rule body element":
	if TP is a negation element:
		add <TP,"negative"> to bodyDependencies
	else:
		add <TP,"positive"> to bodyDependencies	
```

I can suggest a modification, but first I want to make sure that I have correctly understood what is being done here 😅

recalcitrantsupplant:
> If that is the case, is this set initially empty?

body dependencies is not initially empty, perhaps there's a better name than dependencies here, something like "bodyPatternLabels"

> "with dependency labels"

it wasn't intended to imply that there were any _without_ such labels, they are mentioned because it's required for the process (i.e. let depLabel = item.depLabel)

> "In any case, it seems that some explicit initialization of bodyDependencies should be added. In my view, this initialization should rely on the definitions of Rule body element and triplePattern.
>  Something like:
>
>  let bodyDependencies = {}
>  foreach "triplePattern" TP in "Rule body element":
>       if TP is a negation element:
>                add <TP,"negative"> to bodyDependencies
>        else:
>                add <TP,"positive"> to bodyDependencies "

yes this is what `let bodyDependencies = body triple patterns of R1 with dependency labels` does (iterating over elements first though) but we could make it more explicit with the foreach yes.

shacl12-rules/index.html
```diff
@@ -668,8 +668,78 @@ <h3>Rule Dependency</h3>
...
         </dl>
-          
-        
+
+        <section id="dependency-graph-construction-algorithm">
+          <h4>Dependency Graph Algorithm</h4>
+          <p>
+            This section is non-normative. The following algorithm gives one
+            possible method for constructing the [=dependency graph=] from a
+            [=rule set=]. Implementations MAY use any equivalent algorithm.
```
liviorobaldo: Why is "MAY" in all caps? I think this entire sentence could perhaps be considered implicit.

shacl12-rules/index.html
```diff
@@ -668,8 +668,78 @@ <h3>Rule Dependency</h3>
...
+            return "positive"
+        endif
+    endfunction
+
+    foreach rule R1 in ruleSet:
+
+        ## Each item is (triplePattern, depLabel) where depLabel is
+        ## "positive" for a rule body element and "negative" for a
+        ## negation element.
+        let bodyDependencies = body triple patterns of R1 with dependency labels
```
liviorobaldo:
```suggestion
let bodyDependencies = {}
foreach 'triplePattern' TP in 'Rule body element':
    let item be a new pair (triplePattern, depLabel)
    item.triplePattern = TP
	if TP is a negation element:
		item.depLabel="negative"
	else:
		item.depLabel="positive"
	add item to bodyDependencies
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,78 @@ <h3>Rule Dependency</h3>
...
+        else:
+            return "positive"
+        endif
+    endfunction
+
+    foreach rule R1 in ruleSet:
+
+        ## Each item is (triplePattern, depLabel) where depLabel is
+        ## "positive" for a rule body element and "negative" for a
+        ## negation element.
```
liviorobaldo:
```suggestion
```

I think we can now remove this comment. The for loop below now makes explicit what the comment previously described in plain text.

shacl12-rules/index.html
```diff
@@ -668,8 +668,78 @@ <h3>Rule Dependency</h3>
...
+        let bodyDependencies = body triple patterns of R1 with dependency labels
+
+        foreach item in bodyDependencies:
+            let TP = item.triplePattern
+            let depLabel = item.depLabel
+
+            foreach rule R2 in ruleSet:
+                foreach triple template TT in head of R2:
```
afs, line 730: It's a "triple template" in the head.

liviorobaldo: Apologies, I removed my suggestion

shacl12-rules/index.html
recalcitrantsupplant: It's borrowing the RFC style keyword https://datatracker.ietf.org/doc/html/rfc2119. No strong opinion on inclusion of that sentence, happy to remove and keep it concise.

recalcitrantsupplant: Removed in https://github.com/w3c/data-shapes/pull/799/changes/6896c176db5b3743940ec26fd1c3dd5290b093e2

shacl12-rules/index.html
TallTed: I'm confused. "triple termplate" seems like a misspelling, wherever it might occur, but it seems to be a suggestion of a correction in @afs [comment](https://github.com/w3c/data-shapes/pull/799/changes#r2859274063)? I find several occurrences of "triple template" now in this file, and no occurrences of "triple termplate". Perhaps a clarifying comment could be made here, and then this comment be marked resolved?

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
             in the [=dependency graph=] involving `R`.
           </dd>
         </dl>
-          
-        
+
+        <section id="dependency-graph-construction-algorithm">
+          <h4>Dependency Graph Algorithm</h4>
+          <p>
+            This section is non-normative. The following algorithm gives one
```
afs: Respec has support for making sections as "non-normative".

```suggestion
        <section id="dependency-graph-construction-algorithm" class="informative">
          <h4>Dependency Graph Algorithm</h4>
          <p>
            The following algorithm gives one
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+    let edgeLabelMap be a map from (rule, rule) to label
+
+    define mergeLabel(oldLabel, newLabel):
+        ## Negative dependency dominates positive dependency.
+        if oldLabel == "negative" or newLabel == "negative":
+            return "negative"
+        else:
+            return "positive"
+        endif
+    endfunction
```
afs: We have a mix (one each) `enddefine` and `endfunction` but both are indented.


```suggestion
        endfunction
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+
+let bodyDependencies = {}
+foreach 'triplePattern' TP in 'Rule body element':
+    let item be a new pair (triplePattern, depLabel)
+    item.triplePattern = TP
+	if TP is a negation element:
+		item.depLabel="negative"
+	else:
+		item.depLabel="positive"
+	add item to bodyDependencies
```
afs: Indent 8 spaces.

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+
```
afs, line 726:
```suggestion
            endfor

```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+        foreach item in bodyDependencies:
+            let TP = item.triplePattern
+            let depLabel = item.depLabel
```
afs: TP is already in-use from `foreach 'triplePattern' TP`

```suggestion
        foreach pair (triplePattern, depLabel) in bodyDependencies:
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+            return "negative"
+        else:
+            return "positive"
+        endif
+    endfunction
+
+    foreach rule R1 in ruleSet:
+
+let bodyDependencies = {}
+foreach 'triplePattern' TP in 'Rule body element':
```
afs:
```suggestion
foreach triplePatternElement TP in the body of R1:
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+                            edgeLabelMap.set(key, depLabel)
+                        endif
+                    endif
+
+                endfor
+            endfor
+        endfor
+    endfor
+
+    let DP be a graph with vertices = ruleSet
```
afs:
```suggestion
    let DP = { }
```

shacl12-rules/index.html
```diff
@@ -668,8 +668,82 @@ <h3>Rule Dependency</h3>
...
+    foreach entry ((R1, R2), label) in edgeLabelMap:
+        add edge (R1 -&gt; R2) to DP labeled label
```
afs: Treat sets as immutable!
```suggestion
        DP = { edge (R1 -&gt; R2) labeled label } &cups; DP
```

shacl12-rules/index.html
recalcitrantsupplant: There was originally a suggested change from Livio (now deleted) for triple pattern -> triple template. Andy then clarified the existing text with triple template was correct. I couldn't see a way to resolve in the main thread but I have a button in this changes view so will resolve this thread here.

Edit: I cannot resolve it "User is not authorized to resolve the conversation"

recalcitrantsupplant:
are we happy to use "foreach" over "for each" and "enddefine" over "endfunction"? There is some usage of "for each" and "endfunction" in sections this PR wasn't for but I can align them all here as it is not a major change.

shacl12-rules/index.html
recalcitrantsupplant: done in https://github.com/w3c/data-shapes/pull/799/changes/6066111b2cfe86d146e7c8a9f492ccc578b4d524

afs:
> are we happy to use "foreach" over "for each" and "enddefine" over "endfunction"? There is some usage of "for each" and "endfunction" in sections this PR wasn't for but I can align them all here as it is not a major change.

At this stage, getting to a complete document, some differences in presentation shouldn't hold us up. There will be whole-document review later.

afs (approved):

recalcitrantsupplant:
@simonstey @robert-david could I please get one more review prior to merge

shacl12-rules/index.html
```diff
@@ -668,8 +668,81 @@ <h3>Rule Dependency</h3>
...
+            let item be a new pair (triplePattern, depLabel)
+            item.triplePattern = TP
+            if TP is a negation element:
+                item.depLabel="negative"
+            else:
+                item.depLabel="positive"
+            add item to bodyDependencies
+        endfor
+
+        foreach pair (triplePattern, depLabel) in bodyDependencies:
```
```suggestion
        foreach pair (triplePattern bodyTP, depLabel) in bodyDependencies:
```

robert-david (approved):

shacl12-rules/index.html
recalcitrantsupplant: noting this earlier comment https://github.com/w3c/data-shapes/pull/799#discussion_r2879923024 I'll make it bodyTP and change the comment on the following line to refer to bodyTP

PR #807, opened 2026-03-08
#803: Added example and explanation on sh:node between node shapes

### LINKED ISSUE #803: Clarify that sh:node can be used between node shapes to signal "constraint inheritance"
The definition of sh:node only includes an example at property shapes, but its use at node shapes is equally important.

We should make it easier to understand that intent (the old WG even had an explicit discussion and decision to use sh:node between node shapes to signal inheritance).

So I think we should add a second example of sh:node and add a highlighted Note that it can be used similar to rdfs:subClassOf between classes.

See also https://www.linkedin.com/feed/update/urn:li:ugcPost:7434087123420405760?commentUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7434087123420405760%2C7434111141116813312%29&replyUrn=urn%3Ali%3Acomment%3A%28ugcPost%3A7434087123420405760%2C7434383575388520448%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287434111141116813312%2Curn%3Ali%3AugcPost%3A7434087123420405760%29&dashReplyUrn=urn%3Ali%3Afsd_comment%3A%287434383575388520448%2Curn%3Ali%3AugcPost%3A7434087123420405760%29

HolgerKnublauch:
Closes #803

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-803/shacl12-core/index.html#NodeConstraintComponent)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2265,7 +2265,7 @@ ex:SomeClass-alternativePathExample
 					<p><em>The remainder of this section is informative.</em></p>
 					<p>
 						The recursion policy above has been selected to support a large variety of implementation strategies.
-						By leaving recursion undefined, implementations may chose to not support recursion so that they
+						By leaving recursion undefined, implementations may choose to not support recursion so that they
 						can issue a static set of SPARQL queries (against SPARQL end points) without having to support cycles.
 						The Working Group is aware that other implementations may support recursion and that some shapes graphs may
 						rely on these specific characteristics.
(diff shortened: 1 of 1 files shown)


YoucTagh (approved):

bergos (approved):

nicholascar:
Thanks @HolgerKnublauch. I'll start rebuilding my DB of Shapes using `sh:node`-based inheritance to build generic Shapes and then more specialised versions of them, rather than just coping and altering Shapes.

PR #808, opened 2026-03-08
Support for sh:uniqueValuesFor

### LINKED ISSUE #661: Constraint violation on keys
OWL provides the owl:hasKey axiom, which triggers owl:sameAs inferences when individuals share the same values for the specified key properties. I believe it would be beneficial if SHACL could interpret such situations by raising a constraint violation when key property values are duplicated. This feature has arisen for an issue I am currently encountering while modeling relational databases.

HolgerKnublauch:
Closes #661

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-661/shacl12-core/index.html#UniqueValuesForConstraintComponent)


bergos (changes requested):
Maybe it's worth mentioning that other constraints define if there are values for the defined properties and how many. Or it could be shown in the example. Since it has some similarities to composite primary keys from SQL, one may expect the values must be sh:minCount == sh:maxCount == 1.

I think that's nice to have for the specification, but it should be covered in a test.

shacl12-core/index.html
```diff
@@ -5529,6 +5530,124 @@ <h4>sh:rootClass</h4>
...
+							<br />
+							For each <a>value node</a> <code>V</code> for which there exists another <a>node</a> <code>Other</code> in <code>$targetNodes</code>
+							that has exactly the same <a>values</a> for all properties in <code>$properties</code> as <code>V</code>
+							there is a <a>validation result</a> with <code>Other</code> as <code>sh:value</code>.
+							No result is produced if <code>V</code> has no <a>values</a> for any of the properties in <code>$properties</code>.
+						</div>
+					</div>
+					<p><em>The remainder of this section is informative.</em></p>
+					<p>
+						Note that matching of literals needs to be exact, e.g. <code>"04"^^xsd:byte</code> does not match <code>"4"^^xsd:integer</code>.
```
YoucTagh, line 5640:
```suggestion
						Note that matching of literals needs to be exact, e.g. <code>"04"^^xsd:byte</code> does not match <code>"4"^^xsd:integer</code>.
						The same applies to language tags, e.g. <code>"Alice"@en</code> does not match <code>"Alice"@fr</code>.
```
I think it would be a good idea to add a note about language tags, too.
For the [sh:uniqueLang](https://raw.githack.com/w3c/data-shapes/issue-661/shacl12-core/index.html#UniqueLangConstraintComponent) the TEXTUAL DEFINITION mentions also [base direction](https://www.w3.org/TR/rdf12-concepts/#dfn-base-direction).

HolgerKnublauch: I had copied that sentence from sh:in, so if we expand on this topic here, we should do this everywhere.

Given that this particular case is just an example ("e.g.") I honestly don't think we need to give another example.

HolgerKnublauch:
@bergos I think I have now addressed your (valid) feedback. Please review again.

bergos (approved):

Issue #764, opened 2026-02-05
Use case for SHACL profiling of data: multi-model interoperability


ajnelson-nist:
This builds on the "Profiling Data" notes of [PR 754](https://github.com/w3c/data-shapes/pull/754), which is in draft status at the moment, but was discussed on today's profiling working group call.

The draft text at this time describes the conformance of a data graph against a single model/ontology/schema.

Suppose[^1] that I am designing a graph-backed application that will house measurements.  For the measurements "portion" of the graph design, I need to interoperate with two other graph-backed applications that cover different aspect of measurements.

1. One covers instrumentation and experiment design, and is grounded ultimately in BFO.  This particular application reviews that [qualities](http://purl.obolibrary.org/obo/BFO_0000019) link with [inheres in](http://purl.obolibrary.org/obo/BFO_0000197)[^2] in a manner conformant with BFO.
2. One covers measurement of properties and is grounded in SSN/SOSA.  This particular application reviews that [properties](https://www.w3.org/TR/2017/REC-vocab-ssn-20171019/#SSNProperty)[^3] (which I'll just call qualities) attach with [is property of](https://www.w3.org/TR/2017/REC-vocab-ssn-20171019/#SSNisPropertyOf) in a manner conformant with SSN.

To maintain interoperability with these two applications, the measurement application defines alignment classes and properties for storing its records.

```turtle
PREFIX ex: <http://example.org/ontology/>
PREFIX obo: <http://purl.obolibrary.org/obo/>
PREFIX ssn: <http://www.w3.org/ns/ssn/>

ex:Quality
  a owl:Class ;
  rdfs:subClassOf ssn:Property ;
  rdfs:subClassOf obo:BFO_0000019 ;
  .

ex:isQualityOf
  a owl:ObjectProperty ;
  rdfs:subPropertyOf ssn:isPropertyOf ;
  rdfs:subPropertyOf obo:BFO_0000197 ;
  rdfs:domain ex:Quality ;
  .

# Omitted from illustration:
# Other classes and design pertaining to the range of ex:isQualityOf .
```

A shapes graph for this application could trivially import all or some of two other shapes graphs and provide no further shapes:

* A general shapes graph for SSN/SOSA concept review
* A general shapes graph for BFO concept review

This application is an exercise of data profiling to achieve an interoperability goal.

[...]

PR #814, opened 2026-03-10
GH#169: Introduce classes `sh:ShapesGraph` and `sh:DataGraph`

### LINKED ISSUE #169: Have a dedicated class to type SHACL specifications (instead of owl:Ontology)
I would find it useful to have a dedicated type, other than owl:Ontology, to type SHACL specifications.
In particular in systems that manage both OWL ontologies **and** SHACL specifications, to distinguish them.
Also because metadata properties are attached to this entity (versions, dates, etc.)

e.g. `sh:ShapesGraph` (upper case S), probably a subClassOf owl:Ontology.

HolgerKnublauch:
Closes #169

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-169/shacl12-core/index.html#shapes-graph)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2129,7 +2129,7 @@ ex:SomeClass-alternativePathExample
 					<code>owl:imports</code> in the data graph is not enacted, in order to avoid uncontrolled increase of validation work. If you want to validate
 					several related ontologies, pass all of them to the SHACL processor (together or one by one), do not rely on <code>owl:imports</code> links.
 				</p>
-				<p class="note">
+				<p class="note" id="owlImportsOnDataGraph">
 					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a directive to load <code>Y</code>.  See [OWL 2 Web Ontology Language Mapping to RDF Graphs (Second Edition), Section 3.1.1](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/#Resolving_Included_RDF_Graphs) for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .` SHOULD also be included.
 				</p>
 			</section>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2129,6 +2129,9 @@ ex:SomeClass-alternativePathExample
 					<code>owl:imports</code> in the data graph is not enacted, in order to avoid uncontrolled increase of validation work. If you want to validate
 					several related ontologies, pass all of them to the SHACL processor (together or one by one), do not rely on <code>owl:imports</code> links.
 				</p>
+				<p class="note">
+					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a directive to load <code>Y</code>.  See [OWL 2 Web Ontology Language Mapping to RDF Graphs (Second Edition), Section 3.1.1](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/#Resolving_Included_RDF_Graphs) for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .` SHOULD also be included.
+				</p>
 			</section>
 
 			<section id="subClassOfInShapesGraph">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2132,7 +2132,7 @@ ex:SomeClass-alternativePathExample
 				<p class="note" id="owlImportsOnDataGraph">
 					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely
 					on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct
-					presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a
+					presence of <code>X a owl:Ontology .</code> such that <code>X owl:imports Y .</code> will process the <code>owl:imports</code> statement as a
 					directive to load <code>Y</code>.  See <a data-cite="owl2-mapping-to-rdf/#Resolving_Included_RDF_Graphs">OWL 2 Web Ontology Language Mapping
 					to RDF Graphs, Section 3.1.1</a> for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .`
 					SHOULD also be included.
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2130,7 +2130,8 @@ ex:SomeClass-alternativePathExample
 					several related ontologies, pass all of them to the SHACL processor (together or one by one), do not rely on <code>owl:imports</code> links.
 				</p>
 				<p class="note" id="owlImportsOnDataGraph">
-					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a directive to load <code>Y</code>.  See [OWL 2 Web Ontology Language Mapping to RDF Graphs (Second Edition), Section 3.1.1](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/#Resolving_Included_RDF_Graphs) for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .` SHOULD also be included.
+					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a directive to load <code>Y</code>.  See <a data-cite="owl2-mapping-to-rdf/#Resolving_Included_RDF_Graphs">OWL 2 Web Ontology Language
+Mapping to RDF Graphs, Section 3.1.1</a> for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .` SHOULD also be included.
 				</p>
 			</section>
 
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2106,10 +2106,10 @@ ex:SomeClass-alternativePathExample
 					Any RDF graph can be a <dfn>data graph</dfn>.
 				</p>
 				<p>
-					<span data-syntax-rule="DataGraph">The class <code>sh:DataGraph</code> MAY be used as an <code>rdf:type</code> of the <a>IRI</a>
+					<span data-syntax-rule="DataGraph">The <code>sh:DataGraph</code> class MAY be used as an <code>rdf:type</code> of the <a>IRI</a>
 					of a graph that is typically used as a <a>data graph</a>.</span>
-					Note, however, that even a <a>shapes graph</a> may play the role of a <a>data graph</a>, for example
-					when the shapes themselves shall be validated.
+					Note, however, that a <a>shapes graph</a> MAY also play the role of a <a>data graph</a>; for example,
+					when the shapes themselves will be validated.
 				</p>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2085,7 +2085,7 @@ ex:SomeClass-alternativePathExample
 					that may be passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
 				</p>
 				<p>
-					<span data-syntax-rule="ShapesGraph">The class <code>sh:ShapesGraph</code> MAY be used as an <code>rdf:type</code>
+					<span data-syntax-rule="ShapesGraph">The <code>sh:ShapesGraph</code> class MAY be used as an <code>rdf:type</code>
 					of the <a>IRI</a> of a graph that is typically used as a <a>shapes graph</a>.</span>
 				</p>
 				<p><em>The remainder of this section is informative.</em></p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2083,6 +2083,10 @@ ex:SomeClass-alternativePathExample
 					A <dfn>shapes graph</dfn> is an RDF graph containing zero or more shapes
 					that is passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
 				</p>
+				<p>
+					<span data-syntax-rule="ShapesGraph">The class <code>sh:ShapesGraph</code> MAY be used as an <code>rdf:type</code>
+					of the <a>IRI</a> of a graph that is typically used as a <a>shapes graph</a>.</span>
+				</p>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
 					Shapes graphs can be reusable validation modules that can be cross-referenced with the predicate <a data-cite="owl2-syntax/#Imports"><code>owl:imports</code></a>.
@@ -2099,6 +2103,12 @@ ex:SomeClass-alternativePathExample
 				<h3>Data Graph</h3>
 				<p>
 					Any RDF graph can be a <dfn>data graph</dfn>.
+				</p>
+				<p>
+					<span data-syntax-rule="DataGraph">The class <code>sh:DataGraph</code> MAY be used as an <code>rdf:type</code> of the <a>IRI</a>
+					of a graph that is typically used as a <a>data graph</a>.</span>
+					Note, however, that even a <a>shapes graph</a> may play the role of a <a>data graph</a>, for example
+					when the shapes themselves shall be validated.
 				</p>
 				<p><em>The remainder of this section is informative.</em></p>
 				<p>
@@ -2142,7 +2152,7 @@ ex:SomeClass-alternativePathExample
 			</section>
 
 			<section id="sh-shapes-graph">
-				<h3>Linking to shapes graphs (sh:shapesGraph)</h3>
+				<h3>Linking data graphs to shapes graphs (sh:shapesGraph)</h3>
 				<p class="syntax">
 					A <a>data graph</a> can include triples used to suggest one or more graphs to a SHACL processor with the predicate <code>sh:shapesGraph</code>.
 					<span data-syntax-rule="shapesGraph-nodeKind">Every <a>value</a> of <code>sh:shapesGraph</code> is an <a>IRI</a></span> representing a graph that SHOULD be included into the <a>shapes graph</a> used to validate the <a>data graph</a>.
@@ -2155,8 +2165,9 @@ ex:SomeClass-alternativePathExample
 					<div class="data-graph">
 						<div class="turtle">
 &lt;http://example.com/myDataGraph&gt;
-	sh:shapesGraph ex:graph-shapes1 ;
-	sh:shapesGraph ex:graph-shapes2 .
+    a sh:DataGraph ;
+    sh:shapesGraph ex:graph-shapes1 ;
+    sh:shapesGraph ex:graph-shapes2 .
 						</div>
 					</div>
 				</aside>
@@ -6078,6 +6089,7 @@ ex:NameGroup
 				<li>Added support for `rdf:dirLangString` to <a href="#UniqueLangConstraintComponent"></a>, see <a href="https://github.com/w3c/data-shapes/issues/737">Issue 737</a></li>
 				<li>Added the new constraint component <a href="#RootClassConstraintComponent"><code>sh:rootClass</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/792">Issue 792</a></li>
 				<li>Added the new constraint component <a href="#SubsetOfConstraintComponent"><code>sh:subsetOf</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/786">Issue 786</a></li>
+				<li>Added the graph types <code>sh:DataGraph</code> and <code>sh:ShapesGraph</code>; see <a href="https://github.com/w3c/data-shapes/issues/169">Issue 169</a></li>
 			</ul>
 		</section>
   </body>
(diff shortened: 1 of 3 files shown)


shacl12-vocabularies/shacl.ttl
```diff
@@ -176,6 +176,24 @@ sh:TripleTerm
...
+    a rdfs:Class ;
+    rdfs:label "Data graph"@en ;
+    rdfs:comment "This class be used as type of graphs that typically act in the role of a SHACL data graph."@en ;
+    rdfs:subClassOf owl:Ontology ;
+    rdfs:isDefinedBy sh: .
+
+sh:ShapesGraph
+    a rdfs:Class ;
+    rdfs:label "Shapes graph"@en ;
+    rdfs:comment "This class be used as type of graphs that typically act in the role of a SHACL shapes graph."@en ;
```
nicholascar:
```suggestion
    rdfs:comment "The class of ontology that acts in the role of a SHACL shapes graph.""@en ;
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -176,6 +176,24 @@ sh:TripleTerm
 	rdfs:comment "The node kind of all triple terms."@en ;
 	rdfs:isDefinedBy sh: .
 
+
+# Graph types vocabulary ------------------------------------------------------
+
+sh:DataGraph
+    a rdfs:Class ;
+    rdfs:label "Data graph"@en ;
+    rdfs:comment "This class be used as type of graphs that typically act in the role of a SHACL data graph."@en ;
```
nicholascar:
```suggestion
    rdfs:comment "The class of ontology that acts in the role of a SHACL data graph."@en ;
```

nicholascar (changes requested):
Just slight wording changes to make the class definitions read as per their semantics - specialised forms of ontology. The real work is done by the definitions in the written spec of "data graph" and "shapes graph".

I've also edited the definition of "shapes graph" so that it's not conditional on it being passed into a validation process - "may be" not "is" - as we will potentially see shapes graphs defined that are not (only) used for validation, e.g. UI generation.

bergos (approved):

shacl12-core/index.html
```diff
@@ -2083,6 +2083,10 @@ <h3>Shapes Graph</h3>
 					A <dfn>shapes graph</dfn> is an RDF graph containing zero or more shapes
 					that is passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
```
nicholascar:
```suggestion
					that may be passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
```

shacl12-vocabularies/shacl.ttl
HolgerKnublauch: I am less excited about this change: for me the term Ontology is really for class/property definitions, not instance data. It is a historic misnomer that owl:Ontology exists as the only practical alternative. Better would have been rdfs:Graph as base class. I wouldn't want to propagate this misnomer into SHACL TBH.

HolgerKnublauch: What about "The class of graph..."?

mgberg: > for me the term Ontology is really for class/property definitions, not instance data. It is a historic misnomer that owl:Ontology exists as the only practical alternative. Better would have been rdfs:Graph as base class.

Given that the SHACL engine doesn't follow `owl:imports` for data graphs, would it be allowable to not make it a subclass of `owl:Ontology`? Someone could always make something an instance of both if they want.

HolgerKnublauch: Theoretically yes, but what would be the advantage of NOT making this? People can use owl:imports either way, with or without the owl:Ontology (an RDFS engine would infer that type).

Having them under the same root class looks more symmetric to me.

mgberg: I tend to agree that it is a bit unfortunate that `owl:Ontology` is used for instance data for historical reasons, so not making `sh:DataGraph` a subclass of `owl:Ontology` would not be unnecessarily propagating that misnomer. But I agree, having both have the same root class does simplify things.

HolgerKnublauch: We could introduce the abstract base class sh:Graph and use that instead, then say sh:Graph rdfs:subClassOf owl:Ontology. This would somewhat detach the legacy name by moving it one level higher?

ajnelson-nist (commented):
A few remarks on trying to subclass `owl:Ontology`:

OWL 2 DL does not permit encoding `x rdfs:subClassOf owl:Ontology .`.  I don't understand the full motivation from back then, but the precise statement fiatting "no" is in [OWL 2 Syntax, Section 5.1](https://www.w3.org/TR/owl2-syntax/#Classes):

> IRIs from the reserved vocabulary other than _owl:Thing_ and _owl:Nothing_ _MUST NOT_ be used to identify classes in an OWL 2 DL ontology.

`owl:Ontology` is in the reserved vocabulary - see [Table 3 in Section 2.4](https://www.w3.org/TR/owl2-syntax/#IRIs).
(CORRECTION, found by @mgberg below - `owl:Ontology` is **not** in that table, but my understanding is it does have reservations that prevent it from being subclassed.)



So, from the OWL perspective, if SHACL Core's graph included the axiom `sh:ShapesGraph rdfs:subClassOf owl:Ontology .`, then SHACL Core's graph would bump any OWL ontology importing it into OWL Full.  I haven't checked lately, but so far I don't think SHACL Core has issues that outright violate OWL 2 DL rules (an aside: `rdfs:domain` and `rdfs:range` can reference `owl:Ontology`), so this risks being the first statement that prevents SHACL Core from being imported into a big graph that tries to maintain OWL 2 DL conformance.

I think the above is unlikely to cause concern for those who use both OWL and SHACL together, since SHACL's definition graph is likely to be treated like OWL's definition graph.  I forget where this is recorded, but I've seen in the OWL docs that OWL's definition graph can't be imported into an OWL ontology without bumping the ontology to OWL Full.

We could maintain the spirit of `sh:DataGraph` and `sh:ShapesGraph` being sibling classe of OWL Ontology by encoding in SHACL-SHACL rules like this, IRI(s) TBD:

```turtle
shsh:OntologySpecializationShape
	a sh:NodeShape ;
	sh:class owl:Ontology .
	sh:targetClass sh:DataGraph , sh:ShapesGraph ;
	.
```

Then, the range of `sh:shapesGraph` and the like could read:

```turtle
sh:entailment
	rdfs:domain owl:Ontology, sh:ShapesGraph ;
	.
sh:shapesGraph
	rdfs:range owl:Ontology, sh:ShapesGraph ;
	.
sh:suggestedShapesGraph
	rdfs:range owl:Ontology, sh:ShapesGraph ;
	.
```

mgberg:
> `owl:Ontology` is in the reserved vocabulary - see [Table 3 in Section 2.4](https://www.w3.org/TR/owl2-syntax/#IRIs).

I don't see `owl:Ontology` in that table, am I missing it?

ajnelson-nist:
> > `owl:Ontology` is in the reserved vocabulary - see [Table 3 in Section 2.4](https://www.w3.org/TR/owl2-syntax/#IRIs).
> 
> I don't see `owl:Ontology` in that table, am I missing it?

Thank you for catching that.  I had seen an OWL 2 validator flag each of these as outside OWL 2 DL:

```turtle
owl:Ontology
        a owl:Class ;
        .

ex:O
        rdfs:subClassOf owl:Ontology ;
        .

ex:OP-1
        a owl:ObjectProperty ;   # AnnotationProperty is fine.
        rdfs:domain owl:Ontology ;
        rdfs:range owl:Ontology ;
        .
```

Unfortunately, that validator does not link to specific parts of the OWL specification describing why.  (I think my institution's rules would frown on me naming the tool in this instance - apologies for those who wish to reproduce for themselves.)  I had known them before, but cannot recall them at the moment, and I'm about to be in meetings for a good chunk of the day.

So, my references above are not correct, for which I apologize.  I do believe the end effect is true, though, in that subclassing `owl:Ontology` would firmly make SHACL-Core incompatible for *importing* into OWL 2 DL.  In mix-in cases, I don't know how commonly SHACL-Core is imported, rather than its vocabulary just used alongside class specifications in RDFS or OWL.

HolgerKnublauch:
@ajnelson-nist I don't think shsh:OntologySpecializationShape implements what you would want. It basically says that any instance of sh:ShapesGraph MUST also be an instance of owl:Ontology. So it acts as a constraint check, while rdfs:subClassOf would act as an inference. Enforcing this constraint would either require all ShapesGraph instances to be double-typed with owl:Ontology, or the system ontology imported that adds the subClassOf triple.

In general, I am highly doubtful that many SHACL graphs fall under OWL DL anyway. I have lost track if OWL DL is even still relevant. My impression is that the majority of OWL implementations are just OWL RL, which ignores any triples on which an OWL DL engine would block.

shacl12-vocabularies/shacl.ttl
nicholascar: Having a generic class for a Graph would be appropriate helpful for many things beyond SHACL such as multi-graph management.

We could always duck type here and there with a Graph class: if it contains Shapes, it's a ShapesGraph.

HolgerKnublauch: Ok, sh:Graph added to TTL file. That is (still) declared rdfs:subClassOf owl:Ontology, which is probably the only open question here.

ajnelson-nist:
> @ajnelson-nist I don't think shsh:OntologySpecializationShape implements what you would want. It basically says that any instance of sh:ShapesGraph MUST also be an instance of owl:Ontology. So it acts as a constraint check, while rdfs:subClassOf would act as an inference. Enforcing this constraint would either require all ShapesGraph instances to be double-typed with owl:Ontology, or the system ontology imported that adds the subClassOf triple.

That actually IS what I intended.  I'd written my above remarks concerned with subclassing `owl:Ontology` leading to parsing OWL-specific problems.  I'm no longer convinced my original _reason_ holds, i.e., it might actually be fine from the OWL specification to subclass `owl:Ontology` and consider `owl:Ontology` a `owl:Class`.  However, there is a likely problem with the OWL parsing algorithm affecting `owl:imports`, and this could be a pretty broad effect.

The OWL-RDF mapping document, [Section 3.1.1](https://www.w3.org/TR/owl2-mapping-to-rdf/#Resolving_Included_RDF_Graphs), prescribes specifically how to interpret `owl:imports`.  Specifically, there must be a pair of triples to trigger the "reach-out" behavior:  `x a owl:Ontology .` must accompany `x owl:imports y .`.

Hence, this might be a problematic statement for users following this sequence:

1. Their application is built with a data graph as a store, today or earlier.  The data graph houses an identifying IRI, and ties to an ontology "TBox" graph housing classes and properties.

```turtle
# ExampleDataGraph-37.ttl
<http://example.org/ExampleDataGraph-37>
  a owl:Ontology ;
  owl:imports <http://example.org/ExampleOntologyGraph> .
  .
```

2. We merge PR 814 , defining the specialization class including these axioms:

```turtle
sh:DataGraph
  rdfs:subClassOf owl:Ontology ;
  .
```

3. ExampleOrg goes through a classification exercise, and replaces some ABox headers with this:

```turtle
# ExampleDataGraph-37.ttl
<http://example.org/ExampleDataGraph-37>
  a sh:DataGraph ;
  owl:imports <http://example.org/ExampleOntologyGraph> .
  .
```

4. Some applications processing `ExampleDataGraph-37.ttl` **might now stop executing** the `owl:imports` statement.  From reading the OWL-RDF parsing algorithm, it looks like `owl:imports` would be skipped in the ontology header processing step, and just shifted to being a generic `owl:AnnotationProperty` with no effect.  (This should raise no parsing errors because `owl:imports` is another one of those `owl:`-prefixed IRIs that isn't reserved, per the consumptive parsing prior to Table 18 in [the parsing doc](https://www.w3.org/TR/owl2-syntax/#IRIs).  Thanks again to @mgberg for inspiring me to check this more often.)

I should note, I don't know how common this behavior would be.  But, by any strict implementation of the OWL-RDF parsing algorithm, this would be an unexpected failure from the user's perspective.

This is why I suggested that shape-form above.  I agree that `sh:DataGraph` is a useful classification.  However, I do not think it should subsume statements that an OWL parser might require before running any kind of entailment as pre-processing.

I do realize an implementation issue with the way I spelled the shape.  It should do a direct interrogation for an `rdf:type owl:Ontology` triple, rather than doing a `sh:class` review, since `sh:class` would likely look to SHACL-SHACL's knowledge of SHACL-Core.

> 
> In general, I am highly doubtful that many SHACL graphs fall under OWL DL anyway. I have lost track if OWL DL is even still relevant. My impression is that the majority of OWL implementations are just OWL RL, which ignores any triples on which an OWL DL engine would block.

Oh, is ignoring the behavior you're seeing?  My understanding was OWL RL being more restrictive than DL would mean engines would be suggested to halt more often; but, I haven't experienced a wide variety of RL implementations.  (This is an aside, we don't need to dig into it here.)

HolgerKnublauch:
@ajnelson-nist with apologies in advance but I am not an expert of the details of OWL DL parsing algorithms. Nor am I up to date on how many OWL RL tools will reject graphs that use features that go beyond OWL RL.

If this whole topic is causing such complications, I'd rather not touch it and maybe someone else can take ownership of this ticket.

shacl12-core/index.html
```diff
@@ -2143,7 +2153,7 @@ <h3>Graph for rdfs:subClassOf Triples</h3>
 			</section>
 
 			<section id="sh-shapes-graph">
-				<h3>Linking to shapes graphs (sh:shapesGraph)</h3>
+				<h3>Linking data graphs to shapes graphs (sh:shapesGraph)</h3>
 				<p class="syntax">
 					A <a>data graph</a> can include triples used to suggest one or more graphs to a SHACL processor with the predicate <code>sh:shapesGraph</code>.
 					<span data-syntax-rule="shapesGraph-nodeKind">Every <a>value</a> of <code>sh:shapesGraph</code> is an <a>IRI</a></span> representing a graph that SHOULD be included into the <a>shapes graph</a> used to validate the <a>data graph</a>.
```
ajnelson-nist, line 2167: This line appears to have been muddied.  I wanted to suggest this ...

```suggestion
					<span data-syntax-rule="shapesGraph-nodeKind">Every <a>value</a> of <code>sh:shapesGraph</code> is an <a>IRI</a></span> representing a <a>shapes graph</a> that SHOULD be included into the <a>shapes graph</a> used to validate the <a>data graph</a>.
```

... but I see that the IRI-backed "shapes graph" differs a bit in the "total" ("cumulative?") shapes graph already noted in this line.

It'd be fine with me to apply that edit, and also would be fine to replace the latter "the shapes graph" with "the total shapes graph".

shacl12-core/index.html
ajnelson-nist, line 2131: I went looking for how we had addressed the OWL compatibility matter in #212 , and then came upon the note on `owl:imports` not being enacted.  May we include this additional Note?

```suggestion
				</p>
				<p class="note">
					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a directive to load <code>Y</code>.  See [OWL 2 Web Ontology Language Mapping to RDF Graphs (Second Edition), Section 3.1.1](https://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/#Resolving_Included_RDF_Graphs) for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .` SHOULD also be included.
				</p>
```

I'm aware accepting this suggestion will induce a follow-on need to update the informative-references bibliography, but I wanted to see if this suggestion gets taken before going ahead with that split-suggestion.

ajnelson-nist:
I've suggested a revision advising retention of `owl:Ontology`.

My personal preference would also be to include a shape in SHACL-SHACL, targeting `sh:DataGraph`s using `owl:imports`.  Apologies for quick & awkward spelling, this might be more gracefully spelled with `sh:targetWhere` rather than the if-p-then-q disjunctive spelling.

```turtle
shsh:DataGraphImportsShape
  a sh:NodeShape ;
  rdfs:seeAlso <https://www.w3.org/TR/shacl12-core/#data-graph> ;
  sh:message "In light of the OWL RDF mapping, Data Graphs using owl:imports should explicitly include the type owl:Ontology."@en ;
  sh:severity sh:Info ;
  sh:or (
    [
      sh:property [
        sh:maxCount 0 ;
        sh:path owl:imports ;
      ] ;
    ]
    [
      sh:property [
        sh:hasValue owl:Ontology ;
        sh:path rdf:type ;
      ] ;
    ]
  ) ;
  sh:targetClass sh:DataGraph ;
  .
```

I put both of these suggestions forward to maintain functionality for OWL applications that adopt more SHACL functionality.  `sh:DataGraph` looks to be quite useful with work being done with Profiling (e.g., what's currently in #764 ), but I'd like make sure we don't accidentally degrade existing applications that start to adopt `sh:DataGraph`.

If the SHACL-SHACL shape is also OK, I'll push a patch into this branch.

shacl12-core/index.html
HolgerKnublauch: So the fact that `owl:imports rdfs:domain owl:Ontology` is not enough for OWL processors to infer this rdf:type? This would be surprising.

shacl12-core/index.html
HolgerKnublauch: Yes I would suggest to clarify which shapes graph is which, e.g. by using the terms subject and object.

shacl12-core/index.html
ajnelson-nist: @HolgerKnublauch Apologies, I missed your comment.

My reading of the OWL RDF mapping document section I cited is that the exact triple statements are expected to be in the serialized graph, pre-inferencing.  This comes from being in contact with one of the original authors of that document when I was trying to figure out some nuances of the parsing process.  The "Pre-inferencing" part is my own conclusion from a bootstrapping thought exercise: when would inferencing be run in a deserializing process?  I think reasoning is not to be relied on to make the elementary parse work.

So, may we include the additional note I suggested?

If someone commits this, we have another light TODO to fix the citing style that I couldn't remember as I was drafting the text.

ajnelson-nist:
Re: My last comment:

> [...]
> If the SHACL-SHACL shape is also OK, I'll push a patch into this branch.

I'll push a patch after today's call if there are no objections.

shacl12-core/index.html
ajnelson-nist: Apologies, I missed a part: I was discussing with one of the original authors how to handle some of the parsing logic of section 3's tables, and they highlighted to me the very last line of text before section 4:

> At the end of this process, the graph G MUST be empty.

The parsing process is totally consumptive of the triples in the serialized graph.  So, from that I draw there is a firm expectation of hard-coded text in the serializations, without inferencing.  Inferencing can happen after the graph is parsed and hence confirmed to be OWL-DL or OWL-FULL.

shacl12-core/index.html
ajnelson-nist:
```suggestion
					<span data-syntax-rule="shapesGraph-nodeKind">Every <a>value</a> of <code>sh:shapesGraph</code> is an <a>IRI</a></span> representing a <code>sh:ShapesGraph</code> graph that SHOULD be included into the total <a>shapes graph</a> used to validate the <a>data graph</a>.
```

@mgberg , in light of #861 , would you mind either approving & committing this, or trying out an alternative wording?

FWIW, I've always personally been a fan of "gestalt" in cases like this, but I also think that doesn't hit what we need in the last `shapes graph` anchor.  "Shapes graph closure," short for "Shapes graph transitive closure of direct imports," drawn from ["CP 5" here](https://www.w3.org/TR/2012/REC-owl2-syntax-20121211/#Canonical_Parsing_of_OWL_2_Ontologies)?

shacl12-core/index.html
```diff
@@ -2119,6 +2129,14 @@ <h3>Data Graph</h3>
 					<code>owl:imports</code> in the data graph is not enacted, in order to avoid uncontrolled increase of validation work. If you want to validate
 					several related ontologies, pass all of them to the SHACL processor (together or one by one), do not rely on <code>owl:imports</code> links.
 				</p>
+				<p class="note" id="owlImportsOnDataGraph">
+					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely
+					on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct
+					presence of <code>X a owl:Ontology .</code> in order for <code>X owl:imports Y .</code> to process the <code>owl:imports</code> statement as a
+					directive to load <code>Y</code>.  See <a data-cite="owl2-mapping-to-rdf/#Resolving_Included_RDF_Graphs">OWL 2 Web Ontology Language Mapping
+					to RDF Graphs, Section 3.1.1</a> for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .`
+					SHOULD also be included.
```
ajnelson-nist, line 2138: A remark on SHOULD: The syntactic requirements tying `owl:imports` and `owl:Ontology` together are for parsing OWL 2 DL.  SHACL does not require OWL 2 DL (and only indirectly references OWL 2 DL via `sh:entailment`), so "MUST" didn't seem appropriate to me.

shacl12-core/index.html
mgberg: Based on the conversation we had this morning, I'm resolving this comment without applying any of these changes. The issue relating to which definition of shapes graph is which is discussed in #861.

mgberg (approved):

HolgerKnublauch (commented):
As the original author, I cannot officially approve it but I would.

The PR would need to be changed back from Draft to Ready and @nicholascar would need to remove his Blocking for this to proceed.

ajnelson-nist (approved):

bergos (approved):

YoucTagh (approved):

shacl12-core/index.html
```diff
@@ -2082,7 +2082,11 @@ <h2>Validation and Graphs</h2>
 				<h3>Shapes Graph</h3>
 				<p>
 					A <dfn>shapes graph</dfn> is an RDF graph containing zero or more shapes
-					that is passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
+					that may be passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
+				</p>
+				<p>
+					<span data-syntax-rule="ShapesGraph">The class <code>sh:ShapesGraph</code> MAY be used as an <code>rdf:type</code>
```
```suggestion
					<span data-syntax-rule="ShapesGraph">The <code>sh:ShapesGraph</code> class MAY be used as an <code>rdf:type</code>
```

shacl12-core/index.html
```diff
@@ -2101,6 +2105,12 @@ <h3>Data Graph</h3>
 				<p>
 					Any RDF graph can be a <dfn>data graph</dfn>.
 				</p>
+				<p>
+					<span data-syntax-rule="DataGraph">The class <code>sh:DataGraph</code> MAY be used as an <code>rdf:type</code> of the <a>IRI</a>
+					of a graph that is typically used as a <a>data graph</a>.</span>
+					Note, however, that even a <a>shapes graph</a> may play the role of a <a>data graph</a>, for example
+					when the shapes themselves shall be validated.
```
```suggestion
					<span data-syntax-rule="DataGraph">The <code>sh:DataGraph</code> class MAY be used as an <code>rdf:type</code> of the <a>IRI</a>
					of a graph that is typically used as a <a>data graph</a>.</span>
					Note, however, that a <a>shapes graph</a> MAY also play the role of a <a>data graph</a>; for example,
					when the shapes themselves will be validated.
```

shacl12-core/index.html
TallTed, line 2138:
```suggestion
					While SHACL does not enact <code>owl:imports</code> in the data graph, data graph maintainers should remain aware of applications that do rely
					on <code>owl:imports</code> behavior as an import directive.  In particular, some applications may be syntactically reliant on the direct
					presence of <code>X a owl:Ontology .</code> such that <code>X owl:imports Y .</code> will process the <code>owl:imports</code> statement as a
					directive to load <code>Y</code>.  See <a data-cite="owl2-mapping-to-rdf/#Resolving_Included_RDF_Graphs">OWL 2 Web Ontology Language Mapping
					to RDF Graphs, Section 3.1.1</a> for further detail.  Hence, when using statements of the form `X a sh:DataGraph .`, `X a owl:Ontology .`
					SHOULD also be included.
```

nicholascar (approved):

PR #819, opened 2026-03-10
GH #818 - fix target in Example 59 (and add explanation and minCounts)

### LINKED ISSUE #818: Example 59 - bad target
The only target is:

```
ex:PersonShape
    a sh:NodeShape ;
    sh:targetClass ex:Person ;
    sh:property ex:PersonShape-age ;
.
```
but there is no `rdf:type` triple in the data.

Was `sh:targetNode ex:Bob` intended?

HolgerKnublauch:
Also added explanation and minCounts to make it more meaningful.

Thanks for spotting this, Andy!

Closes #818

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-818/shacl12-core/index.html#http://127.0.0.1:5500/shacl12-core/index.html#ReifierShapeConstraintComponent)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5141,7 +5141,8 @@ ex:HandShape
 						</div>
 					</div>
 
-					<aside class="example">
+					<p><em>The remainder of this section is informative.</em></p>
+					<aside class="example" title="Example of sh:reifierShape and sh:reificationRequired">
 						<div class="shapes-graph">
 							<div class="turtle">
 ex:ProvenanceShape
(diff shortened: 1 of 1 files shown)


afs:
The example would be easier to read if the shapes were in the order:

ex:PersonShape
ex:PersonShape-age
ex:ProvenanceShape
ex:ProvenanceShape-date
ex:ProvenanceShape-author

afs:
ex:PersonShape-age should have a sh:minCount 1

(SHACL should have "sh:count N", which expands to "sh:minCount N, sh:maxCount N", or for the common case "sh:required" for "sh:minCount 1, sh:maxCount 1")

ajnelson-nist (approved):

HolgerKnublauch:
@afs the old WG had discussions about sh:count but explicitly decided against it because it was deemed redundant (it's not always good to have too many ways to express the same thing).

Issue #805, opened 2026-03-05
Profile generalization and implications for dcterms:conformsTo reasoning


mgberg:
At the moment, in #790, the Profiling document says the following:

> ### 4.1.2 Implications for data
> If a [data graph](https://www.w3.org/TR/shacl12-core/#data-graph), DG1 is valid according to a [shapes graph](https://www.w3.org/TR/shacl12-core/#shapes-graph), SG1, then using the general [[DCTERMS](https://raw.githack.com/w3c/data-shapes/profiling-standards/shacl12-profiling/index.html#bib-dcterms)] definition of dcterms:conformsTo, as used by [The Profiles Vocabulary](https://www.w3.org/TR/dx-prof/), we can then say:
> 
> `<DG1> dcterms:conformsTo <SG1> .`
>       
> Following the [The Profiles Vocabulary](https://www.w3.org/TR/dx-prof/) axiom of dcterms:conformsTo owl:propertyChainAxiom ( prof:isProfileOf dcterms:conformsTo ) then, if given
> 
> `<DG1> dcterms:conformsTo <SG1> .`
>       
> and
> 
> `<SG1> prof:isProfileOf <S1> .`
>       
> we can conclude that
> 
> `<DG1> dcterms:conformsTo <S1> .`
>       
> This reasoning allows for static conformance claims of data to profiles of specifications and specifications themselves to be made which will have utility for data whose conformance to these things must be known but which cannot be continuously validated.

The possibility of explaining this as a property path instead of a property chain was discussed here: https://github.com/w3c/data-shapes/pull/790#discussion_r2818017873,

The Profiling document also says the following in that same PR:
> ### 3.2 Union Profile
> The Union Profile of SHACL 1.2 is the conceptual whole of SHACL 1.2 which covers all aspects of the [SHACL 1.2 series of Specifications](https://raw.githack.com/w3c/data-shapes/profiling-standards/shacl12-profiling/index.html#specifications). There is no single human-readable Union Profile specification document — you must read all the documents individually, as linked to in the specification section — but ***there is a single validation resource: a [shapes graph](https://www.w3.org/TR/shacl12-core/#shapes-graph) comprising all the individual profiles' validation [shapes graphs](https://www.w3.org/TR/shacl12-core/#shapes-graph)***.

This raises two related issues:
[...]

PR #825, opened 2026-03-16
SHACL RDF Resources metadata additions

nicholascar:
* added some standard properties for all SHACL RDF resources:
    * `owl:versionIRI`
    * `schema:version "1.2" ;`
    * `schema:creator "W3C Data Shapes Working Group" ;`
    * `schema:dateCreated` - placeholder
    * `schema:dateIssued` - placeholder
    * `schema:dateModified` - placeholder
* starts the SHACL Profiling ontology
* adds a SHACL UI ontology stub

### DIFF
--- a/shacl12-vocabularies/shacl-profiling.ttl
+++ b/shacl12-vocabularies/shacl-profiling.ttl
@@ -10,6 +10,7 @@ PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
 
 shpr:
 	a owl:Ontology ;
+	owl:versionIRI shpr:1.2 ;
 	rdfs:label "W3C Shapes Constraint Language (SHACL) Profiling Vocabulary"@en ;
 	rdfs:comment "This vocabulary defines terms used profiling SHACL and profiling RDF data with SHACL, the W3C Shapes Constraint Language."@en ;
 	sh:declare [
@@ -17,7 +18,7 @@ shpr:
 		sh:namespace "http://www.w3.org/ns/shacl#"^^xsd:anyURI ;
 	] ;
     schema:license "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"^^xsd:anyURI ;
-    schema:creator "W3C Data Shapes 1.2 Working Group" ;
+    schema:creator "W3C Data Shapes Working Group" ;
     schema:dateCreated "2026-03-16"^^xsd:date ;
     # schema:dateIssued ""^^xsd:date ;
     schema:dateModified "2026-03-16"^^xsd:date ;
--- a/shacl12-vocabularies/shacl-profiling.ttl
+++ b/shacl12-vocabularies/shacl-profiling.ttl
@@ -10,18 +10,20 @@ PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
 
 shpr:
 	a owl:Ontology ;
-	rdfs:label "W3C Shapes Constraint Language (SHACL) Profiling Vocabulary"@en ;
-	rdfs:comment "This vocabulary defines terms used profiling SHACL and profiling RDF data with SHACL, the W3C Shapes Constraint Language."@en ;
+	owl:imports <http://www.w3.org/ns/shacl#> ;
+	owl:versionIRI shpr:1.2 ;
+	rdfs:label "W3C SHACL Profiling Vocabulary"@en ;
+	rdfs:comment "This vocabulary defines terms used in the SHACL Profiling specification."@en ;
+    schema:creator "W3C Data Shapes 1.2 Working Group" ;
+    # schema:dateCreated ""^^xsd:date ;
+    # schema:dateIssued ""^^xsd:date ;
+    # schema:dateModified ""^^xsd:date ;
+    schema:license "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"^^xsd:anyURI ;
+    schema:version "1.2" ;
 	sh:declare [
 		sh:prefix "shpr" ;
 		sh:namespace "http://www.w3.org/ns/shacl#"^^xsd:anyURI ;
 	] ;
-    schema:license "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"^^xsd:anyURI ;
-    schema:creator "W3C Data Shapes 1.2 Working Group" ;
-    schema:dateCreated "2026-03-16"^^xsd:date ;
-    # schema:dateIssued ""^^xsd:date ;
-    schema:dateModified "2026-03-16"^^xsd:date ;
-    schema:version "1.2" ;
     rdfs:member
         shpr:conformsTo ;
 .
--- a/shacl12-vocabularies/shacl-profiling.ttl
+++ b/shacl12-vocabularies/shacl-profiling.ttl
@@ -32,6 +32,6 @@ shpr:conformsTo
     rdfs:label "conforms to" ;
     rdfs:comment "An established standard to which the described resource conforms." ;
     rdfs:isDefinedBy shpr: ;
-    rdfs:range dcterms:Standard ;
-    skos:scopeNote "This property is just an OWL ObjectProerty version of DCTERM's conformsTo created to allow for OWL reasoning" ;
+    rdfs:range dcterms:Standard , sh:ShapesGraph ;
+    skos:scopeNote "This property's usage is for a special case of conformance, where the subject conforms to the object by merit of SHACL validation when using the object as a shapes graph." ;
 .
--- a/shacl12-vocabularies/shacl-profiling.ttl
+++ b/shacl12-vocabularies/shacl-profiling.ttl
@@ -11,7 +11,7 @@ PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
 shpr:
 	a owl:Ontology ;
 	rdfs:label "W3C Shapes Constraint Language (SHACL) Profiling Vocabulary"@en ;
-	rdfs:comment "This vocabulary defines terms used profiling SHACL and profiling RDF data with SHACL, the W3C Shapes Constraint Language."@en ;
+	rdfs:comment "This vocabulary defines terms applicable when using SHACL, the W3C Shapes Constraint Language, to profile SHACL and to profile RDF data."@en ;
 	sh:declare [
 		sh:prefix "shpr" ;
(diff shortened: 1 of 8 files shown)


shacl12-vocabularies/shacl-profiling.ttl
```diff
@@ -0,0 +1,37 @@
...
+	sh:declare [
+		sh:prefix "shpr" ;
+		sh:namespace "http://www.w3.org/ns/shacl#"^^xsd:anyURI ;
+	] ;
+    schema:license "https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document"^^xsd:anyURI ;
+    schema:creator "W3C Data Shapes 1.2 Working Group" ;
+    schema:dateCreated "2026-03-16"^^xsd:date ;
+    # schema:dateIssued ""^^xsd:date ;
+    schema:dateModified "2026-03-16"^^xsd:date ;
+    schema:version "1.2" ;
```
ajnelson-nist, line 25: `schema:version`, or `owl:versionIRI`?  Both?

shacl12-vocabularies/shacl-profiling.ttl
```diff
@@ -0,0 +1,37 @@
...
+    rdfs:member
+        shpr:conformsTo ;
+.
+
+shpr:conformsTo
+    a owl:ObjectProperty ;
+    rdfs:subPropertyOf dcterms:conformsTo ;
```
ajnelson-nist, line 36: An OWL commitment: If `x a owl:ObjectProperty .` and `x rdfs:subPropertyOf y .`, then the ontology must also contain `y a owl:ObjectProperty .`. This can be verified by looking for `rdfs:subPropertyOf` on [the OWL RDF mapping document](https://www.w3.org/TR/owl2-mapping-to-rdf/), and noting that without `y a owl:ObjectProperty .` the parse would fail because the 4th match (some lines down in Table 16) can't confirm `OPE(y)`.

```suggestion

dcterms:conformsTo
    a owl:ObjectProperty ;
.

shpr:conformsTo
    a owl:ObjectProperty ;
    rdfs:subPropertyOf dcterms:conformsTo ;
```

shacl12-vocabularies/shacl-profiling.ttl
```diff
@@ -0,0 +1,37 @@
...
+    rdfs:label "conforms to" ;
+    rdfs:comment "An established standard to which the described resource conforms." ;
+    rdfs:isDefinedBy shpr: ;
+    rdfs:range dcterms:Standard ;
+    skos:scopeNote "This property is just an OWL ObjectProerty version of DCTERM's conformsTo created to allow for OWL reasoning" ;
```
ajnelson-nist: We can actually make a stronger justification than allowing for an OWL feature: `x shpr:conformsTo y` can be used to describe a special case of `x dcterms:conformsTo y`, that the conformance is by SHACL validation passing when using `y` as a shapes graph.

```suggestion
    rdfs:range dcterms:Standard , sh:ShapesGraph ;
    skos:scopeNote "This property's usage is for a special case of conformance, where the subject conforms to the object by merit of SHACL validation when using the object as a shapes graph." ;
```

My own preference would be to enable generic OWL reasoning applications (as in your original scope note) in the DX-PROF space.  Should DX-PROF adopt such a convention, `shpr:conformsTo` could specialize `prof:conformsTo` instead of `dcterms:conformsTo`.

ajnelson-nist: If we pursue a definition of `shpr:conformsTo` that ties to SHACL validation, we should also be mindful of `conformanceDisallows` from #453 .

shacl12-vocabularies/shacl-profiling.ttl
```diff
@@ -0,0 +1,37 @@
...
+PREFIX schema: <https://schema.org/>
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX sh:   <http://www.w3.org/ns/shacl#>
+PREFIX shpr: <http://www.w3.org/ns/shacl-pr#>
+PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
+
+shpr:
+	a owl:Ontology ;
+	rdfs:label "W3C Shapes Constraint Language (SHACL) Profiling Vocabulary"@en ;
+	rdfs:comment "This vocabulary defines terms used profiling SHACL and profiling RDF data with SHACL, the W3C Shapes Constraint Language."@en ;
```
TallTed, line 15:
```suggestion
	rdfs:comment "This vocabulary defines terms applicable when using SHACL, the W3C Shapes Constraint Language, to profile SHACL and to profile RDF data."@en ;
```

mgberg:
Note that #805 is effectively a follow up to these changes.

shacl12-vocabularies/shacl-profiling.ttl
nicholascar: I wonder then if we'd not be better of just not declaring the `subPropertyOf`? We might just say that `shpr:conformsTo prov:wasDerivedFrom dcterms:conformsTo` so people know there's a link there but don't look to reason over it. After all, we are adding the ObjectProperty contraint where DCTERMS has none. So if we could get away with `dcterms:conformsTo  a owl:ObjectProperty` then we wouldn’t really need the specialised subProperty!

ajnelson-nist: Let's first visit the thread below, on whether we want `shpr:conformsTo` to mean "Conformance is taken to mean SHACL validation with this shape graph passes."
If we accept that as the interpretation, then `shpr:conformsTo` as a subproperty of `dcterms:conformsTo` picks up the benefit of preserving the property chain axiom from PROF.

Designating `dcterms:conformsTo` as an `owl:ObjectProperty` is a decision that I think is best scoped to PROF, not SHACL Profiling.

Another consideration is that OWL DL conformance does not seem to be a prerequisite of OWL-RL expansion; see [this comment](https://github.com/w3c/data-shapes/issues/219#issuecomment-4092693460) where I'd dug into the OWL specs a bit.  So, if PROF is declaring itself an OWL 2 DL conformant ontology, then the `ObjectProperty` extra type designation is a concern.  If SHACL Profiling is declaring itself an OWL 2 DL ontology, then SHACL Profiling needs to have `dcterms:conformsTo a owl:ObjectProperty`, whether by PROF or by itself.

If OWL 2 DL conformance is not a concern, then the property chain expansion from PROF will still work.

shacl12-vocabularies/shacl-profiling.ttl
nicholascar: I've added `owl:versionIRI` as we can use that for ontology work (and added `:1.2` to all the vocabs) and retained `schema:version` as a plain language rendering of the version number.

We could see `owl:versionInfo` used but I tend to think of that as containing, y'know, info, so more than a "1.2" but a sentence about the version which we don't yet have anything meaningful to say about. Perhaps we can add this predicate with values if we see a `1.2.1` appear after this WG and we need to state what's in `1.2.1` v. `1.2`.

shacl12-vocabularies/shacl-profiling.ttl
nicholascar: Great suggestion below, merged in!

shacl12-vocabularies/shacl-shacl-pr.ttl
```diff
@@ -0,0 +1,27 @@
+PREFIX owl:  <http://www.w3.org/2002/07/owl#>
+PREFIX rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX sh:   <http://www.w3.org/ns/shacl#>
+PREFIX shsh: <http://www.w3.org/ns/shacl-shacl#>
+PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
+
```
simonstey:
```suggestion
PREFIX schema: <https://schema.org/>
```

shacl12-vocabularies/shacl-shacl.ttl
```diff
@@ -12,12 +13,22 @@
 @prefix shsh: <http://www.w3.org/ns/shacl-shacl#> .
```
simonstey, line 15:
```suggestion
@prefix schema: <https://schema.org/> .

```

shacl12-vocabularies/shacl-ui.ttl
```diff
@@ -0,0 +1,27 @@
+PREFIX owl:  <http://www.w3.org/2002/07/owl#>
+PREFIX rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX sh:   <http://www.w3.org/ns/shacl#>
+PREFIX shsh: <http://www.w3.org/ns/shacl-shacl#>
+PREFIX xsd:  <http://www.w3.org/2001/XMLSchema#>
+
```
simonstey:
```suggestion
PREFIX schema: <https://schema.org/>
```

shacl12-vocabularies/shacl.ttl
simonstey:
```suggestion
@prefix schema: <https://schema.org/> .
```

shacl12-vocabularies/shnex-sparql.ttl
```diff
@@ -17,13 +17,21 @@
```
simonstey, line 18:
```suggestion
@prefix schema: <https://schema.org/> .

```

shacl12-vocabularies/shnex.ttl
simonstey:
```suggestion
@prefix schema: <https://schema.org/> .
```

simonstey (changes requested):
needs missing prefix declarations for `schema:` 

(sidenote: despite both being technically correct, we should (imo) follow a single style of declaring prefixes, i.e. either `@prefix` or `PREFIX`)

nicholascar:
> needs missing prefix declarations for `schema:`
> 
> (sidenote: despite both being technically correct, we should (imo) follow a single style of declaring prefixes, i.e. either `@prefix` or `PREFIX`)

Thanks @simonstey for fixing the prefixes. I haven't run RDF checking on the files just yet as I expect to add more to them soon.

Yes, I would like to move to the Turtle 1.1 form of `PREFIX` rather than `@prefix` because it matched better with SPARQL, but before converting over, I need to check that this won't cause issues for anyone.

simonstey (approved):

PR #833, opened 2026-03-23
address GH#359 -- Add `sh:describe` predicate

### LINKED ISSUE #359: Add describe and update predicates
[We (mis)use]([https://github.com/sib-swiss/sparql-examples]) the SHACL specified properties to structurally encode SPARQL examples.

So we have a lot of triples with this as a predicate

 * [ask](http://www.w3.org/ns/shacl#ask)
 * [select](http://www.w3.org/ns/shacl#select)
 * [construct](http://www.w3.org/ns/shacl#construct)

It would be nice if the other SPARQL (update) types of queries would also be included.
The graph maintenance options would be interesting for rules even they may not be use able in a validation environment.

We have defined our own predicate for the `describe` queries, which are understandably not in SHACL 1, but it would be a trivial addition.

HolgerKnublauch:
Closes #359

### DIFF
--- a/shacl12-vocabularies/shacl.ttl
+++ b/shacl12-vocabularies/shacl.ttl
@@ -1499,6 +1499,21 @@ sh:construct
 	rdfs:range xsd:string ;
 	rdfs:isDefinedBy sh: .
 
+sh:SPARQLDescribeExecutable
+	a rdfs:Class ;
+	rdfs:label "SPARQL DESCRIBE executable"@en ;
+	rdfs:comment "The class of SPARQL executables based on a SPARQL DESCRIBE."@en ;
+	rdfs:subClassOf sh:SPARQLExecutable ;
+	rdfs:isDefinedBy sh: .
+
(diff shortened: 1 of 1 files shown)


nicholascar (approved):

bergos (approved):

ajnelson-nist:
Is this fine to go in without a unit test?  I'm just unclear on what `sh:describe` is expected to _do_ in SHACL (keeping in mind [@TallTed 's comment](https://github.com/w3c/data-shapes/issues/359#issuecomment-3987660886)).  Is it expected to behave more like `sh:select` or `sh:ask`?  As a SPARQL constraint, how can `sh:describe` induce a result, or is that outside its scope?

HolgerKnublauch:
@ajnelson-nist this new property exists only in the TTL file and has no dedicated meaning in SHACL other than being a syntax the represent custom use cases. It is not used for validation or anything else, therefore it cannot have test cases.

PR #836, opened 2026-03-24
Add that a rule set has import statements

afs:
* [See this section rendered online here](https://raw.githack.com/w3c/data-shapes/afs/ruleset-extra/shacl12-rules/index.html#dfn-rule-set)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -532,6 +532,12 @@ RULE { ?x :name ?FN } WHERE {
             A [=rule body=] is a sequence of [=rule body elements=].
           </dd>
 
+          <dt><dfn data-lt="rule imports|imports">Rule imports</dfn></dt>
+          <dd>
+            [=Rule imports=] (often just "imports") are a collection of URLs
+            for other rule sets that will be included during evaluation.
+          </dd>
+
           <dt><dfn>Rule</dfn></dt>
           <dd>
             A [=rule=] is a pair of a [=rule head=] (often just "head") and
@@ -540,8 +546,9 @@ RULE { ?x :name ?FN } WHERE {
 
           <dt><dfn>Rule set</dfn></dt>
           <dd>
-            A [=rule set=] is a collection of zero or more [=rules=]
-            and a collection of zero or more [=data blocks=].
+            A [=rule set=] is a collection of zero or more [=rules=],
+            a collection of zero or more [=data blocks=],
+            and a collection of zero or more [=rule imports].
           </dd>
 
           <dt><dfn>Base graph</dfn></dt>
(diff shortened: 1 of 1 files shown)


robert-david (approved):

shacl12-rules/index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+rule <!DOCTYPE html>
```
```suggestion
<!DOCTYPE html>
```

shacl12-rules/index.html
```diff
@@ -540,8 +540,10 @@ <h3>Elements of the Abstract Syntax</h3>
 
           <dt><dfn>Rule set</dfn></dt>
           <dd>
-            A [=rule set=] is a collection of zero or more [=rules=]
-            and a collection of zero or more [=data blocks=].
+            A [=rule set=] is a collection of zero or more [=rules=],
+            a collection of zero or more [=data blocks=],
+            and collection of zero or more
```
```suggestion
            and a collection of zero or more 
```

shacl12-rules/index.html
```diff
@@ -540,8 +540,10 @@ <h3>Elements of the Abstract Syntax</h3>
 
           <dt><dfn>Rule set</dfn></dt>
           <dd>
-            A [=rule set=] is a collection of zero or more [=rules=]
-            and a collection of zero or more [=data blocks=].
+            A [=rule set=] is a collection of zero or more [=rules=],
+            a collection of zero or more [=data blocks=],
+            and collection of zero or more 
+            <dfn data-lt="rule imports|imports">imports</dfn> statements.
```
simonstey: couldn't we define imports the same way we did for e.g. data blocks?

something along the lines of:

```html
          <dt><dfn data-lt="import statement|import statements|rule imports|imports">Import statement</dfn></dt>
          <dd>
            An [=import statement=] is an [=IRI=] that identifies another
            [=rule set=] whose [=rules=] and [=data blocks=] are included
            in the importing [=rule set=].
          </dd>
```

and then we could have this be:


```suggestion
            [=import statements=].
```

simonstey (approved):
we should also add a (normative?) note before the eval algo in https://github.com/w3c/data-shapes/blob/685f1ba228ba8c7e0b0010690b34f406305e8c59/shacl12-rules/index.html#L1357 that states _when_ imports should be resolved (e.g., that imports in the rule set are resolved _before_ eval begins; or yet to be decided.. )

afs:
> we should also add a (normative?) note before the eval algo in
> 
> https://github.com/w3c/data-shapes/blob/685f1ba228ba8c7e0b0010690b34f406305e8c59/shacl12-rules/index.html#L1357
> 
> that states _when_ imports should be resolved (e.g., that imports in the rule set are resolved _before_ eval begins; or yet to be decided.. )

that's another PR which is work-in-progress :-)

shacl12-rules/index.html
afs: Change made to make it a description list level item. 

https://raw.githack.com/w3c/data-shapes/afs/ruleset-extra/shacl12-rules/index.html#dfn-rule-imports

It is a collection of URLs to other rulesets included at evaluation time.

Removed "statement" which is more about concrete syntax.

simonstey:
> > we should also add a (normative?) note before the eval algo in
> > https://github.com/w3c/data-shapes/blob/685f1ba228ba8c7e0b0010690b34f406305e8c59/shacl12-rules/index.html#L1357
> > 
> > that states _when_ imports should be resolved (e.g., that imports in the rule set are resolved _before_ eval begins; or yet to be decided.. )
> 
> that's another PR which is work-in-progress :-)

fair enough ;)

simonstey (approved):
perfect!

PR #838, opened 2026-03-24
Dependency graph: Triple patterns inside negation elements

afs:
The dependency graph algorithm needs to look inside negation elements to find triple patterns.

* [See the algorithm rendered online here](https://raw.githack.com/w3c/data-shapes/afs/dependency-graph-negation-elements/shacl12-rules/index.html#dependency-graph-construction-algorithm)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -486,8 +486,8 @@ RULE { ?x :name ?FN } WHERE {
 
           <dt><dfn>Negation element</dfn></dt>
           <dd>
-            A [=negation element=] is a [=rule body element=].
-            It takes a sequence of [=triple patterns=] and
+            A [=negation element=] is a [=rule body element=]
+            comprised of a sequence of [=triple pattern elements=] and
             [=condition elements=].
           </dd>
 
@@ -690,7 +690,7 @@ RULE { ?x :name ?FN } WHERE {
             The following algorithm gives one
             possible method for constructing the [=dependency graph=] from a
             [=rule set=]. Conformance depends on the dependency definitions 
-			in this section, not on use of this procedure.
+            in this section, not on use of this procedure.
           </p>
           <div class="algorithm">
             <pre>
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -711,27 +711,37 @@ <h4>Dependency Graph Algorithm</h4>
...
         let bodyDependencies = {}
-        foreach triplePatternElement TP in the body of R1:
+        foreach rule body element RBE in the body of R1:
             let item be a new pair (triplePattern, depLabel)
-            item.triplePattern = TP
-            if TP is a negation element:
-                item.depLabel="negative"
-            else:
+            if RBE is a negation element:
+                foreach triple pattern element TPE in RBE:
```
simonstey, line 720: In the definition of Negation elements, we say:

```html
          <dt><dfn>Negation element</dfn></dt>
          <dd>
            A [=negation element=] is a [=rule body element=].
            It takes a sequence of [=triple patterns=] and
            [=condition elements=].
          </dd>
```

and since we explicitly define TPE as:

```html
         <dt><dfn>Triple pattern element</dfn></dt>
          <dd>
            A [=triple pattern element=] is a [=triple pattern=]
            used as a [=rule body element=].
          </dd>
```

shouldn't this then actually read:


```suggestion
                foreach triple pattern TPE in RBE:
```

?

shacl12-rules/index.html
```diff
@@ -711,27 +711,37 @@ <h4>Dependency Graph Algorithm</h4>
...
+                    item.triplePattern = TPE
+                    item.depLabel="negative"
+                    add item to bodyDependencies
+                endfor
+            else if RBE is a triple pattern element
```
```suggestion
            else if RBE is a triple pattern element: 
```

just adds a : like for the if branch

shacl12-rules/index.html
```diff
@@ -711,27 +711,37 @@ <h4>Dependency Graph Algorithm</h4>
...
+                    item.triplePattern = TPE
+                    item.depLabel="negative"
+                    add item to bodyDependencies
+                endfor
+            else if RBE is a triple pattern element 
+                item.triplePattern = RBE
                 item.depLabel="positive"
-            add item to bodyDependencies
+                add item to bodyDependencies
+            else if RBE is a condition element
```
```suggestion
            else if RBE is a condition element:
```

shacl12-rules/index.html
```diff
@@ -711,27 +711,37 @@ <h4>Dependency Graph Algorithm</h4>
...
+                ## Do nothing
+            else if RBE is an assignment element
```
```suggestion
            else if RBE is an assignment element:
```

shacl12-rules/index.html
```diff
@@ -711,27 +711,37 @@ <h4>Dependency Graph Algorithm</h4>
...
             let item be a new pair (triplePattern, depLabel)
-            item.triplePattern = TP
-            if TP is a negation element:
-                item.depLabel="negative"
-            else:
+            if RBE is a negation element:
+                foreach triple pattern element TPE in RBE:
+                    item.triplePattern = TPE
+                    item.depLabel="negative"
+                    add item to bodyDependencies
```
simonstey, line 722:
```
let item be a new pair (triplePattern, depLabel)    ← created once per outer iteration
if RBE is a negation element:
    foreach triple pattern element TPE in RBE:
        item.triplePattern = TPE                     ← mutated
        item.depLabel="negative"                     ← mutated
        add item to bodyDependencies                 ← added each iteration
    endfor
```

it's always the same item that gets updated and in theory (depending on the respective copy semantics) only the last mutation "survives" (deep vs shallow copy). Bug for feature?

Couldn't we move this inside the ifs instead?

i.e. something along the lines of:

```
        let bodyDependencies = {}
        foreach rule body element RBE in the body of R1:
            if RBE is a negation element:
                foreach triple pattern TPE in RBE:
                    let item be a new pair (TPE, "negative")
                    add item to bodyDependencies
                endfor
            else if RBE is a triple pattern element:
                let item be a new pair (RBE, "positive")
                add item to bodyDependencies
            else if RBE is a condition element:
                ## Do nothing
            else if RBE is an assignment element:
                ## Do nothing
            endif
        endfor
```

simonstey (approved):
only minor remarks

shacl12-rules/index.html
afs: I think it is better to change negation to be "triple pattern element" - that goes with "condition element".

shacl12-rules/index.html
afs: good idea - done

afs:
Changes incorporated (probably needs a hard refresh to see in the githack rendered PR document).

shacl12-rules/index.html
simonstey: also fine!

PR #851, opened 2026-04-06
Issue 596: Language Resolution

### LINKED ISSUE #596: [UI] Literal language handling
There are different sets of languages that may impact widget handling multi-language literals. We should create a section that describes the logic/priorities of the different sets.

The sets are:
1. values of `sh:languageIn`
2. application languages (based on configuration or browser settings)
3. full ISO list

The order of the list represents the proposed priority of the set used by the widget.

bergos:
Adds the language resolution logic to the UI specification.

Closes #596

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/lang-resolution/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1106,9 +1106,9 @@ ex:PersonShape
 
           <div class="data-graph">
             <div class="turtle">
-              ex:alice a ex:Person ;
-              foaf:name "Alice"@en ;
-              foaf:name "Alice"@fr .
+ex:alice a ex:Person ;
+    foaf:name "Alice"@en ;
+    foaf:name "Alice"@fr .
             </div>
           </div>
 
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1039,7 +1039,91 @@ ex:part-b
 
     <section id="label-and-lang">
       <h2>Label and Language Resolution</h2>
-      <p>Content.</p>
+
+      <p>
+        UI literals — such as labels, descriptions, and other human-readable strings required for
+        rendering a user interface — are selected based on language tags. This section defines how
+        an application determines which language to use when multiple language-tagged literals are
+        available for a given property.
+      </p>
+
+      <section id="lang-resolution">
+        <h3>Language Resolution</h3>
+
+        <p>
+          When selecting a UI literal, the preferred language is determined according to the following order of
+          priority. Each option itself provides an ordered list of languages where earlier entries are preferred over
+          later ones:
+        </p>
+
+        <ol>
+          <li>
+            The values of <code>sh:languageIn</code>, when present in the applicable <a>shapes graph</a> context. This
+            specification extends the semantics of <code>sh:languageIn</code> giving the order of language tags in the
+            list a meaning. Implementations MUST prefer literals according to the order of the
+            <code>sh:languageIn</code> values.
+          </li>
+          <li>
+            A list of languages selected in the application. How this language preference list is configured or
+            expressed is up to the implementation. Possible approaches include but are not limited to:
+            <ul>
+              <li>A UI feature that allows the user to select a preferred display language or languages priorities.</li>
+              <li>A predefined language configuration in the application settings.</li>
+            </ul>
+          </li>
+          <li>
+            The list of preferred languages declared in the user's browser, as expressed through the
+            <code>Accept-Language</code> HTTP header or the <code>navigator.languages</code> Web API. These SHOULD be
+            used as the default values when no application-level language preference has been configured.
+          </li>
+        </ol>
+
+        <p>
+          Language tags are matched according to the basic filtering scheme defined in [[RFC4647]] section 3.3.1.
+          In particular, a language tag such as <code>en-US</code> SHOULD be considered a match for a preferred
+          language of <code>en</code>.
+        </p>
+
+        <p>
+          If no literal matching the preferred language is available, implementations MAY fall back to a literal in
+          another available language or apply the <a>label resolution</a> fallback strategy as defined for the relevant
+          context.
+        </p>
+
+        <aside class="example" title="Language selection based on sh:languageIn">
+          <div class="shapes-graph">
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
```diff
@@ -1039,7 +1039,91 @@ <h4>Other Complex Paths</h4>
 
     <section id="label-and-lang">
       <h2>Label and Language Resolution</h2>
-      <p>Content.</p>
+
+      <p>
+        UI literals — such as labels, descriptions, and other human-readable strings required for
+        rendering a user interface — are selected based on language tags. This section defines how
+        an application determines which language to use when multiple language-tagged literals are
+        available for a given property.
```
smessie, line 1047:
```suggestion
        UI literals, such as labels, descriptions, and other human-readable strings used for
        rendering a user interface, are selected based on language tags. This section defines how
        an application determines which language to use when multiple language-tagged literals are
        available for a given property.
```

shacl12-ui/index.html
```diff
@@ -1039,7 +1039,91 @@ <h4>Other Complex Paths</h4>
...
+          priority. Each option itself provides an ordered list of languages where earlier entries are preferred over
+          later ones:
+        </p>
+
+        <ol>
+          <li>
+            The values of <code>sh:languageIn</code>, when present in the applicable <a>shapes graph</a> context. This
+            specification extends the semantics of <code>sh:languageIn</code> giving the order of language tags in the
+            list a meaning. Implementations MUST prefer literals according to the order of the
+            <code>sh:languageIn</code> values.
```
robert-david, line 1064: , from left to right.

shacl12-ui/index.html
```diff
@@ -1039,7 +1039,91 @@ <h4>Other Complex Paths</h4>
...
+        <p>
+          Language tags are matched according to the basic filtering scheme defined in [[RFC4647]] section 3.3.1.
+          In particular, a language tag such as <code>en-US</code> SHOULD be considered a match for a preferred
+          language of <code>en</code>.
+        </p>
+
+        <p>
+          If no literal matching the preferred language is available, implementations MAY fall back to a literal in
+          another available language or apply the <a>label resolution</a> fallback strategy as defined for the relevant
+          context.
```
robert-david: Preferred default should be a literal with no language?

robert-david (approved):

shacl12-ui/index.html
smessie, line 1064:
```suggestion
            The values of <code>sh:languageIn</code>, when present in the applicable <a>shapes graph</a> context. The use
             of `sh:languageIn` in this document extends its semantics, providing a meaningful order to the  language tags
            in the list. Implementations MUST prefer literals according to the order of the
            <code>sh:languageIn</code> values.
```

shacl12-ui/index.html
```diff
@@ -1039,7 +1039,91 @@ <h4>Other Complex Paths</h4>
...
+          <div class="shapes-graph">
+            <div class="turtle">
+              ex:PersonShape
+              a sh:NodeShape ;
+              sh:property [
+              sh:path foaf:name ;
+              sh:name "Name"@en ;
+              sh:name "Nom"@fr ;
+              sh:languageIn ( "fr" "en" ) ;
+              ] .
```
```suggestion
ex:PersonShape
    a sh:NodeShape ;
    sh:property [
        sh:path foaf:name ;
        sh:name "Name"@en ;
        sh:name "Nom"@fr ;
        sh:languageIn ( "fr" "en" ) ;
    ] .
```

shacl12-ui/index.html
```diff
@@ -1039,7 +1039,91 @@ <h4>Other Complex Paths</h4>
...
+            </div>
+          </div>
+
+          <div class="data-graph">
+            <div class="turtle">
+              ex:alice a ex:Person ;
+              foaf:name "Alice"@en ;
+              foaf:name "Alice"@fr .
```
```suggestion
ex:alice a ex:Person ;
    foaf:name "Alice"@en ;
    foaf:name "Alice"@fr .
```

smessie (approved):
Looks good to me content-wise. I just included some suggestions below to improve the wording to my feeling.

shacl12-ui/index.html
TallTed: Using commas to offset this sub-sentence makes it more difficult to parse, especially but not only for people for whom English is a second or later language. The em-dashes offset the sub-sentence more clearly, such that readers can ignore that section when reading for comprehension, and then return to parse the "such as" details.

shacl12-ui/index.html
TallTed: This rephrasing results in an inaccurate statement.

TallTed: I'm not sure this is necessary, as the lead in to the ordered list states `earlier entries are preferred over later ones`. But if it is...
```suggestion
            <code>sh:languageIn</code> values, from left to right and top to bottom.
```
But this seems to assume one serialization (which seems likely to be Turtle), which I don't think is required anywhere.

shacl12-ui/index.html
TallTed: Starting from @robert-david's suggestion (which only commented on line 1090)
```suggestion
          If no literal matching the preferred language is available, implementations MAY fall back to a literal in
          another available language (or without a declared language) or apply the <a>label resolution</a>
          fallback strategy as defined for the relevant context.
```

shacl12-ui/index.html
bergos: As suggested by @TallTed , I will keep the original.

shacl12-ui/index.html
bergos: As suggested by @TallTed , I will keep the original.

shacl12-ui/index.html
bergos: I've applied the suggestion from @TallTed .

shacl12-ui/index.html
bergos: I prefer the version without left to right as it's serialization specific.

PR #854, opened 2026-04-10
Add Ariadne SHACL EARL report (121/121)

hajovonta:
Ariadne is a graph database implemented in Common Lisp. This report covers all 121 tests in the SHACL test suite (98 Core + 23 SPARQL), all passing.
 
Homepage: https://sr.ht/~hajovonta/ariadne

### DIFF
--- a/data-shapes-test-suite/reports/ariadne-shacl-earl.ttl
+++ b/data-shapes-test-suite/reports/ariadne-shacl-earl.ttl
@@ -14,6 +14,11 @@
   doap:description "Graph database in Common Lisp" ;
   doap:programming-language "Common Lisp" ;
   doap:homepage <https://sr.ht/~hajovonta/ariadne> ;
+  doap:release [
+    rdf:type doap:Version ;
+    doap:revision "0.1.0" ;
+    doap:created "2026-04-15"^^xsd:date ;
+  ] ;
 .
 [
   rdf:type earl:Assertion ;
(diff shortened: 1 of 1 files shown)


TallTed:
As submitted, your report suggests that every version of Ariadne — past, present, and future — has been tested and passed. I suggest modifying your report to include the specific version under test, optimally embedded in the `earl:subject` value if nowhere else. This will make it possible to report different results for each future version, which will be of particular importance if any of the tests are ever failed.

afs:
[DOAP](https://github.com/ewilderj/doap/blob/master/schema/doap.rdf) has `doap:release` --
```
<https://sr.ht/~hajovonta/ariadne>
    rdf:type      doap:Project;
    doap:release  [ rdf:type       doap:Version;
                    doap:created   "......................."^^xsd:date;
                    doap:revision  ".................." ] ;
    ... 
```

hajovonta:
Hi @TallTed @afs,

thanks for the heads up - I have added the version information.

hajovonta:
I believe the change is non-substantive, since it does not modify the specification in any way.

HolgerKnublauch:
I guess the hesitation from our side is that nobody really still wants to update pages for the 1.0 version when 1.2 will be ready by the end of the year anyway. Are you by any chance tracking the spec and will submit an updated report with the 1.2 test cases?

afs:
Ping @nicholascar @PapoutsoglouE.

hajovonta:
@HolgerKnublauch 

I irregularly look at the space what happens there, but if you could suggest a specific place to watch, that would be helpful! Of course, I can update the report with 1.2 later when it is available.

HolgerKnublauch:
The current SHACL 1.2 Core draft (which gets continuously updated) has been quite stable for the last few months, so it may be worth updating your implementation. It would help us a lot to have additional feedback before the freeze deadlines.

https://www.w3.org/TR/shacl12-core/

PR #860, opened 2026-04-17
Review of normative wording and editorial corrections

YoucTagh:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/profile-review-normative-wording/shacl12-profiling/index.html)


shacl12-profiling/index.html
```diff
@@ -664,7 +664,7 @@ <h3>Indicating Element definition</h3>
       a warning is issued that while <a data-cite="shacl12-core#dfn-property-shape">property shapes</a> may be declared
       as <a data-cite="shacl12-core#dfn-blank-node">blank nodes</a>, it is preferable to declare them as
       <a data-cite="rdf12-concepts/#dfn-iri">IRIs</a> as this enables better reuse across graphs. If including information
-      about the shapes as per this section, IRIs really should be used.
+      about the shapes as per this section, IRIs MUST be used.
```
mgberg: A couple questions about this:
1. This note is talking about _property shapes_, but the example above is for a _node shape_. What exactly are we trying to communicate here? Is it:
    - Node shapes shouldn't be blank nodes when used this way despite the reference to property shapes
    - You could also use a property shape as the subject of `rdfs:isDefinedBy` in this manner, and therefore the guidance about property shapes apply
    - Both of the above
    - Something else?

    Regardless, I think it might be good to clarify what things we recommend be subjects of `rdfs:isDefinedBy` in this context. Should all node shapes have that declaration, and you include all associated property shapes along with it? Should you annotate property shapes explicitly?
2. Is it strictly true that you `MUST` use IRIs in this case? I definitely think one _should_ use IRIs in this case (and in general), but is it actually required?

TallTed: It seems to me far more likely that the appropriate change here is from the non-spec "really should" language, to the normative "SHOULD".
```suggestion
      about the shapes as per this section, IRIs SHOULD be used.
```

ajnelson-nist: If IRIs aren't used in this case, then, informally, all the sharing benefits are a wash.

MUST implies there's a failing case presented by using blank nodes here - but it's more of a soft-failing case, because the firm context-free conformance requirements (i.e., blank node usage as syntactically permitted outside this section) are still satisfied.

I think MUST in this context doesn't work.  I agree with @TallTed , "SHOULD" seems right.

shacl12-profiling/index.html
```diff
@@ -664,7 +664,7 @@ <h3>Indicating Element definition</h3>
       a warning is issued that while <a data-cite="shacl12-core#dfn-property-shape">property shapes</a> may be declared
       as <a data-cite="shacl12-core#dfn-blank-node">blank nodes</a>, it is preferable to declare them as
       <a data-cite="rdf12-concepts/#dfn-iri">IRIs</a> as this enables better reuse across graphs. If including information
```
ajnelson-nist, line 666: Non-blocking remark, Re:

> ...it is preferable to declare them as IRIs as this enables better reuse across graphs. ...

I mostly agree with this.  However, property shape IRI identification isn't always a strictly-better benefit.  Consider graph 1 and graph 2 below.

Graph 1:

```turtle
[]
  a sh:PropertyShape ;
  sh:class ex:Foo ;
  sh:path [ sh:inversePath ex:bar ] ;
  .
```

Graph 2:

```turtle
ex:SomePropertyShape
  a sh:PropertyShape ;
  sh:class ex:Foo ;
  sh:path [ sh:inversePath ex:bar ] ;
  .
```

The guidance suggesting IRI-identifying property shapes favors graph 2.  But, graph 2 fails syntax requirements if somehow loaded/imported twice into one graph.

So, graph 1 is "better" in contexts where it's infeasible to try to manage loading every shapes graph exactly once.  Some execution would be redundant, but it would execute instead of failing at parse time.

I wonder if this is another "Profile of SHACL" opportunity - a SHACL syntactic subset that is safe to load multiple times?

ajnelson-nist: Whoops, I missed that the text I highlighted is just barely out of scope of the PR.  Consider this _really_ non-blocking.

shacl12-profiling/index.html
ajnelson-nist: Meanwhile, looking at the context, I think it's not too far a leap to go from a node shape example to property shape syntax.  The overall thrust of the section is on how to identify shapes for reuse.  Property shapes are worth calling out because of the quick drafting pattern, and/or strong-gluing, of inlining property shapes as blank nodes on node shapes.
Maybe it's worth inlining a sentence on that, but I don't think that needs to be a blocker on this PR.

ajnelson-nist (approved):
I think all the other discussion outside of the one word we changed was good to think through, but seems to me to be non-blocking.

shacl12-profiling/index.html
```diff
@@ -865,13 +865,13 @@ <h3>Creating other Profiles</h3>
     <ol>
       <li>Identify the profile with an IRI</li>
       <li>
-        Define the profile as profiles of SHACL 1.2 are defined
+        Define the profile as profiles of SHACL 1.2 as defined in this specification
```
TallTed, line 868: This "sentence" isn't a sentence. I *think* the fix is just to revert the `as` to `are`.
```suggestion
        Define the profile as profiles of SHACL 1.2 are defined in this specification
```

PR #865, opened 2026-04-21
GH-829: Stratification and evaluation for run-once rules

### LINKED ISSUE #829: An approach to handling new RDF terms - "run once" rules.
See also #688, #749

Creating new RDF terms causes problems in several ways:

1. Generator "functions"

* [`BNODE()`](https://www.w3.org/TR/sparql12-query/#func-bnode)
* [`BNODE(str)`](https://www.w3.org/TR/sparql12-query/#func-bnode)
* [`UUID()`](https://www.w3.org/TR/sparql12-query/#func-uuid)
* [`STRUUID()`](https://www.w3.org/TR/sparql12-query/#func-struuid)
* Can be a user-supplied custom functions called by URI

A value from one of these functions changes everytime the function is evaluated.

2. Propagation of new RDF terms to other rules.

```
# Simple propagation dependency
RULE { :x :p1 ?Z1 } WHERE { :x :p ?Z . BIND(?Z+1 As ?Z1) }
RULE { :y :q ?A } WHERE { :x :p1 ?A }
```
```
# Cycle
RULE { :x :p ?Z1 } WHERE { :x :p ?Z . BIND(?Z+1 As ?Z1) }
```

There are three sub-cases for propagation - a chain or tree of propagation, DAGs
(directed acyclic graphs - i.e. shared rules, but not cycles) which for
generator functions are not equivalent to treating DAGs as trees, and cycles/recursion.

This affects negation as well.

Evaluation of a stratum is defined by naïve datalog execution. A stratum is evaluated by simply repeatedly executing a loop over all the rules until no new triples are generated. Each rule is run every time round the loop. semi-naïve evaluation track rules that don't have any more output. 

That means running a rule that does generate one or more triples will be run at least twice, once to get output, and a second time to find out there are no new triples.

With a form of stratification, if a rule only depends on lower strata, then it will always generates the same output and needs to run only once.

Can we use this idea and extend it to have "run once" rules for assignment and generator functions? "run once" would be by static analysis.

A form of stratification (i.e. modify the stratification we currently have) would, loosely speaking, split each stratum into a run-once (done first) rules set,  and the rest of the layer (including cycles) and it can be defined by naïve execution. (And there isn't a single stratification, which by the algorithm ATM, there is.) Alternating layers of run-once, and run-to-completion.

There is implicit creation/assignment in rule head templates which contain blank nodes. so the run once condition needs to apply here as well. This might be the biggest impact on users.

The question is whether this results in a usable system.

It is not clear what affect it has on the top-down/backward chaining implementation style.

It would also be possible to go into a more detailed value-propagation analysis. That might enable supporting some more combination of rules (but how many? Is it useful?) but it's more complicated to implement and harder to explain.

afs:
Closes #829

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/rules-run-once/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -338,7 +338,7 @@
       <section id="desc-basic">
         <h3>Basic Usage</h3>
 
-       <p>In this first example, we have the following data graph and rule set:</p>
+        <p>In this first example, we have the following data graph and rule set:</p>
 
         <div class="data">
 <pre>
@@ -531,11 +531,14 @@ WHERE {
         <p>
           Rules involving assignment are [=run-once rules=]. Rules with assignment
           are run after all the rules that could produce the data that they depend on
-          and before any rules that depend on the data they produce.
+          and before any rules that depend on the data they produce. 
+          Rules that involve blank node in the [=rule head=] are also creating new
+          RDF terms and are run-once rules.
         </p>
         <p>
-          This condition ensures that rules with assignment do not generate new RDF terms
-          multiple times, with potentially different outcomes.
+          This condition ensures that rules that create RDF terms do not generate new terms
+          multiple times, with potentially different outcomes and also that such rules do not 
+          loop back to themselves and cause an unbounded number of RDF terms.
         </p>
       </section>
       
@@ -635,8 +638,13 @@ WHERE {
 
     <section id="rules-abstract-syntax">
       <h2>Shape Rules Abstract Syntax</h2>
-
-      <p>The Shape Rules Abstract Syntax</p>
+      
+      <p>
+        The Shape Rules Abstract Syntax is the logical structure of SHACL Rules.
+        It is used to define the execution algorithm of SHACL Rules.
+        Each of the two concrete syntax forms of SHACL Rules, the SHACL Rules Language (SRL)
+        and the RDF syntax (SRL/RDF), provides a way to express the abstract syntax.
+      /p>
 
       <section id="rules-defns">
         <h3>Elements of the Abstract Syntax</h3>
@@ -651,8 +659,8 @@ WHERE {
           <dt><dfn>Expression</dfn></dt>
           <dd>
             <p>
-              An [=expression=] is a function or a functional form.
-              It's arguments are [=RDF terms=].
+              An [=expression=] is a function or a functional form;
+              the arguments are [=RDF terms=].
               An expression is evaluated with respect to a [=solution mapping=] to give
               an [=RDF term=] as the result.
               <a>Expressions</a> are compatible with
@@ -660,7 +668,13 @@ WHERE {
                  >SHACL list parameter functions</a>
               and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
           </dd>
-
+          <dt><dfn >Condition</dfn></dt>
+          <dd>
+            <p>
+              A [=condition=] is an [=expression=] that evaluates to true or false.
+              [=Conditions=] are used to restrict the values of variables in pattern matching.
+            </p>
+          </dd>
           <dt><dfn data-lt="data-block">Data block</dfn></dt>
           <dd>
             A [=data block=] is a set of triples.
@@ -685,11 +699,10 @@ WHERE {
             as well as inside [=negation elements=].
           </dd>
 
-          <dt><dfn data-lt="condition element|condition">Condition element</dfn></dt>
+          <dt><dfn>Condition element</dfn></dt>
           <dd>
-            A [=condition element=] is an [=expression=]
-            that evaluates to true or false.
-            [=Condition elements=] appear in the [=body=] of a [=rule=].
+            A [=condition element=] is a [=condition=]
+            that appears as a [=rule body element=].
           </dd>
 
           <dt><dfn>Triple pattern element</dfn></dt>
@@ -701,11 +714,11 @@ WHERE {
           <dt><dfn>Negation element</dfn></dt>
           <dd>
             A [=negation element=] is a [=rule body element=]
-            comprised of a sequence of [=triple pattern elements=] and
-            [=condition elements=].
+            comprised of a sequence of [=triple patterns=] and
+            [=conditions=].
           </dd>
 
-          <dt><dfn data-lt="assignment element">Assignment element</dfn></dt>
+          <dt><dfn data-lt="assignment element|assignment">Assignment element</dfn></dt>
           <dd>
             An [=assignment element=] is a pair consisting 
             of a <a>variable</a>, called the <dfn>assignment variable</dfn>, 
@@ -726,8 +739,8 @@ WHERE {
                 >Rule body element</dfn></dt>
           <dd>
             A [=rule body element=] (often just "rule element") is any element that
-            can appear in a [=rule body=],
-            a [=triple pattern=],
+            can appear in a [=rule body=], i.e.,
+            a [=triple pattern element=],
             a [=condition element=],
             a [=negation element=],
             or an [=assignment element=].
@@ -798,7 +811,6 @@ WHERE {
             pattern can be derived from a [=base graph=] using the
             [=rule set=].
           </dd>
-
         </dl>
 
         <p>
@@ -864,92 +876,178 @@ WHERE {
 
       <section id="rule-dependency">
         <h3>Rule Dependency</h3>
+        
+        <p>
+          A rule `R1` depends on a rule `R2` if the output of the second rule
+          affects the evaluation of the body of the first rule. That is, the head of `R2`
+          has a [=triple template=] that might generate a triple that matches
+          a [=triple pattern=] in the body of `R1`, 
+          either as a [=triple pattern element=] or inside a [=negation element=].
+        </p>
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
+          A <em>closed dependency</em> requires that rule `R2`
+          has generated all its possible output before rule `R1` can be executed.
+          This happens when `R1` has a [=triple pattern=] in a [=negation element=]
+          that matches a [=triple template=] of `R2` in a [=negation element=],
+          or if `R1` has any matching [=triple pattern=] depending on `R2` 
+          and `R1` has an [=assignment element=], or `R1` has any matching [=triple pattern=]
+          that matches a [=triple template=] of `R2` that has a blank node.
+          If a rule dependency is not closed, it is an <em>open dependency</em>
+          which allows the first rule `R1` to be executed
+          while the rule `R2` might be run again to generate further triples
+          which can then cause `R1` to be reevaluated with the new triples from `R2`.
+        </p>
 
         <dl>
+          <dt><dfn>Triple pattern matching</dfn></dt>
+          <dd>
+          <p>
+            A [=triple pattern=] matches a [=triple template=] if 
+            the triple template may generate a triple that matches the triple pattern.
+          </p>
+          </dd>
+
           <dt><dfn>Triple pattern dependency</dfn></dt>
           <dd>
-            A [=triple pattern=] in the rule body of rule `R1` depends on a
-            rule `R2` if the head of `R2` can possibly generate triples that
-            match the [=triple pattern=] in `R1`.
-            The dependency is <em>positive</em> 
-            if the [=triple pattern=] is a [=rule body element=]
-            and the dependency is <em>negative</em> if the
-            [=triple pattern=] appears in a [=negation element=].
+            <p>
+              A [=triple pattern=]
+              <dfn data-local-lt="depends-pattern-template">depends on a [=triple template=]</dfn>
+              if the [=triple pattern=] could possibly match the [=triple template=].
+            </p>
+            <p>
+              A [=triple pattern=]
+              <dfn data-local-lt="depends-pattern-rule">depends on a [=rule=]</dfn>
+              if the [=triple pattern=] has dependency on any [=triple template=]
+              in the head of the rule.
+          </p>
           </dd>
-
           <dt><dfn>Rule dependency</dfn></dt>
           <dd>
-            A rule `R1` depends on rule `R2` if any rule element of `R1`
-            depends on `R2`.
-            The dependency is <em>negative</em> if any of the rule elements
-            have a negative dependency on `R2`; otherwise the dependency is
-            <em>positive</em>.
-          </dd>
-
-          <dt><dfn>Dependency graph</dfn></dt>
-          <dd>
-            A [=dependency graph=] of a [=rule set=] is a graph where the
-            vertices are rules and an edge exists where one rule R1 depends on
-            rule R2. An edge is labeled either <em>positive</em> 
-            or <em>negative</em> according to the [=rule dependency=].
-          </dd>
-
-          <dt><dfn
-                data-lt="transitive rule dependency|transitive dependency"
-                >Transitive rule dependency</dfn></dt>
-          <dd>
-            A rule `R1` has a [=transitive dependency=] on rule `R2` 
-            if there is a path in the [=dependency graph=] from `R1` to `R2`.
-          </dd>
-
-          <dt><dfn
-                data-lt="recursive rule dependency|recursive dependency"
-                >Recursive rule dependency</dfn></dt>
-          <dd>
-            A rule `R` has a [=recursive dependency=] if there is a cyclic path
-            in the [=dependency graph=] involving `R`.
+          <p>
+            Rule `R1` depends on rule `R2` if any [=triple pattern=]
+            of `R1` depends on `R2`.
+            The dependency is an [=open dependency=] if `R1` does not have an
+            [=assignment element=] and if all the [=triple patterns=]
+            of `R1` that depend `R2` occur only as [=triple pattern elements=]
+            and not as part of a [=negation element=], and if `R1` does not
+            have any [=triple template=] with a blank node.
+            Otherwise, if `R1` has an [=assignment element=],
+            or if a [=triple pattern=]
+            that might match triples generated by `R2` is present in a
+            [=negation element=], then the dependency is a [=closed dependency=].
+          </p>
           </dd>
         </dl>
 
-        <section id="dependency-graph-construction-algorithm" class="informative">
+        <div class="note">
+          <p>
+            A [=triple template=] with components `ts`, `tp`, `to`
+            can possibly generatce a triple with component RDF terms
+            `s`, `p`, `o` if
+          </p>
+          <ul>
+            <li>`ts` is a variable, or `ts` is the
+              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                 >same RDF term</a> as `s`
+            </li>
+            <li>`tp` is a variable, or `tp` is the
+              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                 >same RDF term</a> as `p`
+            </li>
+            <li>`to` is a variable, or `to` is the
+              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                 >same RDF term</a> as `o`
+            </li>
+          </ul>
+          <p>
+            In addition, if any pair of `ts`, `tp`, and `to` are the same variable,
+            then the corresponding pair of `s`, `p`, and `o` must be the same.
+          </p>
+        </div>
+
+        <section id="dependency-graph">
+          <h4>Dependency Graph</h4>
+          <p>
+            The dependencies between rules are represented as a directed graph,
+            called the [=dependency graph=].
+            The vertices of the graph are the rules of the rule set, and the edges
+            are labeled either <em>open</em> or <em>closed</em> according to
+            whether the dependency is an [=open dependency=] or a [=closed dependency=].
+          </p>
+          <dl>
+            <dt><dfn>Dependency graph</dfn></dt>
+            <dd>
+              A [=dependency graph=] of a [=rule set=] is a directed graph where each
+              vertex is a [=rule=] in the rule set, and an edge exists from rule `R1`
+              to rule `R2` if `R1` depends on `R2`.
+              The edge is labeled either <em>open</em> 
+              or <em>closed</em> according whether the dependency
+              is an [=open dependency=] or a [=closed dependency=].
+            </dd>
+            <dt><dfn
+                  data-lt="transitive rule dependency|transitive dependency"
+                  >Transitive rule dependency</dfn></dt>
+            <dd>
+              A rule `R1` has a [=transitive dependency=] on rule `R2` 
+              if there is a path in the [=dependency graph=] from `R1` to `R2`.
+            </dd>
+            <dt><dfn
+                  data-lt="recursive rule dependency|recursive dependency"
+                  >Recursive rule dependency</dfn></dt>
+            <dd>
+              <p>
+                A rule `R` has a [=recursive dependency=] if there is a cyclic path
+                in the [=dependency graph=] involving `R`.
+              </p>
+            </dd>
+          </dl>
+          <p class="note">
+            The dependency graph is not affected by the data graph.
+          </p>
+        </section>
+
+
+        <section id="dependency-graph-construction-algorithm">
           <h4>Dependency Graph Algorithm</h4>
           <p>
             The following algorithm gives one
             possible method for constructing the [=dependency graph=] from a
-            [=rule set=]. Conformance depends on the dependency definitions 
-            in this section, not on use of this procedure.
+            [=rule set=]. Conformance depends on producing a dependency graph
+            that meets the definitions of a dependency graph,
+            not on the use of this procedure.
           </p>
           <div class="algorithm">
             <pre>
 
+define mergeLabel(oldLabel, newLabel):
+    ## Closed dependency overrides open dependency.
+    if oldLabel == "open" and newLabel == "open":
+        return "open"
+    else:
+        return "closed"
+    endif
+enddefine
+
 ## output -- Dependency graph with rule vertices and labeled edges.
-
 define buildDependencyGraph(ruleSet):
+    ## edgeLabelMap maps (R1, R2) to "open" or "closed"
+    let edgeLabelMap be a map from pair (rule, rule) to label
 
-    ## edgeLabelMap maps (R1, R2) to "positive" or "negative"
-    let edgeLabelMap be a map from (rule, rule) to label
-
-    define mergeLabel(oldLabel, newLabel):
-        ## Negative dependency dominates positive dependency.
-        if oldLabel == "negative" or newLabel == "negative":
-            return "negative"
-        else:
-            return "positive"
-        endif
-        enddefine
-
-    ## Classify all triple patterns in a rule, looking inside negation elements.
     foreach rule R1 in ruleSet:
-
+        ## Classify each triple pattern TP in the rule as requiring "open" or "closed"
+        ## depending on whether it is in a negation element or not.
         let bodyDependencies = {}
         foreach rule body element RBE in the body of R1:
             if RBE is a negation element:
-                foreach triple pattern element TPE in RBE:
-                    let item be a pair (TPE, "negative")
+                foreach triple pattern TP in RBE:
+                    let item be a pair (TP, "closed")
                     add item to bodyDependencies
                 endfor
-            else if RBE is a triple pattern element:
-                let item be a pair (RBE, "postive")
+            else if RBE is a triple pattern element of triple pattern TP:
+                let item be a pair (TP, "open")
                 add item to bodyDependencies
             else if RBE is a condition element:
                 ## Do nothing
@@ -958,14 +1056,19 @@ define buildDependencyGraph(ruleSet):
             endif
         endfor
 
-        foreach pair ( triple pattern TP, depLabel) in bodyDependencies:
-
+        foreach pair (triple pattern TP, depLabel) in bodyDependencies:
+            if R1 has an assignment element:
+              set depLabel to "closed"
+            endif
+            if R1 has a triple template with a blank node:
+              set depLabel to "closed"
+            endif
+            ## Find depenencies for this triple pattern element or negation element.
             foreach rule R2 in ruleSet:
-
                 foreach triple template TT in head of R2:
                     ## "possibly generate" / matching is defined in
-                    ## 3.3 Rule Dependency.
-                    if TT can possibly generate a triple pattern TP:
+                    ## section 3.3 
+                    if TT can possibly match triple pattern TP:
                         let key = (R1, R2)
                         if edgeLabelMap contains key:
                             let oldLabel = edgeLabelMap.get(key)
@@ -976,15 +1079,13 @@ define buildDependencyGraph(ruleSet):
                         endif
                     endif
                 endfor
-
             endfor
-
         endfor
     endfor
 
     let DP = { }
     foreach entry ((R1, R2), label) in edgeLabelMap:
-        DP = { edge (R1 -&gt; R2) labeled label } &cups; DP
+        add edge (R1 -&gt; R2) labeled label to DP
     endfor
 
     the result is DP
@@ -992,37 +1093,6 @@ define buildDependencyGraph(ruleSet):
             </pre>
           </div>
         </section>
-
-        <p>Notes:</p>
-        <ul>
-          <li>
-            <p>
-              A [=triple template=] with components `ts`, `tp`, `to`
-              can possibly generate a triple with component RDF terms
-              `s`, `p`, `o` if 
-              `ts` is a variable or `ts` is the
-              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
-                 >same RDF term</a> as `s`,
-              `tp` is a variable or `tp` is the
-              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
-                 >same RDF term</a> as `p`,
-              and
-              `to` is a variable or `to` is the
-              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
-                 >same RDF term</a> as `o`.
-            </p>
-            <p>
-              In addition, if any pair of `ts`, `tp`, and `to` are the same variable,
-              then the corresponding pair of `s`, `p`, and `o` must be the same.
-            </p>
-            <p class="ednote">
-              Revise
-            </p>
-          </li>
-          <li>
-            The dependency graph is not affected by the data graph.
-          </li>
-        </ul>
 
         <p>
           Examples:
@@ -1044,17 +1114,18 @@ define buildDependencyGraph(ruleSet):
         <p>
           [=Stratification=] is the process of partitioning a [=rule set=]
           into an ordered sequence of [=stratification layers=] (also known
-          as "strata", singular "stratum"), forming a [=stratification=].
-          Rules in lower [=strata=] are evaluated before rules in higher
-          [=strata=].
+          as "strata", singular "stratum"). Rules in lower [=strata=] are
+          evaluated before rules in higher [=strata=].
         </p>
         <p>
           [=Stratification=] imposes constraints on dependencies between [=rules=]
-          to ensure that [=negation elements=] depend only on results computed
-          using earlier (lower) [=strata=]. This guarantees a single well-defined
+          to ensure that [=negation elements=], [=assignment elements=], and
+          blank nodes created in a [=rule head=] depend only on results computed
+          using earlier (lower) [=strata=] and the [=base graph=].
+          This guarantees a single, well-defined, and finite
           outcome from the evaluation of a [=rule set=] over a given [=base graph=].
         </p>
-        <div class="ednote">
+        <div class="note">
           <p>
             A stratification process may also be used to make other evaluation
             decisions. This document describes the necessary conditions for
@@ -1066,18 +1137,26 @@ define buildDependencyGraph(ruleSet):
         </div>
 
         <dl>
-          <dt><dfn>Stratification</dfn></dt>
-          <dd>
-            A [=stratification=] of a [=rule set=] is a sequence of sets of
-            rules. Each rule in a [=rule set=] appears in exactly one
-            [=stratification layer=].
-          </dd>
           <dt><dfn 
                 data-lt="stratification layer|layer|stratum"
                 >Stratification layer</dfn></dt>
           <dd>
-            Each set of rules in a [=stratification=] is called a 
-            [=stratification layer=], also called a "stratum".
+            <p>
+              A [=stratification layer=] `SL`, is a pair of disjoint sets of
+              rules (`SL.once`, `SL.general`) .
+              `SL.once` contains [=run-once rules=], which are
+              rules that use [=assignment elements=] or produce
+              blank nodes in the [=rule head=]; these rules are each evaluated exactly
+              once at the start of evaluation of the [=stratification layer=].
+              `SL.general` contains the remaining rules, which are evaluated 
+              until no new triples are inferred.
+            </p>
+          </dd>
+          <dt><dfn>Stratification</dfn></dt>
+          <dd>
+            A [=stratification=] of a [=rule set=] is a sequence of [=stratification layers=].
+            Each rule in a [=rule set=] appears in exactly one of the sets of one of
+            the [=stratification layers=].
           </dd>
         </dl>
 
@@ -1086,20 +1165,21 @@ define buildDependencyGraph(ruleSet):
           <p>
             [=Stratification=] is only defined when the following condition is
             satisfied. If a [=rule set=] does not meet this condition, then this
-            specification does not define the outcome of [=rule set=] evaluation.
+            specification does not define an outcome for the evaluation of
+            such a [=rule set=].
           </p>
           <dl>
             <dt><dfn>Stratification Condition</dfn></dt>
             <dd>
               The [=stratification condition=] requires that
               there is no [=recursive dependency=] involving a
-              <em>negative</em> dependency
+              [=closed rule dependency=]
               in the [=dependency graph=] for a [=rule set=].
             </dd>
           </dl>
           <p>
-            In other words, there is no `NOT` used in any rule that
-            transitively depends on itself.
+            In other words, there is no `NOT` or run-once rule (assignment or rule [=triple template=]
+            involving a blank node) in any transitive dependency cycle of the [=dependency graph=].
           </p>
         </section>
 
@@ -1120,8 +1200,9 @@ define stratification(ruleSet):
     let stratumMap be a map from rule to integer
 
     ## The dependency graph should satisfy the stratification condition.
-    ## Check unbounded work due to a violation of the stratification condition.
-    let limit = num rules  + 1
+    ## The check for unbounded stratification is a guard 
+    ## due to a violation of the stratification condition.
+    let limit = num rules + 1
     let maxStratum = 0
 
     ## initialize stratumMap
@@ -1138,13 +1219,13 @@ define stratification(ruleSet):
             let qRule = destination of the edge
             let label = edge label
 
-            if label == "positive" :
+            if label == "open" :
                 if stratumMap.get(pRule) < stratumMap.get(qRule) :
                     stratumMap.set(pRule, stratumMap.get(qRule))
                     changed = true;
-                    endif
                 endif
-            if label == "negative" :
+            endif
+            if label == "closed" :
                 if stratumMap.get(pRule) <= stratumMap.get(qRule) :
                     let xStratum = 1 + stratumMap.get(qRule)
                     if ( xStratum > limit )
@@ -1154,38 +1235,42 @@ define stratification(ruleSet):
                     stratumMap.set(pRule, xStratum)
                     maxStratum = max(maxStratum, xStratum)
                     changed = true;
-                    endif
                 endif
-            endfor
-        endwhile
+            endif
+        endfor
+    endwhile
 
     ## Initialize the result map.
-    let stratumLevels be a map from integer to a set of rules
+    let stratumRules be a map from integer to rules.
     for i = 0 to maxStratum
-        stratumLevels.set(i, {})
-        endfor
+        stratumRules.set(i, {})
+    endfor
 
     ## Gather rules in stratumMap with the same level number
     for rule R in map stratumMap:
         let stratumNum = stratumMap.get(R)
-        let S = {R} &cups; stratumLevel.get(stratumNum)
-        stratumLevel.set(stratumNum, S)
-        endfor
+        add R to stratumRules.get(stratumNum)
+    endfor
+
+    ## Partition each level into once and general
+    let stratumLevels be a sequence of pairs of sets of rules.
+    for i = 0 to maxStratum:
+        let rules = stratumLevels.get(i)
+        let once = { R in rules | R is a run-once rule }
+        let general = rules \ once
+        stratumLevels.set(i, pair(once, general))
+    endfor
 
     the result is stratumLevels
-    endfunction
+enddefine
             </pre>
           </div>
-
           <p class="note">
-            A consequence of the [=stratification condition=] is that when 
-            a rule containing a [=negation element=] is
-            <a href="#eval-rule">evaluated</a>, the data used to determine the
-            outcome of that [=negation element=], whether in the
-            [=base graph=] or the [=inference graph=], is fixed and will not
-            change during evaluation.
+            A consequence of the [=stratification condition=] is that once a
+            [=run-once rule=] is <a href="#eval-rule">evaluated</a>,
+            the data used to determine the outcome of the rule will not
+            change during further evaluation.
           </p>
-
         </section>
       </section>
     </section>
@@ -1206,7 +1291,7 @@ define stratification(ruleSet):
       <p>Shape Rules Language:</p>
       <div class="example-rules">
 <pre>
-PREFIX : &lt;http://example/>
+PREFIX : &lt;http://example/&gt;
 
 DATA { :x :p 1 ; :q 2 . }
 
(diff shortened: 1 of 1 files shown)


robert-david:
This is the correct URI for the githack render:
https://raw.githack.com/w3c/data-shapes/rules-run-once/shacl12-rules/index.html

shacl12-rules/index.html
```diff
@@ -833,11 +833,13 @@ <h3>Stratification</h3>
         </p>
         <p>
           [=Stratification=] imposes constraints on dependencies between [=rules=]
-          to ensure that [=negation elements=] depend only on results computed
-          using earlier (lower) [=strata=]. This guarantees a single well-defined
+          to ensure that [=negation elements=], [=assignment elements=], and
+          blank nodes created in a [=rule head=] depend only on results computed
+          using earlier (lower) [=strata=] and the [=base graph=].
+          This guarantees a single, well-defined,
```
```suggestion
          This guarantees a single, well-defined, and finite
```

shacl12-rules/index.html
```diff
@@ -849,18 +851,25 @@ <h3>Stratification</h3>
...
           <dd>
-            Each set of rules in a [=stratification=] is called a 
-            [=stratification layer=], also called a "stratum".
+            A [=stratification layer=], also called a "stratum",
+            is a pair of sets of rules.
+            The pair consists of a set of rules that are executed once, for assignment
+            and blank nodes created in a [=rule head=], and a set of rules
+            that are evaluated, possibly involving recursion, until there are
+            no more inferred triples.
+            A stratum pair, <var>P</var> has components <var>P.once</var>
```
```suggestion
            A stratum pair <var>P</var> has components <var>P.once</var>
```

robert-david (approved):

shacl12-rules/index.html
```diff
@@ -1276,8 +1285,10 @@ <h4>Process Imports</h4>
           <div class="def">
             <p>
             A [=rule set=] has three components: R.rules, R.data and R.imports.
```
TallTed: This might be wrong markup. It's intended to ensure that *some* markup is applied to differentiate special text from the body text.
```suggestion
            A [=rule set=] has three components: `R.rules`, `R.data`, and `R.imports`.
```

shacl12-rules/index.html
```diff
@@ -1276,8 +1285,10 @@ <h4>Process Imports</h4>
...
-            The <dfn>rule set merge</dfn> of two rule sets RS1 and RS2 giving
-            rule set MR:
+
+            The <dfn>rule set merge</dfn> of two rule sets, RS1 and RS2,
+            is a rule set MR defined as follows:
```
```suggestion
            The <dfn>rule set merge</dfn> of two rule sets, `RS1` and `RS2`,
            is a rule set `MR` defined as follows:
```

shacl12-rules/index.html
afs: > This might be wrong markup. It's intended to ensure that _some_ markup is applied to differentiate special text from the body text.

Using `THING.field` needs to be more consistently used and also CSS styling is inconsistent; I've been delaying fixing that while content is being written. `THING.field` is going to be needed more widely.

shacl12-rules/index.html
afs: Both done.

shacl12-rules/index.html
```diff
@@ -849,18 +851,25 @@ <h3>Stratification</h3>
...
+            no more inferred triples.
+            A stratum pair <var>P</var> has components <var>P.once</var>
+            and <var>P.general</var>.
+            <div class="ednote">create a notation section?</div>
+          </dd>
+          <dt><dfn>Stratification</dfn></dt>
+          <dd>
+            A [=stratification=] of a [=rule set=] is a sequence of [=stratification layers=].
+            Each rule in a [=rule set=] appears in exactly one of the sets of one of
+            the [=stratification layers=].
```
simonstey: this def says a "stratum" it is "a pair of sets of rules": one for rules "executed once" (for assignments and blank nodes in head) and one for general rules. However, there is no normative definition anywhere in the spec that specifies **how** a rule is classified into `.once` or `.general`. 

it's somewhat hinted at via: "for assignment and blank nodes created in a [=rule head=]" but I think we should add a precise normative statement such as: 
```
A rule belongs to `P.once` if it contains an [=assignment element=] in its body or a blank node in its [=rule head=]; otherwise it belongs to `P.general`.
```

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
     let finished = false
     while !finished:
         finished = true
-        for each rule in layer:
-            let X = eval(rule, GE)
+        for each rule R in ST.general:
+            let X = eval(R, GE)
             let Y = { t &isin; X | t &notin; in GE }
             if Y is not empty:
                 finished = false
```
simonstey: [The original strat algo returns only flat sets of rules](https://github.com/w3c/data-shapes/pull/865/changes#diff-4cb699e56152bee095a8650906565f110dd7ed444551f33a7b1987a0a92fd214R909-R968) while this eval algo expects each stratum ST to have .once and .general components.

easy fix: require the strat algo to return `Map: Integer -> Pair(once: Set of rules, general: Set of rules),`

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
+let LS be the sequence of layers after stratification
 
 # Inference graph
 let GI = { t &isin; D  | t &notin; in G0 }
 
 # Evaluation graph.
 let GE = G0 &cups; D
 
-for each layer in L:
+for each stratum ST in LS</pre>:
```
```suggestion
for each stratum ST in LS:
```

shacl12-rules/index.html
```diff
@@ -833,11 +833,13 @@ <h3>Stratification</h3>
         </p>
         <p>
           [=Stratification=] imposes constraints on dependencies between [=rules=]
-          to ensure that [=negation elements=] depend only on results computed
-          using earlier (lower) [=strata=]. This guarantees a single well-defined
+          to ensure that [=negation elements=], [=assignment elements=], and
+          blank nodes created in a [=rule head=] depend only on results computed
+          using earlier (lower) [=strata=] and the [=base graph=].
+          This guarantees a single, well-defined, and finite
```
simonstey, line 1125: now that this paragraph got expanded, shouldn't the strat condition section be expanded accordingly as well?

https://github.com/w3c/data-shapes/pull/865/changes#diff-4cb699e56152bee095a8650906565f110dd7ed444551f33a7b1987a0a92fd214R876-R896:

```
   <section id="stratification-condition">
          <h4>Stratification Condition</h4>
          <p>
            [=Stratification=] is only defined when the following condition is
            satisfied. If a [=rule set=] does not meet this condition, then this
            specification does not define the outcome of [=rule set=] evaluation.
          </p>
          <dl>
            <dt><dfn>Stratification Condition</dfn></dt>
            <dd>
              The [=stratification condition=] requires that
              there is no [=recursive dependency=] involving a
              <em>negative</em> dependency
              in the [=dependency graph=] for a [=rule set=].
            </dd>
          </dl>
          <p>
            In other words, there is no `NOT` used in any rule that
            transitively depends on itself.
          </p>
        </section>
```
this still only forbids negative cycles, but says nothing about assignment elements or blank-node-generating rules. There is no condition that requires a rule with assignments or blank nodes in its head to be in a lower stratum than rules it depends on (or that such rules cannot be depending on itsedlf).

bug or feature?

shacl12-rules/index.html
simonstey, line 1254: maybe add:

```
## Partition each level into once and general
for i = 0 to maxStratum:
    let rules = stratumLevels.get(i)
    let once = { R in rules | R is a run-once rule }
    let general = rules \ once
    stratumLevels.set(i, pair(once, general))
endfor
```

```suggestion
let S = {R} &cups; stratumLevels.get(stratumNum)
stratumLevels.set(stratumNum, S)
```

shacl12-rules/index.html
```diff
@@ -849,18 +851,25 @@ <h3>Stratification</h3>
...
                 data-lt="stratification layer|layer|stratum"
                 >Stratification layer</dfn></dt>
           <dd>
-            Each set of rules in a [=stratification=] is called a 
-            [=stratification layer=], also called a "stratum".
+            A [=stratification layer=], also called a "stratum",
+            is a pair of sets of rules.
+            The pair consists of a set of rules that are executed once, for assignment
+            and blank nodes created in a [=rule head=], and a set of rules
+            that are evaluated, possibly involving recursion, until there are
```
simonstey: alternatively also a slightly rephrased version:

```
A [=stratification layer=] is a pair (`P.once`, `P.general`) of disjoint sets of rules. `P.once` contains rules that use [=assignment elements=] or produce blank nodes in the [=rule head=]; these rules are evaluated exactly once. `P.general` contains the remaining rules, which are evaluated iteratively until no new triples are inferred.
```

simonstey: alternatively also a slightly rephrased version:

```
A [=stratification layer=] is a pair (`P.once`, `P.general`) of disjoint sets of rules. `P.once` contains rules that use [=assignment elements=] or produce blank nodes in the [=rule head=]; these rules are evaluated exactly once. `P.general` contains the remaining rules, which are evaluated iteratively until no new triples are inferred.
```

shacl12-rules/index.html
```suggestion
                let item be a pair (RBE, "positive")
```

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
 
 Apply stratification to RS
 
-let L be the sequence of layers after stratification
+let LS be the sequence of layers after stratification
 
 # Inference graph
 let GI = { t &isin; D  | t &notin; in G0 }
```
```suggestion
let GI = { t &isin; D  | t &notin; G0 }
```

`∉ in` doesnt make sense

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
+        endfor
+
     let finished = false
     while !finished:
         finished = true
-        for each rule in layer:
-            let X = eval(rule, GE)
+        for each rule R in ST.general:
+            let X = eval(R, GE)
             let Y = { t &isin; X | t &notin; in GE }
```
```suggestion
            let Y = { t &isin; X | t &notin; GE }
```

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
 let GI = { t &isin; D  | t &notin; in G0 }
 
 # Evaluation graph.
 let GE = G0 &cups; D
 
-for each layer in L:
+for each stratum ST in LS</pre>:
+    for each rule R in ST.once:
+        let X = eval(R, GE)
+        let Y = { t &isin; X | t &notin; in GE }
```
```suggestion
        let Y = { t &isin; X | t &notin; GE }
```

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
 
 # Inference graph
 let GI = { t &isin; D  | t &notin; in G0 }
 
 # Evaluation graph.
 let GE = G0 &cups; D
 
-for each layer in L:
+for each stratum ST in LS</pre>:
+    for each rule R in ST.once:
```
simonstey, line 1873: what if two `.once` rules depend on each other within the same stratum? then the order of evaluation matters (since each is evaluated exactly once).but we do not define whether the order is deterministic or whether this is an error (iirc)

shacl12-rules/index.html
```diff
@@ -1444,20 +1455,27 @@ <h3>Evaluation of a Rule Set</h3>
...
+        GI = Y &cups; GI
+        GE = Y &cups; GE
+        endfor
+
     let finished = false
     while !finished:
         finished = true
-        for each rule in layer:
-            let X = eval(rule, GE)
+        for each rule R in ST.general:
```
simonstey, line 1883: once rules are evaluated before general rules, but what if a once rule depends on the output of a general one in the same stratum? then it won't see that output.. bug or feature?

simonstey (changes requested):
how would following corner cases be classified?

  - A rule with an assignment element AND a negation element -> `.once` or `.general`?
  - A rule with blank nodes in the head AND a recursive positive dependency -> `.once` or `.general`? If `.once`, the recursion is lost; if `.general`, blank nodes generate infinitely.
  - A rule with no assignment elements and no blank nodes in head -> presumably `.general`, but not stated.
  - A rule with `BNODE()` in an assignment expression but not in the head -> `.once` or `.general`?

shacl12-rules/index.html
afs: > but what if a once rule depends on the output of a general one in the same stratum?

It can't depend on the same stratum. This needs to be part of the stratification algorithm.

shacl12-rules/index.html
afs: > what if two .once rules depend on each other within the same stratum? 

A run-once rule can run when it's dependencies from lower strata are ready. Two run-once rules can't depend on each other. Order within `.once` therefore does not matter.

shacl12-rules/index.html
afs: See comment on updating the stratification algorithm.

shacl12-rules/index.html
afs: > now that this paragraph got expanded, shouldn't the strat condition section be expanded accordingly as well?

Yes.

shacl12-rules/index.html
afs: > once rules are evaluated before general rules, but what if a once rule depends on the output of a general one in the same stratum? then it won't see that output.. bug or feature?

Same as point above.

afs:
> how would following corner cases be classified?
> 
> * A rule with an assignment element AND a negation element -> `.once` or `.general`?

"once" - that takes precedence.

> * A rule with blank nodes in the head AND a recursive positive dependency -> `.once` or `.general`? If `.once`, the recursion is lost; if `.general`, blank nodes generate infinitely.

Reject.

> * A rule with no assignment elements and no blank nodes in head -> presumably `.general`, but not stated.

Will update the stratification algorithm

> * A rule with `BNODE()` in an assignment expression but not in the head -> `.once` or `.general`?

Once.

shacl12-rules/index.html
recalcitrantsupplant, line 1179:
```suggestion
            </dd>
            <dd>
              In addition, every rule in `P.once` depends only on rules in lower
              [=stratification layers=].
            </dd>
          </dl>
```

recalcitrantsupplant (approved):
one suggested wording for updating the stratification algorithm

afs:
There is a major revision of dependency calculation and stratification.

This includes renaming "positive" dependencies as "open" dependencies, and "negative" dependencies as "closed" dependencies because the this relationship is more than just use of `NOT`.

shacl12-rules/index.html
```diff
@@ -635,8 +638,13 @@ <h3>Rule Tuples</h3>
 
     <section id="rules-abstract-syntax">
       <h2>Shape Rules Abstract Syntax</h2>
-
-      <p>The Shape Rules Abstract Syntax</p>
+      
+      <p>
+        The Shape Rules Abstract Syntax is the logical structure of SHACL Rules.
+        It is used to define the execution algorithm and of SHACL Rules.
```
```suggestion
        It is used to define the execution algorithm of SHACL Rules.
```

shacl12-rules/index.html
```diff
@@ -652,15 +660,21 @@ <h3>Elements of the Abstract Syntax</h3>
...
               <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
                  >SHACL list parameter functions</a>
               and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
           </dd>
-
+          <dt><dfn >Condition</dfn></dt>
+          <dd>
+            <p>
+              An [=condition=] is a [=expression=] that evaluates to true or false.
+              [=Conditions=] are used to restrict the values of varaibles in pattern matching.
```
```suggestion
              [=Conditions=] are used to restrict the values of variables in pattern matching.
```

shacl12-rules/index.html
```diff
@@ -685,11 +699,10 @@ <h3>Elements of the Abstract Syntax</h3>
             as well as inside [=negation elements=].
           </dd>
 
-          <dt><dfn data-lt="condition element|condition">Condition element</dfn></dt>
+          <dt><dfn>Condition element</dfn></dt>
           <dd>
-            A [=condition element=] is an [=expression=]
-            that evaluates to true or false.
-            [=Condition elements=] appear in the [=body=] of a [=rule=].
+            A [=condition element=] is an [=condition=]
```
```suggestion
            A [=condition element=] is a [=condition=]
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
 
       <section id="rule-dependency">
         <h3>Rule Dependency</h3>
+        
+        <p>
+          A rule `R1` depends on a rule `R2` if the output of the second rule
+          affects the evaluation of the body of the first rule. That is, the head of `R2`
+          has a [=triple template=] that might generate a truiple that matches
```
```suggestion
          has a [=triple template=] that might generate a triple that matches
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+          a [=triple pattern=] in the body of `R1`, 
+          either as a [=triple pattern element=] or inside a [=negation element=].
+        </p>
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
+          A <em>closed dependency</em> requires that rule `R2`
+          has generated all its possible output before rule `R1` can be executed.
+          This is happens when `R1` has a [=triple pattern=] in a [=negation element=]
```
```suggestion
          This happens when `R1` has a [=triple pattern=] in a [=negation element=]
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+          that matches a [=triple template=] of `R2` in a [=negation element=],
+          or if `R1` has any matching [=triple pattern=] depending on `R2` 
+          and `R1` has an [=assignment element=], or `R1` has any matching [=triple pattern=]
+          matches a [=triple template=] of `R2` that has a blank node.
```
```suggestion
          that matches a [=triple template=] of `R2` that has a blank node.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
           <dd>
-            A [=triple pattern=] in the rule body of rule `R1` depends on a
-            rule `R2` if the head of `R2` can possibly generate triples that
-            match the [=triple pattern=] in `R1`.
-            The dependency is <em>positive</em> 
-            if the [=triple pattern=] is a [=rule body element=]
-            and the dependency is <em>negative</em> if the
-            [=triple pattern=] appears in a [=negation element=].
+          <p>
+            A [=triple pattern=] matches a [=triple template=] if there
```
```suggestion
            A [=triple pattern=] matches a [=triple template=] if 
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+          <p>
+            In addition, if any pair of `ts`, `tp`, and `to` are the same variable,
+            then the corresponding pair of `s`, `p`, and `o` must be the same.
+          </p>
+        </div>
+
+        <section id="dependency-graph">
+          <h4>Dependency Graph</h4>
+          <p>
+            The dependencies between rules is represented as a directed graph,
```
```suggestion
            The dependencies between rules are represented as a directed graph,
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+            called the [=dependency graph=].
+            The vertices of the graph are the rules of the rule set, and the edges
+            are labeled either <em>open</em> or <em>closed</em> according to
+            whether the dependency is an [=open dependency=] or a [=closed dependency=].
+          </p>
+          <dl>
+            <dt><dfn>Dependency graph</dfn></dt>
+            <dd>
+              A [=dependency graph=] of a [=rule set=] is a directed graph where each
+              vertex are is a [=rule=] in the rule set, and an edge exists from rule `R1`
```
```suggestion
              vertex is a [=rule=] in the rule set, and an edge exists from rule `R1`
```

shacl12-rules/index.html
```diff
@@ -1160,32 +1236,36 @@ <h4>Stratification Algorithm</h4>
...
           </div>
-
           <p class="note">
-            A consequence of the [=stratification condition=] is that when 
-            a rule containing a [=negation element=] is
-            <a href="#eval-rule">evaluated</a>, the data used to determine the
-            outcome of that [=negation element=], whether in the
-            [=base graph=] or the [=inference graph=], is fixed and will not
-            change during evaluation.
+            A consequence of the [=stratification condition=] is that a
```
```suggestion
            A consequence of the [=stratification condition=] is that once a
```

robert-david (approved):

shacl12-rules/index.html
```diff
@@ -531,11 +531,14 @@ <h3>Assignment and Creating RDF Terms</h3>
         <p>
           Rules involving assignment are [=run-once rules=]. Rules with assignment
           are run after all the rules that could produce the data that they depend on
-          and before any rules that depend on the data they produce.
+          and before any rules that depend on the data they produce. 
+          Rules that involve blank node in the [=rule head=] are also creating new
```
TallTed, line 535: Either this —
```suggestion
          Rules that involve a blank node in the [=rule head=] are also creating new
```

TallTed: — or this (using plural), but I prefer the suggestion above (using singular) —
```suggestion
          Rules that involve blank nodes in the [=rule head=] are also creating new
```

shacl12-rules/index.html
```diff
@@ -531,11 +531,14 @@ <h3>Assignment and Creating RDF Terms</h3>
...
+          RDF terms and are run-once rules.
         </p>
         <p>
-          This condition ensures that rules with assignment do not generate new RDF terms
-          multiple times, with potentially different outcomes.
+          This condition ensures that rules that create RDF terms do not generate new terms
+          multiple times, with potentially different outcomes and also that such rules do not 
+          loop back to themselves and cause an unbounded number of RDF terms.
```
TallTed, line 541:
```suggestion
          multiple times, with potentially different outcomes, and also that such rules do not 
          loop back to themselves, causing creation of an unbounded number of RDF terms.
```

shacl12-rules/index.html
```diff
@@ -635,8 +638,13 @@ <h3>Rule Tuples</h3>
...
     <section id="rules-abstract-syntax">
       <h2>Shape Rules Abstract Syntax</h2>
-
-      <p>The Shape Rules Abstract Syntax</p>
+      
+      <p>
+        The Shape Rules Abstract Syntax is the logical structure of SHACL Rules.
+        It is used to define the execution algorithm and of SHACL Rules.
+        Each of the 2 concrete syntax forms of SHACL Rules, the SHACL Rules Language (SRL)
+        and the RDF syntax (SRL/RDF) provide ways to express the abstract syntax.
```
```suggestion
        Each of the two concrete syntax forms of SHACL Rules — the SHACL Rules Language (SRL)
        and the RDF syntax (SRL/RDF) — provide ways to express the abstract syntax.
```

shacl12-rules/index.html
```diff
@@ -652,15 +660,21 @@ <h3>Elements of the Abstract Syntax</h3>
...
             <p>
               An [=expression=] is a function or a functional form.
-              It's arguments are [=RDF terms=].
+              The arguments are [=RDF terms=].
               An expression is evaluated with respect to a [=solution mapping=] to give
               an [=RDF term=] as the result.
               <a>Expressions</a> are compatible with
               <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
                  >SHACL list parameter functions</a>
               and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
```
TallTed: Line breaks that are used to increase clarity in the ReSpec source are often best also reflected in the HTML rendering, i.e., where line breaks are forced here, they should often be memorialized with `<br>`, `<p>`, `<li>`, or similar. _Short_ (e.g., rendered as one line) paragraphs, list items, etc., should not automatically be treated as a problem. On the other hand, _long_ (e.g., typically those rendered as 10 lines or longer) items _should_ generally be broken into smaller, more easily digested items.

shacl12-rules/index.html
```diff
@@ -652,15 +660,21 @@ <h3>Elements of the Abstract Syntax</h3>
           <dd>
             <p>
               An [=expression=] is a function or a functional form.
-              It's arguments are [=RDF terms=].
+              The arguments are [=RDF terms=].
```
TallTed: NOTE: GitHub does not like a suggestion that includes a line that was already deleted by the PR. To fully apply this suggestion, you'll need to manually drop the line shown here as 664.
```suggestion
              An [=expression=] is a function or a functional form; any arguments
              are [=RDF terms=].
```

shacl12-rules/index.html
```diff
@@ -727,7 +740,7 @@ <h3>Elements of the Abstract Syntax</h3>
           <dd>
             A [=rule body element=] (often just "rule element") is any element that
             can appear in a [=rule body=],
```
```suggestion
            can appear in a [=rule body=], i.e.,
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+        
+        <p>
+          A rule `R1` depends on a rule `R2` if the output of the second rule
+          affects the evaluation of the body of the first rule. That is, the head of `R2`
+          has a [=triple template=] that might generate a truiple that matches
+          a [=triple pattern=] in the body of `R1`, 
+          either as a [=triple pattern element=] or inside a [=negation element=].
+        </p>
+        <p>
+          There are two kinds of dependencies:
```
TallTed, line 888:
```suggestion
          There are two kinds of dependency:
```

shacl12-rules/index.html
afs: It's 3 lines. It's the `data-cite` attribute to `<a ...> ` that make it look longer.

Long `<a data-cite=` are best done on one line or split at an HTML natural place to avoid long lines in the source.

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+        </p>
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
+          A <em>closed dependency</em> requires that rule `R2`
+          has generated all its possible output before rule `R1` can be executed.
+          This is happens when `R1` has a [=triple pattern=] in a [=negation element=]
+          that matches a [=triple template=] of `R2` in a [=negation element=],
+          or if `R1` has any matching [=triple pattern=] depending on `R2`
```
TallTed, line 895:
```suggestion
          or if `R1` has any matching [=triple pattern=] that depends on `R2` 
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
-            <em>positive</em>.
+            <p>
+              A [=triple pattern=]
+              <dfn data-local-lt="depends-pattern-template">depends on a [=triple template=]</dfn>
+              if the [=triple pattern=] could possibly match the [=triple template=].
+            </p>
+            <p>
+              A [=triple pattern=]
+              <dfn data-local-lt="depends-pattern-rule">depends on a [=rule=]</dfn>
+              if the [=triple pattern=] has dependency on any [=triple template=]
```
TallTed, line 923:
```suggestion
              if the [=triple pattern=] depends on any [=triple template=]
```

shacl12-rules/index.html
```diff
@@ -635,8 +638,13 @@ <h3>Rule Tuples</h3>
...
     <section id="rules-abstract-syntax">
       <h2>Shape Rules Abstract Syntax</h2>
-
-      <p>The Shape Rules Abstract Syntax</p>
+      
+      <p>
+        The Shape Rules Abstract Syntax is the logical structure of SHACL Rules.
+        It is used to define the execution algorithm of SHACL Rules.
+        Each of the 2 concrete syntax forms of SHACL Rules, the SHACL Rules Language (SRL)
+        and the RDF syntax (SRL/RDF) provide ways to express the abstract syntax.
```
```suggestion
        and the RDF syntax (SRL/RDF) provides a way to express the abstract syntax.
```

shacl12-rules/index.html
```diff
@@ -635,8 +638,13 @@ <h3>Rule Tuples</h3>
...
+      /p>
```
recalcitrantsupplant, line 647:
```suggestion
      </p>
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+          either as a [=triple pattern element=] or inside a [=negation element=].
+        </p>
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
+          A <em>closed dependency</em> requires that rule `R2`
+          has generated all its possible output before rule `R1` can be executed.
+          This happens when `R1` has a [=triple pattern=] in a [=negation element=]
+          that matches a [=triple template=] of `R2` in a [=negation element=],
```
recalcitrantsupplant, line 894:
```suggestion
          that matches a [=triple template=] of `R2`,
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+          or if `R1` has any matching [=triple pattern=] depending on `R2` 
+          and `R1` has an [=assignment element=], or `R1` has any matching [=triple pattern=]
+          that matches a [=triple template=] of `R2` that has a blank node.
+          If a rule dependency is not closed, it is an <em>open dependency</em>
+          which allows the first rule to be executed
+          while the second rule may still be run again to generate further triples
+          that, in turn, which cause `R1` to be reevaluated with the new triples from `R2`.
```
```suggestion
          while the second rule may still be run again to generate further triples,
          which may then cause `R1` to be reevaluated with the new triples from `R2`.
```

shacl12-rules/index.html
recalcitrantsupplant, line 894:
```suggestion
          that matches a [=triple template=] of `R2`,
```

recalcitrantsupplant (approved):

shacl12-rules/index.html
recalcitrantsupplant: duplicated comment/suggestion

shacl12-rules/index.html
TallTed: Fair enough. Most previews (and even editor's drafts) don't currently render properly for me, for as yet unknown reasons, so I did guess (incorrectly) at how long this would render.

shacl12-rules/index.html
```diff
@@ -652,15 +660,21 @@ <h3>Elements of the Abstract Syntax</h3>
...
               <a>Expressions</a> are compatible with
               <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
                  >SHACL list parameter functions</a>
               and with <a data-cite="sparql12-query#expressions">SPARQL expressions</a>.</p>
           </dd>
-
+          <dt><dfn >Condition</dfn></dt>
+          <dd>
+            <p>
+              An [=condition=] is a [=expression=] that evaluates to true or false.
```
```suggestion
              A [=condition=] is an [=expression=] that evaluates to true or false.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
 
       <section id="rule-dependency">
         <h3>Rule Dependency</h3>
+        
+        <p>
+          A rule `R1` depends on a rule `R2` if the output of the second rule
+          affects the evaluation of the body of the first rule. That is, the head of `R2`
```
TallTed, line 882: Is there a reason to talk about these rules relatively, rather than by name?
```suggestion
          A rule `R1` depends on a rule `R2` if the output of `R2`
          affects the evaluation of the body of `R1`. That is, the head of `R2`
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
+          A <em>closed dependency</em> requires that rule `R2`
+          has generated all its possible output before rule `R1` can be executed.
+          This happens when `R1` has a [=triple pattern=] in a [=negation element=]
+          that matches a [=triple template=] of `R2` in a [=negation element=],
+          or if `R1` has any matching [=triple pattern=] depending on `R2` 
+          and `R1` has an [=assignment element=], or `R1` has any matching [=triple pattern=]
```
TallTed, line 896: Is this ```and `R1` has``` part of a complex condition? I am concerned that the list of  (I think, 3?) conditions will not be clear to readers. I suggest that these (starting after `This happens when` on line 893) be refactored into list items, rather than keeping them in streaming prose.
```suggestion
          and `R1` has an [=assignment element=], or `R1` has any matching [=triple pattern=]
```

shacl12-rules/index.html
TallTed: A few tweaks to @recalcitrantsupplant's suggestion above. Worthy of note, `can` instead of `may` which is too often read as RFC2119's `MAY` even though lowercase.

Also, I think we should be consistent in how we identify these rules — and I suggest `R1` and `R2` are better than `the first rule`, `the second rule`, etc.
```suggestion
          while `R2` can be run again to generate further triples,
          which can then cause `R1` to be reevaluated with the new triples from `R2`.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
-            A [=triple pattern=] in the rule body of rule `R1` depends on a
-            rule `R2` if the head of `R2` can possibly generate triples that
-            match the [=triple pattern=] in `R1`.
-            The dependency is <em>positive</em> 
-            if the [=triple pattern=] is a [=rule body element=]
-            and the dependency is <em>negative</em> if the
-            [=triple pattern=] appears in a [=negation element=].
+          <p>
+            A [=triple pattern=] matches a [=triple template=] if 
+            the triple template may generate a triple that matches the triple pattern.
```
TallTed, line 909: Either —
```suggestion
            the triple template can generate a triple that matches the triple pattern.
```

TallTed: — or —
```suggestion
            the triple template might generate a triple that matches the triple pattern.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+            of `R1` depends on `R2`.
+            The dependency is an [=open dependency=] if `R1` does not have an
+            [=assignment element=] and if all the [=triple patterns=]
+            of `R1` that depend `R2` occur only as [=triple pattern elements=]
+            and not as part of a [=negation element=], and if `R1` does not
+            have any [=triple template=] with a blank node.
+            Otherwise, if `R1` has an [=assignment element=],
+            or if a [=triple pattern=]
+            that might match triples generated by `R2` is present in a
+            [=negation element=], then the dependency is a [=closed dependency=].
```
TallTed, line 940: These complex conditionals are difficult to parse. As above, I suggest refactoring into `<ul>` list items, with nesting as may be appropriate.

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+            <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                >same RDF term</a> as `s`,
+            `tp` is a variable or `tp` is the
+            <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                >same RDF term</a> as `p`,
+            and
+            `to` is a variable or `to` is the
+            <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
+                >same RDF term</a> as `o`.
+          </p>
```
```suggestion
            the following are true:
          </p>
          <ul>
            <li>`ts` is a variable, or `ts` is the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
                >same RDF term</a> as `s`
            </li>
            <li>`tp` is a variable, or `tp` is the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
                >same RDF term</a> as `p`
            </li>
            <li>`to` is a variable, or `to` is the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality"
                >same RDF term</a> as `o`
            </li>
          </ul>
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+            The vertices of the graph are the rules of the rule set, and the edges
+            are labeled either <em>open</em> or <em>closed</em> according to
+            whether the dependency is an [=open dependency=] or a [=closed dependency=].
+          </p>
+          <dl>
+            <dt><dfn>Dependency graph</dfn></dt>
+            <dd>
+              A [=dependency graph=] of a [=rule set=] is a directed graph where each
+              vertex is a [=rule=] in the rule set, and an edge exists from rule `R1`
+              to rule `R2` if `R1` depends on rule `R2`.
```
TallTed: for consistency
```suggestion
              to rule `R2` if rule `R1` depends on rule `R2`.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+              The edge is labeled either <em>open</em> 
+              or <em>closed</em> according whether the dependency
```
TallTed, line 987:
```suggestion
              or <em>closed</em> according to whether the dependency
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,174 @@ <h3>Well-formedness Conditions</h3>
...
+        <section id="dependency-graph-construction-algorithm">
           <h4>Dependency Graph Algorithm</h4>
           <p>
             The following algorithm gives one
             possible method for constructing the [=dependency graph=] from a
-            [=rule set=]. Conformance depends on the dependency definitions 
-            in this section, not on use of this procedure.
+            [=rule set=]. Conformance depends on producing a dependency graph
+            that meets the definitions of a dependency graph,
+            not on the use of this procedure.
```
TallTed, line 1020: While varying words is useful in many forms of prose, specifications like this benefit from using the same word when referring to the same thing, e.g., `algorithm` vs. `procedure`.
```suggestion
            not on the use of this algorithm.
```

shacl12-rules/index.html
```diff
@@ -958,14 +1052,19 @@ <h4>Dependency Graph Algorithm</h4>
...
 
-        foreach pair ( triple pattern TP, depLabel) in bodyDependencies:
-
+        foreach pair (triple pattern TP, depLabel) in bodyDependencies:
+            if R1 has an assignment element:
+              set depLabel to "closed"
+              endif
+              if R1 has a triple template with a blank node:
+                set depLabel to "closed"
+              endif
```
TallTed: This indentation seems off. It seems that `if R1 has a triple template` on line 1059 (and `endif` on both lines 1048 and 1061) should be indented the same as `if R1 has an assignment` on line 1056. There may be similar issues elsewhere in this algorithm and document.
```suggestion
            if R1 has an assignment element:
              set depLabel to "closed"
            endif
            if R1 has a triple template with a blank node:
              set depLabel to "closed"
            endif
```

shacl12-rules/index.html
```diff
@@ -1066,18 +1133,26 @@ <h3>Stratification</h3>
...
-          </dd>
           <dt><dfn 
                 data-lt="stratification layer|layer|stratum"
                 >Stratification layer</dfn></dt>
           <dd>
-            Each set of rules in a [=stratification=] is called a 
-            [=stratification layer=], also called a "stratum".
+            <p>
+              A [=stratification layer=] `SL`, is a pair of disjoint sets of
+              rules (`SL.once`, `SL.general`) .
```
TallTed, line 1146:
```suggestion
              rules (`SL.once`, `SL.general`).
```

shacl12-rules/index.html
```diff
@@ -1066,18 +1133,26 @@ <h3>Stratification</h3>
...
+              `SL.once` contains [=run-once rules=], which are
+              rules that use [=assignment elements=] or produce
+              blank nodes in the [=rule head=]; these rules are each evaluated exactly
+              once at the start of evaluation of the [=stratification layer=].
+              `SL.general` contains the remaining rules, which are evaluated 
+              until no new triples are inferred.
```
TallTed, line 1152:
```suggestion
              and re-evaluated, until no new triples are inferred.
```

shacl12-rules/index.html
```diff
@@ -1086,20 +1161,20 @@ <h4>Stratification Condition</h4>
           <p>
             [=Stratification=] is only defined when the following condition is
             satisfied. If a [=rule set=] does not meet this condition, then this
-            specification does not define the outcome of [=rule set=] evaluation.
+            specification does not define an outcome of [=rule set=] evaluation.
```
```suggestion
            specification does not define an outcome of evaluation of that
            [=rule set=].
```

shacl12-rules/index.html
```diff
@@ -1086,20 +1161,20 @@ <h4>Stratification Condition</h4>
...
-              <em>negative</em> dependency
+              [=closed rule dependency=]
               in the [=dependency graph=] for a [=rule set=].
             </dd>
           </dl>
           <p>
-            In other words, there is no `NOT` used in any rule that
-            transitively depends on itself.
+            In other words, there is no `NOT` or run-once rule (assignment or rule [=triple template=]
+            involving a blank node) in any transitive dependency cycle of the [=dependency graph=].
```
TallTed, line 1182:
```suggestion
            In other words, there is no `NOT` nor run-once rule (assignment or
            rule [=triple template=] involving a blank node) in any transitive
            dependency cycle of the [=dependency graph=].
```

shacl12-rules/index.html
```diff
@@ -1120,8 +1195,9 @@ <h4>Stratification Algorithm</h4>
     let stratumMap be a map from rule to integer
 
     ## The dependency graph should satisfy the stratification condition.
-    ## Check unbounded work due to a violation of the stratification condition.
-    let limit = num rules  + 1
+    ## The check for unbounded stratification is a guard 
+    ## due to a violation of the stratification condition.
```
TallTed, line 1204:
```suggestion
    ## against a violation of the stratification condition.
```

shacl12-rules/index.html
```diff
@@ -1160,32 +1236,36 @@ <h4>Stratification Algorithm</h4>
         endwhile
 
     ## Initialize the result map.
-    let stratumLevels be a map from integer to a set of rules
+    let stratumRules be a map from integer to rules.
```
TallTed, line 1244:
```suggestion
    let stratumRules be a map from integers to rules.
```

shacl12-rules/index.html
afs: The plural form is more appropriate -- singular might be seen to mean "exactly one".

shacl12-rules/index.html
afs: Some change made locally due to overlap.

shacl12-rules/index.html
afs: Some change made locally due to overlap.

shacl12-rules/index.html
afs: Remove the second "rule" as it was all mentioned the lines before.

shacl12-rules/index.html
There are other inconsistences across the doc that need adjusting so this PR is not the right/place for a whole-doc check/fix.

shacl12-rules/index.html
afs: "evaluation of such a [=rule set=]"

shacl12-rules/index.html
afs: That might read as two+evaluations which isn't true.

shacl12-rules/index.html
afs: The web indicates both are acceptable.

shacl12-rules/index.html
```diff
@@ -1598,9 +1683,11 @@ <h4>Process Imports</h4>
           </p>
           <div class="def">
             <p>
-            A [=rule set=] has three components: R.rules, R.data and R.imports.
-            The <dfn>rule set merge</dfn> of two rule sets RS1 and RS2 giving
-            rule set MR:
+            A [=rule set=] has three components: `R.rules`, `R.data` and `R.imports`.
```
```suggestion
            A [=rule set=] has three components: `R.rules`, `R.data`, and `R.imports`.
```

shacl12-rules/index.html
```diff
@@ -1598,9 +1683,11 @@ <h4>Process Imports</h4>
...
+
+            The <dfn>rule set merge</dfn> of two rule sets, `RS1` and `RS2`,
+            is a rule set `MR` defined as follows:
```
```suggestion
            is one rule set, `MR`, defined as follows:
```

shacl12-rules/index.html
```diff
@@ -1773,21 +1860,28 @@ <h3>Evaluation of a Rule Set</h3>
 
 Apply stratification to RS
 
-let L be the sequence of layers after stratification
+let LS be the sequence of layers after stratification
 
 # Inference graph
-let GI = { t &isin; D  | t &notin; in G0 }
+let GI = { t &isin; D  | t &notin; G0 }
```
```suggestion
let GI = { t &isin; D | t &notin; G0 }
```

shacl12-rules/index.html
afs: Added commas - a/one does not read well to me.

afs:
@simonstey -- I believe I have addressed all your comments. 

Please approve this PR to cancel the blocking "requested changes". It can't be dismissed by general WG members due to this repository's setting.

simonstey (approved):
sry, thanks!

shacl12-rules/index.html
TallTed, line 535:
```suggestion
          Rules that involve blank nodes in the [=rule head=] also create new
```

shacl12-rules/index.html
```diff
@@ -531,11 +531,14 @@ <h3>Assignment and Creating RDF Terms</h3>
...
-          and before any rules that depend on the data they produce.
+          and before any rules that depend on the data they produce. 
+          Rules that involve blank node in the [=rule head=] are also creating new
+          RDF terms and are run-once rules.
         </p>
         <p>
-          This condition ensures that rules with assignment do not generate new RDF terms
-          multiple times, with potentially different outcomes.
+          This condition ensures that rules that create RDF terms do not generate new terms
+          multiple times, with potentially different outcomes and also that such rules do not
```
TallTed, line 540:
```suggestion
          multiple times, with potentially different outcomes, and also that such rules do not 
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,178 @@ <h3>Well-formedness Conditions</h3>
...
+          A rule `R1` depends on a rule `R2` if the output of the second rule
+          affects the evaluation of the body of the first rule. That is, the head of `R2`
+          has a [=triple template=] that might generate a triple that matches
+          a [=triple pattern=] in the body of `R1`, 
+          either as a [=triple pattern element=] or inside a [=negation element=].
+        </p>
+        <p>
+          There are two kinds of dependencies: 
+          a <dfn data-lt="closed dependency">closed rule dependency</dfn>
+          and an <dfn data-lt="open dependency">open rule dependency</dfn>.
```
TallTed, line 890: <dfn data-lt="closed dependency">closed rule dependencies</dfn>
          and <dfn data-lt="open dependency">open rule dependencies</dfn>.

shacl12-rules/index.html
```diff
@@ -864,92 +876,178 @@ <h3>Well-formedness Conditions</h3>
...
-            A [=triple pattern=] in the rule body of rule `R1` depends on a
-            rule `R2` if the head of `R2` can possibly generate triples that
-            match the [=triple pattern=] in `R1`.
-            The dependency is <em>positive</em> 
-            if the [=triple pattern=] is a [=rule body element=]
-            and the dependency is <em>negative</em> if the
-            [=triple pattern=] appears in a [=negation element=].
+          <p>
+            A [=triple pattern=] matches a [=triple template=] if 
+            the triple template may generate a triple that matches the triple pattern.
```
TallTed, line 909: "may" can be confusing to readers. "can" is better where RFC2119 "MAY" is not meant.
```suggestion
            the triple template can generate a triple that matches the triple pattern.
```

shacl12-rules/index.html
```diff
@@ -864,92 +876,178 @@ <h3>Well-formedness Conditions</h3>
...
+            of `R1` depends on `R2`.
+            The dependency is an [=open dependency=] if `R1` does not have an
+            [=assignment element=] and if all the [=triple patterns=]
+            of `R1` that depend `R2` occur only as [=triple pattern elements=]
+            and not as part of a [=negation element=], and if `R1` does not
+            have any [=triple template=] with a blank node.
+            Otherwise, if `R1` has an [=assignment element=],
+            or if a [=triple pattern=]
+            that might match triples generated by `R2` is present in a
+            [=negation element=], then the dependency is a [=closed dependency=].
```
TallTed, line 940: lines 927–940 (`<dt><dfn>Rule dependency</dfn></dt>`) appear largely if not completely redundant to lines 877–902 (`<h3>Rule Dependency</h3>`).

shacl12-rules/index.html
```diff
@@ -864,92 +876,178 @@ <h3>Well-formedness Conditions</h3>
...
-                >Transitive rule dependency</dfn></dt>
-          <dd>
-            A rule `R1` has a [=transitive dependency=] on rule `R2` 
-            if there is a path in the [=dependency graph=] from `R1` to `R2`.
-          </dd>
+        <div class="note">
+          <p>
+            A [=triple template=] with components `ts`, `tp`, `to`
+            can possibly generatce a triple with component RDF terms
+            `s`, `p`, `o` if
```
TallTed, line 949: Is this `if` an `and` or an `or`? i.e., `if any of the following is true:` or `if all of the following are true:`?

shacl12-rules/index.html
```diff
@@ -1066,18 +1137,26 @@ <h3>Stratification</h3>
...
-            [=stratification layer=], also called a "stratum".
+            <p>
+              A [=stratification layer=] `SL`, is a pair of disjoint sets of
+              rules (`SL.once`, `SL.general`) .
+              `SL.once` contains [=run-once rules=], which are
+              rules that use [=assignment elements=] or produce
+              blank nodes in the [=rule head=]; these rules are each evaluated exactly
+              once at the start of evaluation of the [=stratification layer=].
+              `SL.general` contains the remaining rules, which are evaluated 
+              until no new triples are inferred.
```
TallTed, line 1152:
```suggestion
              repeatedly until no new triples are inferred.
```

afs:
(I have mentioned this before.)

Comments on merged PR and on individual commits have to be individually cherry-picked. 

They do not relate to current state of the document on main - line numbers are differently aligned, and text has changed.

They are high overhead. A PR against the current document would be appreciated.

TallTed:
Sorry, I didn't intend to make things more difficult. New issues and PRs have been created for the comments made here.

PR #842, opened 2026-03-27
Add support for owl:versionIRI

mgberg:
This adds support for `owl:versionIRI` in shapes graph import resolution and SPARQL prefix declaration lookup. When a shapes graph is imported by its version IRI, processors should now navigate from the version IRI to the ontology IRI before following further `owl:imports` or `sh:declare` paths.

Closes #830

* [See the Core document rendered online here](https://raw.githack.com/w3c/data-shapes/830-support-for-owlversioniri/shacl12-core/index.html)
* [See the SPARQL document rendered online here](https://raw.githack.com/w3c/data-shapes/830-support-for-owlversioniri/shacl12-sparql/index.html)

PR #876, opened 2026-04-30
Handling of possible import closure incompatibilities due to owl:versionIRI

mgberg:
[See this document rendered online here](https://raw.githack.com/w3c/data-shapes/owl-versionIRI-guidance/shacl12-core/index.html)

Since `owl:versionIRI` support was added for shapes graph imports in #842, it was proposed that the spec should address potential incompatibilities that can arise in the import closure. The [OWL 2 Imports specification](https://www.w3.org/TR/owl2-syntax/#Imports) defines constraints on import closures, specifically around conflicting versions and `owl:incompatibleWith` declarations. This is added as part of checking whether the shapes graph is "well-formed", which currently is not strictly required by the spec. Furthermore, the proposed checks are farily simple to perform on the shapes graph after closure is complete so the implementation overhead should not be significant.

This PR includes the following specific changes:

### 1. Expanded well-formed shapes graph definition

Restructured the well-formed/ill-formed definitions so that a shapes graph is well-formed if:
1. It contains no ill-formed nodes (the prior definition of a well-formed shapes graph), AND
2. Its import closure satisfies the constraints defined by OWL 2, namely that no two shapes graphs that are different versions of the same series or where one declares `owl:incompatibleWith` the other.

### 2. Informative note in the Shapes Graph section

Added a note near the `owl:versionIRI` import example warning that versioned imports may introduce incompatibilities.

### 3. Simplified `sh:shapesGraphWellFormed` description

Updated the description to defer to the `well-formed` definition rather than only referencing the syntax rules appendix.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -6423,6 +6423,7 @@ ex:NameGroup
 				<li>Added the new constraint component <a href="#SubsetOfConstraintComponent"><code>sh:subsetOf</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/786">Issue 786</a></li>
 				<li>Added the graph types <code>sh:DataGraph</code> and <code>sh:ShapesGraph</code>; see <a href="https://github.com/w3c/data-shapes/issues/169">Issue 169</a></li>
 				<li>Added the new constraint component <a href="#UniqueValuesForConstraintComponent"><code>sh:uniqueValuesFor</code></a>, see <a href="https://github.com/w3c/data-shapes/issues/661">Issue 661</a></li>
+				<li>Added support for <a data-cite="owl2-syntax/#Ontology_IRI_and_Version_IRI"><code>owl:versionIRI</code></a> in <a href="#shapes-graph">shapes graph</a> imports, and expanded the definition of <a>well-formed</a> shapes graphs to include import closure constraints from [[owl2-syntax]]; see <a href="https://github.com/w3c/data-shapes/issues/830">Issue 830</a></li>
 			</ul>
 		</section>
   </body>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -505,7 +505,7 @@
 					<li>It contains no <a>ill-formed</a> nodes (i.e., all nodes satisfy the syntax rules mentioned above).</li>
 					<li>Its import closure (as determined by transitively following
 						<a data-cite="owl2-syntax/#Imports"><code>owl:imports</code></a> statements)
-						satisfies the constraints defined by [[owl2-syntax]], namely that it does not contain two shapes graphs such that:
+						satisfies the constraints defined by [[owl2-syntax]], namely that it does not contain two shapes graphs where:
 						<ul>
 							<li>
 								they are different versions of the same series
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -497,7 +497,30 @@
 					The complete list of these rules can be found in the <a href="#syntax-rules">appendix</a>.
 					Nodes that violate any of these rules are called <dfn>ill-formed</dfn>.
 					Nodes that violate none of these rules are called <dfn>well-formed</dfn>.
-					A <a>shapes graph</a> is ill-formed if it contains at least one ill-formed node.
+				</p>
+				<p>
+					A <a>shapes graph</a> is <a>well-formed</a> if all of the following conditions hold:
+				</p>
+				<ol>
+					<li>It contains no <a>ill-formed</a> nodes (i.e., all nodes satisfy the syntax rules mentioned above).</li>
+					<li>Its import closure (as determined by transitively following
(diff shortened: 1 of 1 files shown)


nicholascar (approved):
Looks good to me Matt!

shacl12-core/index.html
```diff
@@ -497,7 +497,30 @@ <h3>Document Conventions</h3>
...
-					A <a>shapes graph</a> is ill-formed if it contains at least one ill-formed node.
+				</p>
+				<p>
+					A <a>shapes graph</a> is <a>well-formed</a> if all of the following conditions hold:
+				</p>
+				<ol>
+					<li>It contains no <a>ill-formed</a> nodes (i.e., all nodes satisfy the syntax rules mentioned above).</li>
+					<li>Its import closure (as determined by transitively following
+						<a data-cite="owl2-syntax/#Imports"><code>owl:imports</code></a> statements)
+						satisfies the constraints defined by [[owl2-syntax]], namely that it does not contain two shapes graphs where:
```
ajnelson-nist: Grammar nit - I'd prefer citation style be tweaked (citation-item not be a noun), but won't hold up the PR on this.  I'll suggest text after a scan.

ajnelson-nist: I've had a little trouble getting the in-page anchor syntax right in ReSpec, but there should be a link emphasizing [OWL2-Syntax Section 3.4](https://www.w3.org/TR/owl2-syntax/#Imports), which is from where this guidance is (mostly, and reasonably) quoted.

ajnelson-nist: Ah, nevermind the emphasis link point.  I missed the `data-cite` link just prior.


```suggestion
						satisfies the constraints defined by the syntax specification of OWL 2 [[owl2-syntax]], namely that it does not contain two shapes graphs where:
```

ajnelson-nist (approved):
Approved.  I left one grammar suggestion as a nice-to-have, but it's fine to me to merge without if you'd prefer.

mgberg:
@HolgerKnublauch and other Core editors- I would appreciate a round of feedback on this please when you get a chance.

shacl12-core/index.html
```diff
@@ -497,7 +497,30 @@ <h3>Document Conventions</h3>
...
-					A <a>shapes graph</a> is ill-formed if it contains at least one ill-formed node.
+				</p>
+				<p>
+					A <a>shapes graph</a> is <a>well-formed</a> if all of the following conditions hold:
+				</p>
+				<ol>
+					<li>It contains no <a>ill-formed</a> nodes (i.e., all nodes satisfy the syntax rules mentioned above).</li>
+					<li>Its import closure (as determined by transitively following
+						<a data-cite="owl2-syntax/#Imports"><code>owl:imports</code></a> statements)
+						satisfies the constraints defined by the syntax specification of OWL 2 [[owl2-syntax]], namely that it does not contain two shapes graphs where:
```
HolgerKnublauch: "the constraints" - do you really mean that it needs to satisfy *all* syntax constraints by OWL 2? This doesn't sound right as SHACL doesn't depend on OWL.

shacl12-core/index.html
```diff
@@ -2193,6 +2216,13 @@ <h3>Shapes Graph</h3>
 				<p>
 					The resulting graph forms the input <a>shapes graph</a> for validation and MUST NOT be further modified during the validation process.
 				</p>
+				<p class="note">
+					When using <code>owl:versionIRI</code> to import versioned shapes graphs, care should be taken to
+					avoid importing incompatible versions. In particular, the import closure of a shapes graph
+					SHOULD NOT contain two graphs that are different versions of the same series, or where one
```
HolgerKnublauch, line 2222: Here you're using SHOULD NOT but further above the definition of ill-formed is an implicit MUST. So how strict is it?

shacl12-core/index.html
mgberg: I just meant the two rules listed below, does this change make that clear?
```suggestion
						satisfies the constraints defined by <a data-cite="owl2-syntax/#Imports">Section 3.4</a> of the OWL 2 syntax specification [[owl2-syntax]], namely that it does not contain two shapes graphs where:
```

shacl12-core/index.html
mgberg: I used SHOULD NOT here because section 6.7.1.4 Syntax Checking of Shapes Graph says:
> SHACL validation engines are not strictly required to check whether the shapes graph is well-formed.

If it would be more appropriate to say MUST NOT here even though checking for conformance is not required, I can certainly make that change.

HolgerKnublauch: Ok, hard to say. I'll approve and we can let it sit. It's a rather fuzzy topic in general, with all these weakly defined semantics of ill-formed graphs.

HolgerKnublauch (approved):

Issue #869, opened 2026-04-24
Global configuration for SHACL UI


robert-david:
This is a proposal for global configuration parameters mainly used by the UI
based on the discussions in the SHACL UI task force and the issues
- https://github.com/w3c/data-shapes/issues/855
- https://github.com/w3c/data-shapes/issues/857
- https://github.com/w3c/data-shapes/issues/859

Example
```
:shapes a owl:Ontology .
:shapes shui:hasConfiguration :config .

:config a shui:GlobalConfiguration;
	shui:defaultNamespace "https://somewhere.com";
	shui:hasLanguagePreference ("" "en" "de");
	shui:hasTimeZone "Europe/Vienna" .
```
The configuration node can use the shapes graph URI or any other URI. The configuration is retrieved based on the `owl:Ontology` class membership. This was the original proposal, but we could also change it to be a different class.
`shui:hasConfiguration` links to a `shui:GlobalConfiguration`, which can be locally defined or externally (centralised) provided.
Global parameters are
- `shui:defaultNamespace` is mainly used for generating new URIs for added data via UI.
- `shui:hasLanguagePreference` provides a list of languages in order of priority, with an empty literal `""` representing no language. This is mainly used for defaulting and selecting languages for new added literals in the UI.
- `shui:hasTimeZone` provides a literal time zone representation, which can be used to adapt date and time literals if needed , to log change dates, ...

I agree with @edmondchuc that URI generation is a separate topic (which most likely still uses `shui:defaultNamespace`, though).

ajnelson-nist:
FWIW: Configuration was also described in #431 , though not at the scope of UI.

HolgerKnublauch:
Please let's not start with "hasXY" properties. The "has" is always an implicit default, e.g. we don't need sh:hasMinCount, sh:hasClass etc. The over-use of "has" in OWL is an anti-pattern that doesn't seem to get used in other modeling languages.

Also, I assume the shui:configuration link is redundant if people double-type the sh:Graph, e.g.

```
:shapes a sh:DataGraph, shui:Configuration .
	shui:defaultNamespace "https://somewhere.com";
	shui:languagePreference ("" "en" "de");
	shui:timeZone "Europe/Vienna" .
```

In fact now that we have introduced sh:Graph, sh:DataGraph and sh:ShapesGraph (with a recent PR), we could state sh:Graph rdfs:subClass shui:Configuration to make peoples' lives easier.

robert-david:
[...]

PR #877, opened 2026-05-01
[UI] Added UI label resolution

### LINKED ISSUE #610: [UI] Label resolution
This topic arose from one of the UI TF meetings on language handling, described in https://github.com/w3c/data-shapes/issues/596.

Many SHACL- and RDF-based applications face the challenge of label resolution. This problem exists in applications where a human-friendly label is required to display an RDF resource. This includes displaying resource listings in navigation menus, search or query results, and titles when viewing or editing resources.

The same challenge applies when presenting short descriptions of resources.

Here are some proposed mechanisms for resolving labels, from most specific to more general. In most cases, the selected labels should also undergo language resolution to determine the final display label and to ensure the label matches the preferred language of the user, their operating system, or the shape constraints.

### Property Role Vocabulary

Ticket on property roles: https://github.com/w3c/data-shapes/issues/538

If a resource's node shape contains a property shape annotated with `dash:propertyRole dash:LabelRole`, then the value nodes of this shape should be considered for label use.

```turtle
ex:PropertyShape
    a sh:PropertyShape ;
    sh:path schema:name ;
    dash:propertyRole dash:LabelRole .
```

### `rdfs:label`

`rdfs:label` is widely accepted and supported as the property to use for human-friendly labels in applications.

```turtle
ex:john rdfs:label  "John"@en .
```

### Fallback options

When no options are suitable, implementations may opt to display the full IRI or blank node of the resource, or attempt to make it human-friendly by:

- displaying just the local name of the IRI
- construct a prefixed name if the application supports namespace prefixes.

bergos:
Adds label resolution to the UI specification.

Closes #610

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/label-resolution/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1162,7 +1162,13 @@ ex:alice a ex:Person ;
               as the label.
             </li>
             <li>
-              If <var>P</var> is a predicate IRI and the <a>shapes graph</a> or <a>data graph</a>
+              If <var>P</var> is a predicate IRI and the <a>data graph</a>
+              contains one or more <code>rdfs:label</code> triples with subject <var>P</var>, select
+              the best matching value using <a>language resolution</a>. If a match is found, use
+              that literal as the label.
+            </li>
+            <li>
+              If <var>P</var> is a predicate IRI and the <a>shapes graph</a>
               contains one or more <code>rdfs:label</code> triples with subject <var>P</var>, select
               the best matching value using <a>language resolution</a>. If a match is found, use
               that literal as the label.
@@ -1203,6 +1209,12 @@ ex:alice a ex:Person ;
               that literal as the label.
             </li>
             <li>
+              If the <a>shapes graph</a> contains one or more values for the configured property path for
+              the value node label resolution (defaulting to <code>rdfs:label</code>) for <var>V</var>,
+              select the best matching value using <a>language resolution</a>. If a match is found, use
+              that literal as the label.
+            </li>
+            <li>
               If <var>V</var> is an IRI, use the
               <a href="#label-resolution-local-name">local name resolution</a> of <var>V</var> as the
               label.
@@ -1220,7 +1232,7 @@ ex:alice a ex:Person ;
           <p>
             To determine the label for an IRI the <a>local name</a> <var>L</var> of the IRI, SHOULD be
             used. Implementations SHOULD transform the local name into a human-friendly form, for example
-            by splitting camel-case identifiers into words.
+            by splitting camel case identifiers into words.
           </p>
         </section>
 
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1230,7 +1230,7 @@ ex:alice a ex:Person ;
         <section id="label-resolution-local-name">
           <h4>Local Name Resolution</h4>
           <p>
-            To determine the label for an IRI the <a>local name</a> <var>L</var> of the IRI, SHOULD be
+            To determine the label for an IRI, the <a>local name</a> <var>L</var> of the IRI SHOULD be
             used. Implementations SHOULD transform the local name into a human-friendly form, for example
             by splitting camel case identifiers into words.
           </p>
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1123,6 +1123,175 @@ ex:alice a ex:Person ;
 
       <section id="label-resolution">
         <h3>Label Resolution</h3>
+
+        <p>
+          Label resolution determines the most appropriate human-readable string to display for an
+          RDF resource in a user interface. This section defines the algorithm for determining that
+          label, considering annotations in the <a>shapes graph</a>, values in the <a>data graph</a>,
+          and fallback strategies when no suitable label is found.
+        </p>
+
+        <p>
+          Label resolution is applied in two contexts:
+        </p>
+        <ul>
+          <li>
+            <strong>Property labels</strong>: the label used for the field or column heading of a
+            <a>property UI component</a>, derived from the <a>property shape</a> or from the predicate
+            identified by <code>sh:path</code>.
+          </li>
+          <li>
+            <strong>Value node labels</strong>: the label used to represent a <a>value node</a> (e.g.,
+            an IRI or blank node) when it appears in a UI form or list.
+          </li>
+        </ul>
+
+        <section id="label-resolution-property-labels">
+          <h4>Property Labels</h4>
+
+          <p>
+            To determine the label for a <a>property UI component</a> whose <a>property shape</a> has
+            <code>sh:path</code> pointing to a <a>property path</a> <var>P</var>, implementations MUST
+            apply the following steps in order, stopping at the first step that yields a result:
+          </p>
+          <ol>
+            <li>
+              If the <a>property shape</a> has one or more values for the configured <a>property path</a> for
+              the property label resolution (defaulting to <code>sh:name</code>), select the best
+              matching value using <a>language resolution</a>. If a match is found, use that literal
+              as the label.
+            </li>
+            <li>
+              If <var>P</var> is a predicate IRI and the <a>shapes graph</a> or <a>data graph</a>
+              contains one or more <code>rdfs:label</code> triples with subject <var>P</var>, select
+              the best matching value using <a>language resolution</a>. If a match is found, use
+              that literal as the label.
+            </li>
+            <li>
+              If <var>P</var> is a predicate IRI, use the
+              <a href="#label-resolution-local-name">local name resolution</a> of <var>P</var> as the label.
+            </li>
+            <li>
+              Otherwise, a custom translation algorithm is applied to convert the complex property path
+              <var>P</var> into a human-readable string representation.
+            </li>
+          </ol>
+        </section>
+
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
```diff
@@ -1123,6 +1123,175 @@ <h3>Language Resolution</h3>
...
+        <section id="label-resolution-value-node-labels">
+          <h4>Value Node Labels</h4>
+
+          <p>
+            To determine the label for a <a>value node</a> <var>V</var>, implementations MUST
+            apply the following steps in order, stopping at the first step that yields a result:
+          </p>
+          <ol>
+            <li>
+              If <var>V</var> is a literal, use its lexical form as the label.
```
robert-david, line 1196: Shouldn't this also use the language resolution?

robert-david (approved):

shacl12-ui/index.html
robert-david: Should we also consider here the configuration for the label properties described in https://github.com/w3c/data-shapes/issues/869?

bergos: The describes steps are for a single value node. I made it a bit more explicit by adding "single" in the first paragraph. This step is for the case that the value is a single literal. Language resolution can be only applied if we have multiple values. The next two step cover those cases. One step also covers the property path from the configuration. Once we have a section to refer to, it should be updated.

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,175 @@ <h3>Language Resolution</h3>
...
+              If the <a>property shape</a> has one or more values for the configured <a>property path</a> for
+              the property label resolution (defaulting to <code>sh:name</code>), select the best
+              matching value using <a>language resolution</a>. If a match is found, use that literal
+              as the label.
+            </li>
+            <li>
+              If <var>P</var> is a predicate IRI and the <a>shapes graph</a> or <a>data graph</a>
+              contains one or more <code>rdfs:label</code> triples with subject <var>P</var>, select
+              the best matching value using <a>language resolution</a>. If a match is found, use
+              that literal as the label.
```
smessie: Should all matches in both the shapes graph and the data graph be considered for language resolution, or should first only the shapes graph be considered and only if no match is found, the same approach for the data graph be considered?
Is this clear enough in the current text, and/or do we want to specify this exact behavior?

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,175 @@ <h3>Language Resolution</h3>
...
+            </li>
+            <li>
+              If <var>P</var> is a predicate IRI, use the
+              <a href="#label-resolution-local-name">local name resolution</a> of <var>P</var> as the label.
+            </li>
+            <li>
+              Otherwise, a custom translation algorithm is applied to convert the complex property path
+              <var>P</var> into a human-readable string representation.
```
smessie: Do we want to specify how such a translation algorithm should work?
Otherwise, maybe phrasing like the following would express better that it is implementation specific?

```suggestion
              Otherwise, an implementation-specific translation algorithm should be applied to convert the complex property path
              <var>P</var> into a human-readable string representation.
```

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,175 @@ <h3>Language Resolution</h3>
...
+              whose <code>sh:path</code> is annotated with
+              <code>shui:propertyRole shui:LabelRole</code>, retrieve the values of that path from
+              the <a>data graph</a> for subject <var>V</var>. Select the best matching value using
+              <a>language resolution</a>. If a match is found, use that literal as the label.
+            </li>
+            <li>
+              If the <a>data graph</a> contains one or more values for the configured property path for
+              the value node label resolution (defaulting to <code>rdfs:label</code>) for <var>V</var>,
+              select the best matching value using <a>language resolution</a>. If a match is found, use
+              that literal as the label.
```
smessie, line 1209: If no match is found, should we also consider the shapes graph to search for a match? Additionally, if still not found, should we optionally also allow to consider dereferencing the IRI (in case `V` is an IRI) to search for a match?

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,175 @@ <h3>Language Resolution</h3>
...
+            </li>
+          </ol>
+        </section>
+
+        <section id="label-resolution-local-name">
+          <h4>Local Name Resolution</h4>
+          <p>
+            To determine the label for an IRI the <a>local name</a> <var>L</var> of the IRI, SHOULD be
+            used. Implementations SHOULD transform the local name into a human-friendly form, for example
+            by splitting camel-case identifiers into words.
```
smessie: I believe the correct spelling should be `camelCase` or `camel case`:

```suggestion
            by splitting camelCase identifiers into words.
```

smessie (commented):
Thanks bergi, already looks very nice. I've included some thoughts below.

shacl12-ui/index.html
bergos: That could make sense, but then I would prioritize it the other way round. I guess in most cases the shapes are given (public ontology, another department in a company, ...), and the data graph would be used to overwrite it.

I would just copy that step and only refer to data graph in the first step, and shapes graph in the second step, if that's OK for you?

shacl12-ui/index.html
bergos: Yes, makes sense to make it more explicit that it's implementation-specific. I guess based on the target audience, the implementations may cover a range from close to the RDF definition to close to natural language.

shacl12-ui/index.html
bergos: I was also thinking about it, but thought we may cover that in a separate PR. That would be probably a definition how to customize it since way may not be able to cover all cases. For example, for a use case like Solid, dereferencing the IRI would be the right solution, in an enterprise environment, a SPARQL query or Web Service call could be required.

But adding the shapes graph to the current PR makes sense to cover `rdf:type` triples.

shacl12-ui/index.html
smessie: Sounds good to me, also to prioritize data graph first.

shacl12-ui/index.html
smessie: Separate PR sounds good. Should we open an issue to keep track of that?

bergos: I've opened issue #883 for it.

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,187 @@ <h3>Language Resolution</h3>
...
+              (e.g., an empty string or an identifier derived from the blank node identifier) as
+              the label.
+            </li>
+          </ol>
+        </section>
+
+        <section id="label-resolution-local-name">
+          <h4>Local Name Resolution</h4>
+          <p>
+            To determine the label for an IRI the <a>local name</a> <var>L</var> of the IRI, SHOULD be
```
smessie: If I understand this sentence correctly, I think the comma should be placed differently:
```suggestion
            To determine the label for an IRI, the <a>local name</a> <var>L</var> of the IRI SHOULD be
```

smessie (approved):
One dangling comment I forgot to save. Besides that, the PR looks great!

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,187 @@ <h3>Language Resolution</h3>
 
       <section id="label-resolution">
         <h3>Label Resolution</h3>
+
+        <p>
+          Label resolution determines the most appropriate human-readable string to display for an
```
edmondchuc, line 1128: Should we use `dfn` tags here for both label and language resolution so that their definitions are defined here in a similar manner to how the SHACL Renderer and Node and Property UI Components have been recently changed in PR https://github.com/w3c/data-shapes/pull/885?

edmondchuc (approved):

shacl12-ui/index.html
bergos: If a term is in the Terminology section, the definition should point to that section. But I'm not sure what should go into the Terminology section and the dedicated section. I see only one case in the core specification with an overlap for the property path.

Proposal: Let's create a ticket to revisit the Terminology section once we filled the rest of the specification?

edmondchuc: I could be wrong, but it appears that terms with a dedicated section are defined in those sections, while terms without dedicated sections are defined in the terminology section.

Not a biggie, I don't want this to become a blocker since we can always clean it up later.

bergos: I agree, but in general looking over the terminology section again is maybe not a bad idea. I've created an issue for it #898

shacl12-ui/index.html
```diff
@@ -1123,6 +1123,187 @@ <h3>Language Resolution</h3>
...
+            an IRI or blank node) when it appears in a UI form or list.
+          </li>
+        </ul>
+
+        <section id="label-resolution-property-labels">
+          <h4>Property Labels</h4>
+
+          <p>
+            To determine the label for a <a>property UI component</a> whose <a>property shape</a> has
+            <code>sh:path</code> pointing to a <a>property path</a> <var>P</var>, implementations MUST
```
TallTed, line 1154:
```suggestion
            a <code>sh:path</code> pointing to a <a>property path</a> <var>P</var>, implementations MUST
```

PR #884, opened 2026-05-11
#878: Replaced "is informative" with "is non-normative", SHOULD->should, MAY->may in those sections

### LINKED ISSUE #878: [Core] Difference between "non-normative" and "informative" sections?
I'm not really sure what the difference between "informative" sections like here:
https://github.com/w3c/data-shapes/blob/cd2cdc9fe56f362835957452e2092e898bfeac8c/shacl12-core/index.html#L2092-L2103

vs "non-normative" like here:
https://github.com/w3c/data-shapes/blob/cd2cdc9fe56f362835957452e2092e898bfeac8c/shacl12-core/index.html#L865-L872

is.. If the intention of the former was a mistake (e.g. formal <-> informal) and was actually meant to denote a "non-normative" one then the use of RFC 2119 terms like SHOULD isn't correct (https://www.rfc-editor.org/rfc/rfc2119.html). 

If it was on the other hand indeed meant to say "informative" -> why point this out explicitly? shouldn't the entire spec. be considered "informative" as in "informing about the correct use, definition, ..."?

(/cc our experts @afs @TallTed )

HolgerKnublauch:
Closes #878

* [See the Core document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-878/shacl12-core/index.html)

* [See the SPARQL document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-878/shacl12-sparql/index.html)

* [See the NodeExpr document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-878/shacl12-node-expr/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -861,7 +861,6 @@ ex:Person
 
 		<section id="constraints-section">
 			<h2>Shapes and Constraints</h2>
-
 			<p><em>The following introduction is non-normative.</em></p>
 			<p>
 				The following informal diagram provides an overview of some of the key classes in the SHACL vocabulary.
@@ -1065,7 +1064,7 @@ ex:MultiplePatternsShape
 					<p>
 						An <a>RDF term</a> that is <a>validated</a> against a <a>shape</a> using the triples from a <a>data graph</a> is called a <dfn>focus node</dfn>.
 					</p>
-					<p><em>The remainder of this section is informative.</em></p>
+					<p><em>The remainder of this section is non-normative.</em></p>
 					<p>
 						The set of <a>focus nodes</a> for a <a>shape</a> may be identified as follows:
 					</p>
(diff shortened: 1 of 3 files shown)


afs:
https://www.w3.org/guide/manual-of-style/#Normative

It looks like "informative" and "non-normative" are interchangeable.

HolgerKnublauch:
> https://www.w3.org/guide/manual-of-style/#Normative
> 
> It looks like "informative" and "non-normative" are interchangeable.

Yes but even then it's better to use one term only, and I had to remove the SHOULD/MAY, so I would still appreciate someone approve this PR.

TallTed:
> It looks like "informative" and "non-normative" are interchangeable.

As I said elsewhere, yes, but they can confuse readers. Other groups I'm in have decided to exclusively use the latter, because there's no confusion about the pair of `normative` and `non-normative`. I'm in favor of this change.

For similar reasons of reader confusion, I strongly encourage using `ought` or similar and `can` or similar in non-normative sections, even though `should`, `may`, and other _lowercase_ RFC2119 keywords are defined elsewhere as being non-normative, while `SHOULD`, `MAY`, and other *uppercase* RFC2119 keywords are defined as being normative. I'm not in favor of `SHOULD->should` nor `MAY->may`.

bergos (approved):

PR #886, opened 2026-05-12
Removed SHACL-C references

### LINKED ISSUE #462: Affirm decision to keep SHACL-C snippets in Core
Requested by @jeswr in parent issue #460 .

Do we have consensus to keep the SHACL-C syntax snippets in the Core document?

HolgerKnublauch:
Closes #462

There is probably more SHACL-C parsing code here but I didn't write that - maybe @bergos remembers and can remove that?

### DIFF
--- a/scripts/lib/utils.js
+++ b/scripts/lib/utils.js
@@ -1,5 +1,4 @@
 import rdf from 'rdf-ext'
-import { parse } from 'shaclc-parse'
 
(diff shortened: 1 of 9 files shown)


bergos:
I added a commit to remove SHACL-C from the script that syncs the examples.

bergos (approved):

PR #888, opened 2026-05-14
Add profiling terms

### LINKED ISSUE #887: Add Profiles Vocabulary elements to Core vocab
Add the following new elements to the SHACL Core vocabulary:

* `sh:conformsTo`
* `sh:hasValidationReport` 
* `sh:usedValidationReport`

Definitions for these elements are provided in the [SHACL 1.2 Profiles SHACL Profiling Vocabulary](https://raw.githack.com/w3c/data-shapes/profiling-vocab-start/shacl12-profiling/index.html#vocabulary)

nicholascar:
Adds terms needed for the SHACL Profiling vocab, as per Issue #887.

Fixes a syntax error in the vocab, `##` -> `#`

Closes #887


shacl12-vocabularies/shacl.ttl
```diff
@@ -226,6 +226,15 @@ sh:conforms
 	rdfs:range xsd:boolean ;
 	rdfs:isDefinedBy sh: .
 
+sh:conformsTo
+	a rdf:Property , owl:ObjectProperty ;
```
HolgerKnublauch: Please no unnecessary use of owl: vocabulary. rdf:Property should be enough.

shacl12-vocabularies/shacl.ttl
```diff
@@ -234,6 +243,14 @@ sh:conformanceDisallows
 	rdfs:range sh:Severity ;
 	rdfs:isDefinedBy sh: .
 
+sh:hasValidationReport
+	a rdf:Property , owl:ObjectProperty ;
```
HolgerKnublauch: See above.

shacl12-vocabularies/shacl.ttl
```diff
@@ -234,6 +243,14 @@ sh:conformanceDisallows
...
+	rdfs:label "has validation report"@en ;
+	rdfs:comment "A Validation Report exists for the subject."@en ;
```
HolgerKnublauch: Unclear what this property does (and I didn't read the spec yet, intentionally).

From the comment it sounds like a boolean property, but it has range ValidationReport, so what is it?

If it's not a boolean, I would avoid the prefix hasXY and instead go with sh:validationReport. The "has" is implicit almost everywhere. Any reasons not to?

shacl12-vocabularies/shacl.ttl
```diff
@@ -250,6 +267,14 @@ sh:shapesGraphWellFormed
 	rdfs:range xsd:boolean ;
 	rdfs:isDefinedBy sh: .
 
+sh:usedShapesGraph
+	a rdf:Property , owl:ObjectProperty ;
```
HolgerKnublauch: See above on owl:ObjectProperty.

HolgerKnublauch (changes requested):

shacl12-vocabularies/shacl.ttl
nicholascar: I originally added these OWL bits in so we could use the predicates in a property chain axiom, but we've moved to SHACL Rules for the reasoning so I will remove them.

shacl12-vocabularies/shacl.ttl
nicholascar: I have removed "has". I like to keep predicate names more distinct from class names than just a lowercase / uppercase difference - `validationReport` / `ValidationReport` - but I take your point about the "it looks like a boolean" thing, so done!

HolgerKnublauch: I still see "has" - did you commit this?

nicholascar: Apparently not! Oops. But I have now...

HolgerKnublauch: Good, I will approve. Just minor:

"A Validation Report exists for the subject."

shouldn't this be

"A Validation Report that exists for the subject."

HolgerKnublauch (approved):

bergos (approved):

shacl12-vocabularies/shacl.ttl
ajnelson-nist: I concur with @HolgerKnublauch - "that" would be helpful here.

ajnelson-nist:
```suggestion
	rdfs:comment "A Validation Report that exists for the subject."@en ;
```

ajnelson-nist (approved):
Approving.  The one last wording revision to `sh:validationReport` @HolgerKnublauch had suggested is something I'd like, but won't hold up approving for.

PR #900, opened 2026-05-22
Issue 869: toward a global configuration for SHACL UI

### LINKED ISSUE #869: Global configuration for SHACL UI

robert-david:
Closes #869 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-869/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -682,7 +682,7 @@
 				<p>
 					A shapes graph may contain a <a>SHACL instance</a> of the class <code>shui:Configuration</code> 
           to define global configuration properties for rendering the UI.
-          <code>shui:Configuration</code> is a <a>SHACL subclass</a> of <code>sh:Graph</code>
+          <code>sh:Graph</code> is a <a>SHACL subclass</a> of <code>shui:Configuration</code>.
           The following properties can be defined for a <code>shui:Configuration</code> instance.
 					<div class="parameters">Parameters:</div>
 					<table class="term-table">
@@ -697,7 +697,7 @@
 								<td><code>shui:defaultNamespace</code></td>
 								<td>
 									The default namespace used to construct fresh user-added nodes. 
-                  The value of <code>shui:defaultNamespace</code> is an <a>IRI</a>.
+                  The value of <code>shui:defaultNamespace</code> is a <a>Literal</a> whose value is a valid IRI.
 								</td>
 							</tr>
 							<tr>
@@ -706,29 +706,22 @@
 									The language selection for displaying labels in widgets. 
                   The value of <code>shui:hasLanguagePreference</code> is a well-formed SHACL list where all members are language tags. 
                   The preferred language is determined according to the order of priority in the list from beginning (highest) to end (lowest ). 
+                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
 								</td>
 							</tr>
 							<tr>
 								<td><code>shui:timeZone</code></td>
 								<td>
 									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
-                  The value of <code>shui:timeZone</code> is <code>xsd:string</code>. 
+                  The value of <code>shui:timeZone</code> is of datatype <code>xsd:string</code>. 
 								</td>
 							</tr>
 							<tr>
-								<td><code>shui:shapesLabelPreference</code></td>
+								<td><code>shui:labelPreference</code></td>
 								<td>
-									The label properties used to determine labels to display in widgets. 
-                  Properties are looked up in the <a>shapes graph</a>. 
-                  The value of <code>shui:shapesLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
-								</td>
-							</tr>
-							<tr>
-								<td><code>shui:dataLabelPreference</code></td>
-								<td>
-									The label properties used to determine labels to display in widgets and labels to represent nodes values in the widgets. 
-                  Properties are looked up in the <a>data graph</a>. 
-                  The value of <code>shui:dataLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
+									The label properties used to determine the labels to display in widgets and labels to represent nodes values in the widgets. 
+                  The value of <code>shui:labelPreference</code> is a well-formed SHACL list where all members are IRIs. 
+                  The preferred label property is determined according to the order of priority in the list from beginning (highest) to end (lowest).
 								</td>
 							</tr>
 						</tbody>
@@ -740,8 +733,7 @@ ex:config a shui:Configuration ;
     shui:defaultNamespace &lt;http://example.com/ns#&gt; ;
     shui:hasLanguagePreference ("" "en" "de") ;
     shui:hasTimeZone "Europe/Vienna" ;
-    shui:shapesLabelPreference (rdfs:label, skos:prefLabel) ;
-    shui:dataLabelPreference (skos:prefLabel, dcterms:title, rdfs:label) .
+    shui:labelPreference (skos:prefLabel, dcterms:title, rdfs:label) .
 						</div>
 					</div>
         </p>
@@ -1530,10 +1522,13 @@ ex:alice a ex:Person ;
           </li>
         </ul>
 
-        <p>
-          Label resolution is done according to the property IRIs listed using the <a>global configuration</a> property <code>shui:dataLabelPreference</code>. 
-          It defaults to <code>rdfs:label</code>.
-        </p>
+        <section id="label-property-resolution">
+          <h4>Label Property Resolution</h4>
+          <p>
+            Label property resolution is done according to the preferred property IRIs listed in the <a>global configuration</a> property <code>shui:labelPreference</code>. 
+            Preferred labels default to <code>sh:name</code> for Property Labels and <code>rdfs:label</code> for Value Node Labels.
+          </p>
+        </section>
 
         <section id="label-resolution-property-labels">
           <h4>Property Labels</h4>
@@ -1546,25 +1541,24 @@ ex:alice a ex:Person ;
           <ol>
             <li>
               If the <a>property shape</a> has one or more values for the configured <a>property path</a> for
-              the property label resolution (defaulting to <code>sh:name</code>), select the best
-              matching value using <a>language resolution</a>. If a match is found, use that literal
-              as the label.
+              properties determined by <a>label property resolution</a>, select the best
+              matching value using <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
               If <var>P</var> is a predicate IRI and the <a>data graph</a>
-              contains one or more triples with properties determined by <a>label resolution</a> and with subject <var>P</var>, select
-              the best matching value using <a>language resolution</a>. If a match is found, use
-              that literal as the label.
+              contains one or more triples with properties determined by <a>label property resolution</a> and with subject <var>P</var>, select
+              the best matching value using <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
               If <var>P</var> is a predicate IRI and the <a>shapes graph</a>
-              contains one or more triples with properties determined by <a>label resolution</a> and with subject <var>P</var>, select
-              the best matching value using <a>language resolution</a>. If a match is found, use
-              that literal as the label.
+              contains one or more triples with properties determined by <a>label property resolution</a> and with subject <var>P</var>, select
+              the best matching value using <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
-              If <var>P</var> is a predicate IRI, use the
-              <a>local name resolution</a> of <var>P</var> as the label.
+              If <var>P</var> is a predicate IRI, use the <a>local name resolution</a> of <var>P</var> as the label.
             </li>
             <li>
               Otherwise, an implementation-specific translation algorithm should be applied to convert the complex property path
@@ -1589,19 +1583,20 @@ ex:alice a ex:Person ;
               whose <code>sh:path</code> is annotated with
               <code>shui:propertyRole shui:LabelRole</code>, retrieve the values of that path from
               the <a>data graph</a> for subject <var>V</var>. Select the best matching value using
-              <a>language resolution</a>. If a match is found, use that literal as the label.
+              <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
               If the <a>data graph</a> contains one or more values for the configured property path for
-              the value node determined by <a>label resolution</a> for <var>V</var>,
-              select the best matching value using <a>language resolution</a>. If a match is found, use
-              that literal as the label.
+              the value node determined by <a>label property resolution</a> for <var>V</var>,
+              select the best matching value using <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
               If the <a>shapes graph</a> contains one or more values for the configured property path for
-              the value node  determined by <a>label resolution</a> for <var>V</var>,
-              select the best matching value using <a>language resolution</a>. If a match is found, use
-              that literal as the label.
+              the value node  determined by <a>label property resolution</a> for <var>V</var>,
+              select the best matching value using <a>language resolution</a>. 
+              If a match is found, use that literal as the label.
             </li>
             <li>
               If <var>V</var> is an IRI, use the
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -450,7 +450,7 @@
           </div>
         </div>
         <div class="def">
-          <div class="term-def-header">Language Resolution</div>
+          <div class="term-def-header">Label Property Resolution</div>
           <div>
             <dfn data-lt="label property resolution">Label Property Resolution</dfn> is the process of determining and ordering the
             label property IRIs for label resolution, prioritizing the preferred or most relevant property
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1452,7 +1452,7 @@ ex:part-b
             expressed is up to the implementation. Possible approaches include but are not limited to:
             <ul>
               <li>A UI feature that allows the user to select a preferred display language or languages priorities.</li>
-              <li>A <a>SHACL global configuration</a> using the property <code>shui:hasLanguagePreference</code>.</li>
+              <li>A <a>global configuration</a> using the property <code>shui:hasLanguagePreference</code>.</li>
             </ul>
           </li>
           <li>
@@ -1531,7 +1531,7 @@ ex:alice a ex:Person ;
         </ul>
 
         <p>
-          Label resolution is done according to the property IRIs listed using the <a>SHACL global configuration</a> property <code>shui:dataLabelPreference</code>. 
+          Label resolution is done according to the property IRIs listed using the <a>global configuration</a> property <code>shui:dataLabelPreference</code>. 
           It defaults to <code>rdfs:label</code>.
         </p>
 
@@ -1552,7 +1552,7 @@ ex:alice a ex:Person ;
             </li>
             <li>
               If <var>P</var> is a predicate IRI and the <a>data graph</a>
-              contains one or more <code>rdfs:label</code> triples with subject <var>P</var>, select
+              contains one or more triples with properties determined by <a>label resolution</a> and with subject <var>P</var>, select
               the best matching value using <a>language resolution</a>. If a match is found, use
               that literal as the label.
             </li>
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -433,6 +433,12 @@
               >value node</a
             >
           </dfn>
+          ,
+          <dfn data-lt="global configuration">
+            <a href="https://www.w3.org/TR/shacl/#global-configuration"
+              >SHACL Global Configuration</a
+            >
+          </dfn>
           .
         </p>
         <div class="def">
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -434,7 +434,7 @@
             >
           </dfn>
           ,
-          <dfn data-lt="global configuration">
+          <dfn>
             <a href="#global-configuration"
               >SHACL Global Configuration</a
             >
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -671,6 +671,76 @@
           </div>
       </section>
 
+      <section id="global-configuration">
+      	<h3>SHACL Global Configuration</h3>
+				<p>
+					A shapes graph may contain a <a>SHACL instance</a> of the class <code>shui:Configuration</code> 
+          to define global configuration properties for rendering the UI.
+          <code>shui:Configuration</code> is a <a>SHACL subclass</a> of <code>sh:Graph</code>
+          The following properties can be defined for a <code>shui:Configuration</code> instance.
+					<div class="parameters">Parameters:</div>
+					<table class="term-table">
+						<thead>
+							<tr>
+								<th>Property</th>
+								<th>Summary and Syntax Rules</th>
+							</tr>
+						</thead>
+						<tbody>
+							<tr>
+								<td><code>shui:defaultNamespace</code></td>
+								<td>
+									The default namespace used to construct fresh user-added nodes. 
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
           </div>
       </section>
 
+      <section id="global-configuration">
+      	<h3>SHACL Global Configuration</h3>
+				<p>
+					A shapes graph may contain a <a>SHACL instance</a> of the class <code>shui:Configuration</code> 
+          to define global configuration properties for rendering the UI.
+          <code>shui:Configuration</code> is a <a>SHACL subclass</a> of <code>sh:Graph</code>
```
```suggestion
          <code>sh:Graph</code> is a <a>SHACL subclass</a> of <code>shui:Configuration</code>
```

The order should be reversed. A `sh:Graph` should be a subtype of `sh:Configuration`, making it possible to attach configuration properties directly. However, making every `sh:Configuration` a `sh:Graph` would prevent nested or alternative configurations from coexisting in the same graph.

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+								<th>Property</th>
+								<th>Summary and Syntax Rules</th>
+							</tr>
+						</thead>
+						<tbody>
+							<tr>
+								<td><code>shui:defaultNamespace</code></td>
+								<td>
+									The default namespace used to construct fresh user-added nodes. 
+                  The value of <code>shui:defaultNamespace</code> is an <a>IRI</a>.
```
```suggestion
                  The value of <code>shui:defaultNamespace</code> is a <a>Literal</a> whose value is a valid IRI.
```

It should be a literal whose value is an IRI. The default namespace is not itself a resource — the IRIs derived from it will be.

shacl12-ui/index.html
```diff
@@ -1512,13 +1593,13 @@ <h4>Value Node Labels</h4>
...
               select the best matching value using <a>language resolution</a>. If a match is found, use
               that literal as the label.
             </li>
             <li>
               If the <a>shapes graph</a> contains one or more values for the configured property path for
-              the value node label resolution (defaulting to <code>rdfs:label</code>) for <var>V</var>,
+              the value node  determined by <a>label resolution</a> for <var>V</var>,
               select the best matching value using <a>language resolution</a>. If a match is found, use
               that literal as the label.
             </li>
```
bergos: Defining the logic once and referencing it is a good approach, given it's used four times — but "label resolution" doesn't work as the term here. Since "Property Labels" and "Value Node Labels" are already sub-sections of "Label Resolution" for specific cases, this creates a circular reference. A better approach would be to introduce a new sub-section, as was done for "Local Name Resolution". Additionally, the default should not be defined at the config level but rather within each sub-section: `rdfs:label` as the default for "Value Node Labels" and `sh:name` as the default for "Property Labels".

bergos (changes requested):

shacl12-ui/index.html
smessie: Suggesting an additional trailing dot.

```suggestion
          <code>sh:Graph</code> is a <a>SHACL subclass</a> of <code>shui:Configuration </code>.
```

shacl12-ui/index.html
smessie: @bergos, so do I understand it correctly that you suggest to use rather `shui:defaultNamespace "http://example.com/ns#"` instead of `shui:defaultNamespace <http://example.com/ns#> ;`?

I see now in #869 it was also suggested as a Literal. Beware that we'll then have to update this example in the PR accordingly.

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+									The default namespace used to construct fresh user-added nodes. 
+                  The value of <code>shui:defaultNamespace</code> is an <a>IRI</a>.
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:hasLanguagePreference</code></td>
+								<td>
+									The language selection for displaying labels in widgets. 
+                  The value of <code>shui:hasLanguagePreference</code> is a well-formed SHACL list where all members are language tags. 
+                  The preferred language is determined according to the order of priority in the list from beginning (highest) to end (lowest ).
```
smessie: I find the description of an empty Literal behavior useful to add as well. Also a redundant space.

```suggestion
                  The preferred language is determined according to the order of priority in the list from beginning (highest) to end (lowest). 
                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
```

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+									The language selection for displaying labels in widgets. 
+                  The value of <code>shui:hasLanguagePreference</code> is a well-formed SHACL list where all members are language tags. 
+                  The preferred language is determined according to the order of priority in the list from beginning (highest) to end (lowest ). 
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:timeZone</code></td>
+								<td>
+									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
+                  The value of <code>shui:timeZone</code> is <code>xsd:string</code>.
```
```suggestion
                  The value of <code>shui:timeZone</code> is of datatype <code>xsd:string</code>. 
```

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+								<td><code>shui:timeZone</code></td>
+								<td>
+									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
+                  The value of <code>shui:timeZone</code> is <code>xsd:string</code>. 
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:shapesLabelPreference</code></td>
+								<td>
+									The label properties used to determine labels to display in widgets.
```
```suggestion
									The properties used to determine the labels to display in widgets. 
```

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
+                  The value of <code>shui:timeZone</code> is <code>xsd:string</code>. 
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:shapesLabelPreference</code></td>
+								<td>
+									The label properties used to determine labels to display in widgets. 
+                  Properties are looked up in the <a>shapes graph</a>. 
+                  The value of <code>shui:shapesLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
```
smessie: I guess the order is of relevance here as well?
```suggestion
                  The value of <code>shui:shapesLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
                  A label is determined by evaluating the properties according to the order of priority in the list from beginning (highest) to end (lowest).
```

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:dataLabelPreference</code></td>
+								<td>
+									The label properties used to determine labels to display in widgets and labels to represent nodes values in the widgets. 
+                  Properties are looked up in the <a>data graph</a>. 
+                  The value of <code>shui:dataLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
```
smessie: idem

```suggestion
									The properties used to determine the labels to display in widgets and the labels to represent value nodes in the widgets. 
                  Properties are looked up in the <a>data graph</a>. 
                  The value of <code>shui:dataLabelPreference</code> is a well-formed SHACL list where all members are IRIs.
                  A label is determined by evaluating the properties according to the order of priority in the list from beginning (highest) to end (lowest).
```

shacl12-ui/index.html
```diff
@@ -671,6 +677,76 @@ <h3>Property UI Component</h3>
...
+							</tr>
+						</tbody>
+					</table>
+				<aside class="example" title="Sample Instance Data">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:config a shui:Configuration ;
+    shui:defaultNamespace &lt;http://example.com/ns#&gt; ;
+    shui:hasLanguagePreference ("" "en" "de") ;
+    shui:hasTimeZone "Europe/Vienna" ;
```
```suggestion
    shui:defaultNamespace "http://example.com/ns#" ;
    shui:languagePreference ("" "en" "de") ;
    shui:timeZone "Europe/Vienna" ;
```

shacl12-ui/index.html
```diff
@@ -1376,7 +1452,7 @@ <h3>Language Resolution</h3>
             expressed is up to the implementation. Possible approaches include but are not limited to:
             <ul>
               <li>A UI feature that allows the user to select a preferred display language or languages priorities.</li>
-              <li>A predefined language configuration in the application settings.</li>
+              <li>A <a>global configuration</a> using the property <code>shui:hasLanguagePreference</code>.</li>
```
```suggestion
              <li>A <a>global configuration</a> using the property <code>shui:languagePreference</code>.</li>
```

shacl12-ui/index.html
danielbeeke: I remember vaguely that we also discussed using SHACL property paths here so that we would have support for skosxl:literalForm and others.

shacl12-ui/index.html
```diff
@@ -737,11 +738,10 @@ <h3>SHACL Global Configuration</h3>
 					<div class="shapes-graph">
 						<div class="turtle">
 ex:config a shui:Configuration ;
-    shui:defaultNamespace &lt;http://example.com/ns#&gt; ;
+    shui:defaultNamespace "&lt;"http://example.com/ns#" ;
```
TallTed: I may be misreading, but it seems that this URI markup is broken. I *think* this fixes it.
```suggestion
    shui:defaultNamespace "&lt;http://example.com/ns#&gt;" ;
```

robert-david:
There is still the ex:hasPart property in the example. Should we also rename it to ex:part? @bergos

shacl12-ui/index.html
smessie: The redundant space was not fixed yet.

shacl12-ui/index.html
```diff
@@ -671,6 +685,68 @@ <h3>Property UI Component</h3>
...
+                  The preferred label property is determined according to the order of priority in the list from beginning (highest) to end (lowest).
+								</td>
+							</tr>
+						</tbody>
+					</table>
+				<aside class="example" title="Sample Instance Data">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:config a shui:Configuration ;
+    shui:defaultNamespace "&lt;"http://example.com/ns#" ;
```
smessie: Something is wrong here, I think it should be:
```suggestion
    shui:defaultNamespace "http://example.com/ns#" ;
```

shacl12-ui/index.html
```diff
@@ -433,6 +433,12 @@ <h3>Terminology</h3>
               >value node</a
             >
           </dfn>
+          ,
+          <dfn data-lt="global configuration">
+            <a href="https://www.w3.org/TR/shacl/#global-configuration"
```
afs, line 438: FYI: 
This is a link to SHACL 1.0 which does not have an id for `#global-configuration`.
(all the links in this area are also to SHACL 1.0 but, mostly, the target anchor fragment exists).

shacl12-ui/index.html
```diff
@@ -671,6 +685,69 @@ <h3>Property UI Component</h3>
...
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:languagePreference</code></td>
+								<td>
+									The language selection for displaying labels in widgets. 
+                  The value of <code>shui:languagePreference</code> is a well-formed SHACL list where all members are language tags. 
+                  The preferred language is determined according to the order of priority in the list from beginning (highest) to end (lowest). 
+                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
+                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
```
smessie, line 718:
```suggestion
                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
```
Duplicated sentence.

shacl12-ui/index.html
```diff
@@ -671,6 +685,69 @@ <h3>Property UI Component</h3>
...
+						</tbody>
+					</table>
+				<aside class="example" title="Sample Instance Data">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:config a shui:Configuration ;
+    shui:defaultNamespace "http://example.com/ns#" ;
+    shui:languagePreference ("" "en" "de") ;
+    shui:timeZone "Europe/Vienna" ;
+    shui:labelPreference (skos:prefLabel, dcterms:title, rdfs:label) .
```
smessie, line 745:
```suggestion
    shui:labelPreference (skos:prefLabel dcterms:title rdfs:label) .
```
RDF collections are whitespace-separated, not comma-separated.

shacl12-ui/index.html
smessie:
```suggestion
            <a href="#global-configuration"
```
This comment from Andy was not addressed yet. I believe it should be like this.

shacl12-ui/index.html
```diff
@@ -671,6 +685,69 @@ <h3>Property UI Component</h3>
...
+								<td><code>shui:timeZone</code></td>
+								<td>
+									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
+                  The value of <code>shui:timeZone</code> is of datatype <code>xsd:string</code>. 
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:labelPreference</code></td>
+								<td>
+									The label properties used to determine the labels to display in widgets and labels to represent nodes values in the widgets.
```
smessie, line 731:
```suggestion
									The label properties used to determine the labels to display in widgets and labels to represent node values in the widgets. 
```
Small grammar typo

smessie (commented):
Using property paths for the label preference properties like @danielbeeke commented (https://github.com/w3c/data-shapes/pull/900#discussion_r3354843445) would actually allow more flexibility.

@robert-david I (only) see (now) that you changed in https://github.com/w3c/data-shapes/pull/900/changes/ea14d7ab28fe9e35e009cda2225949c34c8843d1 going from `shui:dataLabelPreference` and `shui:shapesLabelPreference` to the unified `shui:labelPreference`. I think I like how it was first better. I don't immediately see why this change was made, neither in the issue nor in the PR (if I'm not missing something). Can you explain why you did that?

In the issue I see there are still ideas for extra configuration properties such as `shui:inferenceGraph`, but I think they need a bit more discussion. I would be in favor of finishing this PR first, and then handling any other additions in a follow-up PR.

shacl12-ui/index.html
```diff
@@ -671,6 +685,69 @@ <h3>Property UI Component</h3>
...
+                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
+                  An empty literal (`""`) represents no language, i.e., a string literal with no language tag.
+								</td>
+							</tr>
+							<tr>
+								<td><code>shui:timeZone</code></td>
+								<td>
+									The time zone used to construct new terms with <a>datatype</a> <code>xsd:dateTime</code>.
+                  The value of <code>shui:timeZone</code> is of datatype <code>xsd:string</code>. 
+								</td>
```
edmondchuc, line 725: The example below demonstrates using an IANA time zone identifier for `shui:timeZone` values, but it's not explicitly stated here what the lexical value should be.

Should we require the value to be an IANA time zone identifier, or recommend IANA identifiers while leaving the value space open-ended?

robert-david:
> @robert-david I (only) see (now) that you changed in [ea14d7a](https://github.com/w3c/data-shapes/commit/ea14d7ab28fe9e35e009cda2225949c34c8843d1) going from `shui:dataLabelPreference` and `shui:shapesLabelPreference` to the unified `shui:labelPreference`. I think I like how it was first better. I don't immediately see why this change was made, neither in the issue nor in the PR (if I'm not missing something). Can you explain why you did that?
> 
@smessie The merge of shui:dataLabelPreference and shui:shapesLabelPreference into shui:labelPreference was discussed in a UI task force meeting. We did not find a good reason to keep these two configuration properties separated. We can revisit the discussion if you think they should be kept separate.

shacl12-ui/index.html
danielbeeke: I check alternatives like offsets (ISO), but offsets seem to have short comings, in that they do not provide. I either MUST or SHOULD are fine. I think it is helpful to give this guidance else its hard to use it for its actual purpose: "The time zone used to construct new terms with datatype xsd:dateTime."

danielbeeke (changes requested):
Two points (one change requested)

  1. I think we should make it clear what the timezone must be, question Edmund raised. 
  2. And I am wondering how one could have one shapes with a certain `shui:defaultNamespace` and another shape with a different `shui:defaultNamespace`. However from my end this is good to be merged and this addition can be done in a follow up.

bergos (approved):

shacl12-ui/index.html
```diff
@@ -433,6 +433,12 @@ <h3>Terminology</h3>
               >value node</a
             >
           </dfn>
+          ,
+          <dfn data-lt="global configuration">
```
```suggestion
          <dfn>
```
ReSpec generates the plural with appended *s* automatically.

bergos (changes requested):
Ready to merge, if you apply the proposed change for the `dfn` element.

smessie (changes requested):
Thanks for the work on the global configuration, Robert, and thanks for already discussing this with me offline. As we agreed, it would be good to bring the label preference property question to the next meeting before merging.

To summarise the point for everyone: this is about the unification of `shui:dataLabelPreference` and `shui:shapesLabelPreference` into a single `shui:labelPreference` (commit ea14d7a). Looking into the background:

- The decision to use "only one label configuration mechanism" is recorded in the [26 May 2026 UI TF minutes](https://github.com/w3c/data-shapes/tree/agenda/ui-tf/meetings/2026-05-26.md), with Thomas, Robert, and Edmond as the recorded participants.
- The earlier [28 April 2026 minutes](https://github.com/w3c/data-shapes/tree/agenda/ui-tf/meetings/2026-04-28.md) show the group originally agreeing on two separate configuration properties, one for property paths (shapes side) and one for value nodes (data side).
- Daniel explicitly stated on 28 July that he [likes `shui:shapesLabelPreference` and `shui:dataLabelPreference`](https://github.com/w3c/data-shapes/issues/869#issuecomment-5104432656).
- Daniel was not present at the 26 May meeting.
- I was also not present at that meeting.

Since there are members who favor keeping the properties separate, I think it is worth a short discussion in our next meeting so the group can settle this together. There may well be a good reason to unify them, but let us make that call with everyone able to weigh in.

For the rest, the PR looks good to me. This is only about the label preference property split.

robert-david:
@smessie i agree to keep both `shui:dataLabelPreference` and `shui:shapesLabelPreference` if there is an actual need.
However, back then we could not find one. 

Originally the idea was to separate the properties because you might have different label properties for shapes and for data. However, given that the property value of `shui:dataLabelPreference` and `shui:shapesLabelPreference` is a priority list of label properties, you can basically put the label properties used for the shapes and for the data into one list. The shapes and the data will then pick the highest priority match, which does not need to be the same property for both. The only case where we would need to split into `shui:dataLabelPreference` and `shui:shapesLabelPreference` is when the shapes and the data both use multiple label properties, but their priority is in opposite directions. However, i think this is a very rarely occurring case. Therefore we decided to merge `shui:dataLabelPreference` and `shui:shapesLabelPreference` into one `shui:labelPreference` for the sake of simplicity.

smessie (commented):
@robert-david Thanks for the detailed explanation — that resolves it for me. My original concern came exactly from the idea of wanting different label properties in the shapes graph versus the data graph. But as you point out, since the value is an ordered priority list and each graph independently resolves to its own highest-priority property that has an available value, a single `shui:labelPreference` already covers that case: the shapes graph and the data graph can still end up using different properties. The only situation a split would additionally support is when both use multiple label properties with *opposite* priority ordering, which is rare enough that I agree merging into one `shui:labelPreference` is the better trade-off for simplicity.

I've added a change suggestion on the *Label Property Resolution* paragraph to make this explicit (one ordered list, applied independently to both graphs).

@danielbeeke could you confirm you're also okay with keeping the merged `shui:labelPreference` instead of the separate `shui:shapesLabelPreference` / `shui:dataLabelPreference`?

shacl12-ui/index.html
```diff
@@ -1454,6 +1530,14 @@ <h3>Label Resolution</h3>
           </li>
         </ul>
 
+        <section id="label-property-resolution">
+          <h4>Label Property Resolution</h4>
+          <p>
+            Label property resolution (i.e., determining the label properties for retrieving label values) is done according to the preferred property IRIs listed in the <a>global configuration</a> property <code>shui:labelPreference</code>. 
+            Preferred labels default to <code>sh:name</code> for Property Labels and <code>rdfs:label</code> for Value Node Labels.
```
smessie, line 1537: Make explicit that the single `shui:labelPreference` list is applied independently to both graphs (this covers the shapes-vs-data use case that motivated the earlier split).

```suggestion
            Label property resolution (i.e., determining the label properties for retrieving label values) is done according to the ordered list of preferred property IRIs listed in the <a>global configuration</a> property <code>shui:labelPreference</code>. 
            The same list is used regardless of whether the label values are retrieved from the <a>shapes graph</a> or the <a>data graph</a>: in each case the label is taken from the highest-priority property in the list that has an available value, so the <a>shapes graph</a> and the <a>data graph</a> may resolve to different label properties. 
            When no <code>shui:labelPreference</code> is configured, the preferred label property defaults to <code>sh:name</code> for Property Labels and <code>rdfs:label</code> for Value Node Labels.
```

shacl12-ui/index.html
edmondchuc: Perhaps we could recommend IANA time zone identifiers in the same way that DCAT recommends [dataset type classifications](https://www.w3.org/TR/vocab-dcat-3/#classifying-dataset-types).

Best practices change over time, and IANA time zone identifiers may not remain the preferred identifiers in the future.

danielbeeke:
Yes I think shui:labelPreference is fine

shacl12-ui/index.html
smessie: Yes, I believe that a SHOULD for IANA time zone identifiers is an ideal case for this recommendation. We suggest that adopters adopt this format, but we also acknowledge that there may be instances where other identifiers are preferred or necessary, or where best practices evolve over time.

smessie (approved):
I’m already approving this because my requested changes are no longer relevant since I’ve agreed to use the single `shui:labelPreference`.

I still want to see Edmond’s comment addressed, which recommends IANA time zone identifiers; https://github.com/w3c/data-shapes/pull/900#discussion_r3746882865.

shacl12-ui/index.html
robert-david: Agreed to have property paths instead of IRIs (only).

danielbeeke (changes requested):
changes look good, could you also fix the timezone

danielbeeke (approved):
It looks good to me.
Can you also check the other open suggestions? I think they are also good to be added.

bergos (approved):

PR #906, opened 2026-05-28
Prof voc consistency

nicholascar:
A small PR to harmonise the definition of terms in the Profiling doc and the shacl.ttl vocabulary.

The new wording for `sh:conforms` is needed since we are now dependent on an understanding of severity levels for true/false.


HolgerKnublauch (approved):
Approving the core TTL changes. Didn't check the rest.

shacl12-vocabularies/shacl.ttl
```diff
@@ -221,15 +221,15 @@ sh:ValidationReport
 sh:conforms
 	a rdf:Property ;
 	rdfs:label "conforms"@en ;
-	rdfs:comment "True if the validation did not produce any validation results, and false otherwise."@en ;
+	rdfs:comment "True if the validation did not produce any validation results with a severity level of the set of disallowed levels, and false otherwise."@en ;
```
ajnelson-nist, line 224: With a direct mention of disallowed levels, I think it'd be best to see-also the property encoding disallowance.

```suggestion
	rdfs:comment "True if the validation did not produce any validation results with a severity level of the set of disallowed levels, and false otherwise."@en ;
	rdfs:seeAlso sh:conformanceDisallows ;
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -221,15 +221,15 @@ sh:ValidationReport
...
 	rdfs:domain sh:ValidationReport ;
 	rdfs:range xsd:boolean ;
 	rdfs:isDefinedBy sh: .
 
 sh:conformsTo
 	a rdf:Property ;
 	rdfs:label "conforms to"@en ;
-	rdfs:comment "The subject is valid according to Shapes within the object."@en ;
+	rdfs:comment "The subject conforms to the object if validation did not produce any validation results with a severity level of the set of disallowed levels."@en ;
```
ajnelson-nist, line 233: See-also'ing for the same reason as on `sh:conforms`.

```suggestion
	rdfs:comment "The subject conforms to the object if validation did not produce any validation results with a severity level of the set of disallowed levels."@en ;
	rdfs:seeAlso sh:conformanceDisallows ;
```

ajnelson-nist (approved):
Approving, with two non-blocking suggestions provided.

PR #911, opened 2026-05-30
Fix example typing for prof:hasArtifact

ajnelson-nist:
`prof:hasArtifact` is an `owl:ObjectProperty`.  Range unspecified.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/prof_fix_hasArtifact_typing/shacl12-profiling/index.html)

### DIFF
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -788,7 +788,7 @@ ex:ont-N
   prof:hasRole role:Specification ;
   dcterms:format "text/html" ;
   dcterms:conformsTo &lt;http://www.w3.org/TR/html5/> ;
-  prof:hasArtifact "https://www.w3.org/TR/shacl12-core/"^^xsd:anyURI ;
+  prof:hasArtifact &lt;https://www.w3.org/TR/shacl12-core/> ;
 .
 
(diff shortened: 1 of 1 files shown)


nicholascar:
@ajnelson-nist I'm not 100% sure about this one! It's true that PROF says `hasArtifact` is an `owl:ObjectProperty` but it's been used (by Rob a lot and sometimes by me) to indicate a literal (usually an `xds:anyURI`) to step out of the RDF graph and on to a non-semantic object, like a written document, digital image etc. 

This is sometimes necessary for profiles that include such resources, probably a less necessary for pure SHACL things, but the example you are fixing here are cases when the user is being told there is a non-semantic thing there in the end - the HTML doc at <https://www.w3.org/TR/shacl12-core/>. Perhaps the one indicating <http://www.w3.org/ns/shacl-shacl/core> need not even use `hasArtifact` in which case a user could assume that they should just resolve the Resource IRI as a semantic object (an RDF file).

So I think we need PROF to sort this out for us before locking it down here. I've created a ticket for this: https://github.com/w3c/dx-prof/issues/63

nicholascar (approved):

PR #914, opened 2026-05-31
fix: added missing rdfs prefix to targetNode-select-001 test

### LINKED ISSUE #913: Missing rdfs: prefix declaration in targetNode-select-001.ttl test
The test file [targetNode-select-001.ttl](https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-test-suite/tests/sparql/targets/targetNode-select-001.ttl) uses the `rdfs:` prefix in a SPARQL SELECT node expression string but does not declare it. The `rdfs:` prefix declaration should be added to the file.

bergos:
Closes #913


HolgerKnublauch (approved):
Good catch.

Issue #690, opened 2025-12-03
Add well-formness conditions relating to `NOT`


afs:
A variables that does not appear to the left of `NOT` (i.e first appears in `NOT`) should not be allowed to the right of `NOT`.

Example: `?z` in:
```
RULE {} WHERE { 
        :s :p ?x 
        NOT { ?x :data ?z } 
        :y :data ?z }
    }
```

This isn't actually an error - `NOT`  does not give a value to a variable that is available after the `NOT` is executed - it behanves like a condition (= filter)

We could consider rejecting this (`?z` after 'NOT`) on clarity grounds.

afs:
The well-formedness conditions also need to be revised to make FILTER expressions inside NOT have the equivalent conditions as they have when used in outer body.

PR #916, opened 2026-06-01
Improvements from rules document review

afs:
Improvements from rules document review

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/rules-doc-review/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -206,13 +206,11 @@
         RDF triples derived by rule execution.
         Combining the inference graph with the base graph is optional and
         left to users. The <em>query</em> operation determines whether a
-        given goal pattern can be derived from the base graph using the rules
+        given goal pattern can be derived from the base graph using the rules.
       </p>
       <p>
-        The inference graph may contain RDF triples with IRIs, blank nodes or
-        literals that were not present in the base graph. Users are responsible to
-        ensure that iterative applications of rules that generate new blank nodes or
-        literals do not result in infinite loops.
+        SHACL Rules allow the use of new RDF terms, including blank nodes,
+        that can be used in triple templates in the head of rules.
       </p>
       <p>
         SHACL Rules also support constructs, such as negation as failure, 
@@ -231,7 +229,7 @@
         </p>
       </section>
 
-      <section class="conventions">
+      <section id="conventions">
         <h3>Document Conventions</h3>
         <p>
           Some examples in this document use Turtle [[turtle]].
@@ -334,8 +332,8 @@
       </p>
 
       <p class="note">
-        This specification does not define how SHACL Rules processors
-        handle non-conforming input documents.
+        This specification does not define how SHACL Rules
+        processors handle non-conforming input documents.
       </p>
 
       <section id="defined-version-labels">
@@ -363,7 +361,6 @@
           For serializations supporting in-line version announcement,
           the version announcement SHOULD be made early in the document.
         </p>
-
       </section>
 
     </section>
@@ -392,14 +389,22 @@
       </p>
       <p>
         SHACL Rules execution is defined so that the order of rule execution
-        does not lead to to different outcomes when creating new RDF terms,
-        including new blank nodes, nor in the testing for the absence of a
+        does not lead to different outcomes when creating new RDF terms,
+        including new blank nodes, nor when testing for the absence of a
         pattern.  In other words, the same inference graph is produced
         regardless of the order of rule execution.
       </p>
 
+      <p>
+        SHACL Rules has both a RDF syntax, as well as a human-friendly syntax, 
+        inspired by [[[SPARQL12-QUERY]]].
+        Rule set evaluation contains elements similar to SPARQL, with differences
+        in the details to ensure that the same inference graph is produced
+        regardless of the order of rule execution.
+      </p>
+
       <section id="desc-basic">
-        <h3>Basic Usage</h3>
+        <h3>Basic Patterns</h3>
 
         <p>In this first example, we have the following data graph and rule set:</p>
 
@@ -532,8 +537,8 @@ RULE { ?x rdf:type :largeTown } WHERE { ?x :population ?p . FILTER(?p > 1500) }
           ensures that all the rules that could produce triples matching the pattern
           in the negation element have been completed.  This is called
           <i>[=stratification=]</i> and ensures that the negation is based
-          on all the relevant possible triples, whether from the data or inferred
-          from other rules.
+          on all the relevant possible triples, whether from the data or 
+          from inferred triples inferred by other rules.
         </p>
         <div class="data">
 <pre>
@@ -592,16 +597,21 @@ WHERE {
 </pre>    
         </div>
         <p>
-          Rules involving assignment are [=run-once rules=]. Rules with assignment
-          are run after all the rules that could produce the data that they depend on
+          Rules involving [=assignments=] and rules that create blank nodes in 
+          their [=rule head=] are [=run-once rules=].
+          Such rules are run after all the rules that could produce data that they depend on,
           and before any rules that depend on the data they produce. 
           Rules that involve blank node in the [=rule head=] are also creating new
           RDF terms and are run-once rules.
         </p>
         <p>
           This condition ensures that rules that create RDF terms do not generate new terms
-          multiple times, with potentially different outcomes and also that such rules do not 
+          multiple times, with potentially different outcomes, and also that such rules do not 
           loop back to themselves and cause an unbounded number of RDF terms.
+        </p>
+        <p>
+          If evaluating the expression in an [=assignment=] causes an error,
+          then the current solution mapping is rejected by the [=assignment=].
         </p>
       </section>
       
@@ -644,6 +654,25 @@ WHERE {
         </ul> 
       </section>
 
+      <section id="data-blocks">
+        <h3>Data Blocks</h3>
+        <p>
+          Data blocks allow concisely providing RDF triples directly to the rule set evaluation.
+          Triples in datablocks are added to the inference graph and are available for
+          matching in the body of rules.
+        </p>
+        <p>@@examples to be updated</p>
+
+          <div class="rules">
+<pre>
+DATA {
+  :father rdfs:subClassOf :familyRelationship .
+  :mother rdfs:subClassOf :familyRelationship .
+}
+</pre>
+        </p>
+      </section>
+
       <section id="rule-tuples">
         <h3>Rule Tuples</h3>
         <div class="issue" data-number="752">
@@ -683,7 +712,7 @@ WHERE {
         It is used to define the execution algorithm of SHACL Rules.
         Each of the two concrete syntax forms of SHACL Rules, the SHACL Rules Language (SRL)
         and the RDF syntax (SRL/RDF), provides a way to express the abstract syntax.
-      /p>
+      </p>
 
       <section id="rules-defns">
         <h3>Elements of the Abstract Syntax</h3>
@@ -700,7 +729,7 @@ WHERE {
             <p>
               An [=expression=] is a function or a functional form;
               the arguments are [=RDF terms=].
-              An expression is evaluated with respect to a [=solution mapping=] to give
+              An expression is evaluated with respect to a [=solution mapping=], giving
               an [=RDF term=] as the result.
               <a>Expressions</a> are compatible with
               <a data-cite="shacl12-node-expr#dfn-list-parameter-function"
@@ -812,8 +841,14 @@ WHERE {
 
           <dt><dfn>Run-once rule</dfn></dt>
           <dd>
-            A [=run-once rule=] is a rule that is run once at a particular point
-            in the evaluation of a rule set.
+            A [=run-once rule=] is a rule that is run exactly once at a
+            particular point in the evaluation of a rule set.
+          </dd>
+
+          <dt><dfn>General rule</dfn></dt>
+          <dd>
+            A [=general rule=] is a rule that is not a [=run-once rule=].
+            General rules may run more than once during rule set evaluation.
           </dd>
 
           <dt><dfn>Rule set</dfn></dt>
@@ -882,14 +917,25 @@ WHERE {
             of the [=head=] of the [=rule=],
             there is one or more occurrences of a [=variable=]
             of the same name in the [=triple patterns=] in the [=body=],
-            in an [=assignment element=] in the body, or both.
+            or in an [=assignment element=] in the body, or in both.
           </li>
           <li>
             For every [=variable=] in an [=expression=] at position <em>i</em>
             of the [=body=], there is a corresponding [=variable=] of the same
-            name occurring in a [=triple pattern=] at a position <em>j</em>,
-            or occurring as an [=assignment variable=] at a position <em>j</em>,
+            name occurring in a [=triple pattern=] at some position <em>j</em>,
+            or there is an [=assignment variable=] at some position <em>j</em>,
             where <em>i &gt; j</em>.
+          </li>
+          <li>
+            For every [=variable=] in an expression in a [=negation element=] 
+            at position <em>i</em> of the [=body=], either 
+            there is a corresponding [=variable=] of the same
+            name occurring in a [=triple pattern=] 
+            or an [=assignment variable=] at some position <em>j</em>,
+            where <em>i &gt; j</em>,
+            or a  [=variable=] of the same name occurring in some [=triple pattern=] 
+            or [=assignment variable=] of the [=negation element=] 
+            preceeding the expression of the [=negation element=].
           </li>
           <li>
             Each [=assignment variable=] is used in only one [=assignment element=] in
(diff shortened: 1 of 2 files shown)


shacl12-rules/index.html
```suggestion
        let rules = stratumRules.get(i)
```

shouldn't this be stratumRules?

```suggestion
                      <var>μ</var><sub>3</sub> = merge(<var>μ</var><sub>1</sub>, <var>μ</var><sub>2</sub>)
```
broken subscript

shacl12-rules/index.html
```diff
@@ -1859,6 +1900,14 @@ <h3>Evaluation of a Rule</h3>
 
       <section id="eval-rule-set">
         <h3>Evaluation of a Rule Set</h3>
+        <p>
+          Evaluation of a [=rule set=] is defined as the execution of each [=stratum=] of the
+          [=stratification=] of the rule set, where each stratum is executed completely
+          and in order befoer moving onm to the next [=stratum=].
+          A [=stratum=] is evaluated by first evaluating
+          each of the [=run-once rule=] of that stratum, and then evaluating
+          [=general rules=] of thestratum repeatedly until no new triples are produced.
```
simonstey: we do not have a corresponding definition (i.e. <dfn>) for general rules yet

shacl12-rules/index.html
```suggestion
            of `R1` that depend on `R2` occur only as [=triple pattern elements=]
```

shacl12-rules/index.html
```diff
@@ -612,6 +621,25 @@ <h3>Relationship between SHACL Rules and SPARQL</h3>
...
+          Triples in datablocks are added to the inference graph and are available for
+          matching in the body of rules.
+
+          <p>@@examples to be updated</p>
+
+          <div class="rules">
+<pre>
+DATA {
+  :father rdfs:subClassOf:  :familyRelationship .
+  :monther rdfs:subClassOf: :familyRelationship .
```
```suggestion
  :father rdfs:subClassOf  :familyRelationship .
  :mother rdfs:subClassOf :familyRelationship .
```

shacl12-rules/index.html
```diff
@@ -209,10 +209,8 @@ <h2>Introduction</h2>
         given goal pattern can be derived from the base graph using the rules
```
```suggestion
        given goal pattern can be derived from the base graph using the rules.
```

shacl12-rules/index.html
```suggestion
      </p>
```

```suggestion
      <section id="conventions" class="conventions">
```

shacl12-rules/index.html
```diff
@@ -612,6 +621,25 @@ <h3>Relationship between SHACL Rules and SPARQL</h3>
...
       </section>
 
+      <section id="data-blocks">
+        <h3>Data Blocks</h3>
+        <p>
+          Data blocks allow concisely providing RDF triples directly to the rule set evaluation.
+          Triples in datablocks are added to the inference graph and are available for
+          matching in the body of rules.
+
+          <p>@@examples to be updated</p>
```
```suggestion
          <pre>@@examples to be updated</pre>
```

as (afaik) nested <p> aren't valid HTML

shacl12-rules/index.html
```suggestion
<pre class="example-rules">
```
I guess that was meant to be class example-rules

shacl12-rules/index.html
```diff
@@ -1739,12 +1767,18 @@ <h4>Calculate Stratification</h4>
 
       <section id="eval-expression">
         <h3>Evaluation of an Expression</h3>
-        <p class="ednote">Sketch</p>
+        <p>
+          An expression, whether used in a [=condition element=] or 
+          an [=assignment element=], is evaluated with
+          respect to a solutiuon mapping which gives a value which is an RDF term,
```
```suggestion
          respect to a solution mapping which gives a value which is an RDF term,
```

shacl12-rules/index.html
```diff
@@ -1859,6 +1900,14 @@ <h3>Evaluation of a Rule</h3>
 
       <section id="eval-rule-set">
         <h3>Evaluation of a Rule Set</h3>
+        <p>
+          Evaluation of a [=rule set=] is defined as the execution of each [=stratum=] of the
+          [=stratification=] of the rule set, where each stratum is executed completely
+          and in order befoer moving onm to the next [=stratum=].
```
```suggestion
          and in order before moving on to the next [=stratum=].
```

shacl12-rules/index.html
simonstey:
```suggestion
          [=general rules=] of the stratum repeatedly until no new triples are produced.
```
also

shacl12-rules/index.html
simonstey, line 1213:
```suggestion
            @@ Examples of triple pattern dependencies.
```

shacl12-rules/index.html
```diff
@@ -1739,12 +1767,18 @@ <h4>Calculate Stratification</h4>
...
       <section id="eval-expression">
         <h3>Evaluation of an Expression</h3>
-        <p class="ednote">Sketch</p>
+        <p>
+          An expression, whether used in a [=condition element=] or 
+          an [=assignment element=], is evaluated with
+          respect to a solutiuon mapping which gives a value which is an RDF term,
+          for each variable in the expression. 
+          The well-formedness requirements of <a href="#wellformed" class="sectionRef"></a>
+          ensure that that all variables in the expression
```
```suggestion
          ensure that all variables in the expression
```

shacl12-rules/index.html
```diff
@@ -1739,12 +1767,18 @@ <h4>Calculate Stratification</h4>
 
       <section id="eval-expression">
         <h3>Evaluation of an Expression</h3>
-        <p class="ednote">Sketch</p>
+        <p>
+          An expression, whether used in a [=condition element=] or 
+          an [=assignment element=], is evaluated with
+          respect to a solutiuon mapping which gives a value which is an RDF term,
+          for each variable in the expression.
```
simonstey: alternative phrasing suggestion:
```suggestion
          respect to a solution mapping that provides an RDF term value for each variable in the expression.
```

shacl12-rules/index.html
```suggestion
            can possibly generate a triple with component RDF terms
```

```
This happens when `R1` has a [=triple pattern=] in a [=negation element=]
that matches a [=triple template=] of `R2` in a [=negation element=],
```
 does the "in a negation element" qualifier at the end also belong to R1's pattern or does it belong to R2's template? because in the algorithm later on _any_ TP inside R1's negation element is closed.

simonstey:
```suggestion
This happens when `R1` has a [=triple pattern=], occurring inside a in a [=negation element=],
that matches a [=triple template=] of `R2`.
```

shacl12-rules/index.html
```diff
@@ -1739,12 +1767,18 @@ <h4>Calculate Stratification</h4>
...
+          The well-formedness requirements of <a href="#wellformed" class="sectionRef"></a>
+          ensure that that all variables in the expression
+          appear in the solution mapping.
+        </p>
+
         <div class="algorithm">
 <pre>
-@@ Reference SPARQL expression evaluation <a data-cite="SPARQL12-QUERY/#expression-evaluation">Expression Evaluation</a>
-@@ Reference SPARQL EBV <a data-cite="SPARQL12-QUERY/#ebv">Effective Boolean Value (EBV)</a>
-
```
simonstey, line 1779: we don't want to reference those two?

shacl12-rules/index.html
```suggestion
RULE { ?x :q ?o } WHERE { ?x :p :o . SET ( ?o := 18 ) }
```

everywhere else we use SET now. BIND is also not defined in our grammar, but SET is:
```
[27] Assignment                ::= 'SET' '(' Var ':=' Expression ')'
```

simonstey: this is inconsistent with the grammar def.:

```
Declaration               ::= ( 'TRANSITIVE' '(' iri ')' | 'SYMMETRIC' '(' iri ')' | 'INVERSE' '(' iri ',' iri ')' )
```

simonstey: other examples use `srl:filter`. `srl:expr` is **not** declared in the syntax vocab only filter ->

```ttl
srl:filter rdf:type rdf:Property ;
    rdfs:isDefinedBy srl: ;
    rdfs:isDefinedBy <https://www.w3.org/TR/shacl12-rules/#rdf-rules-syntax> ;
    rdfs:label "filter" ;
    rdfs:comment "A filter rule element" .
```

simonstey:
```suggestion
[ srl:filter [ sparql:greater-than ( [ srl:varName "v1" ] 0 ) ] ]
[ srl:subject [ srl:varName "x" ] ; srl:predicate :q ; srl:object [ srl:varName "v2" ] ]
[ srl:filter [ sparql:greater-than ( [ srl:varName "v2" ] 0 ) ] ]
```
also, greatherThan -> greater-than

```ttl
sparql:greater-than
	a sh:ListParameterExpressionFunction ;
	rdfs:label "greater than"@en ;
	rdfs:comment "This operator tests whether the first RDF term is greater than the second RDF term."@en ;
	rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
	rdfs:subClassOf sh:ListParameterExpression ;
	shnex-sparql:category "other" ;
	shnex-sparql:infixOperator ">" ;
	shnex-sparql:example """[ sparql:greater-than ( 10 5 ) ]""" ;
```

```suggestion
<p>The Internet Media Type (formerly known as MIME Type) for @@ is "<code>application/shape-rules</code>".</p>

<p>
The information that follows has been submitted to the Internet Engineering
Steering Group (IESG) for review, approval, and registration with IANA.
</p>
<div class="mediatype"><div id="mime"></div>
<dl>

<dt>Type name:</dt>
<dd>application</dd>

<dt>Subtype name:</dt>
<dd>shape-rules</dd>
```

simonstey: isn't that subsumed by 
>     For every [=variable=] in an [=expression=] at position <em>i</em>
>     of the [=body=], there is a corresponding [=variable=] of the same
>     name occurring in a [=triple pattern=] at a position <em>j</em>,
>     or occurring as an [=assignment variable=] at a position <em>j</em>,
>     where <em>i &gt; j</em>.

since `assignment expression \subset expression` ?

```suggestion
RULE { ?x :q :o } WHERE { ?x :p ?o . FILTER (?o &lt; 18) }
```

```suggestion
RULE { ?x :q ?o } WHERE { ?x :p :o . NOT { ?s :p ?o . FILTER(?o &lt; 18) } }
```

simonstey, line 2456:
```suggestion
          <dd>Data Shapes Working Group &lt;public-shacl@w3.org&gt;</dd>
```

```suggestion
                    μ(v) = merge(<var>μ</var><sub>1</sub>, <var>μ</var><sub>2</sub>)}</pre>
```

add missing `}`

shacl12-rules/index.html
```diff
@@ -560,15 +569,15 @@ <h3>Assignment and Creating RDF Terms</h3>
 </pre>    
         </div>
         <p>
-          Rules involving assignment are [=run-once rules=]. Rules with assignment
-          are run after all the rules that could produce the data that they depend on
+          Rules involving [assignments] and rules that create blank nodes in their [=rule head=] are [=run-once rules=].
```
```suggestion
          Rules involving [=assignment elements=] and rules that create blank nodes in their [=rule head=] are [=run-once rules=].
```

shacl12-rules/index.html
```diff
@@ -1859,6 +1900,14 @@ <h3>Evaluation of a Rule</h3>
 
       <section id="eval-rule-set">
         <h3>Evaluation of a Rule Set</h3>
+        <p>
+          Evaluation of a [=rule set=] is defined as the execution of each [=stratum=] of the
+          [=stratification=] of the rule set, where each stratum is executed completely
+          and in order befoer moving onm to the next [=stratum=].
+          A [=stratum=] is evaluated by first evaluating
+          each of the [=run-once rule=] of that stratum, and then evaluating
```
```suggestion
          each of the [=run-once rules=] of that stratum, and then evaluating
```

shacl12-rules/index.html
simonstey, line 325:
```suggestion
          data-lt="shapes rules language document|shapes rules document|ruleset document|shapes rules language documents|shapes rules documents|ruleset documents"
```
added plural forms

simonstey, line 1953: the latter algo uses a bare `S` for a single solution, while `S` is also used as a one-element sequence in the negation branch => we should align those two

simonstey: earlier we state:
```
let rule R = (H, B) where
             H is the sequence of triple templates in the head
             B is the sequence of triple pattern elements,
                condition elements, negation elements,
                and assignment elements in the body
```
which binds `H` to the head templates. 

But here:
```
# Evaluate rule head
let H = empty set
for each <var>μ</var> in SEQ:
```
we then reuse `H` as the output-triple accumulator.

and here: ` for each triple template TT in head` we refer to "head" (which is in theory not bound anymore)

suggestion:

```suggestion
# Evaluate rule head
let HOut = empty set
for each <var>μ</var> in SEQ:
    for each triple template TT in H:
        add subst(<var>μ</var>, TT) to HOut
result eval(R, G) is HOut
```

simonstey (commented):
also, the namespace of shnex should be `http://www.w3.org/ns/shacl-node-expr#` (at least that's what the nodeexpr test cases also use.

(needs fix to be fixed in `rdf-syntax-shapes.ttl` )

------
We use: 

- "SHACL Rules Language (SRL)" (line 182, line 680), 
- "Shape Rules Language (SRL)" (1317, 1322, 1388), 
- "Shapes Rules Language" (1957), 
- ("Shape Rules Abstract Syntax" (675)) 

what's the canonical _final_ expansion of SRL?

afs:
> also, the namespace of shnex should be `http://www.w3.org/ns/shacl-node-expr#` (at least that's what the nodeexpr test cases also use.

Where is it wrong in the spec? I could only find it in the "Document Conventions" and it is `shacl-node-expr#`.

BTW:

https://github.com/w3c/data-shapes/issues/496#issuecomment-3175295310
says it's `http://www.w3.org/ns/shnex#`

> (needs fix to be fixed in `rdf-syntax-shapes.ttl` )

Added to the changes.

shacl12-rules/index.html
afs: Yes

shacl12-rules/index.html
afs: Done locally because of other review comments/editing.

shacl12-rules/index.html
afs: Section reworked.

shacl12-rules/index.html

shacl12-rules/index.html

shacl12-rules/index.html

afs: s/class=/id=/

shacl12-rules/index.html

shacl12-rules/index.html

shacl12-rules/index.html
afs: They are now used in the text.
A styling change is needed to allow links in what are now `<pre>` blocks.

shacl12-rules/index.html
afs: Fixed

afs: See #918

afs: #918

afs: reSpec does plurals translation.

Is it not it working somewhere?

afs: Yes.

Also - it needs to talk about inside negation elements.

Short term fix applied - need to come back to this section after the document review.

afs: Done.

afs: Done, calling it "OUT".

afs:
This PR could run for a very long time, which isn't helpful.

It can be merged now with two work items to deal with as separate issues:

#930 - running example for section 3 - the description section
#690 - revise the well-formedness conditions for `NOT`

afs:
> We use:
> 
> * "SHACL Rules Language (SRL)" (line 182, line 680),
> * "Shape Rules Language (SRL)" (1317, 1322, 1388),
> * "Shapes Rules Language" (1957),
> * ("Shape Rules Abstract Syntax" (675))
> 
> what's the canonical _final_ expansion of SRL?

My preference (currently!) is:

The document is "SHACL 1.2 Rules".

The language is "Shapes Rules Language" - The "Constraint" part of SHACL would be misleading.
"Shape Rules Language" would be OK but does not read so well IMO.

We probably should use "SRL" more often.
Or just "SRL" and not say what "S" is!

The two syntaxes are "SRL/text" and "SRL/RDF".

simonstey:
> We probably should use "SRL" more often. Or just "SRL" and not say what "S" is!

SRL => SRL Rules Language? ;)

simonstey (approved):

recalcitrantsupplant (approved):

shacl12-rules/index.html
```diff
@@ -206,13 +206,11 @@ <h2>Introduction</h2>
...
         left to users. The <em>query</em> operation determines whether a
-        given goal pattern can be derived from the base graph using the rules
+        given goal pattern can be derived from the base graph using the rules.
       </p>
       <p>
-        The inference graph may contain RDF triples with IRIs, blank nodes or
-        literals that were not present in the base graph. Users are responsible to
-        ensure that iterative applications of rules that generate new blank nodes or
-        literals do not result in infinite loops.
+        SHACL Rules allows the use of new RDF terms, including blank nodes,
```
```suggestion
        SHACL Rules allow the use of new RDF terms, including blank nodes,
```

Issue #929, opened 2026-06-08
sparql:greater-than or sparql:greaterThan ?


HolgerKnublauch:
There seems to be an inconsistency with some SPARQL function URIs:

- sparql:greaterThan: https://github.com/w3c/data-shapes/blob/73e66432074cd7f5cfeb64278bd84a13e1edbfc2/shacl12-rules/index.html#L1338 (also apparently implemented in Jena)
- sparql:greater-than: https://www.w3.org/TR/shacl12-node-expr/#sparql-greater-than

So which one is it? (The same pattern looks rather inconsistent with other local names).

afs:
> (also apparently implemented in Jena)

Where?

HolgerKnublauch:
Here? https://github.com/apache/jena/blob/b9e9874f1a218602ad6096bc6ffcaa43921e0c31/jena-arq/src/main/java/org/apache/jena/sparql/expr/urifunctions/SPARQLDispatch.java#L181

afs:
The primary definitions are https://github.com/w3c/sparql-query/blob/main/sparql-ns.ttl -- `sparql:greater-than`

The rules doc has this changed -  changed in PR #916 - and there isn't `srl:expr`.

The Jena code is out-of-date -- it due for replaced with a different, better implementation.

There is a weak logic to the naming.
Named functions keep their name, operators have lowercase/minus names (F&O style)
If you don't like that, raise an issue on sparql-query.

(shacl12-node-expr does not have links for operators - link to the dispatch table)

HolgerKnublauch:
@afs in case it helps, further mismatches in Jena version 5.6.0:
- public static NodeValue sparql_lang(NodeValue nv)       { return NodeFunctions.str(nv); }  // should be lang
- public static NodeValue sparql_bnode() { return null; }  // should return a node
- public static NodeValue sparql_md5(NodeValue nv)    { return NodeValueDigest.calculateDigest(nv, "MD-5"); } // "md5"
- sparql:hasLang "haslang"   (WHAT SHOULD IT BE?)
- sparql:hasLangdir "haslangdir"   (WHAT SHOULD IT BE?)
- sparql:less-than "lessThan"
- sparql:less-than-or-equal "lessThanOrEqual"
- sparql:logical-and "and"
- sparql:logical-not "not"
- sparql:logical-or "or"

afs:
All the code is going to be replaced. I'll check the new one - which is not in the Jena code base.

[...]

PR #935, opened 2026-06-09
#929: Added SHNEX SPARQL Test Cases

HolgerKnublauch:
We didn't have test cases for the SHNEX SPARQL functions yet, here is a start. We may end up renaming one or two, there are ongoing discussions.

Also updated the TTL file's examples. Noteworthy is the change in syntax from << ... >> to <<( ... )>>


TallTed:
Please give this PR a better, more informative title than `#929`.

nicholascar (approved):
Approved to replace PR #970

PR #937, opened 2026-06-09
GH#936: Add TripleTermExpressions

### LINKED ISSUE #936: Need Node Expression type for Constant Triple Terms
Current Node Expression spec includes URIs and Literals as constants

https://www.w3.org/TR/shacl12-core/#node-expressions
https://w3c.github.io/data-shapes/shacl12-node-expr/#ConstantNodeExpression

But we also need to support RDF 1.2 triple terms such as <<( ex:S ex:P ex:O )>> because they may be legally used as input to other node expressions such as sparql:subject ( <<( ex:S ex:P ex:O )>> )

HolgerKnublauch:
Closes #936

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-936/shacl12-node-expr/index.html#TripleTermExpression)

### DIFF
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -759,7 +759,7 @@ ex:Company-employeeCount
 			</p>
 			<p>
 				The most basic node expression functions are <a>constant node expressions</a>, which are either
-				<a>literals</a>, <a>IRIs</a> or <a>triple terms</a>, and simply evaluate to these constants.
+				<a>literals</a>, <a>IRIs</a>, or <a>triple terms</a>, and simply evaluate to these constants.
 				All other node expressions are represented by <a>blank nodes</a>, and come in the following two variations:
 			</p>
 			<ul>
--- a/shacl12-node-expr/index.html
+++ b/shacl12-node-expr/index.html
@@ -759,7 +759,7 @@ ex:Company-employeeCount
 			</p>
 			<p>
 				The most basic node expression functions are <a>constant node expressions</a>, which are either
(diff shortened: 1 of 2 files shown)


shacl12-node-expr/index.html
```diff
@@ -759,7 +759,7 @@ <h2>Node Expression Syntax</h2>
 			</p>
 			<p>
 				The most basic node expression functions are <a>constant node expressions</a>, which are either
-				<a>literals</a> or <a>IRIs</a>, and simply evaluate to these constants.
+				<a>literals</a>, <a>IRIs</a> or <a>triple terms</a>, and simply evaluate to these constants.
```
TallTed: Omitting the terminal comma can lead to misunderstandings. Including the terminal comma all but never leads to misunderstandings. Thus, my recommendation:
```suggestion
				<a>literals</a>, <a>IRIs</a>, or <a>triple terms</a>, and simply evaluate to these constants.
```

HolgerKnublauch: There was no comma before (when it was just literals and IRIs). So what is the rule here? Use commas only for 3 or more items?

afs: [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma).

It's a convention, not a requirement.
Beware that it can change the meaning, as can its omission. Use as needed.

"There is no universally accepted standard for its use."

HolgerKnublauch:
Kind reminder that this PR has been open for 3 weeks and still needs a review. @simonstey @recalcitrantsupplant @robert-david

recalcitrantsupplant (approved):
looks good to me

PR #941, opened 2026-06-14
Rework well-formedness (rules)

afs:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html)

The V1 way wasn't very good at expressing the conditions on negation.

The idea is that well-formedness is defined on a sequence of rule elements and an initial set of active variables a rule is a body is sequence of rule elements where the initial set of active variables is empty.

A negation is a sequence of rule elements, with restrictions on the rule elements allowed, where the initial set of active  variables is the accumulated of the body-level rule elements preceding the negation element.

Some definition changes - there is is no "rule body element" anymore, "condition" goes away and "condition element" becomes "filter element".

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -161,6 +161,10 @@
       .term2escape-yes { background-color: lightgreen; }
       .darkmode .term2escape-yes { background-color: darkgreen; }
 
+      <!-- lists -->
+      .list-indent     { margin-left: 5px; }
+      .list-no-bullet  { list-style: none; }
+      .list-inner      { list-style: disc; }
       
     </style>
 
@@ -277,37 +281,37 @@
           bindings given above.
         </p>
         <!--
-        <pre class="example-shapes">        # This box represents a shapes graph</pre>
+            <pre class="example-shapes">        # This box represents a shapes graph</pre>
 
-        <pre class="example-data">        # This box represents a data graph.</pre>
+            <pre class="example-data">        # This box represents a data graph.</pre>
 
-        <pre class="example-results">        # This box represents an output results graph</pre>
-        -->
+            <pre class="example-results">        # This box represents an output results graph</pre>
+            -->
 
-        <!--
-        <p>Formal definitions appear in blue boxes:</p>
-        <div class="def def-sparql">
-          <div class="def-header">TEXTUAL DEFINITIONS</div>
-          <pre class="def-sparql-body">          # This box contains textual definitions. </pre>
-        </div>
-        -->
+            <!--
+                <p>Formal definitions appear in blue boxes:</p>
+                <div class="def def-sparql">
+                  <div class="def-header">TEXTUAL DEFINITIONS</div>
+                  <pre class="def-sparql-body">          # This box contains textual definitions. </pre>
+                </div>
+                -->
 
-        <!--
-        <p class="syntax">Grey boxes such as this include syntax rules that apply to the shapes graph.</p>
-        -->
+                <!--
+                    <p class="syntax">Grey boxes such as this include syntax rules that apply to the shapes graph.</p>
+                    -->
 
-        <!--
-        <p>
-          <code>true</code>
-          denotes the RDF term
-          <code>"true"^^xsd:boolean</code>
-          .
-          <code>false</code>
-          denotes the RDF term
-          <code>"false"^^xsd:boolean</code>
-          .
-        </p>
-        -->
+                    <!--
+                        <p>
+                          <code>true</code>
+                          denotes the RDF term
+                          <code>"true"^^xsd:boolean</code>
+                          .
+                          <code>false</code>
+                          denotes the RDF term
+                          <code>"false"^^xsd:boolean</code>
+                          .
+                        </p>
+                        -->
       </section>
     </section>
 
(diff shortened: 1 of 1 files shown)


afs:
Only the new version is now in the PR document.

liviorobaldo:
In V2, these three items:

- If elti is a [triple pattern element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-triple-pattern-element), then varsi is the set of [variables](https://www.w3.org/TR/sparql12-query/#defn_QueryVariable) occurring in the triple pattern element.
- If elti is an [assignment element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-assignment-element), then varsi is the set of the [assignment variable](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-assignment-variable).
- Otherwise, varsi is the empty set.

are repeated twice?

Also, I am quite confused about the difference/relation between a [condition element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition-element) and an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression).

V2 states:

If elti is a [condition element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition-element),
- every [variable](https://www.w3.org/TR/sparql12-query/#defn_QueryVariable) mentioned in the [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) of the condition element is an element of Vi-1.

However, I do not understand what “expression of the condition element” means.

The definitions say:

- A [condition element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition-element) is a [condition](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition)...
- A [condition](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition) is an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression)...

So a condition element is itself an expression. In that case, what does “expression of the condition element” refer to?

I think these definitions should be clearer. For instance:

[Expressions](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) are described as being compatible with [SHACL list parameter functions](https://www.w3.org/TR/shacl12-node-expr/#dfn-list-parameter-function) and with [SPARQL expressions](https://www.w3.org/TR/sparql12-query/#expressions).

What does “compatible with” mean in this context? Does this mean that expressions are either SHACL list parameter functions (e.g., `sparql:plus(38 4)`) or SPARQL expressions (e.g., `FILTER` expressions)? If so, it might be clearer to state directly that expressions form the union of these two categories.

Finally, is it necessary to distinguish between condition, condition element, and SPARQL expressions (essentially `FILTER` expressions)? At the moment, these distinctions feel unclear and potentially confusing.

Overall, the terminology appears difficult to follow in its current form... do we truly need all these definitions, can't some of them be collapses into one?

afs:
> Also, I am quite confused about the difference/relation between a [condition element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition-element) and an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression).
> 
> V2 states:
> 
> If elti is a [condition element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-condition-element),
> 
> * every [variable](https://www.w3.org/TR/sparql12-query/#defn_QueryVariable) mentioned in the [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) of the condition element is an element of Vi-1.
> 
> However, I do not understand what “expression of the condition element” means.

A condition is a special kind of expression - a condition evaluates to true or false. The expression is wrapped in `EBV()` (Effective Boolean Value - from [Functions and Operators](https://www.w3.org/TR/xpath-functions-31/) via SPARQL ). This could be removed, and "condition element" rename "filter element" - the word condition is used in the document as general language as well.

A filter element (`FILTER()` or `srl:filter expr`) is the rule element where the expression is playing a particular role.

It can be confusing because here there is a 1-1 correspondence but it is trying to call out the role being played. Filters have an expression and nothing else - unlike assignments which have expressions and a variable.

> What does “compatible with” mean in this context? Does this mean that expressions are either SHACL list parameter functions (e.g., sparql:plus(38 4)) or SPARQL expressions (e.g., FILTER expressions)? If so, it might be clearer to state directly that expressions form the union of these two categories.

"compatible with" is meant to mean either can be used in the places expressions occur - but also other libraries could be used e.g. domain specific functions. It's not only `shnex:` and SPARQL.

Some SPARQL expressions are more than shnex evaluation. `COALESEC(expr, ....)` is not a function because it does not evaluate all its arguments, and only then call the "coalesce" code; errors matter. These matter for implementers.

liviorobaldo:
Hello!

The new version is better, thanks @afs.

However, I still see some elements of confusion. In particular, I still see:

every [variable](https://www.w3.org/TR/sparql12-query/#defn_QueryVariable) mentioned in the [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) of the [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element) is an element of Vi-1.

I do not understand "the [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) of the [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element)", since "A [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element) is an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression)", according to the definition.

In the definitions, I am also not sure about:

A [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element) is an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) that appears as a [rule element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-rule-element)

and

A [triple pattern element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-triple-pattern-element) is a [triple pattern](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-triple-pattern) that appears as a [rule element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-rule-element).

I am not a native English speaker, so I am not sure whether the expression "that appears as" is appropriate in this case.

I would simply say:

A [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element) is an [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) and a [rule element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-rule-element)

A [triple pattern element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-triple-pattern-element) is a [triple pattern](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-triple-pattern) and a [rule element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-rule-element).

This would be consistent with the definitions of negation element and assignment element.

All the rest looks good to me 👍

afs:
The role in which expression is being used matters. "that appears as" is talking about the role.

A filter element is an expression.

But an expression is not a filter element.
An expression can also be in the role of (the calculation of) an assignment value.

"A is a B" can be read as an equivalence, also meaning "B is an A".

> I do not understand "the [expression](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-expression) of the [filter element](https://raw.githack.com/w3c/data-shapes/rules-wellformed/shacl12-rules/index.html#dfn-filter-element)"

That can be just the filter element. Change made.

(Please say where quoted text is coming from, position within the document and/or an earlier comment or use the "files changed" view. I can't search for text in the HTML because it looks different. Also - documents change so text may be different but the point still valid. /thanks)

afs:
Could I get an approval on this PR, please?

I'm more than happy to further refine definitions.

In order to progress on other sections of the document, I need the definitions to in the editing copy.

And the vacation season approaches - the review cycle gets longer.

I'll use a "working copy" approach to completing the document, not a multiple-PR approach. Then, changes can seen for review and progress made, while not needing approvals on small increments in separate PRs.

simonstey (approved):
looks good for now, thx!

afs:
@liviorobaldo , @simonstey 
Thank you!

PR #942, opened 2026-06-15
#176: Use all prefix declarations if no sh:prefixes are declared, allow xsd:string at sh:namespace

### LINKED ISSUE #176: Make `sh:prefixes` optional in SPARQL queries
Basically every user of SHACL-SPARQL ever has run into this limitation: Not only is the syntax of sh:prefixes complex to use, the expectation is that the namespace prefixes from the shapes graph apply automatically.

When this was designed, the WG decided to not rely on automatic definition of prefixes because prefixes are not persisted in the RDF Graph data model, but are just a temporary feature in during parsing. This is despite the fact that many APIs such as Jena do persist the prefixes together with the graph object. But yeah, the problem remains that when SHACL-SPARQL shapes are moved around the prefixes may get lost.

A solution here may be to automatically prepend all prefixes from the shapes graph when no sh:prefixes triple exists. We need to come up with some creative solution on how to define what these prefixes must be. But it's a huge obstacle, so we cannot just leave the current solution in place. At least I believe we can define a list of default prefixes such as rdf: owl: and sh: that should always be present.

HolgerKnublauch:
Closes #176

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-176/shacl12-sparql/index.html#sparql-prefixes)

This implements the idea at https://github.com/w3c/data-shapes/issues/176#issuecomment-4700587832 plus Andy's suggestion to generalize the datatype of sh:namespace to include xsd:string.

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2035,7 +2035,7 @@ WHERE {
 				<li>SERVICE clauses are conditionally permitted, see <a href="https://github.com/w3c/data-shapes/issues/374">Issue 374</a></li>
 				<li>Removed support for the optional pre-bound variables <code>shapesGraph</code> and <code>currentShape</code>, see <a href="https://github.com/w3c/data-shapes/issues/426">Issue 426</a></li>
 				<li>SPARQL constraints can now directly specify a <code>sh:severity</code>, see <a href="https://github.com/w3c/data-shapes/issues/573">Issue 573</a></li>
-				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
+				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair declared in sh:ShapesGraph, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
 			</ul>
 		</section>
 	</body>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2035,7 +2035,7 @@ WHERE {
 				<li>SERVICE clauses are conditionally permitted, see <a href="https://github.com/w3c/data-shapes/issues/374">Issue 374</a></li>
 				<li>Removed support for the optional pre-bound variables <code>shapesGraph</code> and <code>currentShape</code>, see <a href="https://github.com/w3c/data-shapes/issues/426">Issue 426</a></li>
 				<li>SPARQL constraints can now directly specify a <code>sh:severity</code>, see <a href="https://github.com/w3c/data-shapes/issues/573">Issue 573</a></li>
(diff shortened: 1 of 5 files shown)


shacl12-sparql/index.html
```diff
@@ -2033,6 +2035,7 @@ <h2>Changes between the original SHACL specifications and SHACL 1.2 SPARQL</h2>
 				<li>SERVICE clauses are conditionally permitted, see <a href="https://github.com/w3c/data-shapes/issues/374">Issue 374</a></li>
 				<li>Removed support for the optional pre-bound variables <code>shapesGraph</code> and <code>currentShape</code>, see <a href="https://github.com/w3c/data-shapes/issues/426">Issue 426</a></li>
 				<li>SPARQL constraints can now directly specify a <code>sh:severity</code>, see <a href="https://github.com/w3c/data-shapes/issues/573">Issue 573</a></li>
+				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
```
bergos:
```suggestion
				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair declared in sh:ShapesGraph, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
```

bergos (changes requested):

bergos (approved):

shacl12-sparql/index.html
```diff
@@ -2035,7 +2035,7 @@ <h2>Changes between the original SHACL specifications and SHACL 1.2 SPARQL</h2>
 				<li>SERVICE clauses are conditionally permitted, see <a href="https://github.com/w3c/data-shapes/issues/374">Issue 374</a></li>
 				<li>Removed support for the optional pre-bound variables <code>shapesGraph</code> and <code>currentShape</code>, see <a href="https://github.com/w3c/data-shapes/issues/426">Issue 426</a></li>
 				<li>SPARQL constraints can now directly specify a <code>sh:severity</code>, see <a href="https://github.com/w3c/data-shapes/issues/573">Issue 573</a></li>
-				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
+				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair declared in sh:ShapesGraph, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
```
```suggestion
				<li>If no `sh:prefix`es are present, the system will use any `sh:prefix`/`sh:namespace` pair declared in `sh:ShapesGraph`, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
```
```suggestion
				<li>If no sh:prefixes are present, the system will use any sh:prefix/sh:namespace pair declared in sh:ShapesGraph, see <a href="https://github.com/w3c/data-shapes/issues/176">Issue 176</a></li>
```

HolgerKnublauch: It needs to be `sh:prefixes` (which is a distinct property from `sh:prefix`. I'll edit this manually.

PR #945, opened 2026-06-18
Externalise Profiling's e.g.s

### LINKED ISSUE #926: Externalise RDF examples for validation
* Move all RDF, SPARQL & SHACL Rules examples in Profiles to external files so they can be validated
* incorporate them back into the Spec's HTMl by either:
    * using some ReSpect import function for code - can't yet find this
    * using a pre-proessing (Python) script to build the index.html doc from an index.template.html and the code examples

nicholascar:
Move all the example Turtle in Profiling to external files to allow content syntax validation.

Closes #926

### DIFF
--- a/shacl12-profiling/examples/README.md
+++ b/shacl12-profiling/examples/README.md
@@ -4,7 +4,7 @@ This directory contains example files in pseudo-Turtle format used within the Pr
 
 All examples are included using ReSpec commands like this:
 
-`<div class="turtle" data-include="examples/defining-profiles-of-shacl.ttl"></div>`
+`<div class="turtle" data-include="examples/01.ttl"  data-include-format="text"></div>`
 
 ## Validation
 
--- a/shacl12-profiling/examples/README.md
+++ b/shacl12-profiling/examples/README.md
@@ -1,0 +1,17 @@
+# Examples
+
+This directory contains example files in pseudo-Turtle format used within the Profiling HTML document.
+
+All examples are included using ReSpect commands like this:
+
+`<div class="turtle" data-include="examples/defining-profiles-of-shacl.ttl"></div>`
+
+## Validation
+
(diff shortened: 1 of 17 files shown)


shacl12-profiling/examples/README.md
```diff
@@ -0,0 +1,17 @@
+# Examples
+
+This directory contains example files in pseudo-Turtle format used within the Profiling HTML document.
+
+All examples are included using ReSpect commands like this:
```
ajnelson-nist:
```suggestion
All examples are included using ReSpec commands like this:
```

ajnelson-nist:
I'd thought we would have self-sufficient Turtle files, rather than files requiring some reconstruction and reformatting to be ingestible RDF.  I appreciate that inlining the prefixes in each example would probably make the document pretty busy-looking.

Should the example boxes in the displayed document work when someone copy-pastes the whole box into a new file?

I've had some trouble when learning other languages that have code snippets in HTML documentation focus on the highlights of a coding moment, rather than provide a fully compilable program that just uses comments or something to say "And here's the important part for this documentation page."

All that said, I am happy we have a syntax check now.

shacl12-profiling/examples/check-rdf.py
```diff
@@ -0,0 +1,13 @@
+import sys
+from rdflib import Graph
```
ajnelson-nist, line 2: Confirming - it's OK to have this dependency?  I know we're not running this in CI at the moment.

Hm.  Should we run this in CI?

nicholascar: We might only use this example checking for our doc development and not when the doc is published, so I think it's overkill to automate

ajnelson-nist (approved):
This is good-enough to merge to me.  I see one remaining opportunity to add CI, but that probably doesn't need to happen immediately.

shacl12-profiling/examples/check-rdf.py
ajnelson-nist: Ah, apologies, my approval comment went in between you posting this remark and it displaying for me.

I concur.

PR #949, opened 2026-06-19
Clarify definition of shapes graph

### LINKED ISSUE #861: Definition(s) of "Shapes Graph"
@ajnelson-nist and I were talking about how to phrase certain things regarding shapes graphs yesterday in the Profiling group and we got caught up on the fact that the term "shapes graph" appears in the spec with a few subtly different meanings that are never explicitly distinguished. I've run into this issue before when reading the spec and have had to double check the context to be sure I'm thinking about it the right way.

Here are what I think are the three different usages/definitions related to shapes graphs:

### 1. _Shapes Graph_ (reusable module/artifact)

Any RDF graph that contains SHACL shapes, existing as a publishable, importable unit independent of any validation process. This is what the newly defined `sh:ShapesGraph` represents.

For example:
- The Core abstract uses the plural: *"RDF graphs called shapes graphs"*, and in this context there is no validation happening yet
- `owl:imports` and `sh:shapesGraph` reference individual shapes graphs that get composed together
- The informative text in Core section 6.1 describes shapes graphs as *"reusable validation modules"*

### 2. _Input Shapes Graph_ (what is originally provided)

A specific shapes graph originally provided to a SHACL processor, _before_ `owl:imports` resolution. Includes graphs targeted by `sh:shapesGraph`.

For example:
- The Core spec distinguishes *"the originally provided shapes graph"* from *"the resulting graph"* after transitive import resolution
- These are the starting points of the composition process

### 3. _Expanded Shapes Graph_ (the fully expanded graph used during validation)

The single, unified, immutable graph resulting from composing the input shapes graph with all transitively imported graphs that is actually queried during validation.

For example:
- The immutability requirement in Core section 6.1 applies to this graph: it *"MUST NOT be further modified during the validation process"*
- Shape definitions, constraint components, and validators are all discovered *"in the shapes graph"*, which requires the fully expanded graph to find content from imported modules
- In the SPARQL spec, prefix declarations are currently collected via `sh:prefixes/owl:imports*/sh:declare`. This path works because imported content is already present in the expanded graph including any `owl:imports` triples

### Potential Resolution

Explicitly define these two terms in normative text (potentially with different/better names) and update all references accordingly:

1. **Shapes Graph**: any RDF graph containing shapes (the module/artifact sense, can be plural)
2. **Expanded Shapes Graph**: The fully expanded graph resulting from composing the input shapes graph with all `owl:imports` and `sh:shapesGraph` references. It is the graph actually used by the validation engine and is immutable during validation. Most current normative references to "the shapes graph" or "in the shapes graph" across the spec actually refer to this graph.

Perhaps also define in normative text, if useful to do so:

3. **Input Shapes Graph**: A specific shapes graph originally provided to a SHACL processor before import resolution

Note that any case, a shapes graph in practice may contain more than just shapes (e.g. RDFS/OWL class hierarchies, entailment declarations, SPARQL prefix declarations, etc.) and the processor relies on this broader content being present and queryable for certain features to work.

mgberg:
Closes #861.

### Motivation

The term "shapes graph" was used throughout the spec in subtly different senses (a reusable module/artifact, an input to a processor, and the fully expanded immutable graph used during validation) without these being explicitly distinguished. This change clarifies the definition without introducing new terminology or requiring changes across many documents.

### Changes

Rewrites the normative definition of "shapes graph" in Core to:

1. **Define it as a role** rather than as an intrinsic property of a graph, consistent with the definition of the new `sh:ShapesGraph`

2. **Explicitly acknowledge the potential for composition**: The final immutable shapes graph for a given validation _may_ result from combining multiple shapes graphs (e.g., via `owl:imports` or `sh:shapesGraph`).

3. **Explicitly acknowledge the validation shapes graph in the definition section**: Regardless of how the shapes graph for validation is constructed, it MUST remain immutable during validation. This was previously only stated normatively in the Validation section; it is now also established at the point of definition for clarity (with a reference to the Validation section for more detail).

Additionally, the non-normative sentence following the `owl:imports` resolution examples is simplified since the normative requirement is now stated in the definition.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-861/shacl12-core/index.html)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2089,12 +2089,18 @@ ex:SomeClass-alternativePathExample
 			<section id="shapes-graph">
 				<h3>Shapes Graph</h3>
 				<p>
-					A <dfn>shapes graph</dfn> is an RDF graph containing zero or more shapes
-					that may be passed into a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
+					A <dfn>shapes graph</dfn> is an <a>RDF graph</a> in the role of providing <a>shapes</a> and related information
+					to a SHACL <a>validation</a> process so that a <a>data graph</a> can be validated against the shapes.
 				</p>
 				<p>
 					<span data-syntax-rule="ShapesGraph">The <code>sh:ShapesGraph</code> class MAY be used as an <code>rdf:type</code>
-					of the <a>IRI</a> of a graph that is typically used as a <a>shapes graph</a>.</span>
+					of the <a>IRI</a> of a graph that typically acts in the role of a <a>shapes graph</a>.</span>
+				</p>
+				<p>
+					The <a>shapes graph</a> for a given <a>validation</a> may be the result of combining multiple shapes graphs,
+					for example through <code>owl:imports</code> or <code>sh:shapesGraph</code>.
+					However it is constructed, the <a>shapes graph</a> used for validation MUST remain immutable during the <a>validation</a> process.
+					See <a href="#validation-definition"></a> for further details.
 				</p>
 				<p><em>The remainder of this section is non-normative.</em></p>
 				<p>
@@ -2198,7 +2204,7 @@ ex:CompanyShape
 				</aside>
 
 				<p>
-					The resulting graph forms the input <a>shapes graph</a> for validation and MUST NOT be further modified during the validation process.
+					The resulting graph forms the immutable <a>shapes graph</a> for validation.
 				</p>
 				<p class="note">
 					When using <code>owl:versionIRI</code> to import versioned shapes graphs, care should be taken to
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -2099,7 +2099,7 @@ ex:SomeClass-alternativePathExample
 				<p>
 					The <a>shapes graph</a> for a given <a>validation</a> may be the result of combining multiple shapes graphs,
 					for example through <code>owl:imports</code> or <code>sh:shapesGraph</code>.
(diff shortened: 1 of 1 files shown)


shacl12-core/index.html
```diff
@@ -2089,12 +2089,18 @@ <h2>Validation and Graphs</h2>
...
 				</p>
 				<p>
 					<span data-syntax-rule="ShapesGraph">The <code>sh:ShapesGraph</code> class MAY be used as an <code>rdf:type</code>
-					of the <a>IRI</a> of a graph that is typically used as a <a>shapes graph</a>.</span>
+					of the <a>IRI</a> of a graph that typically acts in the role of a <a>shapes graph</a>.</span>
+				</p>
+				<p>
+					The <a>shapes graph</a> for a given <a>validation</a> may be the result of combining multiple shapes graphs,
+					for example through <code>owl:imports</code> or <code>sh:shapesGraph</code>.
+					However it is constructed, the <a>shapes graph</a> used for validation MUST remain immutable during the <a>validation</a> process.
```
afs: Minor suggestion: 
```suggestion
					However it is constructed, the <a>shapes graph</a> used for validation MUST remain fixed during the <a>validation</a> process.
```
Saying it is "fixed" covers "immutable", and also the case of changing the entire graph for another one.

(An RDF Graph is immutable by definition - but good to stress the point.)

HolgerKnublauch (approved):

nicholascar (approved):

ajnelson-nist (approved):

PR #950, opened 2026-06-19
fix typo where `an` should be `a`

TallTed:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/TallTed-patch-2/shacl12-rules/index.html)


afs:
Done on #952

PR #951, opened 2026-06-19
Fix wording in well-formedness definition

TallTed:
`initial starting` is redundant

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/TallTed-patch-3/shacl12-rules/index.html)


afs:
Done on #952

afs:
Done on #952

PR #958, opened 2026-06-23
Issue 957: Clarify whether sh:values produces triples

### LINKED ISSUE #957: shnex inference
>> What should we do with SHACL-AF Rules? (#766)

https://github.com/w3c/data-shapes/issues/766#issuecomment-3863108674
>I would leave them alone. Most use cases are covered by either sh:values ...

The usage of sh:values for this use case is unclear.

https://www.w3.org/TR/shacl12-node-expr/#getting-started
> Some tools MAY however materialize the triples, for example so that they become available to SPARQL queries.

If even optional, there needs to be  description of how that happens.
e.g. where does the triple go to? is there an output graph or is it only modification of the data graph?
e.g. what are "the triples" for a general `sh:path`?

(that "MAY" is in a non-normative section.)

> This means that if the provided shnex:pathValues path is an [IRI](https://www.w3.org/TR/rdf12-concepts/#dfn-iri), then a path values expression may cause the evaluation of other node expressions, as a simple kind of rule chaining.

Can these produce triples?

HolgerKnublauch:
Closes #957

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-957/shacl12-node-expr/index.html)

I hope this clarifies things?

### DIFF
--- a/shacl12-common/specifications.html
+++ b/shacl12-common/specifications.html
@@ -16,7 +16,7 @@
           <dt><a href="https://www.w3.org/TR/shacl12-sparql/">SHACL 1.2 SPARQL Extensions</a></dt>
           <dd>defines SPARQL-related extensions of SHACL</dd>
           <dt><a href="https://www.w3.org/TR/shacl12-node-expr/">SHACL 1.2 Node Expressions</a></dt>
-          <dd>defines graph expressions used to determine focus nodes in SHACL</dd>
+          <dd>defines expressions used to derive focus nodes and value nodes in SHACL</dd>
           <dt><a href="https://www.w3.org/TR/shacl12-rules/">SHACL 1.2 Rules</a></dt>
           <dd>defines SHACL's methods of rule-based inference</dd>
           <dt><a href="https://w3c.github.io/data-shapes/shacl12-ui/">SHACL 1.2 UI</a></dt>
@@ -51,7 +51,7 @@
         <dt data-transform="noSelfCite">[[[shacl12-sparql]]]</dt>
         <dd>defines SPARQL-related extensions of SHACL</dd>
         <dt data-transform="noSelfCite">[[[shacl12-node-expr]]]</dt>
-        <dd>defines graph expressions used to determine focus nodes in SHACL</dd>
+        <dd>defines expressions used to derive focus nodes and value nodes in SHACL</dd>
         <dt data-transform="noSelfCite">[[[shacl12-rules]]]</dt>
         <dd>defines SHACL's methods of rule-based inference</dd>
         <dt data-transform="noSelfCite">[[[shacl12-cs]]]</dt>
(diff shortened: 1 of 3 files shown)


shacl12-common/specifications.html
```diff
@@ -51,7 +51,7 @@ <h2>SHACL Specifications</h2>
         <dt data-transform="noSelfCite">[[[shacl12-sparql]]]</dt>
         <dd>defines SPARQL-related extensions of SHACL</dd>
         <dt data-transform="noSelfCite">[[[shacl12-node-expr]]]</dt>
-        <dd>defines graph expressions used to determine focus nodes in SHACL</dd>
+        <dd>defines expressions used to derive focus nodes and value nodes in SHACL</dd>
```
afs, line 54: That does not include triples.
Even if optional, triples can be derived.

afs:
This is better wording but it does not cover the point in #957.

It doesn't provide the information of where do materialized the triples go and when are they become visible (node-expr can query for "exists" and "count", and has `shnex:not`).

But removing the possibility would mean that the triples are there (i.e. detectable) but can't be seen (e.g. an explanation of validation can't refer to them).

"SPARQL queries" would include shacl-sparql.

shacl12-common/specifications.html
HolgerKnublauch: Ok before this gets misinterpreted I prefer to take this note out. Node Expressions don't create triples, esp not during materialization. It was a "may" inside of an informative section, and can set wrong expectations.

TallTed:
*PLEASE* put something more informative in this and any other PR currently titled simply "issue ###"! The PR title should give a clear hint of what the PR does (and yes, that means that the PR should be retitled if the action taken thereby is changed substantially).

shacl12-node-expr/index.html
```diff
@@ -738,11 +738,10 @@ <h2>Getting started with Node Expressions</h2>
 				will be fetched.
 			</p>
 			<p>
-				Note that these derived properties do not automatically lead to the creation of triples.
+				Note that derived properties, such as <code>ex:employeeCount</code> from the <a href="#employeeCountExample">example</a> above,
+				do not lead to the creation of triples in the <a>data graph</a> or <a>shapes graph</a>.
 				Tools that use these <code>sh:values</code> expressions typically compute the values on demand
 				only, for example whenever an instance of <code>ex:Company</code> is displayed or queried.
```
```suggestion
				Tools that use these <code>sh:values</code> expressions typically compute the values only on
				demand; for example, whenever an instance of <code>ex:Company</code> is displayed or queried.
```

HolgerKnublauch:
@afs any reason why you are not approving this PR? You opened the issue, you were the only who discussed it (with me), I have addressed your comments and made you a reviewer. This is a harmless change to a non-normative section. The PR has been sitting idle for two weeks.

afs:
The PR does not address the request in the issue (and https://github.com/w3c/data-shapes/pull/958#issuecomment-4787134240) which was to describe where triples go to.

I have no opinion of the change if it is to remove the text as it no longer meets the need I had hoped it covered. #960  is the response.

(The wider discussion on #957 does affect an experimental implementation I have but that's just my experiment.)

HolgerKnublauch:
The text about triples was never normative. If it can be misunderstood, I prefer to remove it, which is what I am doing here by making clear that NO triples are created. I think I have explained this enough now.

Could I please get anyone else to approve this. @nicholascar @simonstey @robert-david - thanks.

simonstey (approved):
minor changes -> approve

PR #965, opened 2026-06-27
Issue 897: Remove sh:value from sh:uniqueValuesFor

### LINKED ISSUE #897: sh:uniqueValuesFor: handling sh:value for multiple duplicates
The current definition for a validation result for `sh:uniqueValuesFor` is as follows:

> For each [value node](https://w3c.github.io/data-shapes/shacl12-core/#dfn-value-node) V for which there exists another [node](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-node) Other in $targetNodes that has exactly the same [values](https://w3c.github.io/data-shapes/shacl12-core/#dfn-value) for all properties in $properties as V there is a [validation result](https://w3c.github.io/data-shapes/shacl12-core/#dfn-validation-result) with Other as sh:value.

In cases where more than two target nodes share the same values for the defined properties, the value of `sh:value` can no longer be determined unambiguously. Since `sh:value` permits only a single value, it is not possible to list all resources participating in the duplicate set as values of `sh:value`. Additionally, such a list could grow very large.

Because a validation result already exists for each duplicate resource, all participating resources are already represented through separate results. Therefore, this proposal suggests using a literal that represents the combination of values of the defined properties. Such a string is likely already generated internally by implementations as part of the duplicate-detection algorithm and would only need to be exposed.

The exact string representation can be implementation-specific. The examples in this proposal use the following logic:

```js
values.map(v => toNT(v)).join(' ')
```

This string could also be made available to result message templates.

Example:

```turtle
@prefix ex: <http://example.com/ns#>.
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix sh: <http://www.w3.org/ns/shacl#>.

ex:InvalidInstance1 a ex:TestShape;
  ex:id1 "A";
  ex:id2 "1".

ex:InvalidInstance2 a ex:TestShape;
  ex:id1 "A";
  ex:id2 "1".

ex:InvalidInstance3 a ex:TestShape;
  ex:id1 "A";
  ex:id2 "1".

ex:TestShape a sh:NodeShape;
  sh:targetClass ex:TestShape;
  sh:uniqueValuesFor (ex:id1 ex:id2).

[ a sh:ValidationReport;
  sh:conforms false;
  sh:result [ a sh:ValidationResult;
      sh:focusNode ex:InvalidInstance1;
      sh:resultSeverity sh:Violation;
      sh:sourceConstraintComponent sh:UniqueValuesForConstraintComponent;
      sh:sourceShape ex:TestShape;
      sh:value
        # "\"A\" \"1\"", proposed value
        ex:InvalidInstance2,
        #ex:InvalidInstance3 currently possible alternative
    ], [ a sh:ValidationResult;
      sh:focusNode ex:InvalidInstance2;
      sh:resultSeverity sh:Violation;
      sh:sourceConstraintComponent sh:UniqueValuesForConstraintComponent;
      sh:sourceShape ex:TestShape;
      sh:value
        # "\"A\" \"1\"", proposed value
        ex:InvalidInstance1,
        # ex:InvalidInstance3 currently possible alternative
    ], [ a sh:ValidationResult;
      sh:focusNode ex:InvalidInstance3;
      sh:resultSeverity sh:Violation;
      sh:sourceConstraintComponent sh:UniqueValuesForConstraintComponent;
      sh:sourceShape ex:TestShape;
      sh:value
        # "\"A\" \"1\"", proposed value
        ex:InvalidInstance1,
        # ex:InvalidInstance2 currently possible alternative
    ]
].
```

bergos:
<!--
# Pull request instructions
(Note that text between the HTML comments will be recorded in this pull request's source for future editing, but will not render on Github.  Text is suggested with portions needing substitution enclosed in curly braces.  The curly braces also need to be removed when substituting.)

Please use "Closing keywords" for the associated Issue if this PR should close the Issue once merged.
https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

If this PR edits a document, please consider including a "Live render."  The non-comment template text suggests one mechanism.  If including the link line(s), please test-load the link(s) while drafting the PR.
-->

Closes #897 
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/{BRANCH_NAME}/{FOLDER}/index.html)

This PR contains two commits:                                                                                                                    
                                                                                                                                                   
  1. Remove `sh:value` from `sh:uniqueValuesFor` validation results and add a test for three nodes sharing the same values.                            
  2. Add a non-normative note that `sh:detail` can be used to provide references to the other duplicate nodes.                                       
                                                                                                                                                   
  Please leave a comment if the second commit should be dropped.

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5782,10 +5782,6 @@ ex:Zoo
 						Also note that <code>sh:uniqueValuesFor</code> does not prescribe the existence of these values, and is therefore
 						often combined with a <code>sh:minCount 1</code> constraint.
 					</p>
-					<p>
-						Implementations can use <code>sh:detail</code> to provide references to the other nodes
-						that share the same values for the constrained properties.
-					</p>
 					<aside class="example" title="Example of sh:uniqueValuesFor on a single property">
 						<div class="shapes-graph">
 							<div class="turtle">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5782,6 +5782,10 @@ ex:Zoo
 						Also note that <code>sh:uniqueValuesFor</code> does not prescribe the existence of these values, and is therefore
 						often combined with a <code>sh:minCount 1</code> constraint.
 					</p>
+					<p>
+						Implementations can use <code>sh:detail</code> to provide references to the other nodes
+						that share the same values for the constrained properties.
+					</p>
 					<aside class="example" title="Example of sh:uniqueValuesFor on a single property">
 						<div class="shapes-graph">
 							<div class="turtle">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5770,9 +5770,9 @@ ex:Zoo
 							If <code>$uniqueValuesFor</code> is a <a>SHACL list</a>, then <code>$properties</code> is the set of the <a>members</a> of that list.
 							Let <code>$targetNodes</code> be the <a>target</a> nodes of <code>S</code>.
 							<br />
-							For each <a>value node</a> <code>V</code> for which there exists another <a>node</a> <code>Other</code> in <code>$targetNodes</code>
+							For each <a>value node</a> <code>V</code> for which there exists another <a>node</a> in <code>$targetNodes</code>
 							that has exactly the same <a>values</a> for all properties in <code>$properties</code> as <code>V</code>
-							there is a <a>validation result</a> with <code>Other</code> as <code>sh:value</code>.
+							there is a <a>validation result</a>.
 							No result is produced if <code>V</code> has no <a>values</a> for any of the properties in <code>$properties</code>.
 						</div>
 					</div>
--- a/shacl12-test-suite/tests/core/node/manifest.ttl
+++ b/shacl12-test-suite/tests/core/node/manifest.ttl
@@ -47,6 +47,7 @@
 	mf:include <uniqueValuesFor-002.ttl> ;
 	mf:include <uniqueValuesFor-003.ttl> ;
 	mf:include <uniqueValuesFor-004.ttl> ;
+	mf:include <uniqueValuesFor-005.ttl> ;
 	mf:include <xone-001.ttl> ;
 	mf:include <xone-duplicate.ttl> ;
 	mf:include <qualified-001.ttl> ;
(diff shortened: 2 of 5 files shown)


shacl12-core/index.html
```diff
@@ -5782,6 +5782,10 @@ <h4>sh:uniqueValuesFor</h4>
 						Also note that <code>sh:uniqueValuesFor</code> does not prescribe the existence of these values, and is therefore
 						often combined with a <code>sh:minCount 1</code> constraint.
 					</p>
+					<p>
+						Implementations can use <code>sh:detail</code> to provide references to the other nodes
```
HolgerKnublauch: Maybe use "may" instead of "can" here.

HolgerKnublauch (approved):

shacl12-core/index.html
bergos: Since this is in the non-normative section, I intentionally used can rather than may. We could move it to the normative section with MAY. I don't have a strong preference either way.

HolgerKnublauch: It is fine to use the term may in lower-case spelling in informative sections. Only MAY is not permitted.

HolgerKnublauch: Reading the spec, we should probably just delete that hint. The values of sh:detail are expected to be instances of sh:AbstractResult, not individual values. 

https://www.w3.org/TR/shacl12-core/#results-detail

So, listing the duplicates as part of the message is probably the only realistic option.

shacl12-core/index.html
```diff
@@ -5782,6 +5782,10 @@ <h4>sh:uniqueValuesFor</h4>
...
+						that share the same values for the constrained properties.
+					</p>
```
YoucTagh:
```suggestion
```

Deleting the hint as suggested by Holger.

shacl12-core/index.html
TallTed: > It is fine to use the term `may` in lower-case spelling in informative sections.

It is permitted, but it is prone to misinterpretation. I think @bergos's choice to use `can` here is appropriate.

PR #972, opened 2026-07-02
#964: Added shnex:conformsToShape

### LINKED ISSUE #964: Add function to check conformance of a node (`shnex:hasShape`, `shnex:conformsTo`)
I believe there should a List Parameter function (e.g. called shnex:hasShape) that takes a node and a shape and returns true if the node conforms to the shape.

Motivation: List parameter functions can potentially be exposed to SPARQL functions, and, similarly, in SHACL 1.2 Rules. They would typically be used in FILTER expressions.

```
    WHERE {
        $this ex:customer ?customer .
        FILTER (shnex:hasShape(?customer, ex:HighRevenueCustomer)) .
    }
```

where the exact details of what makes up a HighRevenueCustomer are encapsulated by the shape (e.g. sum of sales > 100k).

In our product we have such a function called `tosh:hasShape` but it would be better to have this with a standard URI before each vendor proliferates on this.

Does anyone see problems with this? Can this already be achieved in different ways, e.g. in SHACL Rules?

HolgerKnublauch:
Closes #964

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-964/shacl12-node-expr/index.html#conformsTo)


nicholascar (changes requested):
We have already added a property to the SHACL vocabulary called `sh:conformsTo`: https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L226-L234

Will the naming of this function be confusing given this property?

HolgerKnublauch:
Yes I saw that duplication. I originally suggested shnex:hasShape, but Andy suggested the current name (see issue discussion).

What name would you suggest, @nicholascar ?

nicholascar:
`sh:conformsTo` follows on from `dcterms:conformsTo`, so I must claim precedence of form there!

Also, if the expected range value is a boolean, _conforms to_ does't make sense. Conforms to what? In DCTERMS, conforms to indicates the thing that the resource does conform to.

I suggest `shnex:doesConform`. This is more 'active' than `shnex:hasShape` and indicates a boolean with its wording.

HolgerKnublauch:
Other options: shnex:conformsToShape, shnex:nodeConformsToShape
These are a bit longer but shnex:doesConform could mean anything (what conforms to what here)?

shacl12-node-expr/index.html
```diff
@@ -2908,6 +2908,93 @@ <h3>Nodes Matching Expressions</h3>
...
+										<td>
+										</td>
+										<td>
+											The <a>node</a> that is validated.
+										</td>
+									</tr>
+									<tr>
+										<td><b><code>shape</code></b></td>
+										<td>
+											<code>sh:nodeKind sh:IRI</code> and a <a>well-formed</a> <a>shape</a>.
```
afs: what does "and a well-formed shape" mean here?

You can't give the shape as an argument inline, only the IRI for the shape . As a List Parameter Function, argument evaluation happens to blank nodes.

afs:
shnex uses "matches" -- `shnex:matchAll` and also `shnex:nodesMatches`. 

`matchAll` is the stream equivalent so shnex:conformTo

So something like `shnex:nodeMatches`,  `shnex:hasMatch`, `shnex:matchesShape`, might work.

shacl12-node-expr/index.html
HolgerKnublauch: The value must be an IRI and conform to the syntax rules of shapes (in the Core spec). So yes, blank nodes are excluded via the nodeKind.

HolgerKnublauch:
Yes, shnex:nodeMatches or shnex:matchesShape sound good to me. Would you remove your objection, @nicholascar and which one is best?

shacl12-node-expr/index.html
afs: I don't understand the "and" -its' an IRI, not the shape itself.

The layout is broken / different to `shnex:nodesMatching` which is clearer.

HolgerKnublauch: Fixed

nicholascar:
I prefer `shnex:matchesShape` as that isn't forcing info about the domain like `shnex:nodeMatches` does.

I guess you could have `shnex:conformsToShape` as a subPropertyOf `sh:conformsTo`...

robert-david:
@HolgerKnublauch @nicholascar 
To continue with this review, did we agree on a name? It is still `shnex:conformsTo` in the draft.
I prefer `shnex:conformsToShape` because it keeps the similarity with `sh:conformsTo`.

HolgerKnublauch:
It looks like sh:conformsTo rdfs:range sh:ShapesGraph. So I guess one solution here is to rename that to sh:conformsToShapesGraph and the one here to shnex:conformsToShape. Both are new and not in use yet, so it should be fine. Anyone see problems with that? I can do both renamings here.

shacl12-node-expr/index.html
```diff
@@ -2908,6 +2908,94 @@ <h3>Nodes Matching Expressions</h3>
...
+							</table>
+						</span>
+					</p>
+					<div class="def" id="conformsTo-evaluation">
+						<div class="def-header">EVALUATION OF CONFORMSTO EXPRESSIONS</div>
+						<p>
+							Let <code>node</code> be the first argument of a <a>conformsTo expression</a>.
+							Let <code>shape</code> be the second argument of a <a>conformsTo expression</a>.
+							The <a>output nodes</a> of the <a>conformsTo expression</a> are the list of either <code>true</code> or <code>false</code>,
+							whereby <code>true</code> is returned exactly when <code>node</code> <a>conforms</a> to <code>shape</code>.
```
simonstey: 1) Every other evaluation definition in this document uses the `evalExpr(expr, focusGraph, focusNode, scope)` formalism and states which nodes are produced by eval. the arguments. 
Here the arguments are themselves node expressions that MUST be evaluated (in the example later on  `[ shnex:var "focusNode" ]` is used as `node`, which is meaningless unless it is evaluated. Currently, `node` would in this case be the blank node `[ shnex:var "focusNode" ]`, not the focus node.

2) The evaluation says node conforms to shape but does not name the graph the conformance check runs against. Sibling functions state the focus graph explicitly.

shacl12-node-expr/index.html
```diff
@@ -2908,6 +2908,94 @@ <h3>Nodes Matching Expressions</h3>
...
+								</tbody>
+							</table>
+						</span>
+					</p>
+					<div class="def" id="conformsTo-evaluation">
+						<div class="def-header">EVALUATION OF CONFORMSTO EXPRESSIONS</div>
+						<p>
+							Let <code>node</code> be the first argument of a <a>conformsTo expression</a>.
+							Let <code>shape</code> be the second argument of a <a>conformsTo expression</a>.
+							The <a>output nodes</a> of the <a>conformsTo expression</a> are the list of either <code>true</code> or <code>false</code>,
```
simonstey: other boolean expressions (like `shnex:exists`, `shnex:matchAll`) say the output nodes are `( true )` … otherwise `( false )`. Here it is "the list of either `true` or `false`." bug or feature?

shacl12-node-expr/index.html
```diff
@@ -2908,6 +2908,94 @@ <h3>Nodes Matching Expressions</h3>
 					</p>
 				</section>
 
+				<section id="conformsTo">
```
simonstey:
```suggestion
				<section id="ConformsToShapeExpression">
```
this would be more consistent with the ids of other sections

shacl12-node-expr/index.html
```diff
@@ -2908,6 +2908,94 @@ <h3>Nodes Matching Expressions</h3>
...
+										</td>
+									</tr>
+								</tbody>
+							</table>
+						</span>
+					</p>
+					<div class="def" id="conformsTo-evaluation">
+						<div class="def-header">EVALUATION OF CONFORMSTO EXPRESSIONS</div>
+						<p>
+							Let <code>node</code> be the first argument of a <a>conformsTo expression</a>.
```
simonstey: given that node (in theory) has to be evaluated first -> what happens if the node argument produces 0 nodes? what's the conformance of "no nodes?"

simonstey (changes requested):
Corner cases:

- `node` conforms → covered (`( true )`). ✔
- `node` does not conform → covered (`( false )`). ✔
- `node` arg = node expression yielding exactly 1 node → intended (spec example) but **not stated** in the def. 
- `node` arg yielding **>1** node → only implicitly a failure via the general list-parameter rule; what's expected?
- `node` arg yielding **0** nodes → **unhandled/undefined**. ✗
- `shape` = blank node → excluded by `sh:nodeKind sh:IRI` (well-formedness). ✔
- List arity ≠ 2 → not a well-formed conformsTo expression per the syntax rule; 

what happens if the conformance check itself fails (e.g., recursion)?

shacl12-node-expr/index.html
HolgerKnublauch: It needs to be "conformsToShape" because that is the name of the expression. This is a list parameter function and therefore different from the others which are named argument functions.

shacl12-node-expr/index.html
HolgerKnublauch: This and most of the other questions that you raise are covered by section https://w3c.github.io/data-shapes/shacl12-node-expr/#ListParameterFunction which states that such functions can in general only produce one output node and produce an evaluation failure if the arguments produce more than 1 node.

I have tried to clarify all this and added explanatory prose so that these details are more obvious.

Please have another look, your comments certainly made sense!

shacl12-node-expr/index.html
```diff
@@ -2944,7 +2944,7 @@ <h3>ConformsToShape Expressions</h3>
 						Note that <code>shnex:conformsToShape</code> is a <a>list parameter function</a> which means that
 						it returns at most one output node and
 						the members of the arguments list are node expressions that must produce at most one output node each.
-						In other words, if the <code>shape</code> argument is a <a>blank node</a> it will be interpreted as
+						As a consequence, if the <code>shape</code> argument is a <a>blank node</a> it will be interpreted as
```
TallTed: Two word tweaks, plus an additional (technically optional, but recommended by me) comma.
```suggestion
						Consequently, if the <code>shape</code> argument is a <a>blank node</a>, then it will be interpreted as
```

shacl12-node-expr/index.html
```diff
@@ -986,6 +986,29 @@ <h2>Handling of Failures</h2>
...
+				<p>
+					Some node expression functions such as <a href="#FilterShapeExpression"></a> and <a href="#conformsToShape"></a>
+					perform SHACL <a>validation</a> to determine whether a <a>node</a> <a>conforms</a> to a <a>shape</a>.
+					The <a>shapes graph</a> used for this <a>validation</a> is not explicitly stated for each function.
+					Instead, it is derived from the context:
+				</p>
+				<ol>
+					<li>
+						When a node expression is evaluated as part of another SHACL process
+						(e.g. computing value nodes using <code>sh:values</code> or target nodes using <code>sh:targetNode</code>),
```
```suggestion
						(e.g., computing value nodes using <code>sh:values</code> or target nodes using <code>sh:targetNode</code>),
```

shacl12-node-expr/index.html
```diff
@@ -986,6 +986,29 @@ <h2>Handling of Failures</h2>
 				</p>
 			</section>
 
+			<section id="the-shapes-graph">
+				<h2>The Shapes Graph</h2>
+				<p>
+					Some node expression functions such as <a href="#FilterShapeExpression"></a> and <a href="#conformsToShape"></a>
```
```suggestion
					Some node expression functions — such as <a href="#FilterShapeExpression"></a> and <a href="#conformsToShape"></a> —
```

shacl12-node-expr/index.html
```diff
@@ -2901,6 +2924,108 @@ <h3>Nodes Matching Expressions</h3>
...
+						<div class="def-header">EVALUATION OF CONFORMSTOSHAPE EXPRESSIONS</div>
+						<p>
+							Let <code>nodeExpr</code> be the first argument of a <a>conformsToShape expression</a>.
+							Let <code>node</code> be the (single) result of <code>evalExpr(nodeExpr, focusGraph, focusNode, scope)</code>.
+							Let <code>shapeExpr</code> be the second argument of a <a>conformsToShape expression</a>.
+							Let <code>shape</code> be the (single) result of <code>evalExpr(shapeExpr, focusGraph, focusNode, scope)</code>.
+							The <a>output nodes</a> of the <a>conformsToShape expression</a> are the empty list
+							if either <code>node</code> or <code>shape</code> have no value.
+							Otherwise, the <a>output nodes</a> are
+							<code>( true )</code> if and only if <code>node</code> <a>conforms</a> to <code>shape</code> using the <a>focus graph</a> as the <a>data graph</a>,
```
simonstey:
```suggestion
							<code>( true )</code> if and only if <code>node</code> <a>conforms</a> to <code>shape</code> (using the <a href="#the-shapes-graph">derived shapes graph</a> for the shape definition and the <a>focus graph</a> as the <a>data graph</a>),
```

just to clarify where the shape's definition is read from.

shacl12-test-suite/tests/node-expr/shnex/conformsToShape.ttl
```diff
@@ -0,0 +1,54 @@
...
+@prefix sht: <http://www.w3.org/ns/shacl-test#> .
+@prefix sparql:  <http://www.w3.org/ns/sparql#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<>
+  rdf:type mf:Manifest ;
+  mf:entries (
+      <conformsToShape-example-invalid>
+      <conformsToShape-example-valid>
+    ) ;
```
simonstey, line 18:
```suggestion
      <conformsToShape-empty-input>
    ) ;
```

also test cases for empty input  (empty output != false)

```ttl
   <conformsToShape-empty-input>
     rdf:type sht:EvalNodeExpr ;
     rdfs:label "Test of shnex:conformsToShape where the node argument is empty (no focus node bound)" ;
     mf:action [
       sht:nodeExpr [
         shnex:conformsToShape ( [ shnex:var "focusNode" ] ex:Shape ) ;
       ] ;
     ] ;         
     mf:result () ;       
     mf:status sht:approved ;
   .
```

simonstey (approved):

nicholascar (approved):

Issue #930, opened 2026-06-08
Create a running example for the rules description section


afs:

liviorobaldo:
I see there are already some examples in [SHACL 1.2 Rules](https://w3c.github.io/data-shapes/shacl12-rules/), involving fathers, mothers, children, and descendants.

These examples are simple and intuitive enough for everyone to understand, so I would build on them and use a single running example throughout the document.

In particular:

- [3.1 Basic Patterns](https://w3c.github.io/data-shapes/shacl12-rules/#desc-basic)

  The example already there is fine: `RULE { ?x :childOf ?y } WHERE { ?y :fatherOf ?x }`, etc.

- [3.2 Recursion](https://w3c.github.io/data-shapes/shacl12-rules/#desc-recursion)

  The example already there is fine: `RULE { ?x :descendedFrom ?y } WHERE { ?x :childOf ?y }`, etc.

- [3.3 Filtering](https://w3c.github.io/data-shapes/shacl12-rules/#desc-filtering)

  I would replace the `:largeTown` example with one based on the same family scenario. For example:

  `RULE { ?y1 :siblingOf ?y2 . ?y2 :siblingOf ?y1 } WHERE { ?x :childOf ?y1 . ?x :childOf ?y2 . FILTER(?y1 != ?y2) }`

  This example also shows that multiple triples can appear in the rule head, whereas the previous examples only have a single triple in the head.

  After this example, in this section I would also add:

  `RULE { ?y1 :isOlderThan ?y2 } WHERE { ?y1 :siblingOf ?y2 . ?y1 :hasAge ?a1 . ?y2 :hasAge ?a2 . FILTER(?a1 > ?a2) }`

  This rule will fire after the previous one, because the two individuals must first be inferred to be siblings. It also shows that `FILTER` can be applied to datatype values. In the previous example, `FILTER` is only used to prevent the property `:siblingOf` from being reflexive.

- [3.4 Negation](https://w3c.github.io/data-shapes/shacl12-rules/#desc-negation)

  Again, I would replace the `:Place`/`:population` example with one based on the same family scenario. For instance:

  `RULE { ?x :motherOf ?y } WHERE { ?x :parentOf ?y . NOT { ?x :fatherOf ?y } }`

  `RULE { ?x :fatherOf ?y } WHERE { ?x :parentOf ?y . NOT { ?x :motherOf ?y } }`

[...]

PR #870, opened 2026-04-25
RDF syntax for SHACL Rules

afs:
This PR contains:
* `rdf-syntax-shapes.ttl`
   A first pass on shapes for the RDF syntax of SHACL Rules
* Tests for the RDF Syntax. Tests fragments (good and bad) against the shapes in `rdf-syntax-shapes.ttl`
  * `test-terms.ttl` - shapes for terms and variables.
  * `test-triples.ttl` - shapes for triples; data, pattern and template roles
  * `test-elements.ttl` - shapes for rule elements 
  * `test-rules.ttl` - shapes for rule set and rules
 * `rules-ns.ttl` - namespace document containing the SRL RDF syntax vocabulary

Issue #779, opened 2026-02-11
Abbreviations : `TRANSITIVE`, `SYMMETRIC`, and `INVERSE`


afs:
The `TRANSITIVE`, `SYMMETRIC`, and `INVERSE` abbreviations in SHACL Rules give a way to write certain patterns in an immediate and direct and clear way.

`TRANSITIVE` -- this is also useful because it is possible to implement more efficiently than simply following the rule definition shown below. 
More efficient rule-based expression requires e.g. matching for direct property relationship. If built-in this isn't necessary, and the direct triples do not appear in the output graph.

It can also be a useful for caching across rule evaluation. e.g. `rdfs:subClassOf`, `rdfs:subPropertyOf`.

The `SYMMETRIC`  and `INVERSE` abbreviations, when used with care, can help declaring expressivity. However, they can generate large numbers of triples if used with common properties.

See also 
* https://www.oreilly.com/library/view/semantic-web-for/9780123735560/OEBPS/09_chapter07.htm
* http://mlwiki.org/index.php/RDFS-Plus
* #351

----

Illustration - simple implementations:

`TRANSITIVE(:property)`
```
RULE { ?x :property ?z } WHERE  { ?x :property ?y . ?y :property ?z } 
```
or
```
RULE { ?x :property ?z } WHERE  { ?x :property/:property ?z } 
```

`SYMMETRIC(:property)`
```
RULE { ?b :property ?a } WHERE  { ?a :property?b } 
```

`INVERSE(:property1, :property2)`
```
RULE { ?b :property2 ?a } WHERE  { ?a :property1 ?b } 
RULE { ?b :property1 ?a } WHERE  { ?a :property2 ?b } 
```

-----
(<span style="font-size: smaller;"> @TallTed added missing comma + `and` to first line so `SYMMETRIC INVERSE` is not misread as one entity. </span>)

HolgerKnublauch:
What RDF representation would you use for those - or is this only for the native rule format?

afs:
It's still a "rule", it is syntax for a certain case.

It's important to keep the expressivity equivalence of RDF Rules Syntax and SRL.

The RDF syntax represents the structure of the rule set.

The RDF syntax would have entries such as:

```turtle
<http://example/ruleSet1>
    rdf:type srlRuleSet ;
    srl:rules (
        [  rdf:type srl:Rule ;
           srl:transitive true ;
        ]
        [ ... next rule ... ]
     ...
    .
```

[...]

Issue #734, opened 2026-01-24
SHACL Rule grammar - DATA should not allow variables.


afs:
The triple forms for data and the rule head are shared but the data block must not allow variables whereas the head does allow them.

There are two possibilities:

* a new set of productions that disallow variables i.e. duplicate the tree under [TriplesTemplateBlock](https://w3c.github.io/data-shapes/shacl12-rules/#rTriplesTemplateBlock) which is about another 15-20 productions
* Have a note that forbids variables in the DATA block. In effect, have a context flag. 
  (see the [SPARQL Grammar Note 6](https://www.w3.org/TR/sparql12-query/#sparqlGrammar) about
  `INSERT DATA`, `DELETE DATA` ; SPARQL also has context restrictions on blank nodes so  
  "(no) variables x (no) blank nodes" for Turtle forms is a lot of grammar productions.

liviorobaldo:
Idea 💡

We define:

```[14] | Data | ::= 'DATA' vf(TriplesTemplateBlock)```

then we specify in the grammar that `vf(...)`, where vf stands for “variable-free”, is a function that takes the parse tree rooted at a given block and returns the same tree after applying the following transformation:

```[65] | VarOrTerm | ::= Var | iri | RDFLiteral | NumericLiteral | BooleanLiteral | BlankNode | NIL | TripleTerm```

is rewritten as:

```[65] | VarOrTerm | ::= iri | RDFLiteral | NumericLiteral | BooleanLiteral | BlankNode | NIL | TripleTerm```

In this way, the specification remains fully formal, as a grammar should, while avoiding the duplication of 15–20 productions.

I find that using a footnote as the one you suggest is risky, as we mix formal definitions with informal or textual explanations. Moreover, many readers overlook footnotes precisely because they are expected to contain only marginal or supplementary material, which is not the case here. Same considerations of course apply to SPARQL 1.2 as well, but I'm not concerned with that.

afs:
Maybe as a textual marker in the HTML but there are ways to do that without violating [W3C EBNF syntax](https://www.w3.org/TR/xml11/#sec-notation).

Not as a transformation - the grammar is designed to be fed into common parser generators and the concept of transformation does not exist universally.

A textual marker can be done by an intermediate rule `TriplesBlockData` to introduce a name and and an HTML anchor so the grammar requirement can name the point in the grammar.

In SPARQL, these are not foot notes. They are inline and before the grammar table, then reinforced with negative syntax tests.

[...]

PR #976, opened 2026-07-03
Rules draft working copy

afs:
Working copy PR for completing the rules document.

This makes changes visible, while batching them into one for a final approval. This is recognizing that at this time of year, there is less approvals capacity. It is sometimes tricky to continue editing when there are changes in pending PRs.

[See the live editors draft online here](https://raw.githack.com/w3c/data-shapes/rules-draft-2/shacl12-rules/index.html)

Tasks to progress:
* [x] #930
* [x] #734
* [x] #779
* [ ] #870 - more necessary/needed?


afs:
Now would be a good moment for a checkpoint.

afs:
@simonstey -

Updates for `FOR..IN`:

* Grammar change
* syntax tests
* placeholder section

recalcitrantsupplant (approved):
Approving so as not to block, will add to the `FOR .. IN` discussion in the specific issue, otherwise looks good to me

PR #978, opened 2026-07-05
Issue 875: handling disallowing recursive constraints

### LINKED ISSUE #875: SHACL Profile proposal: disallowing recursive constraints
After some discussion with the profiles working group we'd like to propose a profile that focuses on disallowing recursive constraints. The SHACL spec acknowledges that implementations may handle recursion differently, so it seems a reasonable candidate.  A profile that specifically disallows recursive constraints would eliminate this source of non-determinism, ensuring consistent validation outcomes across implementations. The profile would be for SHACL 1.2, but if SHACL 1.0 is defined as a profile of SHACL 1.2, then it would make sense to have a SHACL 1.0 profile that disallows recursive constraints.

While such a profile would necessarily sacrifice some expressiveness, the trade-off seems worthwhile for use cases that prioritize interoperability and predictability. That said, recursion does serve legitimate modeling needs. Three that come to mind are:

### 1. The recursion occurs directly in the shape:

```
ex:PersonShape a sh:NodeShape ;
    sh:targetClass ex:Person ;
    sh:property [
        sh:path ex:hasManager ;
        sh:node ex:PersonShape ;
        sh:minCount 1 ;
    ] .
```
The following data graph produces an infinite validation loop:

```
ex:Alice a ex:Person ; ex:hasManager ex:Bob .
ex:Bob a ex:Person ; ex:hasManager ex:Alice .
```

### 2. The recursion occurs when two shapes reference each other:

```
ex:DepartmentShape a sh:NodeShape ;
    sh:targetClass ex:Department ;
    sh:property [
        sh:path ex:hasMember ;
        sh:node ex:EmployeeShape ;
        sh:minCount 1 ;
    ] .

ex:EmployeeShape a sh:NodeShape ;
    sh:targetClass ex:Employee ;
    sh:property [
        sh:path ex:belongsTo ;
        sh:node ex:DepartmentShape ;
        sh:minCount 1 ;
    ] .
```
The following data graph produces an infinite validation loop:

```
ex:Engineering a ex:Department ; ex:hasMember ex:Alice .
ex:Alice a ex:Employee ; ex:belongsTo ex:Engineering .
```

### 3. Recursive tree validation:

```
ex:FolderShape a sh:NodeShape ;
    sh:targetClass ex:Folder ;
    sh:property [
        sh:path ex:containsFolder ;
        sh:node ex:FolderShape ;
    ] .
```
The following data graph produces an infinite validation loop:

```
ex:FolderA a ex:Folder ; ex:containsFolder ex:FolderB.
ex:FolderB a ex:Folder ; ex:containsFolder ex:FolderA .
```

This is my first pass at examples. There may be others (please let me know). The bottom line is that the data graphs above all lead to an infinite loop and currently it's up to the SHACL implementations to address the issue.

-- Scott

nicholascar:
* text for Issue 875 - Recursion-constrained profile of SHACL

See it at https://raw.githack.com/w3c/data-shapes/ISSUE-875/shacl12-profiling/index.html#disallow-recursive-constraints

### DIFF
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -1535,7 +1535,7 @@ WHERE {
       shapes definitions do not violate OWL constraints on the classes and predicates they target.
     </p>
     <section id="disallow-recursive-constraints">
-      <h4>Profile for Disallowing Recursive Constraints</h4>
+      <h3>Profile for Disallowing Recursive Constraints</h3>
       <div class="issue" data-number="875">&nbsp;</div>
       <p>
         This profile establishes structural validation rules for defining SHACL shapes, explicitly prohibiting 
@@ -1548,7 +1548,7 @@ WHERE {
       </p>
     </section>
     <section id="disallow-recursive-constraints-definition">
-      <h5>Definition of Recursive Constraints</h5>
+      <h4>Definition of Recursive Constraints</h4>
         <p>
           A SHACL shapes graph <strong>SG</strong> conforms to the Non‑Recursive Constraints Profile
          <strong>iff</strong> its <em>shape reference graph</em> (defined below) is <em>acyclic</em>.
@@ -1594,7 +1594,7 @@ WHERE {
        </p>
     </section>
     <section id="disallow-recursive-constraints-shacl">
-       <h5>Validating the Absence of Recursive Constraints</h5>
+       <h4>Validating the Absence of Recursive Constraints</h4>
        <p>
          The following SHACL-SHACL shapes graph can be used to validate that a given shapes graph does not
          contain any recursive constraints.
@@ -1649,11 +1649,11 @@ ex:NoRecursiveShapeReferencesConstraint
          </div>
     </section>
     <section id="disallow-recursive-constraints-examples">
-       <h5>Examples of Disallowed Recursive Constraints</h5>
+       <h4>Examples of Disallowed Recursive Constraints</h4>
        <p>
          The following examples illustrate the types of recursive constraints that are disallowed by this profile.
        </p>
-       <h6>Direct recursion in a single shape example</h6>
+       <h5>Direct recursion in a single shape example</h5>
        <aside class="example" title="Shape raising constraint violation when recursive constraints are present">
          <div class="shapes-graph">
            <div class="turtle">
@@ -1674,7 +1674,7 @@ ex:Bob a ex:Person ; ex:hasManager ex:Alice .
            </div>
          </div>
        </aside>
-       <h6>Mutually referencing shapes example</h6>
+       <h5>Mutually referencing shapes example</h5>
        <aside class="example" title="Two shapes that reference each other, creating a cycle.">
          <div class="shapes-graph">
            <div class="turtle">
@@ -1703,7 +1703,7 @@ ex:Alice a ex:Employee ; ex:belongsTo ex:Engineering .
            </div>
          </div>
        </aside>
-       <h6>Recursive tree example</h6>
+       <h5>Recursive tree example</h5>
        <aside class="example" title="Two shapes that reference each other, creating a cycle.">
          <div class="shapes-graph">
            <div class="turtle">
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -1531,9 +1531,10 @@ WHERE {
   <section id="owl-consistent-profile">
     <h3>OWL Consistent Profile</h3>
     <p>
-      The Owl Consistent profile of SHACL is a subset of the vocabulary elements of SHACL that are selected to ensure
+      The OWL Consistent profile of SHACL is a subset of the vocabulary elements of SHACL that are selected to ensure
       shapes definitions do not violate OWL constraints on the classes and predicates they target.
     </p>
+    <div class="issue" data-number="875">&nbsp;</div>
   </section>
 </section>
 
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -1599,7 +1599,6 @@ WHERE {
          The following SHACL-SHACL shapes graph can be used to validate that a given shapes graph does not
          contain any recursive constraints.
        </p>
-       <aside class="example" title="Shape raising constraint violation when recursive constraints are present">
          <div class="shapes-graph">
            <div class="turtle">
 ex:NonRecursiveConstraintsValidator
@@ -1648,7 +1647,6 @@ ex:NoRecursiveShapeReferencesConstraint
 .
            </div>
          </div>
-       </aside>
     </section>
     <section id="disallow-recursive-constraints-examples">
        <h5>Examples of Disallowed Recursive Constraints</h5>
--- a/shacl12-profiling/index.html
+++ b/shacl12-profiling/index.html
@@ -1547,6 +1547,185 @@ WHERE {
         self-referencing shapes, mutually dependent shapes, and recursive tree structures.
       </p>
     </section>
+    <section id="disallow-recursive-constraints-definition">
+      <h5>Definition of Recursive Constraints</h5>
+        <p>
+          A SHACL shapes graph <strong>SG</strong> conforms to the Non‑Recursive Constraints Profile
+         <strong>iff</strong> its <em>shape reference graph</em> (defined below) is <em>acyclic</em>.
+       </p>
+       <p>
+         Let <strong>R(SG)</strong> be a directed graph where:
+         <ul>
+              <li>
+                Each node is a shape in SG (any RDF node that is a <code>sh:Shape</code>,
+                <code>sh:NodeShape</code>, or <code>sh:PropertyShape</code>, or that is used in a position
+                that denotes a shape as defined below).
+              </li>
+              <li>
+                There is a directed edge s &rightarrow; t in <strong>R(SG)</strong> whenever, in SG, shape
+                <strong>s</strong> references shape <strong>t</strong> via any SHACL term that takes a shape
+                as a value.
+              </li>
+         </ul>
+       </p>
+       <p>
+         At minimum, implementations <em>MUST</em> treat the following as shape reference for conformance with
+         this profile:
+         <ul>
+              <li><code>sh:node</code></li>
+              <li><code>sh:property</code> (edge to the referenced property shape)</li>
+              <li><code>sh:qualifiedValueShape</code></li>
+              <li>the list members of <code>sh:and</code>, <code>sh:or</code>, and <code>sh:xone</code></li>
+              <li><code>sh:not</code></li>
+         </ul>
+       </p>
+       <p>
+         A shapes graph violates this profile if it contains any cycle s &rightarrow; &#x22EF; &rightarrow; s,
+         including:
+         <ul>
+              <li>
+                <strong>direct self-reference</strong> (a shape references itself), and
+              </li>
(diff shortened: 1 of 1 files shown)


shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,194 @@ <h3>Pretty Print Profile</h3>
...
+         </div>
+         <div class="data-graph">
+           <div class="turtle">
+# causes an infinite validation loop
+ex:FolderA a ex:Folder ; ex:containsFolder ex:FolderB .
+ex:FolderB a ex:Folder ; ex:containsFolder ex:FolderA .
+           </div>
+         </div>
+       </aside>
   </section>
```
YoucTagh, line 1722:
```suggestion
   </section>
  </section>
```

I think a missing `</section>` tag is causing the remaining sections to be displayed incorrectly.

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+  sh:targetSubjectsOf sh:not ;
+  sh:targetSubjectsOf sh:and ;
+  sh:targetSubjectsOf sh:or ;
+  sh:targetSubjectsOf sh:xone ;
+
+  # Capture shapes that appear only as list members or property shapes:
+  sh:targetObjectsOf sh:property ;
+  sh:targetObjectsOf rdf:first ;
+
+  sh:sparql nrcp:NoRecursiveShapeReferencesConstraint
```
mgberg, line 1620: Where does `nrcp` come from?

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+              </li>
+              <li>
+                There is a directed edge s &rightarrow; t in <strong>R(SG)</strong> whenever, in SG, shape
+                <strong>s</strong> references shape <strong>t</strong> via any SHACL term that takes a shape
+                as a value.
+              </li>
+         </ul>
+       </p>
+       <p>
+         At minimum, implementations <em>MUST</em> treat the following as shape reference for conformance with
```
mgberg, line 1571: Is it valid to say "at minimum" here? Wouldn't we want this to specify a specific closed set of features to check for? If another system wanted to add additional things to check then I assume that would be treated as a different profile.

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+         <div class="data-graph">
+           <div class="turtle">
+# causes an infinite validation loop
+ex:Engineering a ex:Department ; ex:hasMember ex:Alice .
+ex:Alice a ex:Employee ; ex:belongsTo ex:Engineering .
+           </div>
+         </div>
+       </aside>
+       <h5>Recursive tree example</h5>
+       <aside class="example" title="Two shapes that reference each other, creating a cycle.">
```
mgberg, line 1702: This example has one shape that references itself

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+    sh:path ex:containsFolder ;
+    sh:node ex:FolderShape ;
+  ] .
+           </div>
+         </div>
+         <div class="data-graph">
+           <div class="turtle">
+# causes an infinite validation loop
+ex:FolderA a ex:Folder ; ex:containsFolder ex:FolderB .
+ex:FolderB a ex:Folder ; ex:containsFolder ex:FolderA .
```
mgberg, line 1717: This example is only recursive because a folder can contain other folders _and_ there exists create a loop in the folder tree. In other words, this particular case could be considered a problem because the _data_ is constructed improperly. Is the intent for this profile to catch this case too, i.e. that it wouldn't be allowable under this profile that a folder could contain other folders?

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+       </p>
+       <p>
+         At minimum, implementations <em>MUST</em> treat the following as shape reference for conformance with
+         this profile:
+         <ul>
+              <li><code>sh:node</code></li>
+              <li><code>sh:property</code> (edge to the referenced property shape)</li>
+              <li><code>sh:qualifiedValueShape</code></li>
+              <li>the list members of <code>sh:and</code>, <code>sh:or</code>, and <code>sh:xone</code></li>
+              <li><code>sh:not</code></li>
```
mgberg, line 1578: I'm wondering if this is actually the full set of predicates that would be in scope, or if there are others (e.g. perhaps `sh:condition`, `sh:expression`, `sh:filterShape` if rules are in play?)

shacl12-profiling/index.html
```diff
@@ -1531,11 +1531,195 @@ <h3>Pretty Print Profile</h3>
...
+  sh:targetSubjectsOf sh:property ;
+  sh:targetSubjectsOf sh:qualifiedValueShape ;
+  sh:targetSubjectsOf sh:not ;
+  sh:targetSubjectsOf sh:and ;
+  sh:targetSubjectsOf sh:or ;
+  sh:targetSubjectsOf sh:xone ;
+
+  # Capture shapes that appear only as list members or property shapes:
+  sh:targetObjectsOf sh:property ;
+  sh:targetObjectsOf rdf:first ;
```
mgberg, line 1618: Might need to be careful with this, as it would target every list regardless of context. Maybe a SPARQL-based target would be necessary here. And if it was a SPARQL-based target, then you could filter out shapes that are deactivated (if that was desired).

shacl12-profiling/index.html
TallTed: Also, I'm not sure what "treat the following as shape reference" means. Perhaps there are some missing words, or a reordering of these, that would make it more sensible?

scotthenninger:
@osavkovic could you take a look at this PR? It outlines some problematic recursive use cases and a shack-shacl shape to detect them. It could use your review.

osavkovic:
@scotthenninger 

Mathematically, seems ok.

**Some minor comments:**

- "any SHACL term that takes a shape as a value" — I'm not sure how to interpret this precisely.

- Also: the SHACL stack has constructs like `sh:rule` / `sh:function` etc. where other shape names may occur. I don't know whether these are covered by this definition of the dependency graph (this might actually be relevant to all the profiles, not just this one).

- Possible naming inconsistency: the section is titled "Profile for Disallowing Recursive Constraints," but later refers to itself as the "Non-Recursive Constraints Profile."

**Some remarks on comparisons:**

If we adopt the positive fragment as well, perhaps you would like some comparison between the profiles should probably be given. If the text is too dense to understand, happy to explain better

- *Non-recursive is exactly as expressive as the fragment where no shape references any other shape.* 

Since the dependency graph is acyclic, we can "unfold"/inline all shape definitions starting from the leaves, and end up with a set of shapes that reference no other shapes, with equivalent validation power. Note that this expansion can cause an exponential blow-up (a shape may be referenced several times from within a larger shape), so it might be impractical as an actual rewriting, even though it shows the equivalence in principle.

- *Non-recursive vs. positive.* 

In theory, the non-recursive fragment (and even the recursive fragment with stratified negation) can be rewritten into an "equivalent" positive fragment, at the cost of some changes to the input. We know this from Datalog. However, it wouldn't be practical: for every ground fact not present, e.g. `G = { likes(john, nick) }`, one would have to introduce the complement facts, e.g. `G' = G ∪ { notLikes(nick, john), notLikes(john, john), notLikes(nick, nick) }`, to simulate negation. So: possible in principle, not practical.

- *Full vs. non-recursive.* It's clear that non-recursive is strictly less expressive.

shacl12-profiling/index.html
scotthenninger: s/b ex: - fixed

shacl12-profiling/index.html
scotthenninger: Both valid point. I have reworded this sentence to read:
To determine whether a shapes graph conforms to this profile, implementations <em>MUST</em> treat 
         each of the following properties and list constructs as shape-reference relationships when 
         constructing the shape reference graph and checking it for cycles:

shacl12-profiling/index.html
scotthenninger: This seems to be fixed - let me know

shacl12-profiling/index.html
scotthenninger: A few of points here 1) The data is incorrect, but the shape allows it, and that's what we want to prevent, 2) this is really the same as the self-referential example, but one whose context is illustrative, so I moved up to be a second example in that section, and 3) I added an example on self-referential negation. Again an obvious case, but one that does cause loops.

shacl12-profiling/index.html
scotthenninger: Do you have any examples of using these? I suppose one t=can think of two conditions that contradict each other, but I'm not sure that would cause an infinite loop...

shacl12-profiling/index.html
scotthenninger: I see you point on lists but I'm not exactly clear on using a SPARQL-based target. Would the SPARQL target lists? Can you expound on this for me?

mgberg: No, the SPARQL target wouldn't target lists, it would target shapes. The list traversal stays inside a path expression, so `rdf:first` is only followed in the context of `sh:and`/`sh:or`/`sh:xone` instead of for every list in the shapes graph.

So something like:

```ttl
ex:NonRecursiveConstraintsValidator
  a sh:NodeShape ;

  # Cover both explicitly typed and structurally-used shapes:
  sh:targetClass sh:Shape ;
  sh:targetClass sh:NodeShape ;
  sh:targetClass sh:PropertyShape ;
  sh:targetSubjectsOf sh:node ;
  sh:targetSubjectsOf sh:property ;
  sh:targetSubjectsOf sh:qualifiedValueShape ;
  sh:targetSubjectsOf sh:not ;
  sh:targetSubjectsOf sh:and ;
  sh:targetSubjectsOf sh:or ;
  sh:targetSubjectsOf sh:xone ;
  sh:targetSubjectsOf sh:memberShape ;
  sh:targetSubjectsOf sh:reifierShape ;
  sh:targetSubjectsOf sh:someValue ;

  # Capture shapes that appear only as property shapes:
  sh:targetObjectsOf sh:property ;

  # Capture shapes that appear only as members of sh:and/sh:or/sh:xone lists:
  sh:target ex:LogicalConstraintMemberTarget ;

  sh:sparql ex:NoRecursiveShapeReferencesConstraint
.

ex:LogicalConstraintMemberTarget
  a sh:SPARQLTarget ;
  sh:select """
    SELECT DISTINCT ?this
    WHERE {
      ?shape (sh:and|sh:or|sh:xone)/rdf:rest*/rdf:first ?this .
    }
  """
.
```

That path is the same one already used in the `sh:and|sh:or|sh:xone` branch of `ex:NoRecursiveShapeReferencesConstraint`, which keeps the target and the constraint from drifting apart.

Of course, as I write this, I realize that SPARQL-based targets aren't in the SHACL 1.2 SPARQL-Related Features document, so my original point may be irrelevant. I'll open an issue to see if that's intentional or not.

scotthenninger: @mgberg I responded to some of your comments - can you take a look at the PR so we can try to close this out?

PR #981, opened 2026-07-06
Add/update terminology for validation report provenance

### LINKED ISSUE #963: Additions/rework for validation report provenance
To track how the validity of a graph changes over time and ensure validation results are reproducible, it would be convenient if Validation Report could be self-contained and could carry full context about how it was produced.

### 1. Replace `sh:validationReport` with its inverse

`sh:usedShapesGraph` points outward from the report (`ValidationReport -> ShapesGraph`), but for data graphs the direction is reversed: `sh:validationReport` points inward (`DataGraph -> ValidationReport`).

Remove `sh:validationReport` (domain: `DataGraph`, range: `ValidationReport`) and introduce `sh:validatedDataGraph` or something similar (domain: `ValidationReport`, range: `DataGraph`). This makes the report the subject of both graph references, consistent with `sh:usedShapesGraph`.

The profiling spec's CONSTRUCT rule deriving `sh:conformsTo` would be updated to query from the report outward rather than from the data graph inward.

### 2. Introduce class for validator configuration

The spec notes that reports "MAY contain additional information such as provenance metadata," but defines no structure for it. Without knowing the implementation, its version, entailment regime, or other processor-specific settings, a validation result cannot be reliably reproduced.

Define `sh:ValidatorSettings` or similar as a class whose properties are intentionally opaque to the SHACL spec. Implementations may hang whatever they need on instances of it, such as implementation name, version, entailment regime, etc., then link to it fromthe report via `sh:usedValidatorSettings` or similar (domain: `ValidationReport`, range: `sh:ValidatorSettings`).

mgberg:
Closes #963

* [See Core rendered online here](https://raw.githack.com/w3c/data-shapes/issue-963/shacl12-core/index.html)
* [See Profiling rendered online here](https://raw.githack.com/w3c/data-shapes/issue-963/shacl12-profiling/index.html)

## Validation Report Provenance: Self-Contained Reports

This PR makes validation reports self-contained with respect to provenance by ensuring the report is always the subject of references to its inputs. Previously, the link from a data graph to its report used `sh:validationReport` (domain: `DataGraph`, range: `ValidationReport`), which was inconsistent with `sh:usedShapesGraph` (domain: `ValidationReport`, range: `ShapesGraph`). This also introduces a structured way for processors to describe their configuration at the time of validation.

### Changes

**1. Replace `sh:validationReport` with `sh:usedDataGraph`**

Removed `sh:validationReport` (DataGraph → ValidationReport) and introduced `sh:usedDataGraph` (ValidationReport → DataGraph). The report is now the subject of both graph references, consistent with `sh:usedShapesGraph`. The version IRI of a data graph may also be used as the object, mirroring the existing allowance for `sh:usedShapesGraph`.

**2. Introduce `sh:ProcessorConfiguration` class and `sh:usedConfiguration` property**

Defined `sh:ProcessorConfiguration` as a class whose properties are intentionally not specified by SHACL. This name was chosen to be consistent with the language used elsewhere in the spec. Implementations may attach whatever they need (processor name, version, runtime parameters, etc.). Linked from reports via `sh:usedConfiguration`.

**3. Update the "conforms to" reasoning rule**

The CONSTRUCT and SHACL Rules forms in the profiling spec now query from the report outward rather than from the data graph inward.

**4. Add SHACL-SHACL shape for ValidationReport**

`shsh:ValidationReportShape` constrains `sh:usedDataGraph`, `sh:usedShapesGraph`, and `sh:usedConfiguration` to `sh:maxCount 1`.

### Open Questions

1. ~~**Cardinality of `sh:usedDataGraph` and `sh:usedShapesGraph`:** This currently assumes at most one value per report. Is it always the case that there would only be one identified data graph or shapes graph in this context?~~ Cardinality restrictions removed.

2. ~~**Placement of provenance terms in Core vs. Profiling:** The new provenance properties (`sh:usedDataGraph`, `sh:usedShapesGraph`, `sh:usedConfiguration`) and `sh:ProcessorConfiguration` are currently listed in the Core document because they describe the structure of a validation report regardless of whether profiling is in use. Is this placement agreeable, or should these terms live exclusively in the Profiling document? Also, note that `sh:conformsTo` is currently NOT defined in the Core spec.~~  Added a brief entry for it in Core.

3. ~~**Example placement:** An illustrative example of a self-contained report (including `sh:ProcessorConfiguration` usage) is currently in the Profiling spec. If the terms are defined in the Core document, should the example be moved there as well?~~ Pointer to example put in Core.


nicholascar:
> Cardinality of `sh:usedDataGraph` and `sh:usedShapesGraph`

Tools like pySHACL [allow multiple Data Graphs to be validated all in one go or individually](https://github.com/RDFLib/pySHACL/#command-line-use) but I think we don't need to specify this level of detail. We can just define the properties without cardinalities in this vocab and leave it to implementers to constrain as needed, perhaps using SHACL...

> Placement of provenance terms in Core vs. Profiling

I prefer the terms in Core. It keeps everything in one place and saves further artifact maintenance. And it costs little for Core to keep them there: people can just not use them if not useful. 

[`sh:conformsTo` is defined in the SHACL Core vocab](https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L226-L234) but yes, it currently has no definition in the core Spec.

> Example placement

Perhaps best would just be a note added to Core that Profiling contains description and examples for the profiling terms.

shacl12-vocabularies/shacl.ttl
```diff
@@ -241,14 +248,6 @@ sh:conformanceDisallows
 	rdfs:range sh:Severity ;
 	rdfs:isDefinedBy sh: .
 
-sh:validationReport
```
HolgerKnublauch, line 244: Just to double-check this was deleted intentionally

shacl12-core/index.html
```diff
@@ -2506,7 +2506,12 @@ <h3>Validation Report</h3>
 					<h4>Validation Report (sh:ValidationReport)</h4>
 					<p>
 						The result of a <a>validation</a> process is an RDF graph with exactly one <a>SHACL instance</a> of <code>sh:ValidationReport</code>.
-						The RDF graph MAY contain additional information such as provenance metadata.
+						The RDF graph MAY contain additional provenance metadata.
+						In particular, a validation report MAY use <code>sh:usedDataGraph</code> to identify the <a>data graph</a> that was validated,
```
HolgerKnublauch: Could you turn those into hyperlinks to the sub-sections?

shacl12-core/index.html
```diff
@@ -2558,6 +2563,43 @@ <h5>Syntax Checking of Shapes Graph (sh:shapesGraphWellFormed)</h5>
 							<a>validation</a> process is <a>well-formed</a>.
 						</p>
 					</section>
+					<section id="usedDataGraph">
+						<h5>Used Data Graph (sh:usedDataGraph)</h5>
+						<p>
+							A SHACL instance of <code>sh:ValidationReport</code> in the results graph MAY have a value for the property <code>sh:usedDataGraph</code>.
+							The value identifies the <a>data graph</a> that was validated by the <a>processor</a> to produce the report.
```
HolgerKnublauch: Please state that the value must be nodeKind IRI, as the shacl-shacl is not enough.

shacl12-core/index.html
```diff
@@ -2558,6 +2563,43 @@ <h5>Syntax Checking of Shapes Graph (sh:shapesGraphWellFormed)</h5>
...
+							The version IRI of a data graph MAY also be used as a value for this property.
+						</p>
+					</section>
+					<section id="usedShapesGraph">
+						<h5>Used Shapes Graph (sh:usedShapesGraph)</h5>
+						<p>
+							A SHACL instance of <code>sh:ValidationReport</code> in the results graph MAY have a value for the property <code>sh:usedShapesGraph</code>.
+							The value identifies the <a>shapes graph</a> that was used by the <a>processor</a> during the <a>validation</a> process.
```
HolgerKnublauch: same nodeKind IRI

HolgerKnublauch (approved):
LGTM, although I didn't read the profiling changes because I have no background on that. Hopefully Nick has checked those.

shacl12-vocabularies/shacl.ttl
mgberg: It was. We added that a month or two ago but realized it would be more consistent with other predicates to define its inverse instead.

mgberg:
> [sh:conformsTo is defined in the SHACL Core vocab](https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L226-L234) but yes, it currently has no definition in the core Spec.

@nicholascar I was more concerned about the fact that Profiling explicitly says that `sh:conformsTo` is defined by Core, but Core doesn't actually mention it anywhere.

mgberg:
> [sh:conformsTo is defined in the SHACL Core vocab](https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-vocabularies/shacl.ttl#L226-L234) but yes, it currently has no definition in the core Spec.

@nicholascar based on the conversation this morning, I added a brief section to Core for `sh:conformsTo` to at least ensure the fragment points to something, with a reference to Profiling for more detail.

HolgerKnublauch:
How do you guys feel about renaming sh:conformsTo to sh:conformsToShapesGraph (or sh:conformsToGraph) to keep it self-descriptive and distinct from shnex:conformsToShape that I want to add in https://github.com/w3c/data-shapes/pull/972

mgberg:
> How do you guys feel about renaming sh:conformsTo to sh:conformsToShapesGraph (or sh:conformsToGraph) to keep it self-descriptive and distinct from shnex:conformsToShape that I want to add in #972

As noted in the discussion on #972, `sh:conformsTo` was intended to parallel `dcterms:conformsTo`, but I personally have no issue with renaming it to `sh:conformsToShapesGraph` to disambiguate/improve clarity. 

@nicholascar, @ajnelson-nist what do you think?

ajnelson-nist:
I'm fine with `sh:conformsToShapesGraph`, especially if other `...:conformsTo...` predicates are coming in.

mgberg:
> I'm fine with `sh:conformsToShapesGraph`, especially if other `...:conformsTo...` predicates are coming in.

@nicholascar, @HolgerKnublauch : I went ahead and made that change. Note that I did *not* update the diagrams in the Profiling doc, as I recall we were talking about redoing the diagrams anyway a week or two ago.

Also @nicholascar , @ajnelson-nist : I just noticed that although the Profiling doc says sh:conformsTo(ShapesGraph) is a subproperty of dcterms:conformsTo, that assertion is not present in the shacl.ttl file. I recall debating this quite a while ago, but don't remember what the final outcome was- do we want to add that assertion or not?

HolgerKnublauch:
> Also @nicholascar , @ajnelson-nist : I just noticed that although the Profiling doc says sh:conformsTo(ShapesGraph) is a subproperty of dcterms:conformsTo, that assertion is not present in the shacl.ttl file. I recall debating this quite a while ago, but don't remember what the final outcome was- do we want to add that assertion or not?

Please don't. SHACL doesn't depend on dcterms, and I would caution against unnecessary references.

TallTed:
> Please don't. SHACL doesn't depend on `dcterms`, and I would caution against unnecessary references.

I submit that *either* the assertion that `sh:conformsTo` is a subproperty of `dcterms:conformsTo` should be added to the `shacl.ttl`, *or* Line 908 of `shacl12-profiling/index.html` (i.e., the seemingly equally unnecessary `dcterms` reference, `This predicate is a sub property of the [[DCTERMS]] predicate <code>dcterms:conformsTo</code> used in [[[dx-prof]]].`) should be deleted.

ajnelson-nist:
> > I'm fine with `sh:conformsToShapesGraph`, especially if other `...:conformsTo...` predicates are coming in.
> 
> @nicholascar, @HolgerKnublauch : I went ahead and made that change. Note that I did _not_ update the diagrams in the Profiling doc, as I recall we were talking about redoing the diagrams anyway a week or two ago.
> 
> Also @nicholascar , @ajnelson-nist : I just noticed that although the Profiling doc says sh:conformsTo(ShapesGraph) is a subproperty of dcterms:conformsTo, that assertion is not present in the shacl.ttl file. I recall debating this quite a while ago, but don't remember what the final outcome was- do we want to add that assertion or not?

@mgberg : I recall we were going to delegate that link upstream to PROF.  I'm not sure offhand what we should put into our doc.

mgberg:
@nicholascar @ajnelson-nist I updated the diagrams to account for the new/modified terms. 

However, I realized that If we do NOT define `sh:conformsToShapesGraph` as a subproperty of `dcterms:conformsTo`, then we need to rework some things:

1. Vocabulary definition table: In Section 5, the property definition for `sh:conformsToShapesGraph` contains the following (an easy fix):

    > **Sub-property of:** `dcterms:conformsTo`

2. Section 4.1.2 (Implications for data) explicitly asserts the subproperty relationship (again, an easy fix):

    > This predicate is a sub property of the [[DCTERMS]] predicate `dcterms:conformsTo` used in [[[dx-prof]]].

3. Also in Section 4.1.2, the text references the dx-prof property chain axiom to argue that conformance to a shapes graph transitively implies conformance to the specification the shapes graph profiles:

      > Following the [[[dx-prof]]] axiom of `dcterms:conformsTo owl:propertyChainAxiom ( dcterms:conformsTo prof:isProfileOf )` then, if given
      >
      > `<DG1> sh:conformsToShapesGraph <SG1> .`
      >
      > and
      >
      > `<SG1> prof:isProfileOf <S1> .`
      >
      > we can conclude that
      >
      > `<DG1> sh:conformsTo <S1> .`

     This inference relies on the subproperty relationship existing in order for the property chain to fire. Without the subproperty, this reasoning doesn't work.

    How do you want to deal with this? We could replace this with a rule that accomplishes the same thing without relying on the subproperty assertion, for example.

ajnelson-nist:
I support using a rule, especially so we can dogfood some of the Rules work.

mgberg:
> I support using a rule, especially so we can dogfood some of the Rules work.

@nicholascar @ajnelson-nist take a look at the most recent commit, I took a stab at rewriting that to use the new rules. Note when reading the changes that the new rules do not support variable-length property paths.

shacl12-profiling/index.html
```diff
@@ -904,31 +904,12 @@ <h4>Characterizing the profiling situation</h4>
...
-      <p>
-        we can conclude that
-      </p>
-      <div class="turtle" data-include="examples/12.ttl" data-include-format="text"></div>
-      <p>
-        This reasoning allows for static conformance claims of data to profiles of specifications and specifications
-        themselves to be made which will have utility for data whose conformance to these things must be known but which
-        cannot be continuously validated.
+        Following <a href="#fig-profiling-of-specs"></a>, if we have <code>&lt;DG1> sh:conformsToShapesGraph &lt;SG1></code>
+        and <code>&lt;SG1> prof:isProfileOf &lt;S1></code>, then via the <a href="#rule-conformstospecification">conforms to specification</a>
```
ajnelson-nist:
```suggestion
        and <code>&lt;SG1> prof:isProfileOf &lt;S1></code>, then via the <a href="#rule-conformstospecification"><code>conformsTo</code> specification</a>
```

ajnelson-nist: The English feels a little funny versus the anchor-wrapped text - should "rule" be in the anchor?

shacl12-profiling/index.html
```diff
@@ -1407,31 +1472,67 @@ <h3>Rule: conforms to</h3>
...
+}
+WHERE {
+  ?x sh:conformsToShapesGraph ?y .
+  ?y prof:isProfileOf ?z .
+}
+    </div>
+    <p class="note">
+      The result of this rule also enables the [[[dx-prof]]] property chain axiom
+      <code>dcterms:conformsTo owl:propertyChainAxiom ( dcterms:conformsTo prof:isProfileOf )</code>
+      to propagate conformance claims up a profile hierarchy.
```
ajnelson-nist, line 1533: Somehow in this block, there should be a link to [this Issue](https://github.com/w3c/dx-prof/issues/29), which tracks the noted property chain axiom as at-risk.  Apologies, I forget that tracking practice.

shacl12-profiling/index.html
mgberg: As for the English, would the following be better?
```suggestion
        and <code>&lt;SG1> prof:isProfileOf &lt;S1></code>, then via the <a href="#rule-conformstospecification">'conforms to specification' rule</a>
```

As for the anchor, I just followed the pattern that @nicholascar already used for the other rule. If we want to change this anchor to something else, then we probably should change the other too.

nicholascar (approved):
Approving so we can get all these many edits in. There wlll be some follow-up edits to update diagrams and so on, but we can do another PR for those.

PR #977, opened 2026-07-05
#962 and #954: Added example for sh:deactivated and srl:text

HolgerKnublauch:
I am aware that the RDF Syntax is a stump right now and lacks the prose to explain the mapping. Once that is started, the description should explain how srl:text is interpreted and that sh:deactivated simply makes the rule invisible to the rule set.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-962/shacl12-rules/index.html#rdf-rules-syntax)

Issue #962, opened 2026-06-25
Allow deactivation of individual rules


HolgerKnublauch:
This was discussed starting in https://github.com/w3c/data-shapes/issues/956#issuecomment-4785239744 but is moved into its own ticket for visibility.

In SHACL-AF, rules can be deactivated, see https://w3c.github.io/shacl/shacl-af/#deactivated

The same feature should exist for the new rules syntax, at least in the RDF syntax (because a typical scenario is to combine rules from multiple graphs and have local triples that modify the activation of imported rules).

Proposal: When building a rule set, ignore those rules that have 

```?rule sh:deactivated true```

in the rules graph.

HolgerKnublauch:
I was going to prepare a PR for this but the current section on the RDF syntax is rather incomplete so there was no natural place to describe the semantics. Anyway, I added the syntax to the example in

https://github.com/w3c/data-shapes/compare/rules-draft...issue-962?expand=1

HolgerKnublauch:
Can be closed assuming the SHACL-AF rules get merged into SHACL-SPARQL. I still believe rule deactivation is needed for any RDF syntax, but I don't really care anymore.

Issue #954, opened 2026-06-21
Rules RDF Syntax of individual rules: Do we even need head/body triples?


HolgerKnublauch:
Looking at the current Rules RDF syntax draft:

```
RULE { ?x :bothPositive true . }
WHERE { ?x :p ?v1  FILTER ( ?v1 > 0 )  ?x :q ?v2  FILTER ( ?v2 > 0 )  }
```

is represented in RDF as

```
[
      rdf:type srl:Rule;
      srl:head (
        [ srl:subject [ srl:varName "x" ] ; srl:predicate :bothPositive ; srl:object true ]
      ) ;
      srl:body (
        [ srl:subject [ srl:varName "x" ]; srl:predicate :p ; srl:object [ srl:varName "v1" ] ]
        [ srl:filter [ sparql:greater-than ( [ srl:varName "v1" ] 0 ) ] ]
        [ srl:subject [ srl:varName "x" ] ; srl:predicate :q ; srl:object [ srl:varName "v2" ] ]
        [ srl:filter [ sparql:greater-than ( [ srl:varName "v2" ] 0 ) ] ]
      );
]
```

Question: do we anticipate that applications really want to walk through these triples? Asking because it's not friendly to humans, and also loses the formatting including 7.2 comments. Both issues were typical feedback on the SPIN RDF format (which worked similar to the RDF above).

An alternative would be

```
[
      rdf:type srl:Rule;
      srl:text """
          RULE { ?x :bothPositive true . }
          WHERE { ?x :p ?v1  FILTER ( ?v1 > 0 )  ?x :q ?v2  FILTER ( ?v2 > 0 )  }
      """
]
```

In SPIN we later added a similar property to store the SPARQL queries as text string. The only issue there is about prefix declarations, but #176 hopefully addresses that.

Dropping the RDF details has also the advantage that future changes or extensions to the rule language do not require changes to the RDF syntax.

afs:
The unit is the rule set. It would make more sense to have `srl:text` be a rule set.

A rule on it's own is a restricted `CONSTRUCT` statement, or more properly a SPARQL Update, and does not need the restrictions imposed on rules.

If a system attaches rules in many places in a large graph, there is then the question of the execution model. Probably there are multiple rule sets needed so that adding rules does not break something elsewhere. Each rule can be valid (well-formed) but the collection is invalid (stratification for negation and assignment).

[...]

PR #982, opened 2026-07-07
#954 and #962: Added example for sh:deactivated and srl:text

HolgerKnublauch:
This PR replaces https://github.com/w3c/data-shapes/pull/977 with better base branch

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-954-962/shacl12-rules/index.html#rdf-rules-syntax)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -253,6 +253,10 @@
             <td><code>http://www.w3.org/2000/01/rdf-schema#</code></td>
           </tr>
           <tr>
+            <td><code>sh:</code></td>
+            <td><code>http://www.w3.org/ns/shacl#</code></td>
+          </tr>
+          <tr>
             <td><code>srl:</code></td>
             <td><code>http://www.w3.org/ns/shacl-rules#</code></td>
           </tr>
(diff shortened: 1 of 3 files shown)


robert-david (approved):

HolgerKnublauch:
I will give the reviewers more time before merging this. It does introduce the use of some properties from the sh: namespace. So please veto if this is an issue.

afs:
-1 to merging this PR at this time.

As an editor, I set it to "draft" 2 days ago because ongoing discussions elsewhere are relevant.

`sh:deactivate` is already covered by the emerging SHACL Rules framework.

See discussion https://github.com/w3c/data-shapes/pull/984#discussion_r3565486740 - option 1.

The natural possible properties in the SHACL Rules Framework would be type `sh:SRLType` and `sh:srlRuleSet` for the text, which may include syntax considerations for pre-binding.

HolgerKnublauch:
LOL. I wonder why I anticipated that.

shacl12-rules/index.html
```diff
@@ -1765,8 +1769,32 @@ <h3>RDF Rules Syntax</h3>
         [ srl:subject [ srl:varName "x" ] ; srl:predicate :q ; srl:object [ srl:varName "o" ] ; ]
       )
     ]
+    :ExampleDeactivatedRuleWithURLAndTextSyntax
   ) .
 
+:ExampleDeactivatedRuleWithURIAndTextSyntax
```
simonstey, line 1775: `:ExampleDeactivatedRuleWithURLAndTextSyntax` 

vs 

`:ExampleDeactivatedRuleWithURIAndTextSyntax` 

is it URL or URI?

shacl12-rules/rules-rdf-syntax/rdf-syntax-vocab.ttl
simonstey, line 2:
```suggestion
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
```

since srl:text uses xsd:string

shacl12-rules/index.html
simonstey, line 1692:
```suggestion
PREFIX rdfs: &lt;http://www.w3.org/2000/01/rdf-schema#&gt;

```

shacl12-rules/index.html
```diff
@@ -1765,8 +1769,32 @@ <h3>RDF Rules Syntax</h3>
...
+  rdf:type srl:Rule ;
+  rdfs:label "Example deactivated rule" ;
+  rdfs:comment "This rule illustrates the alternative srl:text syntax." ;
+  sh:deactivated true ;
```
simonstey, line 1779: what this means for SRL Rules isnt defined yet right? we should add a normative statement to the Rule Set Evaluation section that states that rules with `sh:deactivated true` are excluded from evaluation (and from stratification/dependency-graph construction).

simonstey (changes requested):

shacl12-rules/index.html
HolgerKnublauch: I think the PR can be closed because it's blocked with -1 by Andy and I have already closed the corresponding ticket. All these cases are (better) handled with the sh:SPARQLRule/sh:construct syntax in SHACL SPARQL.

Issue #765, opened 2026-02-06
Attaching rules to shapes


afs:
Should we include attaching SHACL (1.2) Rules to shapes?
If so, what does it mean given the difference in execution semantics?

Subsidiary question:
In practice, how often are constriants and AF-rules written on the same shape? 
If they are, how are the rules being used, in practice, to influence the validation?

Sketch:

```turtle
[] rdf:type sh:NodeShape ;
    sh:rule 
      [ a srl:SHACLRule ;
        srl:ruleSet "...";    ## SRL syntax
        sh:prefixes ... ;
      ];
```
```turtle
[] rdf:type sh:NodeShape ;
    sh:rule
      [ a srl:SHACLRule ;
        srl:ruleSet [ ... RDF syntax ... ] ;
      ];
```
Ruleset execution has `$this` bound to the current node shape. This is available to rule body evaluation:

https://www.w3.org/TR/shacl12-rules/#eval-rule
```
let SEQ0: Solution sequence = { ("this", RDF Term of current NodeShape) }
```
Well-formedness changes to compensate for this.

(given the more restricted matching, substitution of `$this` by it's RDF term should also work)

A problem:
`sh:rule` (AF) has it's own execution model. We would have to either modify that (turn it off) or make attached rules "run once". The latter option seems pointless because it is then no more than a way to write a `CONSTRUCT` rule in another syntax.

HolgerKnublauch:
We (and customers) *only* have experience with attaching rules to shapes because that was the default supported technique in SHACL-AF. I definitely think this is a useful approach, because it makes it possible to naturally organize rules by grouping them to the focus nodes that they fit to, similar to OO programming where rules play the role of functions attached to a class. This is IMHO better than the potential spaghetti code of flat rule lists.

I guess attaching a rule to a shape means that the focusNode/this iterates over the targets of the shape. If that is technically too difficult for all target types (e.g. targetWhere, targetNode with node expressions) then I think it would be fine to limit the attachment to classes and shapes that have targetClass. This is by far the most common case anyway.

I assume SHACL 1.2 Rules uses its own namespace, so the property doesn't need to be sh:rule.

Disclaimer: I have not tracked the details of the evolving SHACL 1.2 Rules work yet, so I cannot comment on the technical challenges. But clearly, if something is called SHACL Rules then it should have shapes as a first class citizen. Otherwise, it could just become a stand-alone W3C spec.

afs:
[...]

PR #984, opened 2026-07-08
#766: Added SHACL-AF rules to SHACL-SPARQL

### LINKED ISSUE #766: Handling of SHACL-AF rules
What should we do with SHACL-AF Rules?

Options include:
1. Leave alone
2. Incorporate (some of) them into the SHACL 1.2 Rules document

The AF Rule types:
* [AF Triple Rules](https://w3c.github.io/shacl/shacl-af/index.html#TripleRule)
* [AF SPARQLRules](https://w3c.github.io/shacl/shacl-af/index.html#SPARQLRule)
* [AF Property Value Rules](https://w3c.github.io/shacl/shacl-af/index.html#PropertyValueRule) -- [already in SHACL 1.2]([sh:values](https://www.w3.org/TR/shacl12-core/#syntax-rule-path-defaultValue))

See also:
* #765
* #516

HolgerKnublauch:
Closes #766

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-766/shacl12-sparql/index.html#sparql-rules)


HolgerKnublauch:
This is ready for review now. One open issue that the new section mentions is the handling of iteration. Currently we only specify a single iteration. In practice, it is trivial to do a fix point iteration that keeps running. The main concern there is when new bnodes or infinite amounts of triples are created continuously. This could be defined as an undefined scenario, while in practice implementations may just set a max iteration count and stop with a warning when that happens.

Practically speaking, it would be great to have SOME version of this on the main branch and then make smaller increments from there, with dedicated follow-up PRs. The text that is here right now is essentially what was on SHACL-AF for years.

liviorobaldo:
Hi @HolgerKnublauch, yes, that's exactly what I was pointing out in my email. I was suggesting that we briefly mention here that "SPARQL-based rules may lead to infinite loops when repeatedly executed until saturation, and may also exhibit non-deterministic behaviour due to operators such as `NOT EXISTS`".

To avoid infinite loops, implementations can impose a maximum iteration count. Avoiding non-determinism, however, is the user's responsibility, who must use `sh:order` to that end. 

Nevertheless, based on my experience with use cases I worked on, I would say that relying on `sh:order` is not only verbose and tedious, but also risky and error-prone. Users may run a set of rules on one implementation, observe that everything works as expected, and conclude that there is no need to specify `sh:order`. They then run the same rules on another implementation and obtain different results... because the need for an explicit order was always there: it was simply "masked" by the fact that the first implementation happened to execute the rules in the order the user expected. In other words, users can easily overlook the need to establish an ordering that is portable across implementations.

Therefore, the fact that rule execution yields the same results across implementations must be actually _formally demonstrated_. This is precisely what I had to do in my paper on the Time Ontology, in order to get it published.

SHACL 1.2 rules avoid these issues altogether by defining stratification and run-once rules.

I think we should explain this briefly in this section. What do you both (and anyone else following the discussion) think?

afs:
The name "SHACL Transformation Rules" can reflect that the control is with the user - and include the following changes:

* include INSERT-DELETE-WHERE
* pre-binding is based on a mechanism that is engine neutral, then all standards-based triple stores can be used.

HolgerKnublauch:
@afs SPIN did support INSERT/DELETE. INSERT doesn't add value over CONSTRUCT, so this is all about DELETE. I am not aware that support for deletes was used or requested much. In practice such use cases are better handled using scripting such as our JavaScript integration where updates are determined dynamically.

FWIW since your current draft is essentially a subset of SPARQL, there could be a profile of SHACL-SPARQL Rules that is limited to that same subset.

HolgerKnublauch:
> I think we should explain this briefly in this section. What do you both (and anyone else following the discussion) think?

Yes we should explain and highlight these problems in the document. This language has a design philosophy similar to imperative languages. For example, anyone can easily do endless loops using `while(true);` but this doesn't render these programming languages useless. The responsibility is shifted to the developer and make sure these scenarios are avoided. The rule language itself doesn't prevent them, and people could easily do infinite loops using

```
CONSTRUCT {
    ?b rdf:type rdfs:Resource .
}
WHERE {
    BIND (BNODE() AS ?b) .
}
```

If you have experience and best practices from the other rules document then we should incorporate that.

Practically speaking, I suggest we try to get this first version into the main branch soon and create individual follow-up tickets to address the things that we want to improve, esp the handling of iteration. It will become easier to make incremental changes when the diff becomes smaller and focused.

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+					then it reports a <a>failure</a>.
+				</p>
+				<p>
+					At no time are inferred triples visible to the <a>shapes graph</a>, i.e. it is impossible for rules
+					to modify the definitions of rules or shapes.
+				</p>
+			</section>
+
+			<section id="SPARQLRule">
+				<h3>SPARQL Rules</h3>
```
afs, line 2027:
```suggestion
				<h3>SPARQL CONSTRUCT Rules</h3>
```
to focus on what it does
or
```suggestion
				<h3>SHACL SPARQL Rules</h3>
```
to be clear it is part of the framework.

liviorobaldo:
> I am not aware that support for deletes was used or requested much. In practice such use cases are better handled using scripting such as our JavaScript integration where updates are determined dynamically.

@HolgerKnublauch, I think `DELETE` would be extremely useful for removing redundant triples, as [discussed in the past](https://github.com/w3c/data-shapes/discussions/603#discussioncomment-14798786). Of course, the same can already be achieved with a standard `DELETE WHERE` SPARQL query, JavaScript, or other mechanisms, so I don't see this as a blocker. However, extending SHACL-SPARQL and/or SHACL 1.2 to support it would be a worthwhile enhancement.

Redundant triples are often generated when intermediate results are needed to compute the final results. Once the final results have been obtained, however, those intermediate triples are no longer needed and can safely be removed, even though they remain logically consistent with the final graph. This is the motivation behind supporting `DELETE WHERE` rules. I have already encountered a couple of real-world use cases where this functionality would have been beneficial.

This is something that could still be added in the future, together with aggregates and the other "future work" ideas mentioned by @afs in one of our recent meetings. The main challenge with `DELETE` is that it naturally exacerbates the risk of infinite loops. However, making all `DELETE` rules run once by definition, and placing them in the last stratum by default, could be a reasonable compromise.

HolgerKnublauch:
@liviorobaldo Ok I see what you mean, and also had use cases for temporary triples in the past, for example to remember which source resource maps to which target resource, and whether during OWL2SHACL conversion an OWL axiom was already covered or not.

With RDF 1.2 reification we now have additional ways of marking such triples for later clean up. For example, we could introduce a dedicated property such as <<( A B C )>> sh:tempTriple true, just like we may want to add a generic provenance mechanism that can be used to annotate which rules have produced which triples.

This does sound like a separate issue though, and should become its own thread IMHO.

liviorobaldo:
> For example, anyone can easily do endless loops using `while(true)`; 

Indeed, I also often use `while (true)` as an example to illustrate that avoiding infinite loops is ultimately the responsibility of the programmer (or user).

> If you have experience and best practices from the other rules document then we should incorporate that.

Well, I think run-once rules are the best option. If one needs a fixed number of iterations, e.g., two or three, one can simply define two or three run-once rules explicitly, with each rule triggering on the results produced by the previous one. In real-world use cases, however, one never needs an infinite number of iterations, even though an infinite sequence of entailments may exist in theoretical or abstract terms.

As for other "best practices", as I said, one should formally demonstrate that infinite loops cannot occur. The way I do this (and, in fact, the only way that has come to mind) is to use `NOT EXISTS` clauses that prevent rules from firing once they have already produced their outputs. In effect, however, this is just a SHACL-SPARQL workaround for emulating run-once rules, because you are essentially saying: "run this rule only if its outputs do not already exist in the graph".

> Practically speaking, I suggest we try to get this first version into the main branch soon and create individual follow-up tickets to address the things that we want to improve, esp the handling of iteration. It will become easier to make incremental changes when the diff becomes smaller and focused.

Ok, I was about to create a PR to incorporate the points discussed above, but from what I understand, you'd prefer that we finish this PR before opening any new ones. So I'll wait.

liviorobaldo:
> This does sound like a separate issue though, and should become its own thread IMHO.

@HolgerKnublauch, I completely agree, including with the point about reification. Perhaps for the time being @afs could just add it to the list of future work (assuming there is one, that's what I understood).

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
 		</section>
 
+		<section id="sparql-rules">
+			<h2>SPARQL-based Inference Rules</h2>
+			<p>
+				SHACL defines an RDF vocabulary to describe <a>shapes</a> - collections of <a>constraints</a> that apply to a set of nodes.
+				Shapes can be associated with nodes using a flexible <a>target</a> mechanism, e.g. for all instances of a class.
+				One focus area of SHACL is data validation.
+				However, the same principles of describing data patterns in shapes can also be exploited for other purposes.
+				<a>SHACL rules</a> build on SHACL to form a light-weight RDF vocabulary for the exchange of <a>rules</a> that can be used
```
liviorobaldo, line 1748: I think calling these simply "SHACL rules" is quite confusing. We should clearly distinguish them from the SHACL 1.2 rules:

https://w3c.github.io/data-shapes/shacl12-rules/

Therefore, I propose updating the terminology throughout this document. We could call them **"SHACL-SPARQL rules"** or **"SPARQL-based inference rules"**, as in the heading below. I slightly prefer the latter: it is a bit more verbose, but it clearly distinguishes them from the SHACL 1.2 rules as well as from the "old" SHACL-AF SPARQL rules. In any case, since we decided to drop SHACL-AF Triple Rules, I think the term "SPARQL" should appear somewhere in the name.

Conversely, we can consistently refer to the new rules as **"SHACL 1.2 rules"**. Perhaps we could also link to the specification each time we mention them.

If @HolgerKnublauch, @scotthenninger, and anyone else (e.g., @afs) agree with adopting this terminology consistently, I can create a dedicated PR.

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+				to derive <a>inferred</a> RDF <a>triples</a> from existing <em>asserted</em> <a>triples</a>.
+			</p>
+			<p>
+				The <a>SHACL rules</a> feature defined in this section includes a general framework using the properties
+				such as <code>sh:rule</code> and <code>sh:condition</code>, plus an extension mechanism for specific <a>rule types</a>.
+				This document defines one such rule type: <a>SPARQL rules</a>.
```
liviorobaldo, line 1754: Does this imply that other types of rules may be defined elsewhere? Is this intended to accommodate future rule types (e.g., JavaScript-based rules) that are not currently defined? If so, I think we should add a sentence clarifying this here. When I first read this, my immediate reaction was: "Where can I find the other rule types?"

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+				<p>
+					The following example illustrates a simple use case of a <a>SPARQL rule</a> that applies to all instances of
+					the class <code>ex:Rectangle</code> and computes the values of the <code>ex:area</code> property by multiplying
+					the rectangle's width and height:
+				</p>
+				<aside class="example" title="A rule to compute the area of a Rectangle">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:RectangleShape
+	a sh:NodeShape ;
```
liviorobaldo, line 1768: Ok, so these rules are embedded within SHACL shapes, exactly like the "old" SHACL-AF SPARQL rules. This should also be specified in this document.

In addition, this is, of course, related to [issue #765](https://github.com/w3c/data-shapes/issues/765). Unless I missed some communication, I don't think we have reached an agreement yet on whether and how to address the relation between shapes and rules.

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+					Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
+				</p>
+				<p>
+					All rules may have the properties defined in the rest of this section.
+				</p>
+				<section id="condition">
+					<h3>sh:condition</h3>
+					<p>
+						A <a>rule</a> may have values for the property <code>sh:condition</code> to specify <a>shapes</a>
+						that the <a>focus nodes</a> must conform to before the rule gets executed.
```
liviorobaldo, line 1865: Related to my previous comment, this is the place where we should specify that the shape in which the rule is embedded will be executed _after_ the rule.

afs:
> Perhaps for the time being @afs could just add it to the list of future work (assuming there is one, that's what I understood).

AFAIK the process is that the WG collects future work as issues on this repo. It will at least be recorded if nothing else.

shacl12-sparql/index.html
scotthenninger: I would favor naming consistency. If the SHACL rules are named “SHACL 1.2 Rules” then the SPARQL version should be “SHACL SPARQL 1.2 Rules”. Better yet, drop the 1.2 in both - we would hope these rules live beyond 1.2.

liviorobaldo: I was thinking that the term "SHACL-SPARQL" strongly evokes SHACL-AF. 

I think we should use **"SHACL inference rules"**, or simply **"SHACL rules"** (and yes, let's drop "1.2"), to refer to the rules defined here:

https://w3c.github.io/data-shapes/shacl12-rules/

For the rules described in this section, I would instead use **"SPARQL-based inference rules"**, or simply **"SPARQL-based rules"**.

mgberg: > I think calling these simply "SHACL rules" is quite confusing. We should clearly distinguish them from the SHACL 1.2 rules:

This was my first thought when I started looking at this.

I agree that if the new rules are called "SHACL Rules", then call these "SHACL-SPARQL Rules" or "SPARQL-Based Rules" (whether or not "1.2" is included in them as @scotthenninger mentioned).

I also think it would be valuable somewhere to clearly describe the difference between these and 1.2 Rules- currently there's nothing describing the relationship between these two flavors of rules (or lack thereof) in either this or the 1.2 Rules document, which might confuse implementers and users. I assume the intent is that they would be positioned as two separate features: 1.2 Rules for complex Datalog-style inference, and these SPARQL-based rules for simpler shape-centric inferences. Perhaps describing the motivation for having two somewhat similar features would be beneficial as well.

shacl12-sparql/index.html
mgberg: And if the intent is to support other types of rules, that would presumably have an impact on the above discussion regarding the naming of this feature.

shacl12-sparql/index.html
```diff
@@ -388,7 +390,8 @@ <h3>Document Conventions</h3>
 					This specification describes conformance criteria for:
 				</p>
 				<ul>
-					<li><strong>SHACL-SPARQL processors</strong> as processors that support validation with the SHACL-SPARQL Language</li>
+					<li><strong>SHACL-SPARQL processors</strong> as processors that support validation,
+						the evaluation of node expressions, and the evaluation of inference rules with the SHACL-SPARQL Language</li>
```
mgberg: This bundles inference rules into the same conformance group as validation and node expressions. I would imagine some implementers would include the validation and node expressions but not the rules part. Such implementations would not be considered a SHACL-SPARQL processor in this case. Should there be a separate conformance level for these rules?

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+					<p class="syntax">
+						<span data-syntax-rule="deactivated-maxCount">Each <a>rule</a> may have at most one <a>value</a> for the
+						property <code>sh:deactivated</code>.</span>
+						<span data-syntax-rule="deactivated-in">The <a>values</a> of <code>sh:deactivated</code> are either
+						of the <code>xsd:boolean</code> literals <code>true</code> or <code>false</code>.</span>
+					</p>
+				</section>
+			</section>
+			<section id="Rules">
+				<h3>The sh:Rules Entailment Regime</h3>
```
mgberg, line 1942: Are we intending to have some entailment regime defined for SHACL 1.2 Rules? If so, would this name conflict with that/be confusing?

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+				<p>
+					SHACL defines the property <code>sh:entailment</code>
+					to link a <a>shapes graph</a> with <em>entailment regimes</em>.
+					The <a>IRI</a> <code>sh:Rules</code> represents the <dfn>SHACL rules entailment regime</dfn>.
+					In the following example, the shapes graph indicates to a SHACL validation engine that the SHACL rules
+					inside of the <a>shapes graph</a> need to be executed prior to starting the validation.
```
mgberg, line 1948: `sh:condition` requires checking that a focus node conforms to a shape, but this says rules execute _prior_ to validation.

A few questions about this:
- Does the condition conformance check operate on the original graph, or the graph as modified by rules executed before rules with conditions?
- What if the condition shape itself has rules attached, or references constraints that depend on triples only produced by other rules?

shacl12-sparql/index.html
mgberg: Also, how does `sh:values` interact with these rules, particularly when both appear on the same shape? Are `sh:values`-derived property values visible to SPARQL rules? Are rule-inferred triples visible to `sh:values` evaluation? Is it up to the caller to choose the order in which these features are applied?

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+				<p>
+					Inferred triples:
+				</p>
+				<pre class="example-inferences">	ex:ExampleRectangle ex:area 56 .</pre>
+			</section>
+	
+			<section id="rules-syntax">
+				<h2>General Syntax of SHACL Rules</h2>
+				<p>
+					The <a>values</a> of the property <code>sh:rule</code> at a <a>shape</a> are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
```
mgberg, line 1843: > The <a>values</a> of the property <code>sh:rule</code> at a shape

Is this intentionally using the generalized `shape`, or should it be restricted to node shapes? Can you put a rule on a property shape? I personally don't think I've ever seen that, and I don't see any examples of that in this document.

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+						<td>
+							The prefixes to use to turn the <code>sh:construct</code> into a SPARQL query.
+							<a>SPARQL rules</a> may use the property <code>sh:prefixes</code> to declare a dependency on prefixes based on the
+							mechanism defined in <a href="#sparql-prefixes">Prefix Declarations for SPARQL Queries</a>.
+							This mechanism allows users to abbreviate URIs in the <code>sh:construct</code> strings.
+						</td>
+					</tr>
+				</table>
+				<div class="def def-text">
+					<div class="def-header">EXECUTION OF SPARQL RULES</div>
```
mgberg, line 2059: What happens if there is an error during SPARQL evaluation, e.g.:
- The `sh:construct` string is not a valid CONSTRUCT query
- The query encounters a runtime error during execution (e.g., type error in an expression)
- The CONSTRUCT template produces invalid RDF terms

There is some note of handling failures in SPARQL-based constraints, but I don't see anything like that here.

shacl12-sparql/index.html
```diff
@@ -1704,6 +1707,332 @@ <h3>SPARQL Expr Expressions</h3>
...
+						not visible to each other.
+					</li>
+					<li>
+						If two <a>rules</a> have the same <a>execution order</a> then their newly inferred <a>triples</a> are
+						not visible to each other.
+					</li>
+					<li>
+						If the same <a>rule</a> is executed on multiple target nodes then the newly inferred <a>triples</a> are
+						not visible to the other target nodes.
+					</li>
```
mgberg, line 2013: It may be clearer for implementers if it was additionally explicitly stated if/when inferred triples _can_ be seen.

shacl12-sparql/index.html
HolgerKnublauch: My assumption is that the current SHACL 1.2 Rules draft will be renamed. It has nothing to do with SHACL or shapes, and would be better named RDF Rules. With that, the name SHACL Rules is clearly and uniquely defined here.

shacl12-sparql/index.html
HolgerKnublauch: > Also, how does `sh:values` interact with these rules, particularly when both appear on the same shape? Are `sh:values`-derived property values visible to SPARQL rules? Are rule-inferred triples visible to `sh:values` evaluation? Is it up to the caller to choose the order in which these features are applied?

Rules do not "see" sh:values because those are only used to compute value nodes during validation. However, sh:values CAN see the inferences, assuming that sh:entailment sh:Rules is activated.

shacl12-sparql/index.html
HolgerKnublauch: Agree, split into three conformance parts now.

shacl12-sparql/index.html
HolgerKnublauch: Resolving this comment as duplicate on the general naming topic. I think the other rules should be renamed instead. The term SHACL Rules was already established by the SHACL-AF document and should be brought forward based on that.

shacl12-sparql/index.html
HolgerKnublauch: sh:condition is executed as part of each rule, while the entailment is happening before the whole rule processing starts. Entailment affects the triples that are returned by the data graph.

In a sense, the sh:condition is like a `FILTER shnex:conformsToShape($this, $condition)` in the beginning of the rule. It will therefore see the latest inferences too.

If you have suggestions on how to improve the clarity here, feel free to edit the branch or do a PR against the branch before we merge it.

shacl12-sparql/index.html
HolgerKnublauch: Right, I think this COULD be streamlined to only apply to node shapes. But rules only depend on the target nodes of the shape, so the mechanism is the same.

Property shapes usually don't have their own target but are mostly linked to node shapes that have a target. I was against the ability to have targets at properties in the original WG but the group decided differently.

shacl12-sparql/index.html
HolgerKnublauch: And yes, the intent here is that there can and will be other rule types. The bulk of the syntax and the main algorithm is independent from SPARQL. Therefore the general language here can be named "SHACL Rules", while the specific SPARQL-based rule type can be either SHACL-SPARQL Rules or (shorter) SPARQL Rules (because we are in the context of the SHACL document this should be clear. We also don't spell out "SHACL Shape" everywhere even though ShEx shapes exist too).

shacl12-sparql/index.html
HolgerKnublauch: Just closed #765 - the "other" rules no longer require that linkage assuming this ticket here gets done.

shacl12-sparql/index.html
HolgerKnublauch: What do you mean with "shape will be executed"? Shapes can be validated, but that is a separate process unless explicitly called by the rules engine. For example, a rule may be embedded into a shape that has a sh:targetWhere, in which case the target where computation will happen to determine the target nodes of the rule. And, for example, sh:condition is called while the rule engine runs.

HolgerKnublauch:
> > This does sound like a separate issue though, and should become its own thread IMHO.
> 
> @HolgerKnublauch, I completely agree, including with the point about reification. Perhaps for the time being @afs could just add it to the list of future work (assuming there is one, that's what I understood).

Opened #995 to outsource this topic for now.

shacl12-sparql/index.html
HolgerKnublauch: Outsourced to #996

shacl12-sparql/index.html
liviorobaldo: Sorry, I meant _validated_. So, in my view, we should say that `sh:condition` specifies shapes that are validated before the rule is executed. The rule is then executed. After all the rules embedded in a shape have been executed (according to `sh:order`), the embedding shape is validated.

shacl12-sparql/index.html
liviorobaldo: > And yes, the intent here is that there can and will be other rule types. 

@HolgerKnublauch, ok, so why couldn't these other types of rules also include the [SHACL 1.2 rules](https://w3c.github.io/data-shapes/shacl12-rules/) (or whatever we want to call them: "RDF rules", "stratified run-once rules", etc.)? 

I think this could accommodate everyone's wishes and needs.

SPARQL-based rules are needed for backward compatibility, as well as for the other reasons discussed in [issue #766](https://github.com/w3c/data-shapes/issues/766). The [SHACL 1.2 rules](https://w3c.github.io/data-shapes/shacl12-rules/), on the other hand, are safer because stratification and run-once semantics relieve users of the burden of formally demonstrating that rule execution is both deterministic and terminating.

`sh:condition` allows users to specify shapes that are validated before rules are executed, while the shapes embedding the rules are validated after rule execution. This addresses the question of execution order (i.e., whether shapes should be validated before or after rules).

The only feature that I think is still missing is a mechanism for associating a set of rules with a _set_ of shapes. In the current design, a set of rules can only be associated with a _single_ shape. As a result, if the same rules need to be executed for multiple shapes, they must be duplicated. An implementation can, of course, internally group the rules and execute them only once before validating all associated shapes. However, I think this grouping should also exist at the logical level, meaning that SHACL should provide a mechanism for expressing such associations between sets of rules and sets of shapes explicitly.

shacl12-sparql/index.html
mgberg: Got it. Also, I see the Node Expressions doc does clearly state that derived properties do not lead to the creation of triples and are intended to compute the values only on demand.

shacl12-sparql/index.html
mgberg: Got it, so the rule engine has a dependency on the validator to be able to validate nodes against node shapes specified by `sh:condition`, and the validator will see the latest inferences generated by the rules engine thus far.

But what if the condition shape itself has rules attached, or references constraints that depend on triples only produced by other rules? For example:
```turtle
ex:ProducerShape
    sh:targetClass ex:Thing ;
    sh:rule [
        a sh:SPARQLRule ;
        sh:construct """CONSTRUCT { $this ex:status ex:Ready } WHERE { ... }""" ;
    ] .

ex:ReadinessShape
    sh:property [
        sh:path ex:status ;
        sh:hasValue ex:Ready ;
    ] .

ex:ConsumerShape
    sh:targetClass ex:Thing ;
    sh:rule [
        a sh:SPARQLRule ;
        sh:condition ex:ReadinessShape ;
        sh:construct """...""" ;
    ] .
```

If I understand correctly, because `sh:order` is _not_ specified for these two rules, then it would not be possible for the rule on `ex:ConsumerShape` to ever fire as they have the same execution order and therefore any inferred triples are not visible to each other.

I assume that the solution for this is for the user to understand the dependency between these rules and properly set `sh:order`, as execution order is _only_ dependent on `sh:order`. If that's correct, would it be useful to add an example like this to show the potential interaction between `sh:condition`s dependent on other rules and `sh:order`?

HolgerKnublauch: The current spec only defines the behaviour of a single iteration. Each rule will only be visited once. This was a limitation due to time constraints - the SHACL-AF draft was created in the very last few weeks of the old WG. Surely with the remaining time until the SHACL 1.2 deadlines, we can clarify this and define how iteration should work. Then rules that didn't see certain triples in the first round will see them in the second. In our product we have implemented this iteration and it's simply a fixpoint loop (iterating until no further triples are created).

shacl12-sparql/index.html
HolgerKnublauch: Validation is the process to compute conformance, so I think both would mean the same thing. If you prefer a different wording, feel free to update this with a specific replacement diff. I have no strong opinion, and welcome anything that helps newcomers understand this better.

shacl12-sparql/index.html
HolgerKnublauch: Yes, I see two ways of adding RDF Rules as SHACL Rule types:

1. Add them as an explicit type
```
ex:RectangleShape
    a sh:NodeShape ;
    sh:targetClass ex:Rectangle ;
    sh:rule [
        a rrl:InferenceRule ;
        rrl:text """
            RULE {
                $this ex:area ?area .
            }
            WHERE {
                $this ex:width ?width .
                $this ex:height ?height .
                SET (?area := ?width * ?height) .
            }
        """
    ] .
```

2. Just use what's already there
```
ex:RectangleShape
    a sh:NodeShape ;
    sh:targetClass ex:Rectangle ;
    sh:rule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT {
                $this ex:area ?area .
            }
            WHERE {
                $this ex:width ?width .
                $this ex:height ?height .
                BIND (?width * ?height AS ?area) .
            }
        """
    ] .
```

The latter should work just fine because RDF rules are a subset of SPARQL. So an engine could easily detect if the rules fall into that subset and then can switch to the stratification algorithm. The SHACL engine would do pre-processing and insert a binding clause such as this:

```
            CONSTRUCT {
                $this ex:area ?area .
            }
            WHERE {
                $this rdf:type ex:Rectangle .
                $this ex:width ?width .
                $this ex:height ?height .
                BIND (?width * ?height AS ?area) .
            }
```

or use parameterization to inject the ?this bindings somehow. This approach treats RDF Rules as a profile of SHACL Rules.

I have no clue if RDF rules algorithms would understand how to correctly iterate on more complex targets and if sh:conditions are present, but the most common cases are simple rules that use targetClasses and no sh:conditions anyway.

> As a result, if the same rules need to be executed for multiple shapes, they must be duplicated.

I don't see why duplication. Just give the rule a URI and reference it multiple times:

```
ex:Shape1 sh:rule ex:Rule1 .
ex:Shape2 sh:rule ex:Rule2 .
```

liviorobaldo: > Just give the rule a URI and reference it multiple times:

That's what I meant by duplication 😅

Note also that, in SHACL 1.2 rules, we defined a [RuleSet](https://w3c.github.io/data-shapes/shacl12-rules/#rRuleSet), which is the entry point of [the grammar](https://w3c.github.io/data-shapes/shacl12-rules/#grammar) and the input to the [stratification algorithm](https://w3c.github.io/data-shapes/shacl12-rules/#stratification-algorithm).

Therefore, we should associate shapes with `RuleSet`s, something like this:

```
ex:Shape1 sh:ruleSet ex:RuleSet1 .
ex:Shape2 sh:ruleSet ex:RuleSet1 .
```

P.S. The `RuleSet` (or rule) should be the _same_ in this example. In your previous comment, you wrote `ex:Rule1` and `ex:Rule2`.

However, `sh:rule` (or the hypothetical `sh:ruleSet`) associates a _single_ shape with a set of rules. As a result, you have to _duplicate_ the statement for every shape that should use the same set of rules. In my view, this is both verbose and conceptually confusing. In the use cases I have worked on (and I suspect this holds more generally), the same knowledge graph, inferred by applying a set of rules, must be then validated against a _set_ of shapes, not a single one.

Therefore, IMHO we should introduce something like `sh:ShapeSet` and allow a set of rules to be associated with a set of shapes, i.e., an instance of `sh:ShapeSet`. Likewise, `sh:condition` should apply to a set of rules rather than to an individual rule.

@afs, I would also be interested in hearing your thoughts on this.

afs: Yes, rule sets.

Whether single shape or a groups of shapes, it should be a rule set -- think of it as more like SPARQL Update which allows multiple operations in one document.  SRL rules work together and the SHACL Rules visibility conditions look like they would affect a split into several single rules.

What should be excluded is `DATA`, the (at risk) abbreviations, and `IMPORTS` (on security grounds and because there is `sh:imports`).

afs: I thought, from memory, it was now `sh:imports` after the previous dicussions.
Checking the NS document, I don't see it, and so SHACL uses `owl:imports`.

SRL doesn't use `owl:imports`, or `sh:imports`, because a rule set is not an ontology so the domain is wrong.

liviorobaldo: Yes, I am also a bit confused about the distinction between `owl:imports` and `sh:imports`. I have just opened [a new issue](https://github.com/w3c/data-shapes/issues/998) on this topic. I think we should clarify our position on this matter.

shacl12-sparql/index.html
mgberg: Okay, so should I take this discussion and open a follow up issue for it then?

shacl12-sparql/index.html
HolgerKnublauch: On rule sets, I had suggested elsewhere that the named graphs should count as rule sets by default. This is how SHACL-AF rules worked. So any rule in the shapes graph are in the default rule set. Whether another type of rule set is needed is another question but there is no need for an explicit triple (and users will likely forget it anyway).

shacl12-sparql/index.html
HolgerKnublauch: There is much mix-up potential here. SHACL-AF rules have used the term SPARQL Rules for many years, even declares sh:SPARQLRule and actually uses SPARQL and not RULE...WHERE like the other draft. So at least within the SHACL document SHACL-SPARQL Rules is equivalent to SPARQL Rules. We also don't say "SHACL Shapes" everywhere. SHACL 1.2 Rules should become RDF 1.2 Rules (or some other term, but neither SHACL nor SPARQL).

afs:
It would seem quite natural as a new document and call it "SHACL 1.2 Rules Framework". The framework part is important.
It will likely grow as work happens on the framework execution algorithm.

That said:

"6. SHACL Rules Framework"  
This isn't just about one rule type.

6.1 to 6.4 unchanged.

"6.5 SHACL SPARQL Rules"  
(and informally after that within the document "SPARQL Rules")

Other options:

"6.5 SHACL SPARQL CONSTRUCT Rules"  
It isn't using SPARQL Update. Say which bit of SPARQL it is based one.

or in keeping with the other parts of the SHACL SPARQL Extensions document:  
"6.5 SPARQL-based Inference Rules"

HolgerKnublauch:
The most natural thing, and original plan, is to have a single unified document called SHACL 1.2 Rules that is actually about SHACL Rules.

shacl12-sparql/index.html
liviorobaldo: Hi @HolgerKnublauch,

I'm not sure I fully understand how this would work.

At the moment, `sh:rule` links a single shape to a single rule. A shape can be linked to multiple rules, which together constitute the shape's "set of rules".

Are you suggesting that, instead of linking individual rules, `sh:rule` should link to a named graph, for example:

```
ex:Graph1 {
    ...
}
```

If so, I'm not sure what the "`...`" inside the named graph would contain, i.e., which triples would define the rule set.

By the way, I was also thinking that `sh:rule` should have `maxCount = 1` when it links a shape to a set of rules (regardless of whether the set is represented by a `sh:RuleSet` or by a named graph). In other words, I do not think a shape should be associated with multiple rule sets (possibly even of different types), since this would allow the rule sets to interact with one another, making the inference process much harder (if not impossible) to understand and manage.

Finally, what are your thoughts on introducing sets of shapes?

HolgerKnublauch: There is no need to do anything here. SHACL is based on the data graph and the shapes graph. It doesn't care how these are constructed, but the shapes graph is the named graph containing the shapes and their rules. I see no need to complicate this further. The rule set is simply defined by the triples that exist in the shapes graph. An application could pre-process the shapes graph to hide certain triples, or add sh:deactivated to certain rules to define subsets.

sh:rule links a shape to one or more rules, so there cannot be a sh:maxCount 1. For example, a Rectangle may have one rule to compute the area, and another to compute the circumference.

Sets of shapes are equally unnecessary to me. They are just the shapes graph.

liviorobaldo: The `sh:maxCount 1` constraint on `sh:rule` was only proposed under the assumption that `sh:rule` would link a shape to a *set* of rules. If the objects of `sh:rule` are restricted to *individual* rules, then I agree that we should not impose `sh:maxCount 1`.

In any case, if we assume that the (single) set of shapes and the (single) set of rules are simply those present in the graph, I think we should at least require that all rules are of the same type. This was my main concern in the previous comment. We should not allow SHACL-SPARQL rules and SHACL 1.2 rules to be mixed in the same graph. A graph should contain either only SHACL-SPARQL rules or only SHACL 1.2 rules; otherwise, the inference process would become too difficult to define and implement, if not impossible.

HolgerKnublauch: The discussion on what to do with the other rules requires more bandwidth but is optional from the point of view of the current draft.

To all reviewers: what is missing on this branch before it can be approved? I will record a follow-up ticket to improve the handling of iterations shortly.

mgberg: Assuming the broader questions of the naming of/interaction between these rules and the new rules and whether this should be its own document or not is out of scope for this PR, the comments I made have either been addressed or are nearly covered by the follow-up issues that have been created. The one remaining comment of mine is [this one](https://github.com/w3c/data-shapes/pull/984#discussion_r3561378917) about explicitly stating when inferred triples can be seen, although that may also be addressed as part of #1069.

afs: > the inference process would become too difficult to define and implement, if not impossible.

I don't quite follow this part - could you expand on the concern please. As I read it, each SHACL Rule unit is executed inside the 3 deep for-loop of section 6.4, each time returning a graph, and the framework manages what to do with that graph - when the triples become visible, and when changes go back to the data graph.

shacl12-sparql/index.html
afs: A way to do that is to have the section title use the  name with context i.e. "SHACL SPARQL Rules", and the `<dfn>` link

```
are called <dfn data-local-lt="sparql rules">SHACL SPARQL Rules</dfn>, usually just "SPARQL Rules", 
```
and with any other `data-local-lt` you want.

Then within the document "SPARQL Rules" is then clear and can be `<a>SPARQL Rules</a>` in the HTML.

shacl12-sparql/index.html
liviorobaldo: @afs, My understanding is that the algorithm in Section 6.4 only applies to SHACL-SPARQL rules. In this thread, however, we have been discussing the possibility of mixing SHACL-SPARQL rules with SHACL 1.2 rules (see my question [here](https://github.com/w3c/data-shapes/pull/984#discussion_r3563940880) and Holger's reply [here](https://github.com/w3c/data-shapes/pull/984#discussion_r3565486740)).

What I cannot easily see is how the inference process would work in that case. The algorithm in Section 6.4 does not apply to SHACL 1.2 rules, which instead require the stratification algorithm defined in our section. We would therefore need to define an algorithm that encompasses both kinds of rules... but how?

In any case, this is outside the scope of the present PR.

Regarding this PR, my initial concern was simply whether we should explicitly state that SHACL-SPARQL rules may lead to infinite loops and non-determinism, whereas SHACL 1.2 rules do not. The discussion then went much deeper and moved beyond the scope of this PR. 

I therefore think that my original concern is also outside the scope of this PR: we should first clarify, in a separate discussion, whether SHACL-SPARQL rules and SHACL 1.2 rules are intended to work together or as alternatives to one another. Once that is clear, we can then describe more precisely how they differ.

In summary: I am happy for this PR to be closed, but IMHO the issues I have raised still need to be addressed (elsewhere).

HolgerKnublauch: @mgberg yes I expect the role of sh:order and the visibility of triples between invocations to change as part of #1069 and see that your comment has been automatically linked already. So I consider that handled for now.

HolgerKnublauch: @liviorobaldo @afs one approach to make this distinction cleaner would be to un-generalize sh:rule so that it can, for the purpose of the WG, only point at sh:SPARQLRule. This means it could remain self-contained in the SPARQL document and be understood as a distinct language from the other rules document. They do cover different use cases after all (the SHACL-AF rules have greater expressiveness traded off against less predictable output due to the lack of stratification).

Adding other rule types could then simply become topic for future work and left to implementers. I don't hold my breath, for example, that any official WG will pick up JavaScript-based rules, even though we support them in our product. We need to remain mindful of the limited time frame of the current WG.

mgberg: > This means it could remain self-contained in the SPARQL document and be understood as a distinct language from the other rules document. They do cover different use cases after all (the SHACL-AF rules have greater expressiveness traded off against less predictable output due to the lack of stratification).

If this is the direction we are going to go, as I mentioned in another thread, I think it would be valuable somewhere to clearly describe the difference between these and the new rules and/or the motivation for including two different rule languages. Currently there's nothing describing the relationship between the two rule languages (or lack thereof) anywhere, which might confuse users and implementers.

liviorobaldo: @mgberg, that's precisely what I meant by:

> I therefore think that my original concern is also outside the scope of this PR: we should first clarify, in a separate discussion, whether SHACL-SPARQL rules and SHACL 1.2 rules are intended to work together or as alternatives to one another. Once that is clear, we can then describe more precisely how they differ.

Note that Andy just opened a related issue: https://github.com/w3c/data-shapes/issues/1072

HolgerKnublauch: I have opened #1073 to clarify the design choices between the two languages.

mgberg (approved):
All the comments I had have either been addressed or have been moved to follow-up issues.

Issue #609, opened 2025-10-14
[UI] Possible input combinations to SHACL UI processor


edmondchuc:
SHACL UI may operate in different modes -  such as search or query - to support various UI use cases. Each may require different inputs to the traditional form use case.

This section focuses specifically on the form mode.

At a minimum, a SHACL UI processor requires the following inputs:

- data graph
- shapes graph
- focus node
- node shape

In form mode, SHACL UI uses the focus node as the primary resource to render, and the node shape as the root definition, applying its own constraints - as well as any other connected shapes - to generate the form.

In some scenarios, an application may allow the user to choose which node shape to apply to a form. To support this, the system needs a function that determines which node shapes are valid for, or targets a given focus node, based on the data and shapes graphs. This process may occur before supplying the required inputs to the SHACL UI processor.

Another possible input combination is to provide a data graph, a shapes graph, and a focus node, and let the SHACL UI processor determine the most suitable root node shape to apply to the form. However, this approach might not be feasible in practice, as it introduces significant complexity and places additional burden on UI implementors.

What other input combinations should we consider?

tfrancart:
> What other input combinations should we consider?

In query UI scenario such as [sparnatural](https://sparnatural.eu/):
1. There are no focus node as input
2. there can be multiple "entry point" node shapes
3. the "entry point" node shapes are determined automatically by looking for node shapes for which at least one non-deactivated property shape is attached and that are not themselves deactivated
4. having a standardized mechanism to list "entry point" node shapes would be good

smessie:
Following up on the "no focus node as input" case raised by @tfrancart — I'd like to propose a concrete resolution for one part of it: **what a renderer does when it's given a node shape but no focus node, and multiple focus nodes are applicable.**

### Proposal: a "Rendering Without a Focus Node" subsection

When a SHACL Renderer runs in **automatic mode** without a focus node, the focus nodes to render are the **targets of the given node shape** in the data graph. This set may contain **zero, one, or more** focus nodes.

Proposed behaviour:

1. A renderer **SHOULD** allow the user to view — and, in edit mode, edit — any focus node matching the node shape.
2. When more than one focus node matches, the spec does **not** prescribe presentation. An implementation **MAY**:
   - render a node UI component for every matching focus node at once (e.g. a list);
   - offer a focus node selector showing one matching focus node at a time, with switching;
   - combine both; or
   - use another mechanism, as long as any matching focus node can be viewed/edited.
[...]

PR #994, opened 2026-07-10
Update scoring graph descriptions to clarify use of `WidgetMatcher`

smessie:
This pull request clarifies how widget selection, scoring, and acceptance are defined and processed. It introduces the concept of `shui:WidgetMatcher` as a generalization for both scoring and acceptance matchers, and updates function descriptions to consistently refer to this abstraction. The changes also refine the validation logic and improve terminology for better accuracy and clarity.
The term `shui:WidgetMatcher` was already introduced earlier, but I noticed it was not properly adopted throughout the scoring system.

**Widget matcher generalization and documentation updates:**

* Added a detailed explanation of `shui:WidgetMatcher`, which now serves as the base class for both `shui:WidgetScore` and `shui:WidgetAcceptMatcher`, and described how these are used to associate widgets with matching conditions and scoring/acceptance logic.
* Updated all relevant function descriptions (`Select`, `Matcher`, `Score`, and `Accept`) to consistently refer to the scoring graph as containing `shui:WidgetMatcher` instances, rather than just `shui:WidgetScore`. This clarifies that both scoring and acceptance matchers are involved in widget selection. 

**Validation and algorithmic clarifications:**

* Improved the validation function to specify that the focus node must be a literal or a subject in the target graph, making the acceptance criteria more precise.
* Corrected the matcher and score function steps to use the correct matcher node (`shui:WidgetMatcher`), and clarified that the matcher function is called with the specific matcher instance, not just the accept matcher.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/widget-matcher-descriptions/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -920,7 +920,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
         </aside>
 
         <p class="note">
-          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is 
+          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is
           optional. The `rdfs:label` node expression should therefore handle
           missing values. For brevity, this example assumes that
           `schema:honorificPrefix` is always present.
@@ -1120,7 +1120,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
 
         <p>
           We have only briefly touched on cardinality constraints. Other useful
-          but more complex topics include logical constraints and using 
+          but more complex topics include logical constraints and using
           <a href="../shacl12-node-expr/#dfn-dynamic-shacl">Dynamic SHACL</a>
           to populate the values for the enum and autocomplete editors.
           You can find these in the <a href="#patterns">Patterns</a> section.
@@ -1210,6 +1210,25 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
             example, prefer a Boolean Select widget when the shape requires a
             boolean value via a <code>sh:datatype</code> constraint.</li>
         </ul>
+        <p>
+          The scoring graph is populated with instances of `shui:WidgetMatcher`. A
+          `shui:WidgetMatcher` associates a widget (via `shui:widget`) with the
+          conditions under which it applies, expressed as SHACL shapes referenced through
+          `shui:shapesGraphShape` (validated against the shape node in the shapes graph)
+          and `shui:dataGraphShape` (validated against the focus node in the data graph).
+          Two kinds of matcher are defined:
+        </p>
+        <ul>
+          <li><strong>`shui:WidgetScore`</strong> — a matcher that additionally carries a
+            `shui:score`. When its shapes match, its widget is recorded with that score
+            by the <a href="#scoring-algorithm-score-function">score function</a>.
+          </li>
+          <li><strong>`shui:WidgetAcceptMatcher`</strong> — a matcher used by the
+            <a href="#scoring-algorithm-accept-function">accept function</a> to decide whether a
+            widget is applicable at all. A widget without a `shui:WidgetAcceptMatcher` is
+            accepted unconditionally.
+          </li>
+        </ul>
       </section>
 
       <section id="select-function">
@@ -1227,7 +1246,9 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
           <li><strong>Data graph</strong> — The data graph containing the focus node.</li>
           <li><strong>Shape node</strong> — The SHACL shape IRI.</li>
           <li><strong>Shapes graph</strong> — RDF graph containing the SHACL shapes.</li>
-          <li><strong>Scoring graph</strong> — The RDF graph containing <code>shui:WidgetScore</code> instances that define how widgets are scored.</li>
+          <li><strong>Scoring graph</strong> — The RDF graph containing `shui:WidgetMatcher` instances
+            (`shui:WidgetScore` and `shui:WidgetAcceptMatcher`) that define how widgets are scored and accepted.
+          </li>
           <li><strong>Widget predicate</strong> — The predicate used to specify the widget in the <strong>Shape node</strong>, typically <code>shui:editor</code> or <code>shui:viewer</code>.</li>
         </ul>
 
@@ -1249,7 +1270,9 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
                   <li><strong>Shape node</strong> — The shape node associated with the property shape.</li>
                   <li><strong>Shapes graph</strong> — The shapes graph containing the shape node.</li>
                   <li><strong>Matcher node</strong> — The <strong>Accept Matcher</strong> <code>AM</code>.</li>
-                  <li><strong>Scoring graph</strong> — The RDF graph containing <code>shui:WidgetScore</code> instances that define how widgets are scored.</li>
+                  <li><strong>Scoring graph</strong> — The RDF graph containing `shui:WidgetMatcher` instances
+                    (`shui:WidgetScore` and `shui:WidgetAcceptMatcher`) that define how widgets are scored and accepted.
+                  </li>
                 </ul>
               </li>
               <li>If the return value is <code>true</code>, return the <strong>Widget IRI</strong>.</li>
@@ -1263,7 +1286,9 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
               <li><strong>Data graph</strong> — The data graph containing the focus node.</li>
               <li><strong>Shape node</strong> — The shape node associated with the property shape.</li>
               <li><strong>Shapes graph</strong> — The shapes graph containing the shape node.</li>
-              <li><strong>Scoring graph</strong> — The RDF graph containing <code>shui:WidgetScore</code> instances that define how widgets are scored.</li>
+              <li><strong>Scoring graph</strong> — The RDF graph containing `shui:WidgetMatcher` instances
+                (`shui:WidgetScore` and `shui:WidgetAcceptMatcher`) that define how widgets are scored and accepted.
+              </li>
             </ul>
           </li>
           <li>
@@ -1279,7 +1304,9 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
                   <li><strong>Shape node</strong> — The shape node associated with the property shape.</li>
                   <li><strong>Shapes graph</strong> — The shapes graph containing the shape node.</li>
                   <li><strong>Matcher node</strong> — The <strong>Accept Matcher</strong> <code>AM</code>.</li>
-                  <li><strong>Scoring graph</strong> — The RDF graph containing <code>shui:WidgetScore</code> instances that define how widgets are scored.</li>
+                  <li><strong>Scoring graph</strong> — The RDF graph containing `shui:WidgetMatcher` instances
+                    (`shui:WidgetScore` and `shui:WidgetAcceptMatcher`) that define how widgets are scored and accepted.
+                  </li>
                 </ul>
               </li>
               <li>If the return value is <code>true</code>, return <code>WI</code>.</li>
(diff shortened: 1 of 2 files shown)


robert-david (approved):

smessie:
Extra comments to clarify remaining issues:

We did not fully complete the transition towards a separate `shui:WidgetAcceptMatcher` mechanism instead of using negative `shui:WidgetScore` instances.

As described in the OP above, I updated the text to support the case where both `shui:WidgetScore` and `shui:WidgetAcceptMatcher` instances reside in the scoring graph. These are now covered under the common `shui:WidgetMatcher` concept.

Remaining points:
* [ ] We still need the `shui:WidgetAcceptMatcher` instances to be defined in the Turtle files, replacing the current `-1` score instances.
  * Do we need exactly one accept matcher per widget?
  * This could potentially be handled in a follow-up PR.
* [ ] In the **Matcher Function**, step 5, I believe the last input should be:
  ```
  scoring graph as shapes graph
  ```
  instead of:
  ```
  shapes graph as shapes graph
  ```
* [ ] The **Select Function** currently returns only a single result. However, we previously adjusted the other functions to take a `best` parameter, allowing callers to choose between:
  * only the best matching widget (`best=true`), or
  * all matching widgets (`best=false`).

  The Select Function should also accept this `best` parameter. When `best=false`, it should collect all matching widgets and return them as a list instead of returning only the first match.

danielbeeke:
1. `shui:WidgetAcceptMatcher`
1.1. Do we need exactly one accept matcher per widget? No not by spec, maybe an implementation would do it that way. But it should not be required.
1.2.  Do in follow up P-R? Sounds good to me
2. `matcher` function input
2.1. About renaming shapes graph reference to scoring graph reference: I agree
3. The `select` function
3.1.  I went through the spec and think we are trying something like the following diagram. 
3.2. I think the `score` function can be used to pick up the list of widgets. No need from my perspective to add this to the `select` function.
3.3. Also the text of the `select` function probably needs a line like: "If local state has a value for a selected preference widget return that selected preference widget IRI."

image

danielbeeke (changes requested):
See https://github.com/w3c/data-shapes/pull/994#issuecomment-4969198413

smessie:
> 1. `shui:WidgetAcceptMatcher`
>    1.1. Do we need exactly one accept matcher per widget? No not by spec, maybe an implementation would do it that way. But it should not be required.

To keep the discussion going: is it also acceptable for a widget to have no accept matcher at all? Our current default seems to be that a widget is accepted when no accept matcher is defined. I interpret that behavior from the following part of the Select Function, although we may want to make this more explicit elsewhere in the description as well.

> If there is no Accept Matcher defined for the widget, return the Widget IRI.

> 2. `matcher` function input
>    2.1. About renaming shapes graph reference to scoring graph reference: I agree

Perfect. I will wait before making any further changes to this PR until @edmondchuc, @bergos, and @robert-david have had a chance to comment as well.

> 3. The `select` function
>    3.1.  I went through the spec and think we are trying something like the following diagram.

Your diagram seems to be a correct representation of the current version of the specification.

> 3.2. I think the `score` function can be used to pick up the list of widgets. No need from my perspective to add this to the `select` function.
> 3.3. Also the text of the `select` function probably needs a line like: "If local state has a value for a selected preference widget return that selected preference widget IRI."

Currently, I think the issue with the `score function` (when used as a public API) is that it does not filter results using the `shui:WidgetAcceptMatcher` instances, whereas the `select function` does.

This means the `score function` can return widgets that are not actually allowed to match, since the current `-1` scores will/should be replaced by `shui:WidgetAcceptMatcher` instances.

I think that is precisely why we still have an overarching `select function`, which performs this post-scoring gatekeeping step.

So:

1. You use the `score function`, in which case you can retrieve all matching widgets (not only the best one), but the returned widgets are not checked against the `shui:WidgetAcceptMatcher` instances to verify that they are actually allowed.
2. You use the `select function`, in which case you can retrieve only the best widget, but you do not have the option of retrieving all matching widgets. The widget that is returned is, however, guaranteed to be allowed by the relevant `shui:WidgetAcceptMatcher` instances.

I see two possible solutions:

1. The `score function` also evaluates the `shui:WidgetAcceptMatcher` instances, ensuring that it returns only allowed widgets.
2. The `select function` gains a `best` parameter, allowing it to support both retrieving only the best widget and retrieving all matching widgets.

My preference is **solution 2**, because with option 1 the added value of the `select function` over the `score function` becomes quite limited.

danielbeeke:
Yes from my perspective it is acceptable to not have an `shui:WidgetAcceptMatcher` for a widget.

I agree with the analysis, I had not seen that the score function did not match via the ` shui:WidgetAcceptMatcher`.
I also prefer option 2. That would make the score function an internal function, which I also like very much because it makes the external interface small.

smessie:
Pushed 03234764, resolving the two remaining points.

**2. `matcher` function input** — step 5 now passes `scoring graph as shapes graph`, matching step 2. The `shui:dataGraphShape` value is a shape defined in the scoring graph.

**3. The `select` function** — implemented solution 2. `select` gains a `best` parameter: `best=true` returns a single widget IRI (or `undefined`), `best=false` returns an ordered sequence of score results, all checked against the `shui:WidgetAcceptMatcher` instances. Results are score-result objects (widget IRI, `shui:WidgetScore` IRI, `shui:score`) rather than bare IRIs, so the Widget Selection post-processing guidance still applies. No de-duplication in `select` — that stays with that post-processing step.

**Two clean-ups this surfaced**

The `accept` function had no callers: `select` re-implemented its body inline, twice. Both sites now call it, which also puts the accept-matcher lookup in one place — useful when we specify how multiple accept matchers per widget combine.

I also dropped `best` from the `score` function. `score(best=true)` was never usable by `select`, since the top-scoring widget may be rejected by its accept matcher, so `select` had to pass `false` anyway. `score` now always returns an ordered sequence, which also removes its dual return type. Notes on both functions record that the result order is fixed by the `shui:score` values and `shui:widget` IRIs alone — known before any matcher runs — so implementations MAY produce results lazily, and with `best=true` only the matchers up to the first accepted widget need evaluating. Lazy and eager evaluation give the same sequence, so eager implementations still conform.

@danielbeeke — flagging this one as it goes beyond what we agreed. It follows from `score` becoming internal, but let me know if you'd rather keep the old signature.

**A concern for a future PR**

I'm not happy that widgets declared via a widget predicate are handled as a separate case. It costs us the whole of step 2 with its substeps and makes the algorithm considerably more complex, mainly because that path behaves differently: it produces a different kind of output, and the widget has no explicit score the renderer can use downstream. It also can't be overridden by future or custom widgets. I'd rather fold it into the normal scoring path, but that deserves its own PR — I'll open an issue first, since the answer probably depends on the outcome of #1083.

**Still open** — point 1 (replacing the `-1` score instances with `shui:WidgetAcceptMatcher` instances in the Turtle files), as agreed a follow-up PR.

[Rendered online](https://raw.githack.com/w3c/data-shapes/ui/widget-matcher-descriptions/shacl12-ui/index.html)

smessie:
Pushed 31ae34d2. After a call with @danielbeeke we changed the approach, so this supersedes the design in my previous comment: there is no `select` function any more.

**`score` is now the entry point.** It regains its `best` parameter and calls the `accept` function itself, skipping widgets that their `shui:WidgetAcceptMatcher` instances reject. `best=true` returns the best matching *and accepted* widget, `best=false` the ordered sequence. The `select` function is removed; its Processing rules moved to `score`.

**New step: scoring graph preparation**, mandatory before scoring. It collects the widgets declared with `shui:editor` or `shui:viewer` on a shape and adds a `shui:WidgetScore` for those that have none, using the existing "prefers" pattern (a `shui:shapesGraphShape` testing the declaring property). The score comes from the `shui:defaultWidgetScore` global configuration property, defaulting to 40.

This removes the special case for widgets attached with a widget predicate that I complained about earlier: such widgets now score, rank and get accept-checked like any other, and a widget that already has scores keeps whatever score the scoring graph gives it.

**Note on the build:** `shui:defaultWidgetScore` hangs off the global configuration from #900, which is not merged, so that term does not resolve here yet and ReSpec reports it. #1104 has the same pending reference. `shui:defaultWidgetScore` still needs adding to the property table in #900.

Still open: point 1, replacing the `-1` score instances with `shui:WidgetAcceptMatcher` instances in the Turtle files, as agreed a follow-up PR.

[Rendered online](https://raw.githack.com/w3c/data-shapes/ui/widget-matcher-descriptions/shacl12-ui/index.html)

danielbeeke (approved):
looks good to me, however I would like to request that maybe @edmondchuc , @robert-david  or @bergos has a look, as this contains a big change.

Some background:

@smessie and I had a call, while implementing the previous structure there were many ugly things in the definitions of the functions. The output types were quite overloaded, the "best" argument was finicky and it had no clear public interface. After this change that all gets better, the trade off here is that we added a preparation step to the score graph. I think this is the right decision, however I think its good if there is at least one more reviewer.

shacl12-ui/index.html
```diff
@@ -1210,96 +1219,152 @@ <h3>Introduction</h3>
             example, prefer a Boolean Select widget when the shape requires a
             boolean value via a <code>sh:datatype</code> constraint.</li>
         </ul>
+        <p>
+          The scoring graph is populated with instances of `shui:WidgetMatcher`. A
```
edmondchuc, line 1223: I recall in one of the past task force meeting that we acknowledged the contents of the scoring graph could reside in the shapes graph. Is this still the case?

If not, given that the scoring system defines a separate scoring graph, I wonder whether it should be considered another input parameter to a SHACL UI Renderer (see the discussion in https://github.com/w3c/data-shapes/issues/609).

shacl12-ui/index.html
```diff
@@ -1387,56 +1458,115 @@ <h3>Matcher Function</h3>
...
+            <code>shui:WidgetScore</code> matches and that are accepted according to their
+            <code>shui:WidgetAcceptMatcher</code> instances, ordered from best to worst match.
           </p>
           <p>
             Inputs:
           </p>
         <ul>
-          <li><strong>Best</strong> — A boolean flag; if true, return only the first matching result.</li>
-          <li><strong>Focus node</strong> — The node to validate.</li>
+          <li><strong>Best</strong> — A boolean flag; if true, return only the best matching widget.</li>
```
edmondchuc: I am wondering how useful the `best` parameter is as part of the specified interface. Since the score function already returns results in ranked order, an implementation or caller could select the first result when only the best match is needed.

Could this optimisation be left to implementations instead, with the specification defining a single, consistent return type as an ordered sequence of matching widget results?

shacl12-ui/index.html
smessie: I don’t recall that discussion; perhaps I wasn’t present at that meeting or I simply forgot.
In any case, I interpreted it as if the scoring graph is defined as a separate parameter from the shapes graph.
You’re right that it should then be added to #609’s input parameters.

Shall I proceed adding that in the PR for #609 or do we want to discuss the possibility of having the scoring graph in the shapes graph first?

danielbeeke: I agree also it should be input parameter.

shacl12-ui/index.html
smessie: It is really an optimization case here, as the `best` parameter is there to prevent computing the rest of the matching widgets when only the first one suffices.
I would agree that we leave that optimization to implementation details. An implementation can then choose if they want to add such extra parameter `best` or if they simply want to make `score` a generator that yields widget by widget, and thus only if a next one is requested.
As @danielbeeke already approved the PR, are you fine with removing that `best` parameter in favor of calling it an implementation detail. That would further simplify the output types as well.

danielbeeke: We could do that and keep some of the comment Ieben added about using generators.

shacl12-ui/index.html
smessie: Agreed. Nothing changes here — the scoring graph is already a separate parameter throughout the scoring system. I will add it as a renderer input in #609, referring to the prepared scoring graph this PR introduces.

shacl12-ui/index.html
smessie: Removed in 8c1cba6b — `score` now always returns the ordered sequence, giving it a single return type. Kept the generator note as implementation guidance, and the output now states it is sorted by descending `shui:score`.

edmondchuc (approved):

Issue #1069, opened 2026-07-14
SHACL SPARQL Rules: clarify/improve handling of multiple iterations


HolgerKnublauch:
The SHACL-AF spec (and current draft for SHACL SPARQL Rules) only defines the behavior of a single iteration across all rules. This was mainly due to time constraints and to avoid complexity. This should be revised and clarified. My implementation simple does a fixpoint iteration and that works well in practice.

HolgerKnublauch:
As part of this, we can likely soften up the use of sh:order. This was mainly relevant if there is just a single iteration.

Also probably introduce a sh:singleExecution true flag for rules that are only needed once. Or, sh:maxRuleIterationCount 1.

afs:
In your implementation, what happens with new blank nodes?

`CONSTRUCT { [ ] :p :o } WHERE { ?s :p :o }`

HolgerKnublauch:
It will keep iterating until it reaches a max iteration count or max size of graph or crash. Such scenarios are treated like programming errors in an imperative language. Rule developers will (hopefully) try out the rules and quickly find out.

I guess an alternative approach would be to detect rules that *have* produced blank nodes and limit their execution like your single-run rules. TBD in this ticket here.

afs:
What's the visibility? 
Is "max iteration count or max size of graph or crash" an exception/abort of all the evaluation of this step or a "stop point" or something else?

HolgerKnublauch:
This is up to the implementation. Usually it should fail with an error so that the author of the rule can fix the bug.

HolgerKnublauch:
When we rewrite the iteration logic, I think we should also generalize the ordering algorithm so that algorithms such as stratification can be inserted like plugins. sh:order would then be just one ordering algorithm among others. Stratification could be used whenever all rules fall into the SRL subset.

HolgerKnublauch:
Another thing that I noticed that should be covered here when we reformulate the algorithm is that sh:condition should also include the "inherited" constraints when the node shape (value of sh:condition) is also a class. For example:

```
ex:Person
    a sh:ShapeClass ;
    sh:property ex:Person-name .

ex:Patient
    a sh:ShapeClass ;    # or sh:NodeShape + rdfs:Class
    rdfs:subClassOf ex:Person ;
    sh:property ex:Patient-disease .

ex:Rule1
    a sh:SPARQLRule ;
    sh:condition ex:Patient ;
    ...
```

then the constraints of ex:Person should also apply as condition.

[...]

PR #1078, opened 2026-07-19
#1071: Support for global (or static) rules

### LINKED ISSUE #1071: SHACL SPARQL Rules: Handling of rules that are not attached to shapes
In order to execute "global" rules that don't depend on specific focus nodes, a current trick is to make up a dummy node shape that has exactly one dummy target node:

```
ex:Global
    a sh:NodeShape ;
    sh:targetNode ex:Global ;
    sh:rule ex:GlobalRule .

ex:GlobalRule
    a sh:SPARQLRule ;
    sh:construct """
        CONSTRUCT {
            ?o ?p ?s .
        }
        WHERE {
            ?p a owl:SymmetricProperty .
            ?s ?p ?o .
        }
    """ .
```

However, this is untidy. I think we should change the algorithm to accommodate for rules that are NOT attached to any shape. Then ex:GlobalRule would execute (once per iteration) even without ex:Global.

HolgerKnublauch:
Closes #1071

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1071/shacl12-sparql/index.html#rules-syntax)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2025,7 +2025,7 @@ ex:RuleOrderExampleShape
 				</p>
 				<div class="algorithm">
 					<pre class="nohighlight">
-	for each <a>global rule</a> R in the <a>shapes graph</a>, ordered by <a>execution order</a> {
+	for each non-deactivated <a>global rule</a> R in the <a>shapes graph</a>, ordered by <a>execution order</a> {
 		execute R following the <a>execution instructions</a> of its <a>rule types</a>
 	}
 
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1861,6 +1861,11 @@ ex:InvalidRectangle    # Lacks a value for ex:color, so sh:condition is not met
 					The <a>subjects</a> of <code>sh:rule</code> triples are <a>IRIs</a>.</span>
 				</p>
 				<p>
+				   All rules may have the properties defined in the rest of this section.
+                </p>
+				<section id="ruletypes">
+					<h3>Types of SHACL Rules</h3>
+					   <p>
 					<a>SHACL rules</a> can be divided into two categories:
 				</p>
 				<ul>
@@ -1893,8 +1898,7 @@ ex:SymmetricPropertyRule
 					</div>
 				</aside>
 				<p>
-					All rules may have the properties defined in the rest of this section.
-				</p>
+		</section>
 				<section id="condition">
 					<h3>sh:condition</h3>
 					<p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1840,7 +1840,7 @@ ex:InvalidRectangle    # Lacks a value for ex:color, so sh:condition is not met
 			<section id="rules-syntax">
 				<h2>General Syntax of SHACL Rules</h2>
 				<p>
-					The <a>values</a> of the property <code>sh:rule</code> at a <a>shape</a> are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
+					The <a>SHACL instances</a> of <code>sh:Rule</code> and its subclass <code>sh:SPARQLRule</code> are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
 					SHACL has a flexible design in which (potentially) multiple types of rules can be supported,
 					but this document only defines <a>SPARQL rules</a>.
 					Each <dfn data-lt="rule types">rule type</dfn> is identified by an <a>IRI</a> that is used as <code>rdf:type</code> of rules.
@@ -1855,6 +1855,43 @@ ex:InvalidRectangle    # Lacks a value for ex:color, so sh:condition is not met
 					The creator of such rules needs to make sure that such rules have consistent semantics.
 					Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
 				</p>
+				<p class="syntax">
+					<span data-syntax-rule="rule">The property <code>sh:rule</code> is used to link a <a>shape</a> (<a>subject</a>)
+					with a <a>rule</a> (<a>object</a>).
+					The <a>subjects</a> of <code>sh:rule</code> triples are <a>IRIs</a>.</span>
+				</p>
+				<p>
+					<a>SHACL rules</a> can be divided into two categories:
+				</p>
+				<ul>
+					<li>
+						<dfn>Shape rules</dfn> are <a>rules</a> that are the <a>object</a> of a <code>sh:rule</code> triple.
+						Shape rules are executed for all <a>target</a> nodes of the <a>shapes</a>.
+					</li>
+					<li>
+						<dfn>Static rules</dfn> are <a>rules</a> that are <em>not</em> linked to a <a>shape</a> via <code>sh:rule</code>.
+					</li>
+				</ul>
+				<p>
+					The following example illustrates a <a>static rule</a>.
+				</p>
+				<aside class="example" title="A static rule that is not associated with any shape">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:SymmetricPropertyRule
+	a sh:SPARQLRule ;
+	sh:construct """
+		CONSTRUCT {
+			?o ?p ?s .
+		}
+		WHERE {
+			?p a ex:SymmetricProperty .
+			?s ?p ?o .
+		}
+		""" .</pre>
+						</div>
+					</div>
+				</aside>
 				<p>
 					All rules may have the properties defined in the rest of this section.
 				</p>
@@ -1984,8 +2021,12 @@ ex:RuleOrderExampleShape
 				</p>
 				<div class="algorithm">
 					<pre class="nohighlight">
+	for each <a>static rule</a> R in the <a>shapes graph</a>, ordered by <a>execution order</a> {
+		execute R following the <a>execution instructions</a> of its <a>rule types</a>
+	}
+
 	for each <a>shape</a> S in the <a>shapes graph</a>, ordered by <a>execution order</a> {
-		for each non-deactivated <a>rule</a> R in the <a>shape</a>, ordered by <a>execution order</a> {
+		for each non-deactivated <a>shape rule</a> R of the <a>shape</a>, ordered by <a>execution order</a> {
 			for each <a>target</a> node T of S that <a>conforms</a> to all <a href="#condition">conditions</a> of R {
 				execute R using T as <a>focus node</a> following the <a>execution instructions</a>
 				of its <a>rule types</a>
@@ -2060,9 +2101,18 @@ ex:RuleOrderExampleShape
 					<div class="def-text-body">
 						Let <code>Q</code> be the SPARQL CONSTRUCT query derived from the values of the properties
 						<code>sh:construct</code> and <code>sh:prefixes</code> of the <a>SPARQL rule</a> in the <a>shapes graph</a>.
-						For each <a>focus node</a>, execute the query <code>Q</code>
-						<a>pre-binding</a> the variable <code>this</code> to the <a>focus node</a>,
-						and <a>infer</a> the constructed <a>triples</a>.
+						<ul>
+							<li>
+								If the rule is a <a>shape rule</a>:
+								For each <a>focus node</a>, execute the query <code>Q</code>
+								<a>pre-binding</a> the variable <code>this</code> to the <a>focus node</a>,
+								and <a>infer</a> the constructed <a>triples</a>.
+							</li>
+							<li>
+								If the rule is a <a>static rule</a>:
+								Execute the query <code>Q</code> without any <a>pre-binding</a>.
+							</li>
+						</ul>
 					</div>
 				</div>
 			</section>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1874,13 +1874,13 @@ ex:InvalidRectangle    # Lacks a value for ex:color, so sh:condition is not met
 						Shape rules are executed for all <a>target</a> nodes of the <a>shapes</a>.
 					</li>
 					<li>
-						<dfn>Static rules</dfn> are <a>rules</a> that are <em>not</em> linked to a <a>shape</a> via <code>sh:rule</code>.
+						<dfn>Global rules</dfn> are <a>rules</a> that are <em>not</em> linked to a <a>shape</a> via <code>sh:rule</code>.
 					</li>
 				</ul>
 				<p>
-					The following example illustrates a <a>static rule</a>.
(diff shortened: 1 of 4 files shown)


shacl12-sparql/index.html
```diff
@@ -1840,7 +1840,7 @@ <h3>Examples of SHACL Rules</h3>
 			<section id="rules-syntax">
 				<h2>General Syntax of SHACL Rules</h2>
 				<p>
-					The <a>values</a> of the property <code>sh:rule</code> at a <a>shape</a> are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
+					The <a>SHACL instances</a> of <code>sh:Rule</code> and its subclass <code>sh:SPARQLRule</code> are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
 					SHACL has a flexible design in which (potentially) multiple types of rules can be supported,
 					but this document only defines <a>SPARQL rules</a>.
```
scotthenninger, line 1845:
```suggestion
					<dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn> are defined as any SHACL instance of the core <code>sh:Rule</code> and its subclasses. While SHACL provides an extensible framework capable of supporting multiple rule languages, this document specifically defines <a>SPARQL rules</a> via the <code>sh:SPARQLRule</code> subclass.
```

shacl12-sparql/index.html
```diff
@@ -1855,6 +1855,43 @@ <h2>General Syntax of SHACL Rules</h2>
 					The creator of such rules needs to make sure that such rules have consistent semantics.
 					Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
 				</p>
+				<p class="syntax">
+					<span data-syntax-rule="rule">The property <code>sh:rule</code> is used to link a <a>shape</a> (<a>subject</a>)
+					with a <a>rule</a> (<a>object</a>).
+					The <a>subjects</a> of <code>sh:rule</code> triples are <a>IRIs</a>.</span>
+				</p>
+				<p>
```
scotthenninger, line 1863:
```suggestion
				<section id="ruletypes">
					<h3>Types of SHACL Rules</h3>
					   <p>
```

shacl12-sparql/index.html
```diff
@@ -1855,6 +1855,43 @@ <h2>General Syntax of SHACL Rules</h2>
...
+		}
+		WHERE {
+			?p a ex:SymmetricProperty .
+			?s ?p ?o .
+		}
+		""" .</pre>
+						</div>
+					</div>
+				</aside>
 				<p>
```
scotthenninger, line 1900:
```suggestion
				<p>
		</section>
```

shacl12-sparql/index.html
```diff
@@ -1855,6 +1855,43 @@ <h2>General Syntax of SHACL Rules</h2>
...
 					All rules may have the properties defined in the rest of this section.
 				</p>
```
scotthenninger:
```suggestion

```

shacl12-sparql/index.html
```diff
@@ -1855,6 +1855,43 @@ <h2>General Syntax of SHACL Rules</h2>
 					The creator of such rules needs to make sure that such rules have consistent semantics.
 					Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
 				</p>
+				<p class="syntax">
+					<span data-syntax-rule="rule">The property <code>sh:rule</code> is used to link a <a>shape</a> (<a>subject</a>)
+					with a <a>rule</a> (<a>object</a>).
+					The <a>subjects</a> of <code>sh:rule</code> triples are <a>IRIs</a>.</span>
+				</p>
```
scotthenninger, line 1862:
```suggestion
				</p>
				<p>
				   All rules may have the properties defined in the rest of this section.
                </p>
```

scotthenninger (commented):
A few suggestions for the static SHACL rule additions.

shacl12-sparql/index.html
HolgerKnublauch: The reason why I said "sh:Rule and its subclass sh:SPARQLRule" is that the term SHACL instance is defined to depend on the presence of rdf:type triples AND (for subclasses) rdfs:subClassOf triples in the same graph. This means that if we define SHACL rules to be the SHACL instances of sh:Rule only, everyone would need to include the triple sh:SPARQLRule rdfs:subClassOf sh:Rule into the shapes graph. This would be error-prone.

HolgerKnublauch:
Also: is the name "static" clear enough? In Java it's used to mean fields/methods that are at a class but don't apply to the instances of that class individually. However, these rules here are not attached to anything, so "global" remains an alternative. No strong opinion.

scotthenninger:
> Also: is the name "static" clear enough? In Java it's used to mean fields/methods that are at a class but don't apply to the instances of that class individually. However, these rules here are not attached to anything, so "global" remains an alternative. No strong opinion.

Yes I recall this now. It seems to me that the semantics are different than static in Java. It’s more about the scoping of the rule. Given that, I think global fits better.

shacl12-sparql/index.html
liviorobaldo: OK, I think this PR is a positive step in the right direction. However, I wonder whether it might be a bit premature. It would probably be better to first clarify (hopefully during today's meeting) the direction we want to take with rules.

My understanding was that we were planning to define `sh:Rule` in a dedicated document, possibly the one the Rule TF is currently working on. If so, I think this document should simply point to that specification, rather than repeating text such as "SHACL has a flexible design in which (potentially) multiple types of rules...".

Another point that may need clarification is whether rules of different types can be mixed within the same graph. My intuition is that they should not be, although I may be mistaken. If mixing different rule types is indeed not allowed, should we explicitly state that graphs containing multiple rule types are invalid? By contrast, I believe that mixing local and global rules should be perfectly acceptable, provided they are all of the same rule type.

shacl12-sparql/index.html
```diff
@@ -1855,9 +1855,50 @@ <h2>General Syntax of SHACL Rules</h2>
...
+					   <p>
+					<a>SHACL rules</a> can be divided into two categories:
 				</p>
+				<ul>
+					<li>
+						<dfn>Shape rules</dfn> are <a>rules</a> that are the <a>object</a> of a <code>sh:rule</code> triple.
+						Shape rules are executed for all <a>target</a> nodes of the <a>shapes</a>.
+					</li>
+					<li>
+						<dfn>Global rules</dfn> are <a>rules</a> that are <em>not</em> linked to a <a>shape</a> via <code>sh:rule</code>.
```
mgberg, line 1877: I'm wondering if there would be any situations where a global rule being defined as a rule that is _not_ linked to a shape via `sh:rule` would cause problems vs. some opt-in method of explicitly saying something _is_ a global rule.

shacl12-sparql/index.html
```diff
@@ -1984,8 +2025,12 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 				</p>
 				<div class="algorithm">
 					<pre class="nohighlight">
+	for each <a>global rule</a> R in the <a>shapes graph</a>, ordered by <a>execution order</a> {
```
mgberg:
```suggestion
	for each non-deactivated <a>global rule</a> R in the <a>shapes graph</a>, ordered by <a>execution order</a> {
```

shacl12-sparql/index.html
```diff
@@ -1984,8 +2025,12 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+		execute R following the <a>execution instructions</a> of its <a>rule types</a>
+	}
+
 	for each <a>shape</a> S in the <a>shapes graph</a>, ordered by <a>execution order</a> {
```
mgberg: Out of curiosity, is there an explicit reason why global rules are run before shape rules?

HolgerKnublauch: No specific reason. It was just easier to edit incrementally. But this whole section will be rewritten for #1069 anyway and there it would be best to have single loop over all shapes and then distinguish between shape rules and global rules in each loop.

I just hope we can get this PR out of the way soon so that we can start in the iteration ticket. It's difficult to do when two things are modified at once.

shacl12-sparql/index.html
HolgerKnublauch: Theoretically we could add such a thing but I am worried people will forget that too. For example if we add a flag sh:global true. The real difference that should be obvious on first sight is whether a rule uses $this or not, and I hope this is clear enough. The rest is education.

For house-keeping I'll mark this question resolved but please reopen if you disagree.

shacl12-sparql/index.html
HolgerKnublauch: This question is not really about global/shape rules but more about rule types.

My current preference remains to slightly reorganize the SHACL-SPARQL rules doc to be about SPARQL Rules only, and leave a Note to mention that other rule types MAY be added outside of the spec. The iteration algorithm is easy to adjust, leaving this flexibility in place. It should make editing and explaining this easier if we only focus on SPARQL rules for now.

This, or course, assumes that SRL rules will either share the same syntax (my preference) or remain their own format. In the latter case, we don't need to introduce a second rule type either because there is no need for SRL from SPARQL Rules perspective and it would only drive up implementation dependencies and make compliance harder. We also don't expect SRL to implement all of SPARQL Rules either. SRL would just remain its own language.

With this out of the way, could I get someone to approve this PR so we can move on?

shacl12-sparql/index.html
mgberg: > Theoretically we could add such a thing but I am worried people will forget that too. For example if we add a flag sh:global true.

That's certainly true, someone could easily forget. It feels a bit inconsistent though in the sense that everything else so far in SHACL, shapes or rules, has to be associated with _something_ in order to work; a shape with no target doesn't do anything, it doesn't target everything. Then again, this is the first case I can think of where we're specifically trying to add something that does act globally, not having a flag is simpler, and rules can always be deactivated if you don't want them to fire. 

This may be unnecessary planning for failure, but I'm wondering if there's any value in a flag from the perspective of the consequences if something is missing. If a flag existed but was missing, the rule would simply not fire. In the current proposal, if sh:rule were missing, ?this would match everything. 

If nobody else has any concerns about this, I'm good with it.

HolgerKnublauch: Yes I see your point and agree that so far everything in the SHACL shapes was attached to shapes. But the work-around, to have a singleton node shape with a single sh:targetNode is a well-known hack to essentially express such global things. For example, such global constraints are used to check that the whole graph has certain characteristics, such as exactly one instance of a certain class.

So while the syntax of global rules is something new, the concept of global axioms is not, and it's good to remove the need for the ugly singleton node shape hack.

I'll (optimistically) mark this discussion as resolved so that we can move on for house-keeping, but welcome other opinions to reopen it.

HolgerKnublauch: FWIW we have an implementation of OWL-RL using SHACL SPARQL Rules, consisting of 69 global sh:SPARQLRules. It has been in the product for quite some time now, and gets the job done.

shacl12-sparql/index.html
liviorobaldo: Ok, I have approved the PR because I basically agree with this point:

> My current preference remains to slightly reorganize the SHACL-SPARQL rules doc to be about SPARQL Rules only, and leave a Note to mention that other rule types MAY be added outside of the spec.

I hope we can all reach an agreement soon on the remaining aspects concerning rules.

liviorobaldo (approved):

mgberg (approved):

scotthenninger (approved):
The changes look good.

TallTed:
Submitted on Sunday, merged on Tuesday. Was this speed really necessary?

scotthenninger:
There was no reason to wait

TallTed:
> There was no reason to wait

In point of fact, there was.

> With this out of the way, could I get someone to approve this PR so we can move on?

The request above was made Monday. Roughly 41 hours after the PR was posted, on Sunday.

This sort of speed is certainly not required at this stage in our process, and is only going to lead to problems as we go.

I'm sure I'm not the only one who hadn't reviewed this PR by the time it merged, even if I'm the only one voicing dissatisfaction with this pattern.

scotthenninger:
No one is stopping anyone from creating a PR

TallTed:
@scotthenninger — 

Please stop. I have had many conversations about the heavier lift involved in creating-PRs vs commenting-on-PRs. These are not the only good reasons to wait an extra day or three before merging new PRs.

I have also had a number of conversations about GitHub spontaneously deciding to change indents from spaces to tabs, and tab-stops from 2 spaces to 8 spaces to 4 spaces, and line ends from LF to CRLF or CR. This has made editor nightmares from a number of my PRs and even my suggested changes on other people's PRs.

Today, I have discovered what I *think* are the relevant solutions to these problems, which is for the editors (attn @afs, @HolgerKnublauch, @pchampin, other editors) of each repository to add appropriately populated [`.gitattributes`](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings#per-repository-settings) and [`.editorconfig`](https://github.com/orgs/community/discussions/69330#discussioncomment-7201339) files to the root of each repository. Fixing the existing mixed bag of line endings appears to be fairly quick with --
```bash
git add --renormalize .
git commit -m "Normalize all repository line endings to LF"
```
I've not yet found a similar command to fix the mixed up indents, but I'm hopeful that I or someone else will find it soon.

P.S. I must ask someone else who hopefully already uses `git` locally to run the fix command(s). I do virtually all my editing through GitHub's web interface which does not directly support such commands.

scotthenninger:
Please stop.
  
Keep your temper.

TallTed:
Please make constructive comments, or don't comment.

scotthenninger:
Right back at ya.

HolgerKnublauch:
> Submitted on Sunday, merged on Tuesday. Was this speed really necessary?

This PR had received enough discussion and was approved by all other 3 editors. I had other work in the queue that needed these changes, and for branch housekeeping it is usually best to proceed quickly, to avoid merge complications. There was no reason to wait. As Scott said, anyone can submit follow-up PRs later at any time.

HolgerKnublauch:
@TallTed to illustrate the complications of merging, see what I now have to work through because you have submitted editorial comments while another major branch is still active

image

As I have said multiple times, I would prefer if editorial changes, such as adding Oxford commas, are done at the very end of the process, or when documents have become sufficiently stable. Not in the middle of active work. It is just creating extra work for the editors, and arguably it is best to apply stylistic changes consistently on the whole documents, not individual deltas.

This does not mean that such editorial contributions are not welcome, but please as macro edits, not micro edits.

scotthenninger:
Quite frankly, this is what LLMs are particularly good at. And I would agree that perfecting the grammar of drafts is counterproductive.

PR #1079, opened 2026-07-20
GH-930: Running example and worked evaluation trace

### LINKED ISSUE #930: Create a running example for the rules description section

recalcitrantsupplant:
A concrete proposal for the running example discussed in #930 (theme suggestions from #904): software components / dependency chain / vulnerability exposure.

- §3.1–3.5: one scenario throughout; the rules introduced feature-by-feature form the five-rule set the worked example later traces.
- §4.3: the empty closed-dependency example is filled; the open-dependency example re-themed to match.
- New §6.6 Worked Example (informative): dependency analysis in prose, dependency graph as inline SVG, stratification step-by-step, then pass-by-pass evaluation with the inference graph state shown after each pass and each new triple attributed to its rule. Includes a run-once rule (per @liviorobaldo's request in #904) so the once/general layer split is exercised.

Some of my example updates in S3 are more "running theme" than a single example. They could be changed back to "isolated" examples (SET miles -> km) - I don't think this would change readability or understanding much. 

The worked example at 6 adds a fair amount of text, this is the reason for making it a separate PR. I've tried to keep it as concise as possible. I think it is useful - keen to hear other opinions.

Closes #930 

* [See this document rendered online here](https://raw.githack.com/recalcitrantsupplant/data-shapes/dhabgood/rules-running-example/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -404,52 +404,56 @@
       <section id="desc-basic">
         <h3>Basic Patterns</h3>
 
+        <p>
+          The examples in this section describe software components and
+          their dependencies: a frontend calls an application server, and
+          the application server queries a database that has a known
+          vulnerability.
+        </p>
+
         <p>In this first example, we have the following data graph and rule set:</p>
 
         <aside class="example" title="Basic Example">
           <pre class="block ex-data">
-            :A :fatherOf :X .
-            :B :motherOf :X .
-            :C :motherOf :A .
+            :frontend :callsService :app .
+            :app      :queriesDatabase :db .
+            :db       :hasVulnerability :vuln1 .
           </pre>
           <pre class="block ex-rules">
-            RULE { ?x :childOf ?y } WHERE { ?y :fatherOf ?x }
-            RULE { ?x :childOf ?y } WHERE { ?y :motherOf ?x }
+            RULE { ?x :dependsOn ?y } WHERE { ?x :callsService ?y }
+            RULE { ?x :dependsOn ?y } WHERE { ?x :queriesDatabase ?y }
           </pre>
           <pre class="block ex-inferred">
-            :X :childOf :A .
-            :X :childOf :B .
-            :A :childOf :C .
+            :frontend :dependsOn :app .
+            :app      :dependsOn :db .
           </pre>
         </aside>
 
-        <p>The above rules, applied to the data, will conclude 
-          that: `:X` is the `:childOf` `:A` and `:B`:
+        <p>The above rules, applied to the data, will conclude that
+          `:frontend` depends on `:app`, and that `:app` depends on
+          `:db`, whatever the kind of dependency.
         </p>
 
         <p>
-          We can then derive `:descendedFrom` relationships by adding a rule that depends
-          on `:childOf` triples produced by the other rules:
+          We can then derive `:exposedTo` relationships by adding a rule that depends
+          on `:dependsOn` triples produced by the other rules:
         </p>
-        
-        <aside class="example">
+
+        <aside class="example" title="Depending on inferred triples">
           <pre class="block ex-data">
-            :A :fatherOf :X .
-            :B :motherOf :X .
-            :C :motherOf :A .
+            :frontend :callsService :app .
+            :app      :queriesDatabase :db .
+            :db       :hasVulnerability :vuln1 .
           </pre>
           <pre class="block ex-rules">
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -404,84 +404,102 @@ <h2>SHACL Rules</h2>
...
+        <p>The above rules, applied to the data, will conclude that
+          `:frontend` depends on `:app`, and that `:app` depends on
+          `:db`, whatever the kind of dependency.
         </p>
 
         <p>
-          We can then derive `:descendedFrom` relationships by adding a rule that depends
-          on `:childOf` triples produced by the other rules:
+          We can then derive `:exposedTo` relationships by adding a rule that depends
+          on `:dependsOn` triples produced by the other rules:
```
liviorobaldo, line 439:
```suggestion
          on `:dependsOn` triples produced by the other rules (see also <a href="#rule-dependency" class="sectionRef"></a> below):
```

I think it would be useful to include a forward reference here, since rule dependencies are needed to define the strata (see also my next comment).

shacl12-rules/index.html
liviorobaldo, line 560:
```suggestion
in the negation element have been completed. The rule containing the negation is said to _depend_ on the rules generating these triples (see [§4.3 Rule Dependency](#rule-dependency) below).
```

As above, I would add a forward reference here. I think we should only introduce the concept of "rule dependencies" at this point and defer the discussion of "stratification" to Section 4, where it will be clarified.

In any case, both stratification and rule dependencies apply _not only_ to rules containing negated triples, but _also_ to rules containing non-negated triples. This is why I added a reference to §4.3 both here and in the "Basic Pattern" subsection above.

liviorobaldo, line 626:
```suggestion
This condition ensures that such rules do not 
loop back to themselves and cause an unbounded number of RDF terms.
```

I have removed the part stating that "This condition ensures that rules that create RDF terms do not generate new terms multiple times, with potentially different outcomes".

This is not the purpose of run-once rules. They are intended _only_ to prevent infinite loops. Preventing potentially different outcomes is instead the role of stratification.

liviorobaldo, line 630:
```suggestion
```

I do not understand this sentence. Moreover, this is the first time that "solution mapping" is mentioned in the whole document. For the sake of clarity, I would remove this sentence.

liviorobaldo, line 639: I think an example is also needed here IMHO, consistent with the other examples provided throughout the document.

@recalcitrantsupplant, if you agree, could you kindly add it?

shacl12-rules/index.html
```diff
@@ -640,16 +656,29 @@ <h3>Data Blocks</h3>
           matching in the body of rules.
         </p>
         <p>@@examples to be updated</p>
```
liviorobaldo, line 658:
```suggestion
```

This line can be removed, as the example has been indeed updated.

shacl12-rules/index.html
afs: Respec has ways of including section references.

```
<a href="#rule-dependency" class="sectionRef"></a>
```

shacl12-rules/index.html
liviorobaldo, line 726: I think this section should be moved to a later point in the document, in §4. I would place it immediately before §4.3.

afs:
Please can we put this in after #976 has merged to make the git merging easier?

shacl12-rules/index.html
liviorobaldo: @afs, I've changed it. Is it correct now? If so, I'll update the other occurrence as well.

afs:
Sorry - mix up merging #976. 
This was closed because its base branch was deleted.

shacl12-rules/index.html
afs: If you meant the syntax comparison, then another PR moves it.

liviorobaldo: I thought it was supposed to be moved to §4, but at least we agree it shouldn't be here. :-)

shacl12-rules/index.html
afs: We'll be able to see it if/when the suggestion is included.

shacl12-rules/index.html
TallTed: Please try to (approximately) maintain the line lengths in the source. Doing so makes ongoing review much easier than lines like the above, which is roughly triple the width, and will require horizontal scrolling no matter the width of the reviewer's monitor, significantly slowing review.

PR #1089, opened 2026-07-22
#1069: Iteration of SPARQL rules (adding sh:runOnce and sh:stratum too)

### LINKED ISSUE #1069: SHACL SPARQL Rules: clarify/improve handling of multiple iterations

HolgerKnublauch:
Closes #1069

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1069/shacl12-sparql/index.html#sparql-rules)

This is really just a first pass for now. Test cases will be added and needed for sure.

Note this is also adding support for run-once rules.

And it changes the sh:order to be based on individual rules only, no longer using the sh:order of shapes. This is because it allows us to just group all rules together into a single list before execution, including the global rules.

Also cc @bergos: does this look implementable in your engine?

Others: could this be repurposed to plug in an SRL engine for the rules that fall into the SRL subset? I have added extension points for changing the order and determining the run-once rules.


bergos:
Yes, I should be able to add it to my engine.

One minor thing: Can we define that global rules must be attached to a shapes graph, similar to the prefix declarations? That would require making a rule explicitly global, which would avoid a rule accidentally becoming global. It would also make it possible to import rule libraries and cherry-pick which rules to use.

liviorobaldo:
> One minor thing: Can we define that global rules must be attached to a shapes graph, similar to the prefix declarations?

I think this is a good idea

HolgerKnublauch:
> Can we define that global rules must be attached to a shapes graph

I am not a fan of that. People will forget this. Makes it harder to refactor via copy and paste. We also don't require sh:NodeShapes to be attached to a shapes graph, yet they are "active" by default for the validation engine. The presence of the sh:SPARQLRule in a graph should be enough to count as being in that graph.

I see little chance of people accidentally declaring global rules as (unlike shape rules) they don't use the $this variable.

shacl12-sparql/index.html
```diff
@@ -1894,41 +1898,177 @@ <h3>Types of SHACL Rules</h3>
...
+						</li>
+					</ul>
+					<p>
+						Some SHACL rule engines MAY implement their own function to determine the <a>run-once rules</a>,
+						for example when they automatically detect rules that may cause infinite iterations.
+					</p>
+					<aside class="example" title="Example of run-once rules">
+						<div class="shapes-graph">
+							<div class="turtle">
+ex:RunBeforeRule
```
scotthenninger, line 2076: Did you mean to in include this in the Person shape?

shacl12-sparql/index.html
```diff
@@ -1964,18 +2104,11 @@ <h3>sh:order</h3>
...
-						Rules may be <em>deactivated</em> by setting <code>sh:deactivated</code> to <code>true</code>.
-						Deactivated rules are ignored by the rules engine.
-					</p>
-					<p class="syntax">
-						<span data-syntax-rule="deactivated-maxCount">Each <a>rule</a> may have at most one <a>value</a> for the
-						property <code>sh:deactivated</code>.</span>
-						<span data-syntax-rule="deactivated-in">The <a>values</a> of <code>sh:deactivated</code> are either
-						of the <code>xsd:boolean</code> literals <code>true</code> or <code>false</code>.</span>
+						Some SHACL rule implementations MAY implement different ordering algorithms
+						and those algorithms MAY ignore the explicitly stated <code>sh:order</code> and <code>sh:runOnce</code> values.
```
scotthenninger: I’m unclear on including this statement. It seems to me that you want to detail the approach taken here. Other implementations will exist, of course, but they would be extensions to the recommendation.

shacl12-sparql/index.html
```diff
@@ -2015,53 +2148,47 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
-		}
-	}</pre>
-				</div>
 				<p>
-					The <a>triples</a> that are inferred by a <a>rule</a> do not <em>immediately</em> become part of the <a>data graph</a>,
-					i.e. the <a>triples</a> produced by one <a>rule</a> can not always be queried by other <a>rules</a>.
-					These policies reduce the likelihood of race conditions and better support parallel execution.
+					For a given set of <a>rules</a> in the <a>shapes graph</a>
+					an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
+					specified by <a href="#rules-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
```
scotthenninger:
```suggestion
					specified by <a href="#rules-order">sh:order</a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
```

scotthenninger (commented):
A couple of comments and a correction to consider.

shacl12-sparql/index.html
```diff
@@ -2015,53 +2148,47 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
-						If two <a>rules</a> have the same <a>execution order</a> then their newly inferred <a>triples</a> are
-						not visible to each other.
+						While the previous iteration has produced more inferred triples:<br />
+						execute one <a>iteration</a> over all <a>iterating rules</a>.
 					</li>
 					<li>
-						If the same <a>rule</a> is executed on multiple target nodes then the newly inferred <a>triples</a> are
-						not visible to the other target nodes.
+						Execute one <a>iteration</a> over all <a>run-after rules</a>.
 					</li>
```
mgberg: This removes details about when triples inferred during an iteration become visible to subsequent rules/iterations or not. Would it be worth putting that clarification back in?

mgberg (commented):
Should the addition of the new terms (e.g. sh:runOnce, sh:Before, sh:After) to the vocab.ttl file be part of this PR?

HolgerKnublauch:
> Should the addition of the new terms (e.g. sh:runOnce, sh:Before, sh:After) to the vocab.ttl file be part of this PR?

Yes, plus test cases. This is why the PR is currently still in Draft mode. I wanted to confirm the general direction before going into more details.

shacl12-sparql/index.html
HolgerKnublauch: The hidden visibility was only needed in the old single-pass approach because we thought it is "cleaner" for the rule ordering. It should now go without saying that the results of any rule are immediately visible to the other rules. But I'll add an explicit sentence for that.

shacl12-sparql/index.html
HolgerKnublauch: I had intentionally left out sh:order because some algorithms such as stratification may ignore sh:order and use their own logic. Therefore just the reference to the section.

shacl12-sparql/index.html
HolgerKnublauch: This is an attempt to build a bridge to SRL. If that attempt fails and the two languages remain completely separate then we COULD remove that sentence. However, even when the SRL editors don't like this plugin mechanism, other SPARQL dialects in the future may implement stratification and similar algorithms.

Does this clarify why this is here? I believe the language with a formal MAY is also necessary to ensure that a SHACL SPARQL engine with different algorithms remains compliant.

Note that if people don't use sh:order and some rules actually behave differently depending on the order, then the outcome right now is non-deterministic.

shacl12-sparql/index.html
HolgerKnublauch: No because it is a global rule that doesn't use the variable $this but only produces sample data. In fact before the rule runs there are no instances of Person, so it would never run.

shacl12-sparql/index.html
scotthenninger: OK, ignore this then.

shacl12-sparql/index.html
scotthenninger: I understand why the statement is in there, but in the end there is no telling what a different rules engine will include or exclude. There are other places in the document that alude to alternative implementations and this one feels too specific to me.

HolgerKnublauch: Yes I suggest we wait on whether the SRL supporters can live with the proposed layering (Full vs RL) and whether the current looping is roughly a generalization of SRL. The ideal outcome would be to have stratification as a plugin, but we are not there yet.

I'll mark this branch as ready for review now. I have more work planned on test cases but my implementation is now up to date and the general algorithm seems to work. We can then iterate and fine tune in future PRs as long as we agree on the general direction.

scotthenninger (approved):
Issues raised have been addressed. Looked over again and it looks good.

PR #1090, opened 2026-07-22
Add sh:RulesGraph

### LINKED ISSUE #1087: sh:RulesGraph?
We've defined classes `sh:DataGraph` and `sh:ShapesGraph`, and was wondering if there would be any value in also defining `sh:RulesGraph`.

Some users split their ontology graphs/files and shapes graphs/files and have a shapes graph with shapes that target classes, so I could imagine one might want to do something similar for rules in which a rules graph associates rules with shapes (or otherwise bundles a particular set of rules).

I could imagine some potential use cases for profiling. Perhaps you'd want to use different sets of rules depending on the source of data in order to get it to match a given profile. Or use different sets of rules to ensure the right triples are present in order to support different profiles, e.g., to support content negotiation by profile. 

I added the SRL label because I don't know if this concept is of interest/relevant to SRL rules or not.

mgberg:
Closes #1087

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1087/shacl12-sparql/index.html#RulesGraph)

Initially as draft to wait upon feedback on [this comment](https://github.com/w3c/data-shapes/issues/1087#issuecomment-5045870005).

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1990,10 +1990,10 @@ ex:RuleOrderExampleShape
 				</p>
 				<p><em>The remainder of this section is non-normative.</em></p>
 				<p>
+					The graph type classes (<code>sh:DataGraph</code>, <code>sh:ShapesGraph</code>, <code>sh:RulesGraph</code>)
+					represent roles that are not mutually exclusive; a single graph MAY be typed with more than one of these classes.
 					Rules graphs MAY contain <code>sh:rule</code> triples that reference <a>shapes</a> defined in
 					other graphs, allowing rules to be managed independently of shape and ontology definitions.
-					A graph MAY be typed as both <code>sh:ShapesGraph</code> and <code>sh:RulesGraph</code>
-					if it serves both roles.
 				</p>
 			</section>
 			<section id="Rules">
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1979,6 +1979,23 @@ ex:RuleOrderExampleShape
 					</p>
 				</section>
 			</section>
+			<section id="RulesGraph">
+				<h3>Rules Graph</h3>
+				<p>
+					A <dfn>rules graph</dfn> is an <a>RDF graph</a> that contains <a>SHACL rules</a>.
+				</p>
+				<p>
+					<span data-syntax-rule="RulesGraph">The <code>sh:RulesGraph</code> class MAY be used as an <code>rdf:type</code>
+					of the <a>IRI</a> of a graph that typically acts in the role of a <a>rules graph</a>.</span>
+				</p>
+				<p><em>The remainder of this section is non-normative.</em></p>
+				<p>
+					Rules graphs MAY contain <code>sh:rule</code> triples that reference <a>shapes</a> defined in
+					other graphs, allowing rules to be managed independently of shape and ontology definitions.
+					A graph MAY be typed as both <code>sh:ShapesGraph</code> and <code>sh:RulesGraph</code>
+					if it serves both roles.
+				</p>
+			</section>
 			<section id="Rules">
 				<h3>The sh:Rules Entailment Regime</h3>
 				<p>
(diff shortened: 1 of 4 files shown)


shacl12-sparql/index.html
```diff
@@ -1979,6 +1979,23 @@ <h3>sh:deactivated</h3>
...
+				<p>
+					<span data-syntax-rule="RulesGraph">The <code>sh:RulesGraph</code> class MAY be used as an <code>rdf:type</code>
+					of the <a>IRI</a> of a graph that typically acts in the role of a <a>rules graph</a>.</span>
+				</p>
+				<p><em>The remainder of this section is non-normative.</em></p>
+				<p>
+					Rules graphs MAY contain <code>sh:rule</code> triples that reference <a>shapes</a> defined in
+					other graphs, allowing rules to be managed independently of shape and ontology definitions.
+					A graph MAY be typed as both <code>sh:ShapesGraph</code> and <code>sh:RulesGraph</code>
+					if it serves both roles.
```
ajnelson-nist: The MAY here is a kind reminder, but leaving `sh:DataGraph` unmentioned is a slight headscratch now that might grow to a bigger headscratch in a few years.  Maybe we'd be better off with a more general statement that the roles are non-exclusive, repeated here and in the `sh:DataGraph` and `sh:ShapesGraph` sections.

shacl12-sparql/index.html
```diff
@@ -1979,6 +1979,23 @@ <h3>sh:deactivated</h3>
 					</p>
 				</section>
 			</section>
+			<section id="RulesGraph">
```
ajnelson-nist: For consistency with the core document:

```suggestion
			<section id="rules-graph">
```

shacl12-sparql/index.html
mgberg: @ajnelson-nist I made some changes for this, do they work for you?

ajnelson-nist: I like the text!  Though, it is now odd that the table-row listing of `RulesGraph` is in the profiling document.  This is just another nudge toward what we've been discussing in the profiling group calls - where is the right home(s) for such a table?

mgberg: I didn't add it at first but thought it might be useful with the extra text you recommended, but on second glance that change didn't really affect the profiling document at all. I removed it.

ajnelson-nist (approved):
LGTM, though with the caveat that I'm not one of the SHACL-SPARQL editors.

HolgerKnublauch (commented):
I like it in general. But one thing needs to be generalized then: for prefix resolution, we had recently added that sh:ShapesGraph has a special meaning to declare prefixes without a sh:prefixes triple. This now needs to be extended to be about sh:ShapesGraph and/or sh:RulesGraph.

shacl12-vocabularies/shacl.ttl
```diff
@@ -204,6 +204,13 @@ sh:ShapesGraph
     rdfs:subClassOf sh:Graph ;
     rdfs:isDefinedBy sh: .
 
+sh:RulesGraph
+    a rdfs:Class ;
+    rdfs:label "Rules graph"@en ;
+    rdfs:comment "The class of graphs that typically act in the role of a SHACL rules graph."@en ;
+    rdfs:subClassOf sh:Graph ;
```
HolgerKnublauch: Arguably this could be rdfs:subClassOf sh:ShapesGraph?

mgberg: Based on the language elsewhere in the document regarding rules being in the shapes graph, it would make sense to do that. When I put this together, I was thinking about [this type of use case](https://github.com/w3c/data-shapes/issues/1087#issuecomment-5037607781) in which there exists a graph/file that consists of rules that technically would get loaded into the shapes graph as the rules engine requires but doesn't "contain shapes".

I can certainly make that change if you'd like.

mgberg: > But one thing needs to be generalized then: for prefix resolution, we had recently added that sh:ShapesGraph has a special meaning to declare prefixes without a sh:prefixes triple. This now needs to be extended to be about sh:ShapesGraph and/or sh:RulesGraph.

If I make this change, should I also update the verbiage related to this, or is the fact that sh:RulesGraph is a subclass of sh:ShapesGraph sufficient?

HolgerKnublauch: The triple sh:RulesGraph rdfs:subClassOf sh:ShapesGraph is usually not present so we cannot rely on it. We need to state both URIs in the prefixes section explicitly.

shacl12-sparql/index.html
```diff
@@ -512,7 +512,7 @@ <h3>Prefix Declarations for SPARQL Queries</h3>
 			</p>
 			<p id="no-prefixes">
 				If a SPARQL query has no <a>value</a> for <code>sh:prefixes</code> then the system will use those <a>prefix declarations</a>
-				from the <a>shapes graph</a> that are <a>values</a> of <code>sh:declare</code> at a <a>SHACL instance</a> of <code>sh:ShapesGraph</code>.
+				from the <a>shapes graph</a> that are <a>values</a> of <code>sh:declare</code> at a <a>SHACL instance</a> of <code>sh:ShapesGraph</code> or <code>sh:RulesGraph</code>.
```
HolgerKnublauch: Maybe add 

<code>sh:RulesGraph</code> (which is a subclass of <code>sh:ShapesGraph</code>)

HolgerKnublauch (approved):

PR #1092, opened 2026-07-22
Shacl12 sparql edits srh

scotthenninger:
Closes #{ISSUE_NUMBER}

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/{BRANCH_NAME}/{FOLDER}/index.html)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1058,7 +1058,10 @@ ex:hasPattern
 						<p>
 							The following example illustrates the declaration of a constraint component based on a SPARQL SELECT query.
 							It is a generalized variation of the example from <a href="#sparql-constraints-example"></a>.
-							That SPARQL query included two constants: the specific property <code>ex:germanLabel</code> and the language tag <code>de</code>.
+							Where that SPARQL query included two constants: the specific property <code>ex:germanLabel</code> and the language tag de, 
+							this example generalizes the constraint so that any property can be defined with a language tag. In practice, the constraint component is applied 
+							by adding the parameter predicate(s) to a shape (e.g., ex:lang "de"), which a SHACL-SPARQL processor pre-binds to the corresponding SPARQL 
+							variable(s) (e.g., <code>$lang</code>) before executing the SELECT validator.
 							Constraint components make it possible to generalize such scenarios, so that constants get <a>pre-bound</a> with <a>parameters</a>.
 							This allows the query logic to be reused in multiple places, without having to write any new SPARQL.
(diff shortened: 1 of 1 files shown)


shacl12-sparql/index.html
```diff
@@ -926,7 +892,7 @@ <h3>An Example SPARQL-based Constraint Component</h3>
 				<aside class="example" title="Constraint component based on SPARQL">
 					<div class="shapes-graph">
 						<div class="turtle">
-sh:PatternConstraintComponent
+ex:PatternConstraintComponent
```
HolgerKnublauch: No it really IS sh:PatternConstraintComponent. It even uses sh:pattern and sh:flags and not ex:pattern. And the text above it explicitly states that it implements the component from the Core spec.

If that's confusing (which is seems to be), we should create a distinct example that has nothing to do with pattern.

shacl12-sparql/index.html
```diff
@@ -1092,7 +1090,10 @@ <h5>SELECT-based Validators</h5>
 						<p>
 							The following example illustrates the declaration of a constraint component based on a SPARQL SELECT query.
 							It is a generalized variation of the example from <a href="#sparql-constraints-example"></a>.
-							That SPARQL query included two constants: the specific property <code>ex:germanLabel</code> and the language tag <code>de</code>.
+							Where that SPARQL query included two constants: the specific property <code>ex:germanLabel</code> and the language tag de,
```
HolgerKnublauch: Could we leave this as <code>de</code> or "de" at least?

HolgerKnublauch (commented):
The second part of the PR is fine, the first needs more discussion.

shacl12-sparql/index.html
scotthenninger: Yea I would opt for having an example that doesn't implement a SHACL core construct.

shacl12-sparql/index.html
scotthenninger: Yes

shacl12-sparql/index.html
```diff
@@ -956,7 +922,7 @@ <h3>An Example SPARQL-based Constraint Component</h3>
 			"sh:message": "Value does not match pattern {$pattern}"
 		},
 		{
-			"@id": "sh:PatternConstraintComponent",
+			"@id": "ex:PatternConstraintComponent",
```
HolgerKnublauch: This one needs reverting too

HolgerKnublauch (approved):

PR #1096, opened 2026-07-23
GH-1095: graphMatch

### LINKED ISSUE #1095: `graphMatch` fix
From #1088 

> I think graph match definition written as `graphMatch(G, TP) = { μ | subst(μ, TP) is a triple in G }` leads to sets of infinite size: this definition allows an infinity of solutions because one can add any extra variable not used in the triple set to any RDF term. Adding the restriction that dom(μ) is the set of variables used by TP would fix that.

afs:
Closes #1095

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/gh1095-graphmatch/shacl12-rules/index.html#rule-eval-definitions)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -1986,9 +1986,10 @@ Output: an RDF graph GI of inferred triples
               </p>
             <div class="def">
               <p>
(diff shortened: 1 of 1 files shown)


recalcitrantsupplant (approved):
Looks good to me

PR #1084, opened 2026-07-20
Syntax differences between SRL and SPARQL Query

afs:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/srl-syntax/shacl12-rules/index.html)

PR #1097, opened 2026-07-24
Fix wording and punctuation in rules-sparql-relationship

TallTed:
No issue. Applies to text inserted in #1084.

Changes are very simple, so I have not taken the time to puzzle out a preview link.


afs (approved):
This will conflict with current editing but I make sure these changes do not get lost by the conflict resolution.

PR #1100, opened 2026-07-25
SRL ground data

### LINKED ISSUE #960: Providing default triples
SHACL node expressions is proposing removing text about manifesting derived triples. PR #958.

This was the way to get default triples output and use case motivated by the legacy "triple rule".

We may need a SHACL 1.2 Rules way to deliver on the requirement.

This rule does not work because the `NOT` looks for triples that the rule generates
(and also imagine `:distanceMiles` is possible defaulted/inferred based on, say, average speed and time).

This situation is detected; it violates stratification.

```
    RULE { ?x :distanceKm ?kilometers }
    WHERE { 
        ?x :distanceMiles ?miles
        NOT { ?x :distanceKm ?km }
        SET ( ?kilometers := ?miles * 1.60934 )
    }
```

To be possible, we need a rule element that looks only in the base graph  and not in the growing inferred triples graph.

With an obvious caveat:
```
        NOT EXISTS { ?x :distanceKm ?km }
```
may be more natural:
```
        NOT DATA { ?x :distanceKm ?km }
```
(it would include the rule set level `DATA{ }` used for additional facts/axioms)

or (an early syntax proposal for SPARQL)
```
        UNSAID { ... }
```
or
```
        NOT! { ... }
```

Alternatively, we could direct the rule.

Such rules would run once, before all other rules (this is a consequence of stratification, not a special case) but it make mixing inferred matching and base data matching into a rule programming exercise.

```
RULE { [] :distanceKm ?kilometers }
WHERE DATA {
    ?x :distanceMiles ?miles
    NOT { ?x :distanceKm ?km }
    SET ( ?kilometers := ?miles * 1.60934 )
}
```
or
```
RULE { ?x :distanceKm ?kilometers }
FROM DATA {
    ?x :distanceMiles ?miles
    NOT { ?x :distanceKm ?km }
    SET ( ?kilometers := ?miles * 1.60934 )
}
```

----

Overall, I'm inclined to go for:

```
        NOT DATA { ?x :distanceKm ?km }
```
and
```
        WHERE DATA { ... body ... }
```

with evaluation defined by the graph passed as input to the rule evaluation on each step.

afs:
Closes #960

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/srl-ground-data/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -696,7 +696,7 @@ This condition ensures that such rules do not loop back to themselves and cause 
           Rules are organised into [=rule sets=].  
           A rule set and a data graph (the [=base graph=]) are inputs to evaluation.
           The output is a graph, called the [=inference graph=],
-          which is the set of triples do not appear in the data graph.
+          which is the set of triples that do not appear in the data graph.
         </p>
         <p>
 	        During evaluation, triples inferred based on one rule are 
@@ -1714,6 +1714,54 @@ enddefine
           </p>
         </section>
       </section>
+
+      <section id="process-imports">
+        <h3>Processing Imports</h3>
+        <p>
+          A [=resolved rule set=] is produced from another rule set by 
+          recursively reading all rule sets mentioned
+          in the imports of that other [=rule set=].
+        </p>
+        <div class="def">
+          The <dfn>rule set merge</dfn> of two rule sets, `RS1` and `RS2`,
+          is a rule set, `MR`, defined as follows:
+          <pre>
+define ruleSetMerge(rule set R1,rule Set RS2):
+    MR.rules = RS1.rules &cups; RS2.rules
+    MR.data = merge(RS1.data, RS2.data)
+    MR.imports = {}
+    the result is MR
+enddefine
+          </pre>
+          <p>
+            where `merge` is the 
+            <a data-cite="RDF12-SEMANTICS#dfn-rdf-graph-merge">RDF merge</a>
+            operation.
+          </p>
+        </div>
+
+        <div class="algorithm">
+          <pre>
+define imports(rule set RS, set of URLs V), returning rule set
+    let I = the set of import URLs declared for the rule set RS
+    let RS2 be a rule set formed from RS.rules and RS.data
+    foreach URL x in I:
+        if x &notin; V:
+            V = V &cups; { x }
+            read rule set RS3 from URL x
+            RS2 = rulesetMerge(RS2, imports(RS3, V))
+        endif
+    endfor
+    result is RS2
+enddefine
+
+let RS be a rule set
+let V = {}
+if RS has a location, V = { location of RS }
+result is imports(RS, V)
+          </pre>
+        </div>
+      </section>
     </section>
 
     <section id="concrete-syntax">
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
         <h3>Matching Ground Data</h3>
         <div class="issue" data-number="960">
           <p>
-          Inferring default values can only be expressed by introducing extra
-          vocabulary to express the difference between inferred triples and
-          triples present in the input graph.
+            Inferring default values can only be expressed by introducing extra
```
```suggestion
            Inferring default values can only be expressed by introducing an extra
```

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
           <pre class="none">
             # Rule body matched against the input graph
             WHERE DATA { ... }
           </pre>
         </div>
+
+        <p>
+          Ground data is the triples in the [=base graph=].
+          As <a href="#rule-set-evaluation">rule set evaluation</a> it may be
+          necessary to pattern match on the original and not any inferred
```
```suggestion
          necessary to pattern match only on the original and not on any inferred
```

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
+          triples.
+        </p>
+        <p>
+          An example of this is setting a default value.
+          The rule must test whether the base graph already a value and, if not,
```
```suggestion
          The rule must test whether the base graph already contains a value and, if not, 
```

shacl12-rules/index.html
```diff
@@ -833,43 +887,45 @@ <h3>Relationship between SHACL Rules and SPARQL</h3>
           <li>
             The syntax of a rule body is similar to a SPARQL `CONSTRUCT` query, but has
             restrictions such as not having `UNION` or `OPTIONAL` syntax.
-            These restrictions ensure that variables are always bound, whereas in SPARQL, some
-            variables might not be bound in all pattern solutions.
             The effect of these can be achieved using [=well-formed rules=]
             so that the [=rule set=] can be analysed.
+            These restrictions ensure that variables are always bound, whereas in SPARQL, some
+            variables might not be bound in all pattern solutions
+            which could cause incomplete inferences.
```
robert-david: I think this requires a description, i.e. what incomplete means, or an example.

shacl12-rules/index.html
```diff
@@ -1029,24 +1085,50 @@ <h3>Elements of the Abstract Syntax</h3>
 
           <dt><dfn>Inference graph</dfn></dt>
           <dd>
-            An [=inference graph=] is an [=RDF Graph=] produced by evaluating a
-            [=rule set=].  It contains all triples not present in the [=base
+            An [=inference graph=] is an [=RDF Graph=] produced by 
+            a <a>rule set evaluation</a>. 
+            It contains all triples not present in the [=base
```
```suggestion
            It contains all inferred triples not present in the [=base
```

shacl12-rules/index.html
```diff
@@ -1029,24 +1085,50 @@ <h3>Elements of the Abstract Syntax</h3>
...
             graph=] that are inferred by applying the [=rule set=] to the
-            [=base graph=].
+            [=base graph=].  
+          </dd>
+
+          <dt><dfn>Rule set evaluation</dfn></dt>
+          <dd>
+            A [=rule set evaluation=] is the process of applying a [=rule set=]
+            to a [=base graph=] to produce an [=inference graph=].
+            A <dfn>rule evaluation</dfn> is the process evaluating a rule once.
```
```suggestion
            A <dfn>rule evaluation</dfn> is the process of evaluating a rule once.
```

shacl12-rules/index.html
```diff
@@ -1029,24 +1085,50 @@ <h3>Elements of the Abstract Syntax</h3>
...
+            Evaluating a rule produces all the triples given by the [=rule head=]
+            given the evaluation of the [=rule body=], regardless of whether a 
+            triple is in the [=base graph=], already inferred during [=rule set evaluation=],
+            or is a new inferrer triple.
```
```suggestion
            or is a new inferred triple.
```

robert-david (approved):

shacl12-rules/index.html
```diff
@@ -685,10 +696,10 @@ <h3>Evaluation of a Rule Set</h3>
           Rules are organised into [=rule sets=].
```
```suggestion
          Rules are organized into [=rule sets=].  
```

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
+        <p>
+          An example of this is setting a default value.
+          The rule must test whether the base graph already a value and, if not, 
+          the rule can then calculate a default value.
+        </p>
+        <aside class="example" title="Ground Data with NOT">
+          <pre class="block ex-data">
+            :A :distanceMiles 5 .
+            :A :distanceKm 8 .
+            :B :distanceMile 10 .
```
```suggestion
            :B :distanceMiles 10 .
```

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
+        </p>
+        <p>
+          An example of this is setting a default value.
+          The rule must test whether the base graph already a value and, if not, 
+          the rule can then calculate a default value.
+        </p>
+        <aside class="example" title="Ground Data with NOT">
+          <pre class="block ex-data">
+            :A :distanceMiles 5 .
+            :A :distanceKm 8 .
```
```suggestion
            :A :distanceKilometers 8 .
```

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
+          It is also possible to perform all [=rule body=] matching against the [=base graph=] by 
+          using `WHERE DATA`.
+        </p>  
+        <aside class="example" title="Ground Data with WHERE">
+          <pre class="block ex-rules">
+            RULE  { ?x :distanceKilometers ?kilometers }
+            WHERE DATA {
+                ?x :distanceMiles ?originalMiles .
+                NOT { ?x :distanceKilometers ?km } 
+                SET ( ?kilometers := xsd:integer(?originalM`iles *  1.60934))
```
```suggestion
                SET ( ?kilometers := xsd:integer(?originalMiles *  1.60934))
```

shacl12-rules/index.html
```diff
@@ -2255,7 +2350,8 @@ <h3>Evaluation of a Rule</h3>
             SEQ1 = {}
             for each solution <var>μ</var> in SEQ:
                 S = sequence{ <var>μ</var> }
-                NEG = evalRuleElements(N, S, G)
+                if rElt with DATA,
+                    NEG = evalRuleElements(N, S, G, GD)
```
recalcitrantsupplant: I think this should be like the `if R.groundData:` condition further down
```suggestion
                if rElt with DATA:
                    NEG = evalRuleElements(N, S, GD, GD)
                else:
                    NEG = evalRuleElements(N, S, G, GD)
```

recalcitrantsupplant (changes requested):
One material comment in the evaluation section

shacl12-rules/index.html
afs: Done. Note the issue box will get removed sometime soon.

shacl12-rules/index.html
afs: Reworded.

>  These restrictions ensure that variables are always bound, whereas in a SPARQL
            `CONSTRUCT` query template, triples involving an unbound variable
            are omitted from results yet triples in the same template, with all 
            variables bound, still generate output.

shacl12-rules/index.html
afs: Yes, good catch.

HolgerKnublauch:
Instead of introducing the DATA syntax, an alternative would be to align with SPARQL and allow GRAPH with certain "special" URIs. These special URIs could be either "prebound" variables or special URNs. There could be one graph for the asserted data, and one for the inferred triples only, maybe others in the future.

It wouldn't need to rely on SPARQL pre-binding but the grammar could understand cases like GRAPH $assertedGraph { ... } and switch to that graph under the hood for the inner block.

afs:
I didn't go with that idea because:

1. GRAPH should be reserved for general datasets. A specific use of datasets to handle base+evaluation graphs will get in the way.
1. Various triple store already have "special URIs" [SEP-000](https://github.com/w3c-cg/sparql-dev/blob/main/SEP/SEP-0004/sep-0004.md)

recalcitrantsupplant (approved):

afs:
BTW:

`WHERE DATA { }` rules will always be in stratum 0 and only need to be run once.

A rule set with only `NOT DATA` and no `NOT { }` is related to "semi-positive datalog" [Alice book, section 15.2](http://webdam.inria.fr/Alice/pdfs/Chapter-15.pdf) which does not need stratification (the RDF features, assignment and head template bnodes, get in the way though).

shacl12-rules/index.html
```diff
@@ -726,22 +737,65 @@ <h3>Evaluation of a Rule Set</h3>
...
             # Rule body matched against the input graph
             WHERE DATA { ... }
           </pre>
         </div>
+
+        <p>
+          Ground data is the triples in the [=base graph=].
+          As <a href="#rule-set-evaluation">rule set evaluation</a> it may be
+          necessary to pattern match only on the original and not on any inferred
+          triples.
```
TallTed, line 761:
```suggestion
          During <a href="#rule-set-evaluation">rule set evaluation</a> it may be
          necessary to pattern match only on the original triples, the ground data
          — i.e., not on any inferred triples.
```

afs: Thank - corrections made in my working copy.

Issue #1083, opened 2026-07-20
How open/closed should the SHACL UI specification be?


bergos:
## Context
This issue is prompted by our recent discussion on shape rendering order, but it's not actually about that specific topic — it's about something underlying it that I think we should resolve once, rather than argue about again in every future issue.
During the discussion on rendering order, I found myself questioning why we were spending time on it at all. Rendering order can reasonably depend on:
- user preferences
- screen size / device form factor
- user roles
- other contextual factors we haven't even considered yet

Given that variability, and given the limited time the task force has, I don't think we can (or should try to) pin down a single "correct" answer for cases like this in the spec itself. I suspect this pattern will come up again across many other topics we haven't gotten to yet.

## The underlying question
Before we go further into individual feature discussions, I'd like the task force (and the wider WG) to weigh in on a more basic question:

## How tightly should the SHACL UI specification constrain implementations?
My concern is that we are currently trending toward a specification that is too tightly defined, prescribing specific behaviors where multiple valid approaches exist and could reasonably coexist.

## Proposed position
I'd like to propose the following as a guiding principle for the spec going forward:
- Everything defined in shapes should end up rendered on screen — that's the contract with the author of the shapes.
- The spec should provide **building blocks**, not a single rendering algorithm. These building blocks should let an implementation: 
  - find the data to render
  - find matching widgets
  - determine the order in which shapes are rendered
  - apply other layout tools, potentially derived from existing properties/classes in other SHACL WG specifications
- Where multiple valid approaches exist for solving the same problem, the spec should **not** lock implementations into one specific approach.
- If one approach is strongly favored by a majority of the task force, it can still be included — but as a **SHOULD**, not a MUST, leaving room for implementations to differ with good reason.

## Ask
I'd like to get feedback from the rest of the task force (and the WG more broadly) on this general direction, before we keep resolving individual feature discussions like rendering order one at a time. If we can agree on a shared approach here, it should make future discussions faster and more consistent.

@nicholascar @PapoutsoglouE @HolgerKnublauch

HolgerKnublauch:
[...]

PR #1104, opened 2026-07-29
Add sh:order semantics for UI rendering

smessie:
Property groups and ungrouped property shapes are members of a **single ordered sequence**. For a property shape that specifies `sh:group`, its position in the top-level sequence comes from the `sh:order` of the referenced group; its own `sh:order` only positions it *within* that group. The rules apply recursively to nested groups where a renderer supports them.

The **effective order** of a property shape or property group is:

1. the value of its `sh:order`, if present;
2. otherwise the value of `shui:defaultOrder` in the global configuration, if present;
3. otherwise none.

Members are ordered by ascending numeric effective order, and members with **no** effective order are placed **after** all members that have one.

Ties must be broken deterministically. Renderers SHOULD order by resolved label, then lexicographically by identifier, and MAY use any other algorithm that is deterministic within that renderer.

## Why the default is "render last" rather than 0

An earlier draft assumed a default of `0` for elements without `sh:order`. This version instead treats them as having no effective order, so they are appended at the end of the form; this matches better with what existing renderers already do. It also means adding an unordered property shape to a fully ordered shape never shifts the existing properties.

Interoperability is preserved because the default is overridable **from the shapes graph**, not per-implementation: `shui:defaultOrder 0` sets the alternative behaviour, including the ability to use negative `sh:order` values to place properties before otherwise-unordered ones. Because the global configuration travels with the shapes, every implementation produces the same order.

## Depends on #900

`shui:defaultOrder` is a property of the `shui:Configuration` instance introduced by #900, which is not yet merged. Two consequences:

- ReSpec will report the `global configuration` links in this section as unresolved until #900 lands.
- **When #900 merges, a `shui:defaultOrder` row must be added to its global configuration parameter table:** *the effective order of property shapes and property groups that have no value for `sh:order`; its value is a literal with datatype `xsd:decimal` or `xsd:integer`.*

## Depends on #1083 

The exact wording related to the RFC2119 keywords will still be updated once the discussion in #1083 lands.
Let's use this PR to already advance on the SHACL UI contents and keep the conformance discussion in the other issue.

## Also included

Cross-links to the new section from Getting Started, `shui:ValueTableViewer`, and Property Roles. The Property Roles cross-link also states explicitly that its use of `sh:order` determines role precedence and is unrelated to presentation order, as the two uses were previously undistinguished. 
Adds `sh:order`, `sh:group`, and `sh:PropertyGroup` rows to the SHACL Core syntax index (the first populated "Section used" cells in that table).

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/order-semantics/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -808,7 +808,9 @@ ex:PersonShape a sh:NodeShape ;
           <a href="#displaying-validation-results">Displaying Validation
           Results</a> pattern. The renderer also uses
           <code>sh:group</code> and <code>sh:order</code> annotations to
-          control the form layout. In this example, the name-related properties
+          control the form layout, as defined in the related
+          <a href="#grouping-and-ordering">Grouping, Ordering, and Layout
+          Hints</a> pattern. In this example, the name-related properties
           are grouped into a <strong>Name</strong> section and ordered
           according to their <code>sh:order</code> values, while the birth
           date is rendered separately because it does not belong to a group.
@@ -920,7 +922,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
         </aside>
 
         <p class="note">
-          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is 
+          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is
           optional. The `rdfs:label` node expression should therefore handle
           missing values. For brevity, this example assumes that
           `schema:honorificPrefix` is always present.
@@ -1120,7 +1122,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
 
         <p>
           We have only briefly touched on cardinality constraints. Other useful
-          but more complex topics include logical constraints and using 
+          but more complex topics include logical constraints and using
           <a href="../shacl12-node-expr/#dfn-dynamic-shacl">Dynamic SHACL</a>
           to populate the values for the enum and autocomplete editors.
           You can find these in the <a href="#patterns">Patterns</a> section.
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
```diff
@@ -2158,10 +2160,143 @@ <h3>Enumerations, Select Lists, and Autocomplete</h3>
...
+        their <a>effective order</a>. Members that have no <a>effective order</a> MUST be placed
+        after all members that have one.
+      </p>
+
+      <p>
+        The <a>global configuration</a> property <code>shui:defaultOrder</code> declares the
+        <a>effective order</a> assumed for <a>property shapes</a> and property groups that do not
+        specify <code>sh:order</code>. Its value is a <a>literal</a> with <a>datatype</a>
+        <code>xsd:decimal</code> or <code>xsd:integer</code>, which are the datatypes that
+        <code>sh:order</code> permits.
```
danielbeeke, line 2203: what if we make a MUST or a SHOULD here instead of permits? 

And I wonder if we should allow xsd:string. Fine if we use numbers here, just suggesting as alternative.

shacl12-ui/index.html
```diff
@@ -2158,10 +2160,143 @@ <h3>Enumerations, Select Lists, and Autocomplete</h3>
...
+        </li>
+        <li>
+          Members that have the same resolved label are ordered lexicographically by the identifier
+          of the resource on which the <code>sh:order</code> is defined, that is, its IRI or blank
+          node identifier.
+        </li>
+      </ol>
+
+      <p>
+        Renderers MAY use an alternative tie-breaking algorithm, provided that the algorithm yields
```
danielbeeke, line 2223: Did we have multiple positions in the task force about the details of the tie breaking algo? I wonder if we could just have one instead.

danielbeeke (commented):
I have two questions

shacl12-ui/index.html
smessie: Well, I think I can answer both of your questions here with the fact that this is defined by SHACL Core and UI should not say anything extra about that.
If we want to allow xsd:string as well, we should escalate that to the SHACL 1.2 Core document.

shacl12-ui/index.html
smessie: I don't think we had a concrete idea for another tie breaking algorithm besides Bergi that wanted to keep the spec open.
I'm in favor of having exactly one as well, such that we can guarantee interoperability with this standard across implementations. 
We should revisit this as well however once #1083 lands.

danielbeeke: okay from my perspective we could first merge this and later pick one.

danielbeeke (approved):

Issue #999, opened 2026-07-12
Update pre-binding to align with SPARQL 1.2


HolgerKnublauch:
> The prebinding needs updating. While the AFR mechanism I drafted was the best at the time, SPARQL 1.2 has evolved.
> 
> The syntax restrictions are necessary (though they need checking).
> 
> The Values Insertion interacts with SPARQL 1.2 for EXISTS. It isn't necessary - the way the TQ product does it nowadays does not interact.
> 
> The benefit means pre-binding isn't tied to the engine, and specifically a local engine, and only requires lightweight parsing (for security). It also means it can work on a triple store over the SPARQL HTTP protocol. 

 _Originally posted by @afs in [#766](https://github.com/w3c/data-shapes/issues/766#issuecomment-4932864614)_

See also #647

HolgerKnublauch:
@afs do you have some specific input here?

Looking at https://www.w3.org/TR/sparql12-query/#changes-1-1 I don't see what specifically has changed in SPARQL 1.2.

Is Jena substitution still the "correct" implementation?

afs:
We need a system-neutral solution for a W3C standard.

There are several issues with pre-binding.  We need to decide what the group consensus is.
Some are below, but there may well other concerns from user experiences.

1.
Clash with SPARQL 1.2: `EXISTS` (and, in the future, `LATERAL`) uses an approach that overlaps with the current pre-binding but it is different. This would avoided if 3 is dealt with.

2.
Modifying the algebra is not something SPARQL engines allow through service APIs if they allow it at all.
They are "query in, results out".

3.
At the moment, pre-binding is code-API specific, which is making it local only.

They are multiple ways to do an neutral parameterization-style solution. It would still be SHACL-custom but not engine specific. I suspect that multiple SHACL-SPARQL engines take this approach already.

HolgerKnublauch:
@afs Do you have suggestions on how to proceed here? I agree that modifying the algebra isn't supported by the SPARQL protocols, but then I guess query string pre-processing would be the only alternative? For example Injecting VALUES clauses in all relevant branches of the syntax tree? That wouldn't support passing in blank nodes, I guess.

Are there any new SPARQL 1.2 features that help to reformulate this?

We are also not supposed to break backwards compatibility too much...

afs:
First, see what implementations are currently actually doing.
Check the existing test case for coverage and add some more if needed.

Second, decide requirements.
[...]

PR #1106, opened 2026-07-30
SPARQL: Delete outdated pre-binding tests

HolgerKnublauch:
https://github.com/w3c/data-shapes/issues/999#issuecomment-5137136574


scotthenninger (approved):
Seems like a safe and simple change

afs:
Asking for a review at 00:06 my time and merging at 02:42 is unlikely to get a review.

HolgerKnublauch:
I won't ask you again. Thanks.

TallTed:
>> [@afs] Asking for a review at 00:06 my time and merging at 02:42 is unlikely to get a review.
>
> [@HolgerKnublauch] I won't ask you again. Thanks.

I don't think @afs meant "don't ask me for reviews". (Please correct me if I've misunderstood you, @afs.)

Rather, I think he was suggesting that you allow a reasonable interval of waking and working hours, between your request for a review and your treating a lack of response as meaningful."

I believe I have asked for the same several times, not only on those PRs where you explicitly ask for a review, but on all PRs.

Hereby, I make the same request, again.

HolgerKnublauch:
There are different reasons to ask someone for reviews, and here it was more for awareness and for the case that the other reviewers are not confident about the implications. I felt the change itself was uncontroversial enough. Surely I would have waited much longer otherwise.

Also, repeating myself: an approved and merged PR does not need to be the end of the story. It is normal practice to have follow-up PRs to change or fine tune previous commits. Even GitHub issues can be reopened at any time. So I don't think we need to be overly strict here when we feel that moving fast is the best thing to do.

TallTed:
> here it was more for awareness and for the case that the other reviewers are not confident about the implications

Awareness and other purposes *other than PR reviews as they are generally understood* are better accomplished by simply tagging the people in question. I'm quite sure you, @HolgerKnublauch, know how to do this. For others, note that you just need to name the user whose attention you want with the leading `@`, as I've just done, and include whatever other details may be relevant.

Then keep actual PR review requests for their intended purpose and general use.

PR #1107, opened 2026-07-31
#1103: Added sh:sourceRule

### LINKED ISSUE #1103: Vocabulary to record provenance of inferred triples
For debugging, tracing and governance purposes, it should be useful to be able to record the rule that has inferred a given triple. RDF 1.2 reifiers could be used for that, allowing this info to be stored in the same inferences graph as an option, without affecting the main inferences:

```
ex:s ex:p ex:o {| sh:sourceRule ex:SomeRule |} .
```

The name sourceRule is suggested as it is consistent with sh:sourceConstraintComponent and sh:sourceShape.

HolgerKnublauch:
Closes #1103

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1103/shacl12-sparql/index.html#sourceRule)


shacl12-sparql/index.html
```diff
@@ -2245,6 +2246,26 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
 				</p>
 			</section>
+			<section id="sourceRule">
+				<h3>Tracking the Rule that has produced a Triple (sh:sourceRule)</h3>
+				<p>
+					Rule engines MAY have an option to generate additional triples that can be used to track
+					which rules have produced the inferred triples.
+					The property <code>sh:sourceRule</code> can be used in a <a>reifier</a> of a triple in the inferences graph
+					to link the triple with the <a>rule</a>.
+					The values of <code>sh:sourceRule</code> SHOULD be <a>IRIs</a> and the <a>reifier</a> SHOULD be a <a>blank node</a>.
```
```suggestion
					The values of <code>sh:sourceRule</code> SHOULD be <a>IRIs</a>.
```
because a generated UUID URN would do just fine, and indeed as reading file causes new blank nodes, sometime better generated at allocation time is better.

shacl12-sparql/index.html
```diff
@@ -2245,6 +2246,26 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+				</p>
+				<p>
+					The following example indicates that the triple <code>ex:Alice ex:friend ex:Bob</code> has been
+					inferred by the rule <code>ex:SymmetricPropertyRule</code>.
+				</p>
+				<pre>
+ex:Alice ex:friend ex:Bob {| sh:sourceRule ex:SymmetricPropertyRule |} .</pre>
+				<p>
+					If a rule engine adds these triples into the same graph as the actual inferences,
+					it MUST do so as a post-processing step so that the new triples do not interfere with the executing rules.
```
```suggestion
					If a rule engine adds these triples, the triples MUST NOT be visible to executing rules.
```
Minor suggestion - state the requirement. There are several ways to implement.

afs (commented):
Suggestions.

nicholascar (approved):

shacl12-vocabularies/shacl.ttl
```diff
@@ -1969,6 +1969,13 @@ sh:runOnce
 	rdfs:range xsd:boolean ;
 	rdfs:isDefinedBy sh: .
 
+sh:sourceRule
+	a rdf:Property ;
+	rdfs:label "source rule"@en ;
+	rdfs:comment "Can be used in reifiers to track the rule that has produced a given inferred triple."@en ;
+	rdfs:range sh:Rule ;
```
ajnelson-nist, line 1976: I'm a little fuzzy on the typing - do reified nodes have an implicit `rdf:type` we could use here for `rdfs:domain`?  I recall [`rdfs:Proposition`](https://www.w3.org/TR/rdf12-schema/#ch_proposition), but forget if there's some nuance where it's not appropriate to chance inferring those by RDFS inferencing.

Relatedly, SHACL-SHACL would benefit from a shape constraining `nodeKind` to `TripleTerm`---if I'm not mistaken?

HolgerKnublauch: As these sh:sourceRule annotations can appear on any triple, there is not much that SHACL-SHACL can say about them. It can, for example, not use sh:reifierShape because that would require individual properties.

And implicit type triples are IMHO not desirable because they are just noise that would break the compact rendering of sh:sourceRule triple reifiers using the {| ... |} syntax.

So all that SHACL-SHACL can do is say that sh:nodeKind sh:BlankNodeOrIRI.

HolgerKnublauch:
I am merging this into main now. Before people complain that the time frame was too short, I want to work on sh:tempTriple next and that also requires reified triples. Best to move quickly before there is a pile of dependent PRs that make merging difficult.

If anyone has further comments, I can apply them in a follow-up PR later. Thanks for your attention to this matter.

PR #1111, opened 2026-08-03
Persisting validation results

### LINKED ISSUE #1117: Persisting outcomes of validation as provenance in RDF
I would like to persist the results of validation so that I can know the validity of data graphs in relation to shapes graphs over time. I'd like to
persist the results in RDF so I can put it in the same database used for the data validated and the Shapes Graphs. I'd like to have a [PROV](https://www.w3.org/TR/prov-o/)-like way of representing the information so it's compatible with other provenance recorded in a W3C-standardised way.

SHACL 1.0 has a validation outcome that is characterised in RDF with a `sh:ValidtionReport` object linking to validation outcomes, but this is not linked back to a representation of the validation event as a `prov:Activity` nor are details of the validation tool recorded.

Things I'd like to see recorded:

* the validation event
* the inputs to the event
    * the data graph validated
    * the shapes graph used to do the validation
    * any other inputs
* the outputs from the event
* details of the tool used to do the validation

nicholascar:
* removed duplicated content about profiling background which is now in [PROF](https://www.w3.org/TR/dx-prof/)
* added a new _Packaging SHACL_ motivation and subsection to do PROV-O-style provenance of validation

See this rendered at https://raw.githack.com/w3c/data-shapes/refs/heads/profiling-multi-issue/shacl12-profiling/index.html


shacl12-profiling/index.html
```diff
@@ -242,25 +224,24 @@ <h3>What is profiling?</h3>
...
+      [[[dx-prof]]] defines a <em><a data-cite="dx-prof#dfn-profile">profile</a></em> to be:
     </p>
     <p style="padding: 0 50px 0 50px;">
       <em>a specification that constrains, extends, combines, or provides guidance or explanation about the use of
         other specifications</em>
     </p>
     <p>
       If a <a data-cite="shacl12-core#shapes-graph">shapes graph</a> is taken to
-      be a "specification," then not only is the data that is valid according to
+      be a "specification", then not only is the data that is valid according to
```
ajnelson-nist, line 235: Grammar nit - while I prefer this style when referencing code artifacts where every character counts, English-for-humans should quote the way you revised away from.

```suggestion
      be a "specification," then not only is the data that is valid according to
```

nicholascar: Sorry but I just can't agree to that! It's the word "specification" that's quoted, not "specification," which includes grammar for the sentence in which the word is used, in this case.

Jamie-SA: Not sure if this is well known, but I think it is a country specific thing. US English writing style guidelines say to include the comma (and periods) in the quote. Even though I am a US English native speaker this has always seemed wrong to me and I don't follow it. I have heard that some other English speaking countries do not follow this style guideline (which I was happy to hear so I can point this out next time someone tries to correct me :).

mgberg:
Section 5 of the document still needs to be updated, e.g., to account for the new classes and the removal of ProcessorConfiguration.

nicholascar:
> Section 5 of the document still needs to be updated, e.g., to account for the new classes and the removal of ProcessorConfiguration.

I've done this Matt. Next on to Sect 4

shacl12-profiling/index.html
```diff
@@ -830,145 +853,123 @@ <h3>Creating other Profiles</h3>
...
+    This section describes how to create profiles of RDF resources with SHACL.
   </p>
   <p>
     General profiling terms and elements from [[[dx-prof]]] are used as well as more
     specialised SHACL profiling elements defined in the <a href="#vocabulary"></a>.
   </p>
+  <section id="profiling-with-shacl-intro" class="informative">
+    <h3>Introduction</h3>
+    <p>
+      Since SCHAL was released in 2017, it has been used extensively to make closed-world RDF data models for use by
```
YoucTagh:
```suggestion
      Since SHACL was released in 2017, it has been used extensively to make closed-world RDF data models for use by
```

shacl12-profiling/index.html
```diff
@@ -830,145 +853,123 @@ <h3>Creating other Profiles</h3>
...
+      applications, building on open-world RDF models such as W3C ontologies.
+    </p>
+    <p>
+      A typical example, the <a href="https://linked.data.gov.au/org/idn">Indigenous Data Network</a> defines a
+      <a href="https://www.w3.org/TR/dx-prof-1.0/#dfn-profile">profile</a> of <a href="https://schema.org">schema.org</a> which constrains
+      that general-purpose RDF vocabulary for use in the field of Indigenous data cataloguing. The IDN provides an
```
YoucTagh:
```suggestion
      that general-purpose RDF vocabulary for use in the field of Indigenous data cataloguing. The IDN provides a
```

ajnelson-nist:
After discussion on the figure heading Section 5, I had a question on cardinalities: Say I have a data graph that uses two shapes graphs:

```turtle
ex:DG sh:shapesGraph ex:SG1, ex:SG2 .
```

How many `sh:ValidationReport`s should there be?  1 or 2?

On the call this morning, we confirmed just 1.  Suppose `SG2` just contains `sh:deactivated` for some shapes in `SG1`.  This is also consistent with @mgberg 's separation of the various meanings of "shapes graph" - the shapes graph used for validation being built from all shapes graphs referenced by linking `sh:shapesGraph`.  The 1 `ValidationReport` should link both shapes graphs with `sh:usedShapesGraph`.

It could be helpful to list cardinalities in the edges of Figure 3.

mgberg:
> It could be helpful to list cardinalities in the edges of Figure 3.

And if there are cardinalities, add them into the shapes.

shacl12-profiling/index.html
```diff
@@ -1304,103 +1363,92 @@ <h3>Property: used data graph</h3>
...
+            <td><a data-cite="shacl12-core#usedShapesGraph"><code>sh:usedShapesGraph</code></a></td>
           </tr>
           <tr>
             <th>Is defined by:</th>
             <td>[[[shacl12-core]]]</td>
           </tr>
           <tr>
             <th>Definition:</th>
-            <td>The subject Validation Report was generated by a processor described by the object Processor Configuration.</td>
+            <td>The property that is used to indicate the utilization of a shapes graph by a validation activity.</td>
```
mgberg, line 1390: We may want to provide more guidance/clarity around the usage of `sh:usedShapesGraph`, i.e., would it be used to reference:

- The provided shapes graph (and any graphs that are referenced by `sh:shapesGraph`), i.e. the root(s) of import closure
- The set of all graphs in the shapes graph after import closure starting at the above root(s)
- Dealer's choice

nicholascar: I had badly copied and pasted material there. Fixed with Commit https://github.com/w3c/data-shapes/pull/1111/commits/98a2e3607f016acbaeda10ca97e75336dfa624cd.

I don't think we should mention anything to do with the import closures here as the advice should be given in Core as this document proposes to just follow the Core mechanism for validation.

nicholascar:
> After discussion on the figure heading Section 5, I had a question on cardinalities: Say I have a data graph that uses two shapes graphs:
> 
> ```turtle
> ex:DG sh:shapesGraph ex:SG1, ex:SG2 .
> ```
> 
> How many `sh:ValidationReport`s should there be? 1 or 2?
> 
> On the call this morning, we confirmed just 1. Suppose `SG2` just contains `sh:deactivated` for some shapes in `SG1`. This is also consistent with @mgberg 's separation of the various meanings of "shapes graph" - the shapes graph used for validation being built from all shapes graphs referenced by linking `sh:shapesGraph`. The 1 `ValidationReport` should link both shapes graphs with `sh:usedShapesGraph`.
> 
> It could be helpful to list cardinalities in the edges of Figure 3.

As per my comment to @mgberg's point above: we should defer to Core for all these kinds of things and only offer advice beyond what Core is doing. Core contains the information about performing validation, assembling inputs etc. so we shouldn't touch those.

mgberg:
> we should defer to Core for all these kinds of things and only offer advice beyond what Core is doing. Core contains the information about performing validation, assembling inputs etc. so we shouldn't touch those.

Hmm. Where would that go in Core? This provenance feature is not discussed in Core, and I don't really think there's any language around tracking the graphs that were imported or not in Core, just the procedure used to collect them and that they get merged into one union graph. Unless that changes, I don't really see how `sh:usedShapesGraph` could be used for anything other than the root(s) of imports.

Actually, it may be worse than that- if you provide a graph of shapes to a validator (i.e. not via `sh:shapesGraph`), then would there even be a URI associated with the shapes graph based on the current language in Core? Similarly, would there be a URI for the data graph based on the language in Core (as it currently assumes that the data graph is complete and does no closure of its own, meaning you would have to do any merging necessary in advance)?

nicholascar:
> After discussion on the figure heading Section 5, I had a question on cardinalities: Say I have a data graph that uses two shapes graphs:
> 
> ```turtle
> ex:DG sh:shapesGraph ex:SG1, ex:SG2 .
> ```
> 
> How many `sh:ValidationReport`s should there be? 1 or 2?
> 
> On the call this morning, we confirmed just 1. Suppose `SG2` just contains `sh:deactivated` for some shapes in `SG1`. This is also consistent with @mgberg 's separation of the various meanings of "shapes graph" - the shapes graph used for validation being built from all shapes graphs referenced by linking `sh:shapesGraph`. The 1 `ValidationReport` should link both shapes graphs with `sh:usedShapesGraph`.
> 
> It could be helpful to list cardinalities in the edges of Figure 3.

I have added Usage Notes to the definitions of the `usedDataGraph` and `usedShapesGraph` predicates to indicate multiple objects are combined. I think this is better than diagram clutter and it's more precise too.

nicholascar:
> > we should defer to Core for all these kinds of things and only offer advice beyond what Core is doing. Core contains the information about performing validation, assembling inputs etc. so we shouldn't touch those.
> 
> Hmm. Where would that go in Core? This provenance feature is not discussed in Core, and I don't really think there's any language around tracking the graphs that were imported or not in Core, just the procedure used to collect them and that they get merged into one union graph. Unless that changes, I don't really see how `sh:usedShapesGraph` could be used for anything other than the root(s) of imports.
> 
> Actually, it may be worse than that- if you provide a graph of shapes to a validator (i.e. not via `sh:shapesGraph`), then would there even be a URI associated with the shapes graph based on the current language in Core? Similarly, would there be a URI for the data graph based on the language in Core (as it currently assumes that the data graph is complete and does no closure of its own, meaning you would have to do any merging necessary in advance)?

Core may get the provenance work as per Issue #1245. If so, we can defer more. If not, we have to cover things here.

I have also put in usage notes on the cardinality of Data and Shapes graphs, as seen by the ValidationActivity/Report.

> Actually, it may be worse than that- if you provide a graph of shapes to a validator (i.e. not via `sh:shapesGraph`), then would there even be a URI associated with the shapes graph based on the current language in Core? 

Buyer be ware! Packaging SHACL gives you suggested methods for packaging SHACL but doesn't mandate you or anyone follow them. But if you don't then yeah, perhaps you can't nicely persist results. So the provenance reporting is dependent on packaging and I think this is clear as it's the last element in the Packaging SHACL section.

shacl12-profiling/index.html
```diff
@@ -1342,19 +1342,29 @@ <h3>Property: used data graph</h3>
           </tr>
           <tr>
             <th>Definition:</th>
-            <td>The subject Validation Report was generated by validating the object Data Graph.</td>
+            <td>The property that is used to indicate the utilization of a data graph by a validation activity.</td>
           </tr>
           <tr>
             <th>Domain:</th>
-            <td><a href="#cls-validationreport">Validation Report</a></td>
+            <td><a href="#cls-validationactivity"><code>owl:ValidationActivity</code></a></td>
```
mgberg, line 1349: Here and a few other places you have `owl:ValidationActivity` instead of `sh:ValidationActivity`

nicholascar: Found and replaced!

mgberg (commented):
Looks good to me

mgberg (approved):
Looks good to me

YoucTagh (approved):

PR #1112, opened 2026-08-03
[UI] Referencing external dynamic IRIs in sh:in for federated SHACL validation and UI option lists. 

### LINKED ISSUE #846: [Core & UI] Referencing external dynamic IRIs in sh:in for federated SHACL validation and UI option lists.
This proposal allows someone to have federated references in their dataset and select/view those in a rich user interface.

### Ingredient 1: Dynamic SHACL with a SERVICE clause

Given that:
- [Dynamic SHACL](https://www.w3.org/TR/shacl12-node-expr/#dynamic-shacl) allows for a [select expression](https://www.w3.org/TR/shacl12-sparql/#SelectExpression) with `sh:in`
- The SERVICE clause limitation has been made [less strict](https://github.com/w3c/data-shapes/issues/374)

We have:
```turtle
ex:propertyCity
  sh:path ex:city ;
  sh:in [
	  sh:select """
		select ?city where {
			SERVICE <http://example.org/sparql> {
				?city a ex:City .
			}
		}
		""" ;
  ] .  
```

__Notice:__
- This query returns all valid items. When validating, the one allowed binding ?city can be bound with the selected value, if nothing returns we have a validation error.

### Ingredient 2, Proposal  for `shui:DisplayProvider`

A mechanism to fetch labels and possibly other properties of a resource.

```turtle
ex:propertyCity
  sh:path ex:city ;
  sh:in ( ex:Amsterdam ex:Rotterdam ex:NewYork ) ;
  shui:displayProvider ex:myDisplay ;
].
	
ex:myDisplay
	a shui:DisplayProvider ;
	sh:select """
		select $value $label $depiction $description where {
			$value rdfs:label $label .
			optional { $value schema:depiction $depiction }
			optional { $value schema:description $description }
		}
	""" .
```

- `$value` will be pre-bound with the value node of which the SHACL renderer needs to render a label
- This concept can be used with the `shui:EnumSelectEditor` and the `shui:AutocompleteEditor` and other places where it might apply.
- If no `shui:displayProvider` is given the SHACL renderer may fall back to it's own label resolution algorithm.
- $label is a special binding. You must use this binding. If not bound the implementation could fallback to the local name of the IRI.
- We also allow $depiction and $description.
### Ingredient 3, Proposal  for `shui:ResourceFilter`

A mechanism to filter resources with a search term.

```turtle
ex:propertyCity
  sh:path ex:city ;
  sh:in ( ex:Amsterdam ex:Rotterdam ex:NewYork ) ;
  shui:resourceFilter ex:myFilter ;
].

ex:myFilter
	a shui:ResourceFilter ;
	sh:select """
		select $value where {
			$value rdfs:label ?label .
			filter(contains(?label, $searchTerm))
		}
	""" .
```

- `$value` will be pre-bound with the value node of which the SHACL renderer needs to if it should be shown.
- `$searchTerm` will be bound with the search term the user is entering.
- This concept can be used with the `shui:AutocompleteEditor` and other places where it might apply.
- If no `shui:resourceFilter` is given the SHACL renderer may fall back to it's own resource filter. Which in turn can use the display provider if defined or the fallback label resolution algorithm.

## Example using all ingredients together

```turtle
ex:propertyCity
  sh:path ex:city ;
  shui:displayProvider ex:myDisplay ;
  shui:resourceFilter ex:myFilter ;
  sh:in [
	  sh:select """
		select ?city where {
			SERVICE <http://example.org/sparql> {
				?city a ex:City .
			}
		}
		""" ;
  ] .  

ex:myFilter
	a shui:ResourceFilter ;
	sh:select """
		select $value where {
			SERVICE <http://example.org/sparql> {
				$value rdfs:label ?label .
				$value ex:synonym ?synonym .
				$value rdfs:comment ?description .
				bind (concat(
					?label, " ", 
					?synonym, " ", 
					?description
				) as ?allText)

				filter(contains(?allText, $searchTerm))
			}
		}
	""" .


ex:myDisplay
	a shui:DisplayProvider ;
	sh:select """
		select $value $label $depiction $description where {
			SERVICE <http://example.org/sparql> {
				$value rdfs:label $label .
				FILTER(lang($label) = "en")
				
				optional { $value schema:depiction $depiction }
				optional { $value schema:description $description }
			}
		}
	""" .
```

### Notes

- Implementations can choose to combine the queries all together, see this [PoC](https://poc-label-resolution-resource-filter.shapething.com/).
- Implementations are free to optimize these queries however they like, for example by batching, as long as the outcome matches as described by the spec.
- When an endpoint has support for something like: `bif:contains` the end user could use that mechanism inside the `shui:ResourceFilter` select query.

danielbeeke:
## Abstract

This proposal introduces a standardized syntax for extending SHACL shapes to handle the display of the selected value and the options of the [AutocompleteEditor](https://www.w3.org/TR/shacl12-ui/#AutoCompleteEditor) and the [EnumSelectEditor](https://www.w3.org/TR/shacl12-ui/#EnumSelectEditor) and possibly others. It mostly piggybacks on SHACL core, Dynamic SHACL and Node Expressions. It also introduces one new concept `shui:searchQuery`, which can be used when the shape writer wants to have full control over the query that is used to fetch the autocomplete items.

Non-SPARQL or hybrid backends can consume the roles static metadata directly without needing SPARQL expression evaluations.

### Using Dynamic SHACL and shui:propertyRole

__Goal:__ Have a nice display of the selected value and the options of the [AutocompleteEditor](https://www.w3.org/TR/shacl12-ui/#AutoCompleteEditor) and the [EnumSelectEditor](https://www.w3.org/TR/shacl12-ui/#EnumSelectEditor) and possibly others.

__How to apply:__ 
1. Specify `sh:node` on a `sh:PropertyShape`  pointing to shape properties containing`shui:propertyRole` declarations so that a SHACL renderer can pick these up and show a depiction, a label and a sub label.
2. Use a [SelectExpression](https://www.w3.org/TR/shacl12-sparql/#SelectExpression) to specify the values of the `sh:in`.
3. Possibly use the keyword `SERVICE` to make a federated SPARQL call.

```turtle
ex:BookShape
    a sh:NodeShape ;
    sh:targetClass ex:Book ;
    sh:property [
        sh:name "Author" ;
        sh:path dct:creator ;
        sh:class ex:Person ;
        sh:node ex:PersonDisplayShape ;
        sh:in [
            sh:select """
                PREFIX ex: <http://example.com/>
                PREFIX schema: <https://schema.org/>
                SELECT ?value
                WHERE {
                     SERVICE <http://example.com/sparql> {
                         ?value a ex:Person ;
                                   schema:name ?name .
                    }
                }
            """
        ]
    ] .

ex:PersonDisplayShape
    a sh:NodeShape ;
    sh:property [
        sh:path schema:name ;
        shui:propertyRole shui:LabelRole ;
        sh:minCount 1; 
        sh:order 1 ;
    ] ;
    sh:property [
        sh:path ex:pseudonym ;
        shui:propertyRole shui:SubLabelRole ;
        sh:order 2 ;
    ] .
```

When this combination of `sh:node` and `sh:in` with a nested `sh:select` is detected by the SHACL renderer it MAY use these declarations to resolve the property role values. For now I see the following important property roles: `shui:LabelRole`, `shui:SubLabelRole`, `shui:DepictionRole`. Implementations can create their own roles for a more customized display. 

When multiple property shapes contain the same `shui:PropertyRole` the one with the highest order wins. Other ties are broken via codepoint comparision.

### Using a concrete SPARQL query to enable fulltext search

Some vendors have fulltext search. We want to be able to use that when prefered by the shape writer.

```turtle
ex:BookShape
    a sh:NodeShape ;
    sh:targetClass ex:Book ;
    sh:property [
        sh:name "Author" ;
        sh:path dct:creator ;
        sh:class ex:Person ;
        sh:node ex:PersonDisplayShape ;
        sh:in [
            sh:select """
                PREFIX ex: <http://example.com/>
                PREFIX schema: <https://schema.org/>
                
                SELECT ?value
                WHERE {
                     SERVICE <http://example.com/sparql> {
                         ?value a ex:Person ;
                    }
                }
            """ ;

			# This query will then be wrapped by the auto generated query that adds the property shapes with property roles. If this is not given the SHACL renderer generates it.
			shui:searchQuery """
				PREFIX ex: <http://example.com/>
				PREFIX text: <http://jena.apache.org/text#>
	
				SELECT ?value WHERE { 
					SERVICE <http://example.com/sparql> {
						?value text:query (?searchTerm ?uiLanguage) . 
						?value a ex:Person . 
					}
				}
			"""    
        ] ;
    ] .

ex:PersonDisplayShape
    a sh:NodeShape ;
    sh:property [
        sh:path schema:name ;
        shui:propertyRole shui:LabelRole ;
        sh:order 1 ;
        sh:minCount 1 ;
    ] ;
    sh:property [
        sh:path ex:pseudonym ;
        shui:propertyRole shui:SubLabelRole ;
        sh:order 2 ;
    ] .
```

The value of `shui:searchQuery` MUST NOT contain a `LIMIT` or an `OFFSET` clause.

`?searchTerm` and `?uiLanguage` are reserved string parameters bound at execution time by the renderer.  ?searchTerm contains the string entered by the end user in the UI. ?uiLanguage binds the language currently prefered by the user context.

If no `shui:LabelRole` is present, renderers SHOULD fall back to the [label resolution mechanism](https://www.w3.org/TR/shacl12-ui/#label-and-lang).

When the SHACL renderer detects `shui:searchQuery` it must use that query to filter the data down. Implementations MAY combine the query of filtering with resolving the propertyRoles. Below is a possible solution:

```SPARQL
prefix text: <http://text.com/>
prefix ex: <https://example.com/>
prefix schema: <https://schema.org/>

select ?value (sample(?prop_1) as ?prop1) (sample(?prop_2) as ?prop2) where {
  {
    select distinct ?value where {
      service <http://example.com/sparql> {
        ?value text:query (?searchTerm ?uiLanguage) .
        ?value a ex:Person .
      }

    }
    limit 10 # added by the renderer
  }
  service <http://example.com/sparql> {
    ?value schema:name ?prop_1
    FILTER(lang(?prop_1) = ?uiLanguage || lang(?prop_1) = "")
    
    optional {
      ?value ex:pseudonym ?prop_2
      FILTER(lang(?prop_2) = ?uiLanguage || lang(?prop_2) = "")
    }
  }
}
group by ?value
```

The rationale behind including `shui:searchQuery` as a subquery is that it is quite complicated to write an efficient  SPARQL query to gather the autocomplete item meta data. For this reason this is up to the implementions of the SHACL renderers instead of the shape writers.

---

_The rest of this issue are examples that do not add to the spec:_
### Example: custom implementation of a REST service

The following is an example how fetching from a REST service could be added in a custom implementation of a [Custom List Parameter Function](https://www.w3.org/TR/shacl12-node-expr/#custom-list-parameter-functions).

The goal of this example is to show that custom implementations can do this. I do not want to put anything about this example in the formal spec.

This example has been requested by @bergos. His perspective, if interpreted correctly was that the SPARQL functionality should have feature parity with Node expressions.

```turtle
ex:BookShape
    a sh:NodeShape ;
    sh:targetClass ex:Book ;
    sh:property [
        sh:name "Author" ;
        sh:path dct:creator ;
        sh:class ex:Person ;
        sh:node ex:PersonDisplayShape ;
        sh:in [
          ex:myApi (
			# Parameter 1:
			"search" 
			# Parameter 2:
			shui:searchTerm
			# Parameter 3: This ex:myApi support JQ
			"[].id"
          )
        ]
    ] .

ex:PersonDisplayShape
    a sh:NodeShape ;
    sh:property [
        sh:path schema:name ;
        shui:propertyRole shui:LabelRole ;
        sh:order 1 ;
        sh:values [
          ex:myApi (
			# Parameter 1:
			"resource" 
			# Parameter 2: The node to fetch via the custom REST implementation
			shnex:focusNode 
			# Parameter 3: This ex:myApi support JQ
			"[0].author.label"
          )
        ] ;
    ] ;
    sh:property [
        sh:path ex:pseudonym ;
        shui:propertyRole shui:SubLabelRole ;
        sh:order 2 ;
        sh:values [
          ex:myApi (
			# Parameter 1:
			"resource" 
			# Parameter 2: The node to fetch via the custom REST implementation
			shnex:focusNode 
			# Parameter 3: This ex:myApi support JQ
			"[0].author.pseudonym"
          )
        ] ;
    ] .
```
### Example: custom implementation of `ex:dereference`

The following is an example how de-referencing could be added in a custom implementation of a [Custom List Parameter Function](https://www.w3.org/TR/shacl12-node-expr/#custom-list-parameter-functions).

With dereferencing I mean that RDF data is fetched from the IRI itself.

The goal of this example is to show that custom implementations can do this. I do not want to put anything about this example in the formal spec.

This example has been requested by @smessie. He wanted to see a way to de-reference data in SHACL UI.

```turtle
ex:BookShape
    a sh:NodeShape ;
    sh:targetClass ex:Book ;
    sh:property [
        sh:name "Author" ;
        sh:path dct:creator ;
        sh:class ex:Person ;
        sh:node ex:PersonDisplayShape ;
    ] .

ex:PersonDisplayShape
    a sh:NodeShape ;
    sh:property [
        sh:path schema:name ;
        shui:propertyRole shui:LabelRole ;
        sh:order 1 ;
        sh:values [
          ex:dereference (
			# Parameter 1: The node to dereference (e.g., the focus node) 
			shnex:focusNode 
			# Parameter 2: The property path to fetch from the resolved graph
			( [ sh:alternativePath ( schema:name rdfs:label ) ] )
          )
        ] ;
    ] ;
    sh:property [
        sh:path ex:pseudonym ;
        shui:propertyRole shui:SubLabelRole ;
        sh:order 2 ;
        sh:values [
        ex:dereference (
			# Parameter 1: The node to dereference (e.g., the focus node) 
			shnex:focusNode 
			# Parameter 2: The property path to fetch from the resolved graph
			( ex:pseudonym )
          )
        ] ;
    ] .
```

Closes #846

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/pattern-using-property-roles-and-shui-searchQuery/shacl12-ui/index.html)


shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+          <code>SERVICE</code>. The accompanying <code>sh:node ex:PersonDisplayShape</code> constraint declares that the
+          <code>schema:name</code> of each candidate serves as its <a href="#LabelRole"><code>shui:LabelRole</code></a> and that
+          <code>ex:pseudonym</code> serves as its <a href="#SubLabelRole"><code>shui:SubLabelRole</code></a>.
+        </p>
+      </aside>
+
+      <p>
+        For the purposes of resolving item display metadata as described in this section, when multiple property shapes on the
+        <code>sh:node</code> shape are annotated with the same <code>shui:propertyRole</code>, the property shape with the
+        <em>highest</em> <code>sh:order</code> value wins. Remaining ties are broken using codepoint comparison. Implementations
```
smessie: "with the highest sh:order" contradicts what is currently specified under [Qualified Role Annotation](https://raw.githack.com/w3c/data-shapes/ui/pattern-using-property-roles-and-shui-searchQuery/shacl12-ui/index.html#qualified-role-annotation)
> For property shapes annotated with the same role, user interfaces should prefer the shape with the smallest sh:order value. If multiple shapes share the same role, they must be sorted and processed in ascending order of their sh:order values.

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+        sh:node ex:PersonDisplayShape ;
+        sh:in [
+            sh:select """
+                PREFIX ex: &lt;http://example.com/&gt;
+                PREFIX schema: &lt;https://schema.org/&gt;
+
+                SELECT ?value
+                WHERE {
+                     SERVICE &lt;http://example.com/sparql&gt; {
+                         ?value a ex:Person ;
```
```suggestion
                         ?value a ex:Person .
```
Invalid SPARQL query.

shacl12-ui/index.html
smessie: The tie-break rule says remaining ties are "broken using codepoint comparison," but it doesn't say codepoint comparison of what. The property path IRI, the shape node, something else? For blank-node shapes this would be non-deterministic. Relatedly, `sh:order` is optional; the rule doesn't define what happens when it's missing.

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+
+      <p>
+        Within the value of <code>shui:searchQuery</code>, <code>?searchTerm</code> and <code>?uiLanguage</code> are reserved
+        string parameters bound at execution time by the SHACL Renderer: <code>?searchTerm</code> is bound to the string entered
+        by the end user in the user interface, and <code>?uiLanguage</code> is bound to the language currently preferred by the
+        user context.
+      </p>
+
+      <p>
+        When a SHACL Renderer detects a <code>shui:searchQuery</code>, it MUST use that query, instead of the query given by
```
smessie: `sh:in` is a validation constraint whose valid value set derives from `sh:select`. This section says the renderer uses `shui:searchQuery` instead of `sh:select` to populate the list while the user is searching. Since nothing requires the fulltext query to return a subset of the `sh:select` results, a user could pick a candidate that then fails `sh:in` validation. Should we require `shui:searchQuery` to return a subset of the `sh:select` value set (or at least call out this risk)?

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+        The example below illustrates one possible way for a SHACL Renderer to combine the <code>shui:searchQuery</code> from the
+        previous example with the property roles declared on <code>ex:PersonDisplayShape</code>, using the filtering query as a
+        sub-query and adding, grouped and sampled by the candidate value, the property paths for each role in the outer query.
+        This combined query is generated by the SHACL Renderer; shape authors are not expected to author it themselves, as writing
+        an efficient combined query typically requires backend-specific knowledge.
+      </p>
+
+      <pre class="example">
+PREFIX text: &lt;http://jena.apache.org/text#&gt;
+PREFIX ex: &lt;https://example.com/&gt;
```
smessie: Your other examples use http instead of https for example.org. We might want to align there.

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+    sh:property [
+        sh:path schema:name ;
+        shui:propertyRole shui:LabelRole ;
+        sh:order 1 ;
+        sh:minCount 1 ;
+    ] ;
+    sh:property [
+        sh:path ex:pseudonym ;
+        shui:propertyRole shui:SubLabelRole ;
+        sh:order 2 ;
```
smessie: The tie-break rule is per-role. Should we set an sh:order of 1 here as well such that this example does not seem to imply otherwise?

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+      SERVICE &lt;http://example.com/sparql&gt; {
+        ?value text:query (?searchTerm ?uiLanguage) .
+        ?value a ex:Person .
+      }
+    }
+    LIMIT 10 # Added by the SHACL Renderer.
+  }
+  SERVICE &lt;http://example.com/sparql&gt; {
+    ?value schema:name ?prop_1 .
+    FILTER (lang(?prop_1) = ?uiLanguage || lang(?prop_1) = "")
```
smessie: In this illustrative combined query, `FILTER(lang(?prop_1) = ?uiLanguage …)` uses exact string equality, so a `?uiLanguage` of `"en"` won't match an `"en-US"` literal. It's only an example, but readers tend to copy these. Should we consider `langMatches(...)` or a note?

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,229 @@ <h4>shui:propertyRole Property</h4>
...
+    sh:property [
+        sh:path schema:name ;
+        shui:propertyRole shui:LabelRole ;
+        sh:order 1 ;
+    ] ;
+    sh:property [
+        sh:path ex:pseudonym ;
+        shui:propertyRole shui:SubLabelRole ;
+        sh:order 2 ;
+    ] .
```
smessie, line 3088: The examples in this section use the direct role-annotation form with `sh:order` placed directly on the property shape.
But per the existing spec, the direct form leaves duplicate-role behavior undefined, and `sh:order`-based precedence is defined for the qualified form, where `sh:order` lives inside the annotation node `( shui:propertyRole [ shui:propertyRole …Role ; sh:order 0 ] )`. So the new tie-break rule references an `sh:order` mechanism that the examples don't use the way the spec defines it.
Could we align the examples with one of the two established forms, so we're not implicitly introducing a third convention?

shacl12-ui/index.html
```diff
@@ -2952,13 +2965,30 @@ <h3>Built-in Property Roles</h3>
...
 
-      <h4>shui:LabelRole</h4>
+      <h4 id="LabelRole">shui:LabelRole</h4>
 
       <p>
         The <code>shui:LabelRole</code> is used to identify properties whose values serve as human-readable display labels.
         Common examples of display label predicates include <code>rdfs:label</code>, <code>skos:prefLabel</code>, and <code>schema:name</code>.
       </p>
 
+      <h4 id="SubLabelRole">shui:SubLabelRole</h4>
```
smessie: I find the term SubLabelRole a bit awkward. Do you have a strong opinion on that name or could we explore alternatives?
Maybe `shui:DescriptionRole`, `shui:NoteRole`, or `shui:CommentRole`.

smessie (commented):
I still need to further think about this PR, but I already included some comments below.

bergos (changes requested):
Requesting changes to hold this PR until the "How open/closed should the SHACL UI specification be?" issue is resolved, since its outcome may affect this PR. I'll approve once I've reviewed it and confirmed there's no conflict.

shacl12-ui/index.html
```diff
@@ -2148,11 +2148,15 @@ <h3>Conditional Shape Selection</h3>
       </p>
 
       <h3>Enumerations, Select Lists, and Autocomplete</h3>
+      <p>
+        Using <code>sh:in</code> together with a <a href="../shacl12-sparql/#SelectExpression">select expression</a>, possibly with
+        a SPARQL <code>SERVICE</code> clause, and a <code>sh:node</code> shape annotated with <a href="#property-roles">property
+        roles</a> is one way of sourcing and displaying the items of a dynamic selection list; see
+        <a href="#selection-list-items">Resolving Selection List Items</a>.
```
TallTed: It's not clear to me what the `possibly with` applies to. Is it just `a SPARQL <code>SERVICE</code> clause` or does it also include `a <code>sh:node</code> shape annotated with <a href="#property-roles">property roles</a>`?

When this is made clear to me, I will offer a rephrasing for this paragraph.

shacl12-ui/index.html
```diff
@@ -2217,6 +2221,11 @@ <h4>shui:AutoCompleteEditor</h4>
             For example, if the `sh:class` of the property is `ex:Country` and the user starts
             typing "Nig", then "Niger" and "Nigeria" would show up as possible choices.
           </p>
+          <p>
+            See <a href="#selection-list-items">Resolving Selection List Items</a> for how a shape can control the display of
+            the selected value and of the candidate options, and how a shape author can supply a dedicated fulltext search query
```
TallTed:
```suggestion
            See <a href="#selection-list-items">Resolving Selection List Items</a> for ways a shape can control the display of
            the selected value and of the candidate options, as well as ways a shape author can supply a dedicated full text search query
```

shacl12-ui/index.html
```diff
@@ -2421,6 +2430,10 @@ <h4>shui:EnumSelectEditor</h4>
             <b>Rendering:</b>
             A drop-down editor for enum fields (based on the `sh:in` list, in that order).
           </p>
+          <p>
+            See <a href="#selection-list-items">Resolving Selection List Items</a> for how a shape can control the display of
+            the selected value and of the candidate options when the `sh:in` list is sourced dynamically.
```
TallTed:
```suggestion
            See <a href="#selection-list-items">Resolving Selection List Items</a> for ways a shape can control the display of
            the selected value, and of the candidate options when the `sh:in` list is sourced dynamically.
```

shacl12-ui/index.html
```diff
@@ -2857,7 +2870,7 @@ <h2>Property Roles</h2>
       <h3>Direct Role Annotation</h3>
 
       <p>
-        Property roles can be annotated on property shapes by using the <code></code>shui:propertyRole</code> predicate and linking directly to a property role instance.
+        Property roles can be annotated on property shapes by using the <code>shui:propertyRole</code> predicate and linking directly to a property role instance.
         SHACL renderers may use such direct annotations to drive the way specific user-interface elements are displayed.
```
TallTed:
```suggestion
        SHACL renderers can use such direct annotations to drive the way specific user-interface elements are displayed.
```

shacl12-ui/index.html
```diff
@@ -2952,13 +2965,30 @@ <h3>Built-in Property Roles</h3>
...
         The <code>shui:LabelRole</code> is used to identify properties whose values serve as human-readable display labels.
         Common examples of display label predicates include <code>rdfs:label</code>, <code>skos:prefLabel</code>, and <code>schema:name</code>.
       </p>
 
+      <h4 id="DescriptionRole">shui:DescriptionRole</h4>
+
+      <p>
+        The <code>shui:DescriptionRole</code> is used to identify properties whose values serve as secondary, less prominent display
+        text shown alongside a <a href="#LabelRole"><code>shui:LabelRole</code></a> value, for example to help a user disambiguate
+        between resources that otherwise share the same label.
```
TallTed:
```suggestion
        The <code>shui:DescriptionRole</code> identifies properties that serve as secondary, less prominent display
        text shown alongside a <a href="#LabelRole"><code>shui:LabelRole</code></a> value. For example, it helps users disambiguate
        between resources that share the same label.
```

shacl12-ui/index.html
```diff
@@ -2952,13 +2965,30 @@ <h3>Built-in Property Roles</h3>
...
+      </p>
+
+      <h4 id="DepictionRole">shui:DepictionRole</h4>
+
+      <p>
+        The <code>shui:DepictionRole</code> is used to identify properties whose values serve as a depiction, such as a portrait,
+        logo, or icon, of a resource. The value node of a property shape annotated with <code>shui:DepictionRole</code> is expected
+        to be an <a>IRI</a> that can be dereferenced to obtain the depiction, for example the value of a <code>foaf:depiction</code>
+        or <code>schema:image</code> property.
```
TallTed:
```suggestion
        The <code>shui:DepictionRole</code> identifies properties that serve as a depiction of a resource, such as a portrait,
        logo, or icon. The value node of a property shape annotated with <code>shui:DepictionRole</code> is expected
        to be an <a>IRI</a> that can be dereferenced to obtain the depiction. This is true, for example, for the value 
        of a <code>foaf:depiction</code> or <code>schema:image</code> property.
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+    <section id="selection-list-items">
+      <h2>Resolving Selection List Items</h2>
+
+      <p>
+        Widgets such as the <a href="#AutoCompleteEditor"><code>shui:AutoCompleteEditor</code></a> and the
+        <a href="#EnumSelectEditor"><code>shui:EnumSelectEditor</code></a>, and possibly others, present the user with a list of
+        candidate values and need to render both the currently selected value and each candidate as more than a bare <a>IRI</a>.
+        This section defines how a <a>shapes graph</a> can combine <a href="../shacl12-node-expr/#dfn-dynamic-shacl">Dynamic
+        SHACL</a> with <a href="#property-roles">Property Roles</a> to declare, alongside the source of a value list, the
+        additional display metadata — such as a label, a description, or a depiction — that a SHACL Renderer may use for each item.
```
TallTed:
```suggestion
        Widgets such as the <a href="#AutoCompleteEditor"><code>shui:AutoCompleteEditor</code></a> and the
        <a href="#EnumSelectEditor"><code>shui:EnumSelectEditor</code></a>, and possibly others, present the user with a list of
        candidate values. These widgets need to render both the currently selected value and each candidate as more than a bare <a>IRI</a>.
        This section defines how a <a>shapes graph</a> can combine <a href="../shacl12-node-expr/#dfn-dynamic-shacl">Dynamic
        SHACL</a> with <a href="#property-roles">Property Roles</a> to declare, alongside the source of a value list, the
        additional display metadata — such as a label, a description, or a depiction — that a SHACL Renderer can use for each item.
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+      </p>
+
+      <p>
+        Because the roles described in this section are declared as static annotations on a <a>node shape</a>, non-SPARQL and
+        hybrid backends can consume them directly to drive the same display, without needing to evaluate any SPARQL expression.
```
TallTed:
```suggestion
        The roles described in this section are declared as static annotations on a <a>node shape</a> so that non-SPARQL and
        hybrid backends can consume them directly to drive the same display, without needing to evaluate any SPARQL expression.
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+
+      <p>
+        A <a>property shape</a> that constrains its values using <code>sh:in</code> together with a
+        <a href="../shacl12-sparql/#SelectExpression">select expression</a> MAY also declare a <code>sh:node</code> constraint
+        pointing at a <a>node shape</a> whose property shapes are annotated with <a href="#property-roles">property roles</a>. When
+        a SHACL Renderer detects this combination of <code>sh:node</code> and <code>sh:in</code> with a nested
+        <code>sh:select</code> on a property shape, it MAY use the property roles declared on the <code>sh:node</code> shape to
+        resolve a label, a description, a depiction, or other display metadata for each candidate value, including the value
+        currently selected. The query given as the value of <code>sh:select</code> MAY use the SPARQL <code>SERVICE</code> keyword
+        to make a federated call.
```
TallTed:
```suggestion
        currently selected.
      </p>
      <p>
        The query given as the value of <code>sh:select</code> MAY use the SPARQL <code>SERVICE</code> keyword
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+        ] ;
+    ] .
+          </div>
+        </div>
+        <p>
+          The <code>sh:in</code> constraint on <code>dct:creator</code> uses a <a href="../shacl12-sparql/#SelectExpression">select
+          expression</a> to fetch candidate <code>ex:Person</code> resources from a federated SPARQL endpoint using
+          <code>SERVICE</code>. The accompanying <code>sh:node ex:PersonDisplayShape</code> constraint declares that the
+          <code>schema:name</code> of each candidate serves as its <a href="#LabelRole"><code>shui:LabelRole</code></a>, with
+          <code>schema:alternateName</code> declared as a lower-priority fallback for the same role, and that
```
TallTed:
```suggestion
          <code>schema:name</code> of each candidate serves as its <a href="#LabelRole"><code>shui:LabelRole</code></a> — with
          <code>schema:alternateName</code> declared as a lower-priority fallback for the same role — and that
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+        (<code>0</code>, versus <code>1</code> for <code>schema:alternateName</code>), it is the one used to resolve the
+        label. More generally, when multiple property shapes on the <code>sh:node</code> shape are annotated with the same
+        <code>shui:propertyRole</code>, the same rule as <a href="#property-roles">qualified role annotation</a> applies: the
+        property shape with the smallest <code>sh:order</code> value wins. A property shape that declares
+        <code>sh:order</code> is always preferred over a tied property shape that does not. If a tie remains — because
+        the tied property shapes share the same <code>sh:order</code> value, or because none of them declares
+        <code>sh:order</code> — it is broken by comparing the Unicode codepoints of the string form of each property
+        shape's <code>sh:path</code> (its IRI, for the common case of a simple path); the property shape whose path
+        sorts first wins. Implementations MAY define
+        additional, custom property roles, beyond <a href="#LabelRole"><code>shui:LabelRole</code></a>,
```
TallTed:
```suggestion
        sorts first wins.
      </p>
      <p>
        Implementations MAY define additional, custom property roles, beyond <a href="#LabelRole"><code>shui:LabelRole</code></a>,
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+      </p>
+
+      <h3 id="searchQuery">shui:searchQuery</h3>
+
+      <p>
+        Some SPARQL backends offer fulltext search capabilities over their data that cannot be expressed efficiently, or at all,
+        using the <a href="../shacl12-sparql/#SelectExpression">select expression</a> given as the value of <code>sh:in</code>. The
+        property <code>shui:searchQuery</code> lets a shape author give full control over the query used to fetch autocomplete
+        items whenever the shape author prefers to use such a fulltext search facility.
+      </p>
```
TallTed:
```suggestion
      <p>
        Some SPARQL backends offer full-text search capabilities over their data that cannot be expressed efficiently, or at all,
        using the <a href="../shacl12-sparql/#SelectExpression">select expression</a> given as the value of <code>sh:in</code>. The
        property <code>shui:searchQuery</code> lets a shape author give full control over the query used to fetch autocomplete
        items whenever the shape author prefers to use such a full-text search facility.
      </p>
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+
+      <p>
+        <code>shui:searchQuery</code> is specified as an additional value on the same node expression that carries the
+        <code>sh:select</code> property, i.e. the value of <code>sh:in</code>. Its value MUST be a valid SPARQL 1.2 SELECT query,
```
TallTed:
```suggestion
        <code>sh:select</code> property, i.e., the value of <code>sh:in</code>. Its value MUST be a valid SPARQL 1.2 SELECT query,
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+        following the same prefix handling and single-variable projection rules as <a href="../shacl12-sparql/#SelectExpression"
+        >select expressions</a>. The value of <code>shui:searchQuery</code> MUST NOT contain a <code>LIMIT</code> or an
+        <code>OFFSET</code> solution modifier.
+      </p>
+
+      <p>
+        Within the value of <code>shui:searchQuery</code>, <code>?searchTerm</code> and <code>?uiLanguage</code> are reserved
+        string parameters bound at execution time by the SHACL Renderer: <code>?searchTerm</code> is bound to the string entered
+        by the end user in the user interface, and <code>?uiLanguage</code> is bound to the language currently preferred by the
```
TallTed:
```suggestion
        by the end user in the user interface; and <code>?uiLanguage</code> is bound to the language currently preferred by the
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+        user context.
+      </p>
+
+      <p>
+        When a SHACL Renderer detects a <code>shui:searchQuery</code>, it MUST use that query, instead of the query given by
+        <code>sh:select</code>, to filter the candidate values while the end user is searching. Implementations MAY combine the
+        query used for filtering with the queries needed to resolve the property roles declared on the <code>sh:node</code> shape.
+      </p>
```
TallTed:
```suggestion
      <p>
        When a SHACL Renderer detects a <code>shui:searchQuery</code>, it MUST use that query to filter
        candidate values during an end-user search, overriding the query given by <code>sh:select</code>.
        Implementations MAY combine this filtering query with the queries needed to resolve the property
        roles declared on the <code>sh:node</code> shape.
      </p>
```

shacl12-ui/index.html
```diff
@@ -2976,6 +3006,263 @@ <h4>shui:propertyRole Property</h4>
...
+
+      <p>
+        Because <code>shui:searchQuery</code> is evaluated independently of <code>sh:select</code>, nothing guarantees that a
+        candidate matched by <code>shui:searchQuery</code> is also present in the value set produced by <code>sh:select</code>.
+        Shape authors SHOULD design <code>shui:searchQuery</code> so that it only returns values within the same scope as
+        <code>sh:select</code> (e.g., by reusing the same type and graph constraints), to avoid offering the end user candidates
+        that would fail the <code>sh:in</code> constraint. Regardless of how a value was located, it remains subject to the
+        <code>sh:in</code> constraint, and a SHACL Renderer or validating engine MAY reject a value returned by
```
TallTed:
```suggestion
        <code>sh:select</code> (e.g., by reusing the same type and graph constraints). This avoids offering the end user candidates
        that would fail the <code>sh:in</code> constraint. Regardless of how a value was located, it remains subject to the
        <code>sh:in</code> constraint; a SHACL Renderer or validating engine MAY reject a value returned by
```

shacl12-ui/index.html
danielbeeke: I removed it thanks

bergos:
I've reviewed the PR, but I need more time to reflect on it.

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+        <a href="#LabelRole"><code>shui:LabelRole</code></a>, SHACL Renderers SHOULD fall back to the
+        <a href="#label-and-lang">label resolution</a> mechanism to determine a label for each value.
+      </p>
+
+      <h3 id="searchQuery">shui:searchQuery</h3>
+
+      <p>
+        Some SPARQL backends offer full-text search capabilities over their data that cannot be expressed efficiently, or at all,
+        using the <a href="../shacl12-sparql/#SelectExpression">select expression</a> given as the value of <code>sh:in</code>. The
+        property <code>shui:searchQuery</code> lets a shape author give full control over the query used to fetch autocomplete
```
edmondchuc:
```suggestion
        property <code>shui:searchQuery</code> gives the shape author full control over the query used to fetch autocomplete
```

shacl12-ui/index.html
```diff
@@ -2857,8 +2869,8 @@ <h2>Property Roles</h2>
       <h3>Direct Role Annotation</h3>
 
       <p>
-        Property roles can be annotated on property shapes by using the <code></code>shui:propertyRole</code> predicate and linking directly to a property role instance.
-        SHACL renderers may use such direct annotations to drive the way specific user-interface elements are displayed.
+        Property roles can be annotated on property shapes by using the <code>shui:propertyRole</code> predicate and linking directly to a property role instance.
+        SHACL renderers can use such direct annotations to drive the way specific user-interface elements are displayed.
```
edmondchuc: Why was **may** changed to **can** here? This makes it less clear whether the feature is optional or not.

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+      <p>
+        Within the value of <code>shui:searchQuery</code>, <code>?searchTerm</code> and <code>?uiLanguage</code> are reserved
+        string parameters bound at execution time by the SHACL Renderer: <code>?searchTerm</code> is bound to the string entered
+        by the end user in the user interface; and <code>?uiLanguage</code> is bound to the language currently preferred by the
+        user context.
+      </p>
+
+      <p>
+        When a SHACL Renderer detects a <code>shui:searchQuery</code>, it MUST use that query to filter
+        candidate values during an end-user search, overriding the query given by <code>sh:select</code>.
```
edmondchuc: It's not clear to me whether a SHACL Renderer must implement `shui:searchQuery` or not. It sounds like for the shape author, `shui:searchQuery` is optional, but implementations must support the filter behaviour if it detects its presence. Is this correct? If that's the case, does this mean all SHACL Renderers must support `shui:searchQuery`?

edmondchuc (commented):
I like the general direction of this PR. I just have a few clarifying questions and an editorial comment.

shacl12-ui/index.html
danielbeeke: This was @TallTed's proposal. Ill go back to MAY.

shacl12-ui/index.html
danielbeeke: Good point you are touching on. Ill take some time today to process this further.

danielbeeke: I have changed it to a MAY.

edmondchuc (approved):

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+        ] ;
+    ] ;
+    sh:property [
+        sh:path ex:pseudonym ;
+        shui:propertyRole [
+            shui:propertyRole shui:DescriptionRole ;
+            sh:order 0
+        ] ;
+    ] .
+          </div>
```
bergos, line 3552: Why did you define the other properties in the example? Is the intention to modify the SPARQL query so it can be reused across properties, or to find the matching query based on the `schema:name` pattern? Either way, I think it's too error-prone. One query should be defined per property.

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+        string parameters bound at execution time by the SHACL Renderer: <code>?searchTerm</code> is bound to the string entered
+        by the end user in the user interface; and <code>?uiLanguage</code> is bound to the language currently preferred by the
+        user context.
+      </p>
+
+      <p>
+        When a SHACL Renderer detects a <code>shui:searchQuery</code>, it MAY use that query to filter
+        candidate values during an end-user search, overriding the query given by <code>sh:select</code>.
+        Implementations MAY combine this filtering query with the queries needed to resolve the property
+        roles declared on the <code>sh:node</code> shape.
```
bergos, line 3617: Generating the query on the fly is error-prone: SPARQL syntax is order-sensitive, and it's not obvious where the auto-generated pattern should be inserted. Requiring the query to be fully qualified up front would avoid this problem.

bergos (changes requested):

shacl12-ui/index.html
danielbeeke: The properties with shui:propertyRole declare what properties MAY be used to draw the autocomplete item.
It is up to the implementer to resolve the properties.

shacl12-ui/index.html
danielbeeke: I disagree, first the properties that have a minCount should be placed. Further more it is not specified who to resolve it. Implementations can just get these properties from the data graph. If they wish to support service clauses they can implement a SPARQL query to get the properties.

shacl12-ui/index.html
```diff
@@ -2857,8 +2869,8 @@ <h2>Property Roles</h2>
       <h3>Direct Role Annotation</h3>
 
       <p>
-        Property roles can be annotated on property shapes by using the <code></code>shui:propertyRole</code> predicate and linking directly to a property role instance.
-        SHACL renderers may use such direct annotations to drive the way specific user-interface elements are displayed.
+        Property roles can be annotated on property shapes by using the <code>shui:propertyRole</code> predicate and linking directly to a property role instance.
+        SHACL renderers MAY use such direct annotations to drive the way specific user-interface elements are displayed.
```
robert-david, line 3334: Why 'direct'? Are there also indirect annotations?

shacl12-ui/index.html
```diff
@@ -2952,13 +2964,30 @@ <h3>Built-in Property Roles</h3>
         by SHACL Renderers. Additional property roles may be defined in other namespaces.
       </p>
 
-      <h4>shui:LabelRole</h4>
+      <h4 id="LabelRole">shui:LabelRole</h4>
```
robert-david, line 3430: How does this tie in with the global configuration, i.e. the shui:labelPreference property?

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+        A <a>property shape</a> that constrains its values using <code>sh:in</code> together with a
+        <a href="../shacl12-sparql/#SelectExpression">select expression</a> MAY also declare a <code>sh:node</code> constraint
+        pointing at a <a>node shape</a> whose property shapes are annotated with <a href="#property-roles">property roles</a>. When
+        a SHACL Renderer detects this combination of <code>sh:node</code> and <code>sh:in</code> with a nested
+        <code>sh:select</code> on a property shape, it MAY use the property roles declared on the <code>sh:node</code> shape to
+        resolve a label, a description, a depiction, or other display metadata for each candidate value, including the value
+        currently selected.
+      </p>
+      <p>
+        The query given as the value of <code>sh:select</code> MAY use the SPARQL <code>SERVICE</code> keyword
```
robert-david, line 3500: sh:select points to a SPARQL SELECT query. SERVICE is part of the spec. It is confusing that it is 'allowed' explicitly. I suggest to remove this sentence.

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+        hybrid backends can consume them directly to drive the same display, without needing to evaluate any SPARQL expression.
+      </p>
+
+      <h3>Using sh:node and shui:propertyRole to Resolve Item Display</h3>
+
+      <p>
+        A <a>property shape</a> that constrains its values using <code>sh:in</code> together with a
+        <a href="../shacl12-sparql/#SelectExpression">select expression</a> MAY also declare a <code>sh:node</code> constraint
+        pointing at a <a>node shape</a> whose property shapes are annotated with <a href="#property-roles">property roles</a>. When
+        a SHACL Renderer detects this combination of <code>sh:node</code> and <code>sh:in</code> with a nested
```
robert-david, line 3494: I don't understand the combination of sh:in/sh:select and sh:node. Also, in the example both seem to be independent.

robert-david: I suggest to avoid such implicit semantics added on top of what the core spec defines. I would prefer that the situational relation/dependency between sh:in and sh:node is explicitly represented.

shacl12-ui/index.html
```diff
@@ -2976,6 +3005,268 @@ <h4>shui:propertyRole Property</h4>
...
+      <p>
+        If none of the property shapes on the <code>sh:node</code> shape declare a
+        <a href="#LabelRole"><code>shui:LabelRole</code></a>, SHACL Renderers SHOULD fall back to the
+        <a href="#label-and-lang">label resolution</a> mechanism to determine a label for each value.
+      </p>
+
+      <h3 id="searchQuery">shui:searchQuery</h3>
+
+      <p>
+        Some SPARQL backends offer full-text search capabilities over their data that cannot be expressed efficiently, or at all,
```
robert-david, line 3592: Why can't i express this as a SPARQL query? FTS support usually is configured on some labels and executed with magic predicates, which are part of the query. The user has the freedom to return whatever values are needed. I miss to see the reason for having a dedicated shui:searchQuery.

robert-david (changes requested):

shacl12-ui/index.html
danielbeeke: Yes, this is outside of the scope of my changes. I only made it uppercase. See existing text: https://w3c.github.io/data-shapes/shacl12-ui/#property-roles.

I am only using the existing structure here to declare that a certain property has a certain role.

shacl12-ui/index.html
danielbeeke: This is well defined in the label resolution:
https://w3c.github.io/data-shapes/shacl12-ui/#label-resolution-value-node-labels

shacl12-ui/index.html
danielbeeke: The `sh:node` here points to a shape that constraints the result values further than the `sh:in`. I think only the intersection of the `sh:in` / `sh:select` and the `sh:node` are valid. This is SHACL core / Dynamic SHACL if I understand correct.

Further more the `sh:node` has properties that have `shui:propertyRole` annotations. Giving the UI a way to understand which property path or predicate is preferred for the label and also other roles, such as the depiction.

There is nothing new here that I add to the existing spec, from my view point.

shacl12-ui/index.html
danielbeeke: Jena uses the following syntaxes:

```
?s text:query 'word'                              # query
?s text:query ('word' 10)                         # with limit on results
?s text:query (rdfs:label 'word')                 # query specific property if multiple
?s text:query (rdfs:label 'protégé' 'lang:fr')    # restrict search to French
(?s ?score) text:query 'word'                     # query capturing also the score
(?s ?score ?literal) text:query 'word'            # ... and original literal value
(?s ?score ?literal ?g) text:query 'word'         # ... and the graph

```

Some of these can not be expressed other than inside a SPARQL query. My the sentence is not fully clear. The use case is exactly about expressing it in a SPARQL query.

shacl12-ui/index.html
danielbeeke: In the link you will find there are direct role annotations and qualified role annotations.

shacl12-ui/index.html
bergos: I would propose skipping this section if it duplicates definitions already covered in this spec or in node expressions. If there is something new in this section, the wording should be revised to make this clearer, and optional items should be omitted.

shacl12-ui/index.html
bergos: Defining logic that injects triple patterns for the property labels into a SPARQL query is hard. For example, variables might intentionally be bound before the triple pattern, and there are probably more such cases that would need to be covered. I would propose instead simply defining a `shui:searchQuery` property — perhaps under a different name like `shui:searchQueryBuilder` — that has no fixed structure, leaving it up to the implementation to decide how to use it.

danielbeeke: what do you mean with variables might be intentionally be bound?

danielbeeke:
The alternative pull request has been merged I think we can close this for now.

PR #1114, opened 2026-08-04
#995: Added support for sh:tempTriple

### LINKED ISSUE #995: SHACL-SPARQL Rules should support temporary triples
See https://github.com/w3c/data-shapes/pull/984#issuecomment-4933413656 it should be possible to infer "temporary" triples that are only visible while the rules execute but don't end up in the final result graph.

Some options:

1. Have a dedicated named graph that is part of the data graph during inference but will be discarded later
2. Use RDF 1.2 reifiers to mark temp triples (my preference right now)
3. Have rules that can DELETE instead of CONSTRUCT/INSERT

HolgerKnublauch:
Closes #995

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-995-tempTriple/shacl12-sparql/index.html#tempTriple)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2196,7 +2196,7 @@ ex:Granddaughter
 				<p>
 					It is sometimes useful for inference rules to produce triples that are only visible during the execution
 					of other rules but do not end up in the final inferences.
-					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inferences graph contains a
+					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inference graph contains a
 					<a>reifier</a> with the <a>value</a> <code>sh:tempTriple true</code>.
 					Rules can produce such triples as illustrated in the following example:
 				</p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2191,67 +2191,6 @@ ex:Granddaughter
 					<a href="#rules-execution">rules execution instructions</a> prior to performing the actual validation. 
 				</p>
 			</section>
-			<section id="rules-execution">
-				<h3>General Execution Instructions for SHACL Rules</h3>
-				<p>
-					A <dfn data-lt="rules engine">SHACL rules engine</dfn> is a computer procedure that takes as input
-					a <a>data graph</a> and a <a>shapes graph</a> and is capable of adding <a>triples</a> to the <a>data graph</a>.
-					The new <a>triples</a> that are produced by a rules engine are called the <dfn data-lt="inferring|infer">inferred</dfn> triples.
-				</p>
-				<p>
-					Note that, from a logical perspective, the <a>data graph</a> will be <em>modified</em> if <a>triples</a> get inferred.
-					This means that rules can trigger after other triples have been inferred.
-					However, in cases where the original data should not be modified, implementations may construct a logical <a>data graph</a>
-					that has the original data as one subgraph and a dedicated inferences graph as another subgraph, and where
-					the inferred triples get added to the inferences graph only.
-				</p>
-				<p>
-					An <dfn>execution</dfn> of a <a>rule</a> is the process that produces <a>inferred</a> triples from the <a>rule</a>
-					based on the <a>execution instructions</a> of its <a>rule types</a>.
-					If the <a>rule</a> is a <a>shape rule</a> (i.e., it is linked to at least one <a>shape</a> via <code>sh:rule</code>),
-					then the rule is executed for each <a>target node</a> of each of the linked and <a data-cite="shacl12-core#deactivated">non-deactivated</a>
-					<a>shapes</a> that <a>conform</a> to all <a data-cite="shacl12-core#deactivated">non-deactivated</a> <a href="#condition">conditions</a> of the <a>rule</a>.
-					These <a>target nodes</a> become the <a>focus nodes</a> of the executing <a>shape rule</a>.
-				</p>
-				<p>
-					For a given set of <a>rules</a> in the <a>shapes graph</a>
-					an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
-					specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-					Within an iteration, the inferred triples of one rule become immediately visible to the next rule.
-				</p>
-				<p>
-					The <dfn>execution of a rule set</dfn> is defined as follows:
-				</p>
-				<div class="algorithm">
-					<pre class="text">
-						For all <a>layers</a> in the rule set (in ascending order):
-							Execute one <a>iteration</a> over all <a>run-once rules</a> in the layer
-							do
-								Execute one <a>iteration</a> over all <a>iterating rules</a> in the layer
-							while the iteration has produced newly inferred triples
-
-						Delete the <a>temporary triples</a> and their <a>reifiers</a>
-					</pre>
-				</div>
-				<p>
-					If any of the <a>rules</a> reports a <a>failure</a> during <a>execution</a>,
-					then the <a>execution of a rule set</a> also produces a <a>failure</a>.
-				</p>
-				<p>
-					If a <a>rules engine</a> is not able to execute a given <a>rule</a>
-					because it does not support any of the <a>rule types</a> of the <a>rule</a>,
-					then it reports a <a>failure</a>.
-				</p>
-				<p>
-					Rule engines MAY also report a failure after a pre-configured maximum iteration count has been exceeded or
-					a pre-configured maximum number of inferred triples has been produced.
-					This can help as a guard to avoid out-of-memory problems and infinite loops.
-				</p>
-				<p>
-					At no time are inferred triples visible to the <a>shapes graph</a>, i.e. it is impossible for rules
-					to modify the definitions of rules or shapes.
-				</p>
-			</section>
 			<section id="tempTriple">
 				<h3>Temporary Triples</h3>
 				<p>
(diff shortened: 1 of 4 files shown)


shacl12-sparql/index.html
```diff
@@ -2229,6 +2229,8 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 							do
 								Execute one <a>iteration</a> over all <a>iterating rules</a> in the layer
 							while the iteration has produced newly inferred triples
+
+						Delete the <a>temporary triples</a> and their <a>reifiers</a>
```
liviorobaldo, line 2303: This is the first occurrence of the term "temporary triples" in the document, yet they are introduced and explained only in the following subsection. I would therefore suggest swapping Sections 7.5 and 7.6.

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,76 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+				LIMIT 1
+			}
+		}
+	""" .
+						</div>
+					</div>
+					<p>
+						The first rule computes <code>ex:offspring</code> triples as the (transitive)
+						children of each Person.
+						The second rule uses the <code>ex:offspring</code> triples to infer the <code>ex:oldestOffspring</code>
```
liviorobaldo:
```suggestion
						The second rule uses the <code>ex:offspring</code> triples to infer the <code>ex:youngestOffspring</code>
```

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,76 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+						of each Person.
+						The final inferences only include the <code>ex:oldestOffspring</code> triples.
```
liviorobaldo:
```suggestion
						The final inferences only include the <code>ex:youngestOffspring</code> triples.
```

liviorobaldo: Here I would also add that temporary triples are automatically deleted by the engine (see *General Execution Instructions for SHACL Rules*). For example:

"The final inferences include _only_ the `ex:youngestOffspring` triples: all temporary triples, i.e., triples whose reifier is marked with `sh:tempTriple true`, are _automatically_ deleted by the SHACL engine at the end of the inference process."

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,76 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+					</p>
+					<p>
+						Note the example is for illustration purposes only.
+						The same result could be achieved with a single rule that does everything in one go.
```
liviorobaldo:
```suggestion
```

I would remove this sentence, as it sounds ambiguous. It seems to suggest that a single rule can both infer the youngest offspring and delete all temporary triples required for that inference.

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,76 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+						Creating temporary triples can be helpful if it avoids repetitive computations that are
+						needed by multiple rules down the road.
```
liviorobaldo:
```suggestion
Creating temporary triples can be helpful when a computation is too complex to be carried out by a single rule. The complex computation can then be decomposed into _multiple_ rules: some rules infer _partial_ results, from which other rules derive the _final_ results. The partial results are only functional to the computation of the final results and are no longer needed once the latter have been obtained. By marking these partial results as temporary triples, they can be automatically removed after the final results have been computed.
```

I think that the claim "it avoids repetitive computations that are needed by multiple rules down the road" is correct, but it can be generalised as above: temporary triples represent _partial_ results that become unnecessary once the _final_ results have been computed. Multiple final results may indeed be derived from the same partial results, but this is only a particular case of the more general principle described above.

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,76 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+		CONSTRUCT {
+			$this ex:offspring ?offspring .
+			?reifier sh:tempTriple true .
+			?reifier rdf:reifies ?tt .
+		}
+		WHERE {
+			$this ex:child+ ?offspring .
+			BIND (BNODE() AS ?reifier) .
+			BIND (TRIPLE($this, ex:offspring, ?offspring) AS ?tt) .
+		}
```
liviorobaldo, line 2222: A question: could we simply write the following here?

```
CONSTRUCT{
    $this ex:offspring ?offspring .
    <<$this ex:offspring ?offspring>> sh:tempTriple true .
}
WHERE{
     $this ex:child+ ?offspring .
}
```

This is what I would do, mainly to reduce verbosity (and, in my view, it is also more intuitive). My understanding is that RDF 1.2 provides support for RDF-star syntax.

HolgerKnublauch: Yes I assume so. The (old) version of Jena 5 that I tried this with seems to have an unfinished coverage of this syntax though so I wasn't getting the expected results and for now went with the verbose syntax.

I assume this should also work and is perhaps best:

```
$this ex:offspring ?offspring {| sh:tempTriple true |} .
```

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,75 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
+						The first rule computes <code>ex:offspring</code> triples as the (transitive)
+						children of each Person.
+						The second rule uses the <code>ex:offspring</code> triples to infer the <code>ex:youngestOffspring</code>
+						of each Person.
+						The final inferences only include the <code>ex:youngestOffspring</code> triples.
+					</p>
+					<p>
+						Note the example is for illustration purposes only.
+						The same result could be achieved with a single rule that does everything in one go.
+Creating temporary triples can be helpful when a computation is too complex to be carried out by a single rule. The complex computation can then be decomposed into _multiple_ rules: some rules infer _partial_ results, from which other rules derive the _final_ results. The partial results are only functional to the computation of the final results and are no longer needed once the latter have been obtained. By marking these partial results as temporary triples, they can be automatically removed after the final results have been computed.
```
mgberg: The Markdown-style emphasis (_multiple_, _partial_, _final_) does not seem to render as italics.

shacl12-sparql/index.html
```diff
@@ -2250,6 +2252,75 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 					to modify the definitions of rules or shapes.
 				</p>
 			</section>
+			<section id="tempTriple">
+				<h3>Temporary Triples</h3>
+				<p>
+					It is sometimes useful for inference rules to produce triples that are only visible during the execution
+					of other rules but do not end up in the final inferences.
```
mgberg, line 2198: It says below that `sh:sourceRule` triples MUST NOT be visible to executing rules. Are `sh:tempTriple` triples visible to rules?

Also, to clarify, since `sh:sourceRule` triples are not visible to rules, I assume there would be no reason to add `sh:sourceRule` triples to temporary triples.

shacl12-sparql/index.html
mgberg: Would it be worth adding the annotation syntax as a comment in the query so users can see both options?

shacl12-sparql/index.html
liviorobaldo: Indeed. It shows "\_multiple\_", "\_partial\_", and "\_final\_" in the document rendered online.

shacl12-sparql/index.html
HolgerKnublauch: Yes they are visible, and I think that is OK and even a feature.

And yes, the engine will not add the sourceRule triples to them because they are deleted at the end of inferencing, before the sourceRule additions would take place. In my implementation I collect them in a HashMap until everything has finished.

HolgerKnublauch:
Thanks I have addressed all comments and would appreciate another round of reviews.

shacl12-sparql/index.html
HolgerKnublauch: Yes I have updated the text but left the test case with the verbose syntax for now. I suspect a small glitch in the current Jena version https://github.com/apache/jena/issues/4129

shacl12-test-suite/tests/sparql/rules/temp-triples-example.ttl
```diff
@@ -0,0 +1,106 @@
...
+		CONSTRUCT {
+			$this ex:offspring ?offspring .
+			?reifier sh:tempTriple true .
+			?reifier rdf:reifies ?tt .
+		}
+		WHERE {
+			$this ex:child+ ?offspring .
+			BIND (BNODE() AS ?reifier) .
+			BIND (TRIPLE($this, ex:offspring, ?offspring) AS ?tt) .
+		}
```
mgberg, line 46: The above comment says that this example appears in the SHACL 1.2 SPARQL spec, but the spec was updated to use annotation syntax. This is equivalent, but I wanted to double check if this should be updated accordingly.

shacl12-sparql/index.html
mgberg: Would it be valuable to explicitly state that they are visible to rules in the document?

shacl12-sparql/index.html
```diff
@@ -2191,6 +2191,75 @@ <h3>The sh:Rules Entailment Regime</h3>
...
+						All temporary triples, i.e., triples whose reifier is marked with <code>sh:tempTriple true</code>,
+						are automatically deleted by the SHACL engine at the end of the inference process.
+					</p>
+				</aside>
+				<p>
+					Creating temporary triples can be helpful when a computation is too complex to be carried out by a single rule.
+					The complex computation can then be decomposed into <em>multiple</em> rules:
+					some rules infer <em>partial</em> results, from which other rules derive the <em>final</em> results.
+					The partial results are only functional to the computation of the final results and are no longer needed once the latter have been obtained.
+					By marking these partial results as temporary triples, they can be automatically removed after the final results have been computed.
```
mgberg: What if a rule creates triples with predicate `sh:tempTriple` with an object other than `true`?

shacl12-sparql/index.html
```diff
@@ -2191,6 +2191,75 @@ <h3>The sh:Rules Entailment Regime</h3>
 					<a href="#rules-execution">rules execution instructions</a> prior to performing the actual validation. 
 				</p>
 			</section>
+			<section id="tempTriple">
+				<h3>Temporary Triples</h3>
+				<p>
+					It is sometimes useful for inference rules to produce triples that are only visible during the execution
+					of other rules but do not end up in the final inferences.
+					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inferences graph contains a
```
liviorobaldo:
```suggestion
					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inference graph contains a
```

shacl12-sparql/index.html
liviorobaldo:
```suggestion
					By marking these partial results as temporary triples, they are automatically removed after the final results have been computed.
```

shacl12-sparql/index.html
```diff
@@ -2229,6 +2298,8 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 							do
 								Execute one <a>iteration</a> over all <a>iterating rules</a> in the layer
 							while the iteration has produced newly inferred triples
+
+						Delete the <a>temporary triples</a> and their <a>reifiers</a>
```
liviorobaldo, line 2303:
```suggestion
Delete all <a>temporary triples</a>, together with the triples that mark their <a>reifiers</a> with `sh:tempTriple true`.
```

"and their reifiers" could give the misleading impression that *all* triples involving those reifiers are deleted, which is not what we intend here.

liviorobaldo:
> Thanks I have addressed all comments and would appreciate another round of reviews.

done 👍

shacl12-sparql/index.html
HolgerKnublauch: This risks overcomplicating the algorithm. I don't even know how to describe that because there can be all kinds of corner cases:

```
ex:a ex:b ex:c {| sh:tempTriple true, false |}
```
In that case, I guess two triples would survive, but what purpose do they serve:

```
_:b0 sh:tempTriple false .
_:b0 rdf:reifies <<( ex:a ex:b ex:c )>>
```

Then there can be the case

```
ex:a ex:b ex:c {| sh:tempTriple true |}
ex:a ex:b ex:c {| dcterms:created "..."^^xsd:dateTime |}
```
Why should the dcterms:created reifier survive when the temp triple is deleted?

I would like to keep this algorithms simple and just delete any reifiers when the actual triple gets deleted. If people really really want to store other reifiers, they can do so by other means, e.g. using rdf:Statement.

shacl12-sparql/index.html
HolgerKnublauch: The spec only talks about sh:tempTriple true, so sh:tempTriple false has no meaning and will count as a normal triple. If a triple has both true and false, true wins because that triple exists.

I believe the spec is correct in this respect but please reopen if you disagree.

shacl12-test-suite/tests/sparql/rules/temp-triples-example.ttl
HolgerKnublauch: Agreed. I have updated it.

shacl12-sparql/index.html
HolgerKnublauch: Could you suggest a specific change here? I am unclear what you mean.

liviorobaldo (approved):

shacl12-sparql/index.html
mgberg: Perhaps modifying this paragraph to include one extra sentence like the following, just to clarify it's a little different from `sh:sourceRule` (I didn't highlight the whole paragraph in the original comment so I can't do it as a code suggestion):
```html
					It is sometimes useful for inference rules to produce triples that are only visible during the execution
					of other rules but do not end up in the final inferences.
					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inference graph contains a
					<a>reifier</a> with the <a>value</a> <code>sh:tempTriple true</code>.
					<a>Temporary triples</a> and their <a>reifiers</a> are visible to executing rules.
					Rules can produce such triples as illustrated in the following example:
```

PR #1118, opened 2026-08-05
[UI] #609: Separate focus node and node shape resolution from rendering

### LINKED ISSUE #609: [UI] Possible input combinations to SHACL UI processor

smessie:
Closes #609

Resolves the open TODO on the SHACL Renderer definition by splitting one conflated concept into three, each with its own contract.

### Concepts

**Focus Node and Node Shape Resolution** — a pure algorithm, no presentation rules. Inputs: data graph, shapes graph, focus node (may be absent), node shape (may be absent). Output: a flat, possibly empty list of fully determined (focus node, node shape) pairs. All four input combinations are now defined:

1. Both given → the single pair.
2. Node shape only → one pair per target of that node shape.
3. Focus node only → one pair per non-deactivated node shape whose targets include that focus node.
4. Neither → one pair per target of every non-deactivated node shape.

Targets are computed as defined in the [Targets](https://w3c.github.io/data-shapes/shacl12-core/#targets) section of SHACL 1.2 Core, including implicit class-based targets; resolution introduces no new machinery. Steps 3 and 4 enumerate rather than choose: several node shapes may target the same focus node, and deciding which the user meets first is presentation, not resolution.

**SHACL UI Application** (name under discussion in #1177) — the concept sitting between resolution and rendering. It drives one or more SHACL Renderer instances, decides how the user encounters the resolved pairs (all at once, behind a selector, paginated, or navigated), and owns the logic for the initial focus node and node shape plus any cross-pair selectors.

**SHACL Renderer** — receives a single, fully determined pair and renders it into a Node UI Component. It performs no resolution and is never handed absent inputs.

### Normative requirements

- Supporting step 1 is a **MUST**: an implementation that accepts only fully determined inputs conforms.
- Supporting steps 2, 3 and 4 is **OPTIONAL**; an implementation that supports one MUST produce exactly the pairs that step defines.
- **Reachability**: a SHACL UI Application MUST allow the user to reach every resolved pair and MUST NOT expose only a subset. Reachable means navigable, not displayed all at once. There is no filtering step anywhere in the pipeline.

### Other changes to the document

- SHACL Renderer inputs now include the [prepared scoring graph](https://w3c.github.io/data-shapes/shacl12-ui/#scoring-algorithm-scoring-graph-preparation) alongside the data graph and shapes graph, with a paragraph explaining what the scoring graph holds and why the renderer consumes the *prepared* one.
- The manual/automatic mode distinction and its TODO are removed.
- `deactivated` added to the terminology list.
- Two non-normative examples: an alerts dashboard (several pairs) and page-rank-style entry points (many pairs — the navigable tree comes from the data graph edges, not from the order of the output).
- The appendix pattern TODO is narrowed to what genuinely remains open: how an application nominates which node shapes serve as entry points.

### Changed since the first review round

For reviewers who read the original version:

- Resolution became a pure algorithm — multiplicity, reachability, selectors, navigation and the examples all moved out onto the SHACL UI Application.
- Steps 3 and 4 were "out of scope in the current revision"; they are now defined.
- The "Zero, one, or several pairs" subsection is gone; its content folded into the SHACL UI Application section.
- Output is a flat list — the ordered-list rationale and the "defines what is computed, not how it is invoked" paragraph are dropped.
- Conformance split: previously all-OPTIONAL, now MUST for step 1 and OPTIONAL for steps 2–4.
- Editorial fixes from review: `target` is defined as a term rather than a raw link, SHACL 1.2 Core is cited by its Targets section rather than by the non-existent phrase "target resolution", the redundant data graph mention is dropped, the renderer section no longer borrows resolution's "pair" wording, and the editor's note on #994 is removed now that #994 has merged.

### Out of scope / follow-ups

- Multiple node shapes for one focus node — #1119
- Naming of the new concept — #1177
- Profiles, and where the conformance statements ultimately belong
- Entry-point node shape nomination (tracked by the remaining appendix TODO)
- Query-mode rendering and the `undefined` focus node in the Select / matcher / score functions

---

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-609-focus-node-node-shape-resolution/shacl12-ui/index.html)


danielbeeke (approved):
Looks good to me.
One question though, is it the case that __node shape__ may be multiple IRIs? 
I am asking because there is this use case of one resource with multiple rdf:type and there is this use case of a hierarchy of shapes via rdfs:subClassOf where it might make sense to give the IRIs of all the shapes in the applicable hierarchy.

smessie:
@danielbeeke no, currently not. The document is written now to accommodate exactly one focus node and node shape.
You know the use case best. How would you want to adopt scenarios like that in the specification?
And to understand it better myself: how should an implementation behave in that case?

danielbeeke:
I would make it plural: __node shapes__.
Technically it doesn't do much, it only changes a little bit how you gather property shapes that belong to the current Node UI component.

danielbeeke:
I have created https://github.com/w3c/data-shapes/issues/1119 to deal with my question.

bergos (changes requested):
Requesting changes to hold this PR until the "How open/closed should the SHACL UI specification be?" issue is resolved, since its outcome may affect this PR. I'll approve once I've reviewed it and confirmed there's no conflict.

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
       <p>This section explains the UI concepts and how they work together.</p>
 
+      <section id="focus-node-and-node-shape-resolution">
+        <h3>Focus Node and Node Shape Resolution</h3>
+
+        <p>
+          <dfn data-lt="focus node and node shape resolution" id="dfn-focus-node-and-node-shape-resolution" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn">Focus Node and Node Shape Resolution</dfn>
+          determines, from the available inputs, the (<a>focus node</a>, <a>node shape</a>) pairs to be rendered.
+          It is the only part of this document in which the <a>focus node</a> or the <a>node shape</a> may be absent;
+          every other part, including the <a>SHACL Renderer</a>, operates on fully determined inputs.
```
bergos: I don't think it's a good idea to make claims about the rest of the document within this section. Either remove the reference to the rest of the document, or move it into the introduction section. Otherwise, if things change later, an unrelated section would need to be updated too. I'd prefer removing the reference.

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
+        <p>Steps:</p>
+        <ol>
+          <li>
+            If both the <strong>focus node</strong> and the <strong>node shape</strong> are present, return a list
+            containing the single pair (<strong>focus node</strong>, <strong>node shape</strong>).
+          </li>
+          <li>
+            If the <strong>node shape</strong> is present and the <strong>focus node</strong> is absent, compute the
+            <a href="../shacl12-core/#targets">target</a> of the <strong>node shape</strong> in the
+            <strong>data graph</strong> using target resolution as defined in [[SHACL-CORE]], and return one pair per
```
bergos: - **target** should be defined as term like **focus node**. That href link should be removed.
- **target resolution** is not an official term used in the core spec. Currently it reads like opening link to SHACL Core would guide me to target resolution. Maye use something like: compute the targets of the node shape as defined in the target section of the SHACL Core spec.
- data graph is redundant - it's already defined in the SHACL Core spec.

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
+              paginating, or any other arrangement are all conforming. This latitude is deliberate rather than an
+              omission.
+            </li>
+          </ul>
+          <p>
+            An implementation that supports <a>Focus Node and Node Shape Resolution</a> MUST allow the user to reach
+            every pair in the output; it MUST NOT expose only a subset. Reaching a pair means being able to act on it in
+            whichever mode the <a>SHACL Renderer</a> is operating in: viewing the pair in view mode, or viewing and
+            editing it in edit mode. The requirement constrains which pairs are reachable, not which modes an
+            implementation offers.
```
bergos, line 1242: I guess this should cover use cases like page-rank and only return the root nodes. That ties defined algorithm to the render logic. But earlier there is this paragraph:

> This algorithm defines what is computed, not how it is invoked. An implementation MAY expose it as a single entry point that accepts optional inputs, as separate entry points, or as a standalone function reused by other software. None of these arrangements is assumed here.

I would propose to describe it more like a pipe. Describe that all pairs MUST be returned, and there is an optional filter step to remove pairs for use cases like tree root nodes / page-rank, aligned to the renderer logic.

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
+        </section>
+
+        <section id="focus-node-and-node-shape-resolution-conformance">
+          <h4>Conformance</h4>
+          <p>
+            Support for <a>Focus Node and Node Shape Resolution</a> is OPTIONAL. An implementation MAY support it. An
+            implementation that supports it MUST produce the output pairs as defined above and MUST satisfy the
+            reachability requirement stated above.
+          </p>
+        </section>
```
bergos, line 1264: Skips this. Profiles should handle that.

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
-            <a>focus node</a> and a <a>node shape</a>. When all four inputs are provided, the renderer
-            operates in <strong>manual mode</strong>. When the focus node or node shape is not provided,
-            the renderer may operate in <strong>automatic mode</strong>, in which case the application determines
-            appropriate values for the missing inputs.
+        <p>
+          Rendering is defined over exactly one (<a>focus node</a>, <a>node shape</a>) pair. Both members of the pair are
+          always present: at this step there is no undefined case and no multiplicity. Determining which pairs to render,
+          including any case in which the <a>focus node</a> or the <a>node shape</a> is absent, or in which zero or
+          several pairs result, is the responsibility of <a>Focus Node and Node Shape Resolution</a>. Where resolution
+          yields several pairs, rendering is performed once per pair.
```
bergos: This is redundant and the language of the resolution leaked into it. Why use the term **pair** when the renderer requires more parameters?

shacl12-ui/index.html
```diff
@@ -1138,32 +1138,170 @@ <h2>Rendering Concepts</h2>
...
+        <p>
+          The scoring graph is the RDF graph that holds the widget matcher definitions — the
+          <code>shui:WidgetScore</code> and <code>shui:WidgetAcceptMatcher</code> instances — that drive widget
+          selection, as defined in the <a href="#scoring-system">Scoring System</a> section. When selecting
+          widgets, the renderer passes down the prepared scoring graph produced by the scoring graph preparation step,
+          which is what the scoring functions operate on.
+        </p>
+
+        <p class="ednote">
+          The prepared scoring graph and the mandatory scoring graph preparation step referenced here are introduced by
```
```suggestion
          The prepared scoring graph and the optional scoring graph preparation step referenced here are introduced by
```

shacl12-ui/index.html
```diff
@@ -2140,6 +2278,10 @@ <h3>Root Shapes and Entry-Point Forms</h3>
       <p class="ednote">
         TODO: Add a pattern describing how applications choose the initial node shape and focus node for rendering.
         This pattern should explain that applications may nominate one or more root node shapes for entry-point forms.
+        Where a <a>node shape</a> is nominated but no <a>focus node</a> is given, the resulting pairs are those computed
+        by <a>Focus Node and Node Shape Resolution</a>. Choosing a root <a>node shape</a> from a given <a>focus node</a>,
+        and discovering entry-point <a>node shapes</a> when neither input is given, are out of scope in the current
+        revision and are open questions there.
```
bergos, line 2402: Is this a proposal to let the renderer call the resolution? We should define a lean flow resolution -> renderer, where the resolution is optional. That flow doesn't belong into the renderer section.

bergos (changes requested):
The PR's language isn't consistent about whether Focus Node and Node Shape Resolution is a function or a component. It's initially framed as a function, but later described as more tightly coupled to the SHACL Renderer — which implies a component. Even if the intent is for it to be a component, the boundary with the SHACL Renderer needs to be drawn more clearly; right now the two concepts bleed into each other's sections throughout the document.

Suggestion: Define it explicitly as a component, and give it more flexibility in what it returns. This would:

- Cover the page-rank use case
- Also allow it to be used as a navigator between focus nodes and node shapes

shacl12-ui/index.html
smessie: Done. The reference to the rest of the document is removed. The section now states only its own contract: its focus node and node shape inputs may be absent, and it resolves them into pairs whose members are always present.

shacl12-ui/index.html
smessie: Done, all three. Step 2 now uses the defined `targets` term with no raw href, cites the Targets section of SHACL 1.2 Core rather than the phrase "target resolution", and drops the redundant data graph mention. The document `target` definition now points at the SHACL 1.2 Core anchor.

shacl12-ui/index.html
smessie: Could you describe the concrete page rank use case you have in mind, so I address the real scenario? On the optional filter itself: the section already requires that an implementation supporting resolution allows the user to reach every pair and does not expose only a subset. A normative filter that drops pairs from the output would contradict that reachability requirement. If the intent is a presentation choice where every pair stays reachable, that is already allowed. If it is removing pairs entirely, that works against reachability, so I would rather not add it in this revision.

shacl12-ui/index.html
smessie: I would keep the conformance statement here for now and revisit it in the Profiles PR. Profiles and conformance requirements belong together, since Profiles let an implementation report which parts it conforms to. I would rather move this deliberately there than drop it now.

shacl12-ui/index.html
smessie: Reworked. The renderer section no longer uses "pair". The SHACL Renderer is now described as software that operates on a data graph, a shapes graph, and a prepared scoring graph, together with a focus node and node shape per component, and produces one or more Node UI Components. The single focus node and node shape is the per component unit, not the whole renderer.

shacl12-ui/index.html
smessie: The editor's note is removed now that #994 is merged, so the suggestion no longer applies. Note that the merged Scoring Graph Preparation section states a scoring graph MUST be prepared before it is passed to the score function, so describing the step as optional would not match the normative text now in the document. The renderer references that section and uses the prepared scoring graph.

smessie:
@bergos Whether resolution is a function or a component is deliberately left open. It is an implementation detail that does not change what is computed or what a user can rely on; the aim is to fix what each step does, not how it is exposed. To reduce the bleed you noted, the renderer section now keeps the SHACL Renderer as the software that produces one or more Node UI Components, with the single focus node and node shape as the per component unit, and resolution as the separate step that decides which and how many. On navigation: resolving focus nodes from a node shape is step 2 and already supported; resolving a node shape from a focus node is step 3 and explicitly out of scope this revision. Could you also clarify what you mean by the page rank use case? That is still unclear to me, and I want to understand it before adjusting anything.

shacl12-ui/index.html
smessie: This note pre-existed; I only extended it to reference the new Focus Node and Node Shape Resolution section. The cases it leaves open, a node shape from a focus node and entry points when neither input is given, are the two points marked out of scope there (steps 3 and 4). Solving those could let resolution cover this pattern, though whoever works on that will need to investigate, since multiplicity and reachability already handle surfacing several entry points. So I would keep this as a reference for now and decide in a follow-up whether to fold it in or keep it separate.

bergos:
@smessie I've proposed an alternative solution for the issue in #1133. The page-rank example is included, and the PR description outlines the other differences.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
+
+        <p>
+          <dfn data-lt="focus node and node shape resolution" id="dfn-focus-node-and-node-shape-resolution" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn">Focus Node and Node Shape Resolution</dfn>
+          determines, from the available inputs, the (<a>focus node</a>, <a>node shape</a>) pairs to be rendered. Its
+          <a>focus node</a> and <a>node shape</a> inputs may be absent; it resolves them into pairs in which both
+          members are always present.
+        </p>
+
+        <p>
+          This algorithm defines what is computed, not how it is invoked. An implementation MAY expose it as a single
```
robert-david: I don't understand this paragraph. I suggest to remove it if it is not essential.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
+          entry point that accepts optional inputs, as separate entry points, or as a standalone function reused by
+          other software. None of these arrangements is assumed here.
+        </p>
+
+        <p>Inputs:</p>
+        <ul>
+          <li><strong>Data graph</strong> — The <a>data graph</a>.</li>
+          <li><strong>Shapes graph</strong> — The <a>shapes graph</a>.</li>
+          <li><strong>Focus node</strong> — A <a>focus node</a>, which may be absent.</li>
+          <li><strong>Node shape</strong> — A <a>node shape</a>, which may be absent.</li>
```
robert-david: How can the node shape be absent? If there is no shape, there are no target nodes.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
+              <strong>Zero pairs.</strong> Nothing is rendered. The implementation decides what, if anything, the user
+              sees in its place.
+            </li>
+            <li>
+              <strong>One pair.</strong> The pair is rendered by the <a>SHACL Renderer</a>.
+            </li>
+            <li>
+              <strong>Several pairs.</strong> How the pairs are surfaced is deliberately left to the implementation.
+              Rendering all of them at once as a list, rendering one at a time with a means of selecting between them,
+              paginating, or any other arrangement are all conforming. This latitude is deliberate rather than an
```
robert-david: I don't understand the last sentence.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
+              same pairs one at a time behind a selector, or paginating them, would be equally conforming: the only
+              requirement is that every alert can be reached.
+            </p>
+          </aside>
+        </section>
+
+        <section id="focus-node-and-node-shape-resolution-conformance">
+          <h4>Conformance</h4>
+          <p>
+            Support for <a>Focus Node and Node Shape Resolution</a> is OPTIONAL. An implementation MAY support it. An
```
robert-david: Why is this OPTIONAL? If we want to make SHACL UI portable, i think this should be mandatory, right?

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
       <section id="renderer">
         <h3>SHACL Renderer</h3>
         <p>
           A <dfn data-plurals="SHACL Renderers" id="dfn-shacl-renderer" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn">SHACL Renderer</dfn>
           is software that processes SHACL
-            <a>shapes</a> and data and dynamically generates a user-interface using one
-            or more <a>node UI components</a>. Implementations may add supplementary UI elements, such as a submit
+            <a>shapes</a> and data and dynamically generates a user-interface using one or more
+            <a>Node UI Components</a>. Implementations MAY add supplementary UI
+            elements, such as a submit
```
robert-david, line 1186: Should we really define it on such a granular level? Better to stay focused on the SHACL elements.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,162 @@ <h2>Rendering Concepts</h2>
...
+        <p>
+          A SHACL Renderer operates on a <a>data graph</a>, a <a>shapes graph</a>, and a
+          <a href="#scoring-algorithm-scoring-graph-preparation">prepared scoring graph</a>, together with the
+          <a>focus node</a> and <a>node shape</a> for each <a>Node UI Component</a> it renders. From these it produces
+          one or more <a>Node UI Components</a>.
         </p>
 
-        <p class="ednote">TODO: review SHACL Renderer definition, particularly the manual and automatic mode, now that we have the scoring system defined.</p>
+        <p>
+          The scoring graph is the RDF graph that holds the widget matcher definitions — the
```
robert-david, line 1206: Should this be optional? I guess users want to be able to just use their (validation) shapes to render a UI as a starting point.

robert-david (changes requested):

smessie:
Thanks @bergos. I have now read through #1133 in full, so let me respond with that in mind and then ask two things I need clarified.

On the page-rank example: as far as I can tell it is already supported by this PR. Resolution returns all the pairs, and the multiplicity section explicitly allows surfacing them in any arrangement, including presenting a small set of starting points and letting the user navigate to the rest, as long as every pair stays reachable. Your own example keeps every node reachable by traversal from the selected roots, so it appears to meet the reachability requirement here without additional machinery. In #1133 you lower reachability from MUST to SHOULD and let the manager filter pairs, which to me weakens the guarantee this PR intends to give rather than enabling the example.

On resolving the two steps currently marked out of scope, a node shape from a focus node and the case where neither input is given: we deliberately left these open because we discussed them in the task force and concluded they uncover considerably more complex cases. In the meetings of 14 July and 21 July 2026 we looked at exactly these scenarios, in particular how a renderer should decide which node shape or focus node to select when several candidates match, and agreed they need further discussion. On 4 August 2026 we then aligned on keeping the SHACL Renderer focused on the inputs it receives, with target resolution and any higher-level selection components, such as a multi-item list or search, handled separately. In #1133, step 3 returns all non-deactivated node shapes whose targets include the focus node, and step 4 does the same across the whole graph. That avoids the selection question rather than resolving it, and it is a substantial design decision I do not think we should settle inside this PR. I am open to exploring further how we can support these two cases, but I believe we should do that as follow-up work in a separate issue rather than in this PR.

I also see that #1133 bundles further changes that go beyond issue #609, such as the normative SHACL Render Manager component, a new Profiles section, and changing the ordering requirements from MUST to SHOULD and MAY.

**To have a good discussion, I would really appreciate direct answers to these two questions:**

**1. Concretely, how is the page-rank use case not supported by this PR? Please point to the specific step or sentence that blocks it.**

**2. Which changes do you consider actually required for #609, and which of those do you see as blocking this PR?**

These are not clear to me from the PR description, and therefore I think it is good if we can first focus on this PR, resolve any blocking issues, and then look into follow-up PRs for other separate improvements.

shacl12-ui/index.html
smessie: This paragraph is intentional and, I would argue, essential. It states the governing constraint we agreed on: the specification defines what resolution computes, never how it is exposed. It must remain equally valid to implement one entry point taking optional inputs, separate entry points, or resolution as a standalone function reused by a separate renderer. Without this note, readers tend to assume a specific API or component boundary, which the spec deliberately does not impose. I am happy to make it more concrete if it reads too abstractly, but I would prefer to keep it.

shacl12-ui/index.html
smessie: These are two different cases. When the node shape is absent you do not compute targets from it. Instead you would resolve a node shape from a given focus node, based on the properties that node has in the data graph. In the task force we have raised ideas along exactly these lines, but it is a complex case (it is a ranking problem, since several node shapes may target one node), so it is explicitly marked out of scope in this revision, in steps 3 and 4. The input is declared as possibly absent so the algorithm's contract is complete, with the unhandled cases deferred rather than silently broken.

shacl12-ui/index.html
smessie: Reworded to: "This freedom is intentional: the specification does not prescribe how several pairs are presented, and the absence of a prescribed arrangement must not be read as an oversight."

shacl12-ui/index.html
smessie: Handling missing inputs is the more complex case, and we do not want to force every implementation to support it. The minimum an implementation MUST support to be conforming is the case where all inputs are well defined. Supporting certain missing inputs is optional, but if an implementation chooses to support it, the semantics are strict. We discussed and agreed this in the meetings (see the 21 July 2026 discussion on rendering without a focus node, framed as an added capability, and the 4 August 2026 discussion where resolution is handled by higher-level components that pass focus nodes to the renderer, keeping the renderer's mandatory contract to the inputs it receives).

shacl12-ui/index.html
smessie: I want to make sure I understand your comment correctly. This block (submit button, edit/view toggle, selectors, state management, widget management) already existed in the document; in this PR I only corrected the letter case of the normative keywords. These are non-normative examples (MAY) of what a renderer may add on top of the Node UI Components, not requirements. If you think this pre-existing list should be changed, I would suggest handling that in a follow-up PR (or an issue first), since it is outside the scope of this PR. Could you clarify what change you have in mind?

shacl12-ui/index.html
smessie: The scoring graph is a required input, not optional: the scoring system needs it to select widgets. It is not something a user generally authors, though. It holds the widget score and widget accept matcher instances, which are already defined in this repository for the built-in widgets and may be extended by implementations. So "just use validation shapes as a starting point" already works: scoring graph preparation runs on those shapes and produces the prepared scoring graph the renderer consumes.

bergos:
Mixing the algorithm with the filtering for use cases like tree roots from page-rank, which depend on the renderer, makes it hard to understand and adds an unnecessary cross-reference between them.

"Zero, one, or several pairs" doesn't fit the approach of defining only what is required.

I can't recall the decision to skip the case where node shape == null. I think introducing an ordered list for the pairs is the actual problem. This again illustrates the issue of mixing the algorithm with rendering tasks.

I think all of them should be addressed.

smessie:
@bergos let me take your three points in turn.

**On mixing the algorithm with filtering for tree roots / page-rank:** #1118 has no filtering step. Resolution returns all pairs, and the "Zero, one, or several pairs" section only requires that every pair is reachable. There is no step that removes pairs based on renderer logic. That optional filter is the one introduced in #1133, not here. So I do not think this PR ties the algorithm to render logic in the way you describe. **Could you point to the exact text in #1118 that you read as a filter tied to the renderer?**

**On "Zero, one, or several pairs" not fitting "define only what is required":** I believe this section helps the specification by making explicit what the possible outputs are (zero, one, or several pairs) and how each cardinality is treated. It does not impose any requirement beyond the reachability guarantee; everything else is deliberately left to the implementation. Its purpose is to give the reader context, so the silence on presentation is not misread as an omission. I do not see how it conflicts with defining only what is required. **Which parts of that section do you think conflict with that approach, and how exactly would you propose to resolve it?**

**On the case where the node shape is absent:** I left this out of scope for this revision and marked it as an open question in the document. Resolving a node shape from a given focus node is a ranking problem, since several node shapes may target one node, and we flagged it as complex in earlier meetings (see the discussions on 14 July 2026 on selecting which matching entity to use as the focus node, and on 21 July 2026 on the edge cases when no focus node is provided). I would like to keep the door open to handle it: I am open to support these more complex cases, but I think they deserve a dedicated GitHub issue with recorded use cases and a follow-up PR once we agree on an approach, rather than widening the scope of this PR.

**On the ordered list:** I take it you mean the paragraph explaining why the output is a list rather than a set. I have no strong opinion there. I used a list so that a single invocation yields a definite sequence, but the order is explicitly not prescribed. If it reads better to you, I am happy to drop the ordering and define the output as a set instead. Just let me know and I will change it.

To keep this PR moving, I would like to focus here on the concrete blocking issues, and handle the more complex node-absent cases and any broader design philosophy as separate follow-up work.

bergos:
I don't understand how "has no filtering step" and "only requires that every pair is reachable" fit together. If there's no filter, then the "every pair is reachable" requirement shouldn't need to be mentioned at all. And if there truly is no filter, the page-rank example won't work, since a tree structure can't be represented as an ordered list.

More generally, the text is hard to follow, and even the explanations given so far haven't made it clearer. I'd like to raise this on the next task force call, since I don't see how it can be fixed without a major restructuring.

smessie:
@bergos I am open to discussing this on a call, but I would prefer if we could resolve it here in the thread first. It is useful to have the reasoning written down, and the agendas for our next calls are already quite full. Let me explain the two points in full, because I think there are two separate misunderstandings here, and once they are cleared up the change needed is much smaller than a restructuring.

**1. "No filtering step" and "every pair is reachable" do not contradict each other, because they are about two different layers.**

The algorithm and the reachability requirement operate at different levels:

- The **algorithm** produces the complete collection of (focus node, node shape) pairs. It never removes any pair. That is what I mean by "no filtering step."
- The **reachability requirement** is a conformance rule on the *implementation*, not on the algorithm. It says that once resolution has produced those pairs, the implementation must not silently drop some of them at the presentation layer.

So reachability is not evidence of a hidden filter. It is the guarantee that there is *no* filter, including at the UI layer. It is worth stating precisely *because* the algorithm does not filter: without it, an implementation could resolve, say, 100 pairs, render only the first one, and still claim conformance. The requirement is what forbids that. The two statements reinforce each other: the algorithm returns everything, and the implementation must not quietly discard part of it.

**2. The page-rank / tree use case does work, because the output is a flat collection of pairs and the tree is not encoded in it.**

You are right that a tree cannot be represented as an ordered list. It does not need to be. Resolution returns a flat collection of pairs, nothing more. The tree in the page-rank case is built by the implementation from the **edges in the data graph**, not from the order of the resolution output. The output is the set of nodes to be made reachable; the tree structure is a separate, implementation-side presentation concern derived from the data.

Concretely, page-rank works like this under #1118: resolution yields the pairs to be made reachable; the implementation computes page-rank and the tree edges from the data graph; it renders the top-ranked node(s) as entry points and lets the user traverse the edges to reach the rest. Every pair stays reachable, so it conforms. "Reachable" means the user can navigate to a pair, not that all pairs are shown at once. Rendering only the roots up front and revealing the rest through navigation is explicitly allowed.

The one thing #1118 does not allow is permanently discarding pairs so the user can never reach them at all. If your page-rank use case requires rendering only the top nodes and making the remainder permanently unreachable, then that is a deliberate design difference we should discuss, but it is a scoping choice, not a contradiction in the text. Note also that switching the output from an ordered list to a set, which I already offered to do, changes none of this, which is exactly the point: the container was never meant to carry the tree.

**3. On readability.**

I do hear that the section reads as abstract, and I am glad to improve that. I think two local edits address most of it, without a restructuring:

- Drop the ordering and define the output as a set, as I offered above, so there is no "ordered list" to misread as a presentation order.
- Sharpen the "Zero, one, or several pairs" section to state explicitly that the output is a flat collection of pairs, that any presentation structure such as a tree or pagination is derived separately by the implementation, and that "reachable" means reachable through navigation rather than shown all at once.

These are wording changes to the existing two-step structure. The split of resolution and rendering is what closes the open TODO from #609, and I do not think the perceived contradiction requires reworking it. If we can align on the wording here in the thread, I am glad to turn it into a concrete change, and we can always fall back to a call if we get stuck.

shacl12-ui/index.html
```diff
@@ -1149,34 +1149,163 @@ <h2>Rendering Concepts</h2>
...
+        </ul>
+
+        <p>Output:</p>
+        <ul>
+          <li>
+            An ordered list of (<a>focus node</a>, <a>node shape</a>) pairs, which may be empty. Every pair in the output
+            is fully determined: neither member of any pair is absent. A list is used, rather than a set, only so that a
+            single invocation yields the pairs in a definite sequence. The focus nodes derive from a <a>node shape</a>'s
+            SHACL Core <a>targets</a>, which form a set with no inherent order, so the order of the pairs is an
+            implementation detail and is not fixed here.
```
edmondchuc: It's not clear to me why this section distinguishes between sets and lists. Is ordering or uniqueness important here?

shacl12-ui/index.html
danielbeeke: Shall we accept this piece of text for now? I am in favor of that. The current spec also contains MUST / MAY statement. If this becomes a blocker, then all PR's will be halted until we have made a decision on profiles.

smessie: Yes, changes resulting from other discussions such as the profiles should lead to follow-up PRs once settled.
It would be great if those with changes requested could review the PR again such that we can move forward: @robert-david @bergos (if I'm not mistaken).

smessie:
@bergos Restructured along the lines of your review and #1133.

* **Resolution is now a pure algorithm**: definition, inputs, output, steps. Multiplicity, reachability, the examples etc. have all moved out. I believe this was your main point.
* **Added the separate concept between resolution and the renderer.** It owns multiplicity, selectors, navigation, and the logic for determining the initial focus node and node shape (removed from the SHACL Renderer's list). The renderer now takes one fully determined pair and performs no resolution.
* **Steps 3 and 4 are defined**, following #1133. The output is every pair, not one, so enumerating the non-deactivated node shapes is a lookup rather than a selection. `deactivated` added to terminology.
* **"Zero, one, or several pairs" is gone** as a section; some text has folded into the enclosing one.
* **The ordered list is gone**: output is a list, rationale removed.
* **Your page-rank example is included**, with the point that resolved the confusion between us: the output is flat, and the tree comes from the data graph edges.
* **Conformance is split**: step 1 MUST, steps 2 to 4 OPTIONAL.

Name: proposal and reasoning in #1177. One-line change either way, so it doesn't need to hold up review of the structure.

One thing I hope you can agree to though: #1133 has MAY filter with SHOULD reach; this PR keeps MUST reach and permits no filtering. I don't think your page-rank example needs the weaker form: it states that root selection keeps every node reachable by traversal, which satisfies MUST as written. If you have a case where resolved pairs must be *permanently* unreachable, could you describe it?

bergos:
@smessie  it would be great if we could close this PR and switch over to the original, #1133. Your latest commit essentially duplicates my alternative proposal, so I take it you're on board with that approach. At this point, the title, description, and discussion here no longer reflect what the PR actually contains.

Separate from my PR specifically: for a change of this magnitude, it would generally make sense to open a new PR rather than keep evolving this one.

I'm also not sure why some parts differ from my proposal. It would be much easier to discuss if you left comments directly on #1133, since that's the source of your recent major changes. In the meantime, I've only changed the target branch to gh-pages and resolved the conflicts.

smessie:
All previous comments have been addressed and no new comments or requests for changes have been addressed since my last comment 3 weeks ago. 
Shall we proceed with merging this PR?

bergos (changes requested):
> All previous comments have been addressed and no new comments or requests for changes have been addressed since my last comment 3 weeks ago. Shall we proceed with merging this PR?

We should proceed with the original PR, as mentioned in the comment you haven't addressed:  https://github.com/w3c/data-shapes/pull/1118#issuecomment-5422265651

PR #1120, opened 2026-08-05
#1110: Add `sh:expectedPredicate` to link rules to `sh:defaultValue`/`sh:values`

### LINKED ISSUE #1110: SHACL SPARQL Rules should integrate with sh:defaultValue and sh:values
> The SHACL Rules framework could have a node expressions step (c.f. #957, #958), or have a first step of "apply defaults" (and maybe post -processing? Not sure that's needed).
> 
> Unless triples are materialized, rather than just declared, later layer CONSTRUCT pattern matching doesn't see a default.
>  

 _Originally posted by @afs in [#960](https://github.com/w3c/data-shapes/issues/960#issuecomment-5151988730)_

HolgerKnublauch:
Closes #1110

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1110/shacl12-sparql/index.html#expectedPredicate)

This is a draft PR for now because this branch is based on the sh:tempTriple branch which isn't merged into main yet. Apart from that I believe it's good to be reviewed and I have implemented the algorithm.

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2196,7 +2196,7 @@ ex:Granddaughter
 				<p>
 					It is sometimes useful for inference rules to produce triples that are only visible during the execution
 					of other rules but do not end up in the final inferences.
-					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inferences graph contains a
+					A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inference graph contains a
 					<a>reifier</a> with the <a>value</a> <code>sh:tempTriple true</code>.
 					Rules can produce such triples as illustrated in the following example:
 				</p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2261,8 +2261,10 @@ ex:SmallRectangle ex:isSmall true .
 						</div>
 					</aside>
 					<p>
-						Unless told otherwise, a SHACL rules engine will remove the derived triples from
-						the inferences at the end of the execution process.
+						A SHACL rules engine will remove the derived triples from the inferences at the end of
+						processing each layer, except those triples that have also been inferred by rules.
+						Some engines may have a setting to keep all derived triples, for example
+						when data is exported to systems that do not support SHACL.
 					</p>
 				</section>
 			</section>
@@ -2411,7 +2413,8 @@ ex:SetYoungestOffspringsRule
 							do
 								Execute one <a>iteration</a> over all <a>iterating rules</a> in the layer
 							while the iteration has produced newly inferred triples
-							Delete the <a>derived triples</a> and their <a>reifiers</a>
+							Delete the <a>derived triples</a> (except those that were also inferred by rules)
+							  and their <a>reifiers</a>
 
 						Delete the <a>temporary triples</a> and their <a>reifiers</a>
 					</pre>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2161,7 +2161,7 @@ ex:Granddaughter
 					<ul>
 						<li>
 							<a data-cite="shacl12-core#syntax-rule-path-defaultValue"><code>sh:defaultValue</code></a> defines what values should be used
-							when no other other values are present for a property.
+							when no other values are present for a property.
 							For example, the default value of <code>ex:childCount</code> could be <code>0</code>.
 						</li>
 						<li>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2191,67 +2191,6 @@ ex:Granddaughter
 					<a href="#rules-execution">rules execution instructions</a> prior to performing the actual validation. 
 				</p>
 			</section>
-			<section id="rules-execution">
-				<h3>General Execution Instructions for SHACL Rules</h3>
-				<p>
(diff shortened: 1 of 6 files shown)


shacl12-sparql/index.html
```diff
@@ -2150,9 +2149,122 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
 				</section>
+				<section id="expectedPredicate">
+					<h3>Expected Derived Triples (sh:expectedPredicate)</h3>
+					<p>
+						SHACL Core includes the following properties to <em>derive</em> property values that are not necessarily asserted in the <a>data graph</a>:
+					</p>
+					<ul>
+						<li>
+							<a data-cite="shacl12-core#syntax-rule-path-defaultValue"><code>sh:defaultValue</code></a> defines what values should be used
+							when no other other values are present for a property.
```
mgberg:
```suggestion
							when no other values are present for a property.
```

shacl12-sparql/index.html
```diff
@@ -2150,9 +2149,122 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+						<li>
+							<a data-cite="shacl12-core#syntax-rule-path-defaultValue"><code>sh:values</code></a> defines general instructions for computing
+							derived values for a property.
+							For example, the <code>ex:area</code> of a rectangle may be derived by multiplying its <code>ex:width</code> and <code>ex:height</code>.
+						</li>
+					</ul>
+					<p>
+						Both properties can use <a>node expressions</a> including those defined in <a data-cite="shacl12-node-expr"></a>.
+						Both properties can only be used to compute the <a>objects</a> of a given <a>predicate</a> for the
+						<a>subjects</a> that are targetted by <a>shapes</a>.
```
mgberg:
```suggestion
						<a>subjects</a> that are targeted by <a>shapes</a>.
```

shacl12-sparql/index.html
```diff
@@ -2150,9 +2149,122 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+
+# Inferred triples:
+ex:IncompleteRectangle ex:isSmall true .
+ex:SmallRectangle ex:isSmall true .
+							</div>
+						</div>
+					</aside>
+					<p>
+						Unless told otherwise, a SHACL rules engine will remove the derived triples from
+						the inferences at the end of the execution process.
```
mgberg: How would one tell the rules engine otherwise? Do we need a standardized way to do that?

shacl12-sparql/index.html
```diff
@@ -2225,10 +2406,14 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 				<div class="algorithm">
 					<pre class="text">
 						For all <a>layers</a> in the rule set (in ascending order):
+							Compute the <a>expected derived triples</a> of the layer
```
mgberg: The following definition is above:
> The expected derived triples of a rule are the derived triples for all values of the property sh:expectedPredicate at the rule.

For consistency, might it be better to say something like
```suggestion
							Compute the <a>expected derived triples</a> for all rules in the layer
```

shacl12-sparql/index.html
```diff
@@ -2225,10 +2406,14 @@ <h3>General Execution Instructions for SHACL Rules</h3>
...
 							Execute one <a>iteration</a> over all <a>run-once rules</a> in the layer
 							do
 								Execute one <a>iteration</a> over all <a>iterating rules</a> in the layer
 							while the iteration has produced newly inferred triples
+
+						Delete the <a>temporary triples</a> and their <a>reifiers</a>
+						Delete the <a>derived triples</a> and their <a>reifiers</a>
```
mgberg: Should this be moved into the per-layer loop? If rules in multiple layers declare the same `sh:expectedPredicate` for the same shape, and if expected derived triples are computed on a per-layer basis, one could presumably end up with different expected derived values at each layer.

shacl12-sparql/index.html
HolgerKnublauch: The spec does not define the API for calling a rules engine. Such a flag would need to go into such an API.

My goal here was just to indicate that some engines may want to keep the derived triples, e.g. when someone wants to export a complete dump of all inferences.

If you have a specific suggestion on how to make this clearer, I'd appreciate a diff.

shacl12-sparql/index.html
HolgerKnublauch: Excellent point! I was wondering how to solve this. For example, one batch of rules may produce new instances for which no derived triples exist yet. In that case the same predicate may be an expectedPredicate in multiple layers. A similar case is when the sh:values computation has changed in between layers.

So I'll move this into each layer for now. This may not be the final word though - potentially this recomputation could happen for every rule, but that could become a performance issue.

shacl12-sparql/index.html
afs: What happens if a rule adds the same triple - is it kept or removed?

HolgerKnublauch: I have no strong opinion, but for now made these edits (to address both questions)

https://github.com/w3c/data-shapes/commit/ac91d970371ac4acb3995fceef69f595f9a9f481

HolgerKnublauch:
Now ready for general review as the base branch has been merged into main and current feedback has been applied.

mgberg (approved):

liviorobaldo (approved):

PR #1122, opened 2026-08-06
#1105: Custom SPARQL functions

### LINKED ISSUE #1105: Support user-defined SPARQL functions
> Working group : 
> Are SHACL SPARQL-based functions planned for SHACL 1.2?
> Or are they superseded by node expressions?
>  

 _Originally posted by @afs in [#999](https://github.com/w3c/data-shapes/issues/999#issuecomment-5118265483)_

I am wondering about the best way forward, too. The current solution feels unfinished:

There is sh:Function vocabulary still in the TTL file

```
sh:Function
	a rdfs:Class ;
	rdfs:label "Function"@en ;
	rdfs:comment "The class of SHACL functions."@en ;
	rdfs:subClassOf sh:Parameterizable ;
	rdfs:isDefinedBy sh: .

sh:returnType
	a rdf:Property ;
	rdfs:label "return type"@en ;
	rdfs:comment "The expected type of values returned by the associated function."@en ;
	rdfs:domain sh:Function ;
	rdfs:range rdfs:Class ;
	rdfs:isDefinedBy sh: .

sh:SPARQLFunction
	a rdfs:Class ;
	rdfs:label "SPARQL function"@en ;
	rdfs:comment "A function backed by a SPARQL query - either ASK or SELECT."@en ;
	rdfs:subClassOf sh:Function ;
	rdfs:subClassOf sh:SPARQLAskExecutable ;
	rdfs:subClassOf sh:SPARQLSelectExecutable ;
	rdfs:isDefinedBy sh: .
```

This has been in production in our product for years and works quite well.

But now we also have https://www.w3.org/TR/shacl12-node-expr/#custom-list-parameter-functions as an alternative syntax for functions that can be used within node expressions. Those COULD have a sh:select as their body and essentially have the same interpretation for SPARQL functions.

A difference is that sh:Function relies on Parameterizable which means that it is expected to have the arguments as direct property values, not in an RDF list like for node expressions. Node expressions OTOH have no syntax yet to declare parameters at all.

My gut feeling is to let sh:Function expire (we'll continue to have them in our product) and add a section to the SPARQL spec on how to interpret node expressions as SPARQL functions. This would have the benefit that functions could be written using the general SHACL node expressions framework and not only with a SPARQL body. But we should have a proper parameter vocab for them.

Discussion and input welcome.

HolgerKnublauch:
Closes #1105

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/shacl12-sparql/issue-1105/index.html#sparql-functions)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1514,6 +1514,8 @@ ex:AnnotationExample
 						variable that is projected from the <code>SELECT</code> clause when the query is evaluated against the <a>focus graph</a>.
 						The value of <code>focusNode</code> is <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
 						The value of each scope variable is <a>pre-bound</a> as a SPARQL variable with the same name and value.
+						If the name of a variable in scope is not a string literal, use <code>"arg" + str(name)</code>.
+						For example, if the variable name is <code>"0"^^xsd:integer</code>, use <code>arg0</code>.
 						A <a>failure</a> is produced when one of the scope variables is called <code>this</code>.
 						<br/>
 						<br/>
@@ -1699,6 +1701,8 @@ ex:Bernd
 						when the query is evaluated against the <a>focus graph</a>.
 						The value of <code>focusNode</code> is <a>pre-bound</a> as the value of the SPARQL variable <code>this</code>.
 						The value of each scope variable is <a>pre-bound</a> as a SPARQL variable with the same name and value.
+						If the name of a variable in scope is not a string literal, use <code>"arg" + str(name)</code>.
+						For example, if the variable name is <code>"0"^^xsd:integer</code>, use <code>arg0</code>.
 						A <a>failure</a> is produced when one of the scope variables is called <code>this</code>.
 						<br/>
 						<br/>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1858,9 +1858,12 @@ WHERE {
 				</p>
 				<p>
 					Evaluate each SPARQL expression from <code>args</code>, yielding a new list of <a>nodes</a> <code>nodesList</code>.
-					Where the evaluation caused an <a data-cite="sparql12-query#sparql-error">error</a>,
+					Where the evaluation has caused an <a data-cite="sparql12-query#sparql-error">error</a>,
 					the result of the SPARQL function call is also an error,
 					unless the corresponding <code>sh:parameter</code> for the argument has <code>sh:optional true</code>.
+					In that case, the argument will be unbound in the scope.
+				</p>
+				<p>
 					Let <code>rs</code> be the <a data-cite="shacl12-node-expr#dfn-output-nodes">output nodes</a>
 					of <code>exprEval(function, focusGraph, f, scope)</code> where <code>focusGraph</code> is the
 					currently active query graph from the SPARQL context, and <code>scope</code> is the map of evaluated
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1907,13 +1907,15 @@ ex:langLabelCount
     rdfs:label "lang label count"@en ;
     rdfs:subClassOf sh:ListParameterExpression ;
     sh:bodyExpression [
-		sh:sparql """
+		sh:select """
+			PREFIX ex: &lt;http://example.com/ns#&gt;
+			PREFIX rdfs: &lt;http://www.w3.org/2000/01/rdf-schema#&gt;
 			SELECT (COUNT(?label) AS ?result)
 			WHERE {
 				?predicate rdfs:subPropertyOf* rdfs:label .
 				$arg0 ?predicate ?label .
 				BIND (COALESCE($arg1, 'en') AS ?matchLang) .
-				FILTER langMatches(?matchLang) .
+				FILTER langMatches(lang(?label), ?matchLang) .
 			}
 		""" ;
 	] ;
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1862,7 +1862,7 @@ WHERE {
 					the result of the SPARQL function call is also an error,
 					unless the corresponding <code>sh:parameter</code> for the argument has <code>sh:optional true</code>.
 					Let <code>rs</code> be the <a data-cite="shacl12-node-expr#dfn-output-nodes">output nodes</a>
-					of <code>exprEval(function, focusGraph, f, scope)</code> where <code>focusGraph</code> is the
+					of <code>evalExpr(function, focusGraph, f, scope)</code> where <code>focusGraph</code> is the
 					currently active query graph from the SPARQL context, and <code>scope</code> is the map of evaluated
 					arguments from <code>nodesList</code> with the index of the argument as the variable name (map key).
 				</p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -1791,23 +1791,75 @@ ex:Resource-uriLength
 			<p>
 				<a data-cite="shacl12-node-expr"></a> defines <a data-cite="shacl12-node-expr#custom-list-parameter-functions">Custom List Parameter Functions</a>
 				as a mechanism to declare new node expression functions.
-				These can be used as part of other node expressions.
-				This section introduces a mechanism for SPARQL processors to make such list parameter functions
-				available as SPARQL functions, using the <a data-cite="sparql12-query/#extensionFunctions">Extensible Value Testing</a>
-				extension point of SPARQL.
+				These can be evaluated as part of other SHACL node expressions, but may also be useful to other engines
+				that support function-like execution.
 			</p>
-			<aside class="example" title="An example custom list parameter expression function">
+			<p>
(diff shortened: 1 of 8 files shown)


shacl12-sparql/index.html
```diff
@@ -1778,6 +1786,99 @@ <h3>SPARQL Expr Expressions</h3>
...
+				as a mechanism to declare new node expression functions.
+				These can be used as part of other node expressions.
+				This section introduces a mechanism for SPARQL processors to make such list parameter functions
+				available as SPARQL functions, using the <a data-cite="sparql12-query/#extensionFunctions">Extensible Value Testing</a>
+				extension point of SPARQL.
+			</p>
+			<aside class="example" title="An example custom list parameter expression function">
+				<p>
+					This function takes two parameters and uses a <a>SPARQL expr expression</a> as its body expression.
+					In that SPARQL expression, the variables `$arg0` and `$arg1` are used to access the list parameter values.
```
scotthenninger: It's not clear how $arg0 and $arg1 get bound to the variables passed to the function. I understand that the order the variables are called in e.g. ex:spacedConcat('John', 'Doe'), have an impact on the order. But what if the variables are named $x and $y - or is only $arg{n} allowed. I think we should be explicit here about how the binding happens.

scotthenninger (commented):
Consider clarifying how function parameter ordering id defined.

shacl12-sparql/index.html
```diff
@@ -1778,6 +1786,99 @@ <h3>SPARQL Expr Expressions</h3>
...
+					Let <code>function</code> be the corresponding custom list parameter function in the <a>graph</a>
+					where it has been defined, with <code>f</code> as its <a>IRI</a>.
+				</p>
+				<p>
+					Evaluate each SPARQL expression from <code>args</code>, yielding a new list of <a>nodes</a> <code>nodesList</code>.
+					Where the evaluation caused an <a data-cite="sparql12-query#sparql-error">error</a>,
+					the result of the SPARQL function call is also an error,
+					unless the corresponding <code>sh:parameter</code> for the argument has <code>sh:optional true</code>.
+					Let <code>rs</code> be the <a data-cite="shacl12-node-expr#dfn-output-nodes">output nodes</a>
+					of <code>exprEval(function, focusGraph, f, scope)</code> where <code>focusGraph</code> is the
```
mgberg: Is this supposed to be
```suggestion
					of <code>evalExpr(function, focusGraph, f, scope)</code> where <code>focusGraph</code> is the
```

shacl12-sparql/index.html
```diff
@@ -1778,6 +1786,99 @@ <h3>SPARQL Expr Expressions</h3>
 			</section>
 		</section>
 
+		<section id="sparql-functions">
+			<h2>Declaring SPARQL Functions based on Node Expressions</h2>
+			<p>
+				<a data-cite="shacl12-node-expr"></a> defines <a data-cite="shacl12-node-expr#custom-list-parameter-functions">Custom List Parameter Functions</a>
+				as a mechanism to declare new node expression functions.
```
mgberg, line 1793: I only see List Parameter Functions discussed here. Should Named Parameter Functions also be available as SPARQL functions? Or are they explicitly excluded for some reason, e.g. because their args aren't constrained to evaluate to a single value?

shacl12-sparql/index.html
```diff
@@ -1778,6 +1786,99 @@ <h3>SPARQL Expr Expressions</h3>
...
+					Let <code>f</code> be the <code>iri</code> of a SPARQL <a data-cite="sparql12-query#ririOrFunction">function call</a>
+					and <code>args</code> be the list of <code>Expression</code> arguments.
+					Let <code>function</code> be the corresponding custom list parameter function in the <a>graph</a>
+					where it has been defined, with <code>f</code> as its <a>IRI</a>.
+				</p>
+				<p>
+					Evaluate each SPARQL expression from <code>args</code>, yielding a new list of <a>nodes</a> <code>nodesList</code>.
+					Where the evaluation caused an <a data-cite="sparql12-query#sparql-error">error</a>,
+					the result of the SPARQL function call is also an error,
+					unless the corresponding <code>sh:parameter</code> for the argument has <code>sh:optional true</code>.
```
mgberg: If a parameter is optional and errors, what gets sent in for that parameter?

shacl12-sparql/index.html
mgberg, line 1793: https://github.com/w3c/data-shapes/issues/1105#issue-5014334387 says you're thinking about retiring `sh:Function`- is that why SPARQL-based SHACL-AF-style functions using `sh:select` are not included here either? Presumably the new select expressions would make that older function type redundant.

shacl12-sparql/index.html
```diff
@@ -1778,6 +1786,99 @@ <h3>SPARQL Expr Expressions</h3>
...
+				If a function with the same <a>IRI</a> is already registered, SHACL engines MUST ignore the
+				attempt to redefine it unless the function was previously added as a custom SPARQL function.
+				A SPARQL engine MUST NOT modify the registered SPARQL functions while any SPARQL query is executing.
+			</p>
+			<div class="def" id="SPARQLFunctions-evaluation">
+				<div class="def-header">EVALUATION OF CUSTOM SPARQL FUNCTIONS</div>
+				<p>
+					Let <code>f</code> be the <code>iri</code> of a SPARQL <a data-cite="sparql12-query#ririOrFunction">function call</a>
+					and <code>args</code> be the list of <code>Expression</code> arguments.
+					Let <code>function</code> be the corresponding custom list parameter function in the <a>graph</a>
```
mgberg: Is this supposed to be
```suggestion
					Let <code>function</code> be the corresponding custom list parameter function in the <a>shapes graph</a>
```

shacl12-sparql/index.html
mgberg: Would it be sufficient to add to the EVALUATION OF CUSTOM SPARQL FUNCTIONS section some more detail about how `scope` is constructed?

shacl12-sparql/index.html
HolgerKnublauch: Named Parameter Functions cannot be mapped to SPARQL functions because they can take lists as input parameters and also produce lists of output nodes. They would require something like magic properties/multifunctions but such things don't exist in SPARQL.

shacl12-sparql/index.html
HolgerKnublauch: It would become an unbound scope variable. This is a bit like COALESCE, bound and similar SPARQL features that tolerate unbound variables. From my understanding, SPARQL's extension point does allow such functional forms to have this behavior. I'll clarify this in the text.

shacl12-sparql/index.html
HolgerKnublauch: Exactly. The new mechanism is more general and allows any node expression, including sh:select expressions to be used as body. We should add more examples of such functions, starting with test cases.

(Editorial: due to the limited remaining time to get the normative features into the spec, I am not adding as many test cases as I should. I think this can be done later and then we can also clarify aspects of the main docs once we have created more tests).

shacl12-sparql/index.html
HolgerKnublauch: I had "shapes graph" there first but then generalized it because the spec does not explicitly state where the functions are coming from. They are usually shapes graphs, but could also be rules graphs for example. The term shapes graph is just one role that such graphs typically play. If that's not good enough, let me know but I'll mark this as resolved for house keeping purposes.

shacl12-sparql/index.html
HolgerKnublauch: Yes this is a tricky topic. The complication here is that this is going through various levels of indirection, where a SPARQL query calls a SPARQL function and that function may call any node expression including SPARQL-based node expressions. 

Now, the general mechanism to pass in variables into a node expression is through the scope, but SPARQL-based node expressions have their own mechanism on how to map the scope vars to SPARQL variables. This PR actually clarifies that: the scope has variables such as 0 and 1 but these are not valid SPARQL var names, so I prepend "arg" yielding $arg0 etc.

I agree this needs to be expanded upon, and there should be enough examples so people and agents end up just doing copy and paste without having to understand all these complexities.

HolgerKnublauch: I added more examples and prose here, please recheck if anything remains unclear. Marking this resolved for now.

HolgerKnublauch:
@simonstey @robert-david 

Note this PR also includes a slight generalization of shnex:instancesOf in the Node Expression document so that the value of shnex:instancesOf is not limited to URIs but can be a node expression by itself, to produce the classes. This allows users to pass in the class as argument and otherwise dynamically compute that class. I don't remember why this was limited to URIs until now, but it feels counter intuitive and I bumped into this issue myself here. Could you kindly take a look if you see problems with that change?

mgberg:
Looks good to me, provided the above comment about `shnex:instancesOf` is cleared by the other node expressions editors.

HolgerKnublauch:
> Looks good to me, provided the above comment about `shnex:instancesOf` is cleared by the other node expressions editors.

Awesome, I would then just need an official approval by someone. We can always use follow-up PRs to address bugs etc.

The remaining open PRs are the last new features that I was hoping to get in before "code freeze". And FWIW I will be offline this coming week.

mgberg (approved):

PR #1124, opened 2026-08-06
Blank nodes are pattern variables

afs:
* [See the section rendered online here](https://raw.githack.com/w3c/data-shapes/pattern-blank-nodes/shacl12-rules/index.html#eval-rule)


recalcitrantsupplant (approved):
looks good to me

Issue #1094, opened 2026-07-23
Editorial comments on SRL document


afs:
From #1088

- [ ] 2.1 Version Labels section position is weird
- [ ] 4.3.2 Dependency Graph Algorithm: the algorithm is unreadable in dark mode
- [ ] The function EBV(x) returns the effective boolean value for an RDF term. line is not close to the algorithm using it
- [ ] In section 6.4, I think the if SEQ1 is empty condition is an optimization not useful for the algorithm validity

PR #1125, opened 2026-08-07
#1094: Fix dark mode contrast of algorithm blocks

recalcitrantsupplant:
Editorial item from #1094: the algorithm pseudocode is unreadable in dark mode.
Was able to reproduce the issue with system dark mode.

I'm not familiar with css for darkmode. A two line LLM change seems to resolve this. Tested on chrome and firefox.

`.darkmode .algorithm` set a `#808080` background. The highlight.js dark palette
is designed for a near-black background, so its text sits at 1.85:1 against
mid-grey; WCAG AA requires 4.5:1.

Darkened to `#2b2b2b` — the dark background already used by `.syntax` — with the
border adjusted to match. Measured after: 6.6:1. Affects all seven algorithm
blocks; light mode unchanged.

Does not close issue #1094; the other three items are untouched.

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -129,8 +129,8 @@
           padding-right:     0.5em;
       }
       .darkmode .algorithm {
-          background:        #808080 ;
-          border-color:      #aaaaaa ;
+          background:        #2b2b2b ;
+          border-color:      #55595f ;
       }
 
       /* Block to go inside <aside> "example" */


recalcitrantsupplant:
FYI @Tpt - I can't assign you as a reviewer - are you able to confirm the fix

afs:
Edited description so it does not contain the string "closes #nnnn" because that seems to be all that GH looks for, and ignores the "Does not". A problem with negation!

afs (approved):

afs:
* [See the document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1094-algorithm-darkmode/shacl12-rules/index.html#eval-rule)

Tpt (approved):
Thank you! Looks much better

PR #1132, opened 2026-08-10
[UI] #1083: Define profiles as compositions of concepts in SHACL UI

### LINKED ISSUE #1083: How open/closed should the SHACL UI specification be?

bergos:
Introduces a "Profiles" section that lets an application or ecosystem bundle a selection of the spec's concepts (SHACL Renderer, Scoring System, Label and Language Resolution, Grouping/Ordering/Layout Hints, Built-in Widgets) and optionally tighten their requirements, illustrated by an example "Very strict profile". To make room for profile-level tightening, the base tie-breaking and ordering requirements are weakened from MUST to SHOULD/MAY.

Closes #1083

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/profiles/shacl12-ui/index.html)


robert-david:
Some comments on the profiles proposal.
- Will there be a vocabulary to express the profiles? this might be necessary for portability.
- The term 'concept' is commonly used in the context of SKOS. This might be confusing. I am ok with keeping 'concept' if we don't find a better fit, though.
- If every concept is optional, we do not have any portability if no MUST is used and if the profile is not formalised (see above).

PR #1134, opened 2026-08-10
Remove redundant dfn attributes in shacl12-ui

bergos:
Respec automatically generates the plural linking text, so data-lt values that were just the term with an "s" appended are dropped. The data-plurals, id, tabindex, aria-haspopup, and data-dfn-type attributes on the SHACL Renderer, Node UI Component, and Property UI Component dfns were leftover copies from the render version and are not needed here.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/dfn-cleanup/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -258,11 +258,11 @@
         </dl>
         <p>
           The SHACL &amp; RDF terms include:
-          <dfn data-lt="bindings">
+          <dfn>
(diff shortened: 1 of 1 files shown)


danielbeeke:
Would it make sense to have this pull request merge to `gh-pages`? Or is it already based on some other branch. The changes I see are good for `gh-pages`, is that a correct observation?

bergos:
Superseded by https://github.com/w3c/data-shapes/pull/1135

PR #1138, opened 2026-08-11
[SRL] Fix reSpec validation errors

afs:
Fix errors in valifation from the GH action

"Duplicate definition(s) of 'shape rules language document'"

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/srl-validation/shacl12-rules/index.html)

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -298,23 +298,20 @@
 
       <p>This specification defines conformance criteria for:</p>
       <ul>
-        <li><a href="#rule-set-evaluation">SHACL Rule Set evaluation</a></li>
-        <li><a href="#shape-rules-syntax">Shape Rules Language</a></li>
+        <li><a href="#rule-set-evaluation">SRL Rule Set evaluation</a></li>
+        <li><a href="#shape-rules-syntax">Shape Rules language syntax</a></li>
       </ul>
 
-      <p>A conforming 
-        <dfn 
-          data-lt="shape rules document|ruleset document"
-          >Shape Rules Language Document</dfn> is an
+      <p>A conforming [=SRL document=] is an
         <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a> that
-        conforms to the grammar, starting with the
+        conforms to the <a href="#grammar">grammar</a> starting with the
         <a href="#rRuleSet"><code>RuleSet</code></a>
-        production, and conforming to the additional constraints defined in
-        <a href="#shapes-rules-grammar" class="sectionRef"></a>
+        production as defined in
+        <a href="#shapes-rules-grammar" class="sectionRef"></a>.
       </p>
 
       <p class="note">
-        This specification does not define how SHACL Rules
+        This specification does not define how SPARQL-RL
         processors handle non-conforming [=rule sets=].
       </p>
     </section>
@@ -2088,7 +2085,7 @@ PREFIX srl:    &lt;http://www.w3.org/ns/shacl-rules#&gt;
 
         <p>
           Multiple <a href="#rVersionDecl"><code>VERSION</code></a> directives
-          may appear in a [=Shape Rules Language Document=].
+          may appear in a [=SPARQL-RL Document=].
           Each directive applies to the part of the document following the directive,
           until another directive is encountered or the end of the document is reached.
         </p>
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```suggestion
          There are three forms of escapes used in a [=Shape Rules language document=]:
```

```suggestion
```

robert-david (changes requested):
There was still a ReSpec error shown for me in the rendered doc. Tried to fix it.

shacl12-rules/index.html
```diff
@@ -2941,9 +2938,11 @@ <h4>Evaluating the Strata</h4>
       <h2>Shape Rules Language Grammar</h2>
 
       <p>
-        A <dfn data-lt="SRL document">Shape Rules Language document</dfn>
+        A <dfn data-lt="SRL document">Shape Rules language document</dfn>
         is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>
-        encoded in UTF-8 [[!RFC3629]].
+        encoded in UTF-8 [[!RFC3629]] and starting with the
+        <a href="#rRuleSet"><code>RuleSet</code></a>
+        production.
```
simonstey: this here now says: 

``` 
a Shape Rules language document *is* an RDF string encoded in UTF-8 and starting with the `RuleSet` production
```
while in https://github.com/w3c/data-shapes/pull/1138/changes#diff-4cb699e56152bee095a8650906565f110dd7ed444551f33a7b1987a0a92fd214R305-R311 it says 

```
a *conforming* Shape Rules language document is an RDF string that conforms to the grammar starting with the `RuleSet` production
```

so which statement is now the normative conformance criterion?

(the RDF version keeps the encoding requirement in the definition and the grammar requirement in the conformance clause)

so what about changing this here to:

```html
 A <dfn data-lt="SRL document|Shape Rules document">Shape Rules Language document</dfn>
        is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>
        encoded in UTF-8 [[!RFC3629]].
```

and tighten the conformance clause at 305-311 to:

```html
      <p>A conforming [=Shape Rules Language document=] is an
        <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a> that
        conforms to the  <a href="#grammar" class="sectionRef"></a> starting with the
        <a href="#rRuleSet"><code>RuleSet</code></a>
        production and satisfies the additional constraints defined in
        <a href="#shapes-rules-grammar" class="sectionRef"></a>.
      </p>
```

shacl12-rules/index.html
```diff
@@ -298,23 +298,20 @@ <h3>Document Conventions</h3>
...
+        conforms to the <a href="#grammar">grammar</a> starting with the
         <a href="#rRuleSet"><code>RuleSet</code></a>
         production, and conforming to the additional constraints defined in
-        <a href="#shapes-rules-grammar" class="sectionRef"></a>
+        <a href="#shapes-rules-grammar" class="sectionRef"></a>.
       </p>
 
       <p class="note">
-        This specification does not define how SHACL Rules
+        This specification does not define how Shape Rules
```
simonstey: in various places throughout the document, such as https://github.com/w3c/data-shapes/pull/1138/changes#diff-4cb699e56152bee095a8650906565f110dd7ed444551f33a7b1987a0a92fd214R179, https://github.com/w3c/data-shapes/pull/1138/changes#diff-4cb699e56152bee095a8650906565f110dd7ed444551f33a7b1987a0a92fd214R188, (211,220,224,...) we still refer to SHACL Rules. in the abstract we say "This document defines SHACL Rules", a

=> so what are Shape Rules processors now? 

what about:

```suggestion
        This specification does not define how SRL 
```

shacl12-rules/index.html
simonstey: this puts a spec-plus-anchor value in data-lt, where data-cite was intended. 

(ReSpec takes data-lt on an <a> as the lookup term, so it will search for a local definition titled `sparql12-query#defn_sparqlsolutionmapping`, find none, and report an unresolved auto-link.

```suggestion
            A [=solution mapping=], 
```

simonstey (approved):
approved when changes are applied

shacl12-rules/index.html
afs: @robert-david Fixed.

This PR does not attempt complete renaming in the HTML to the latest name of SRL. The working group has not completed the process of name change yet.  The document does use the proposed name in some places as changes are made.

shacl12-rules/index.html
afs: Change made (to the proposed full name - this the the conformance section).

There will need to be a PR to do systematic renaming after the WG votes on the short name etc.

shacl12-rules/index.html

shacl12-rules/index.html
afs: The `<dfn>` is the one around L2945.

At L305 there is a link to that definition. 
I think around L305 needs to say a little more than "see below".

I've changed the text arounf L305 to refer the to grammar section, hence it includes the detailed conditions.

It also says: "... as defined in 7. Shape Rules Language Grammar."

Nearby: The grammar table now enforces "Variables are not allowed in a `DATA` block." but it is worth mentioning still.

robert-david (approved):

PR #1154, opened 2026-08-14
Add SRL evaluation tests

afs:
Some more tests for evaluation and a few more syntax tests.


afs:
If it is more practical for implementers, then this PR could be merged, and then new issues/PRs raised later as implementations catch up with new tests.

recalcitrantsupplant (approved):
will provide any feedback based on implementing

PR #1158, opened 2026-08-18
GH#1157: SHACL Rules: Add rule sets to RDF vocabulary

### LINKED ISSUE #1157: SHACL Rules: Add rule sets to RDF vocabulary
It may be useful to group rules into sets that are supposed to be executed together. SRL relies on that concept.

Suggestion:

```
ex:Rule1
    a sh:SPARQLRule ;
    sh:ruleSet ex:RuleSet1 .

ex:RuleSet1
    a sh:RuleSet ;
    sh:importsRuleSet ex:RuleSet2 .

ex:RuleSet2
    a sh:RuleSet . 
```

The naming is chosen to be consistent with SRL which has an IMPORTS keyword and owl:imports.

We point from Rules to RuleSets similar to how property shapes point to property groups. Rules may be in multiple sets.

There is an unnamed default rule set per graph which consists of all rules that do not have a sh:ruleSet triple.

HolgerKnublauch:
Closes #1157

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1157/shacl12-sparql/index.html#ruleSet)

CC @afs

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2133,13 +2133,13 @@ ex:ExampleRectangle ex:area 56 .
 					<p>
 						The input to a SHACL rules processor is a set of rules called a <dfn>rule set</dfn>.
 						A <a>rule set</a> is identified by an <a>IRI</a>.
-						The property <code>sh:memberRule</code> can be used to declare that a <a>rule set</a> has a given <a>rule</a> as a member.
+						The property <code>sh:hasRule</code> can be used to declare that a <a>rule set</a> has a given <a>rule</a> as a member.
 						The <dfn>default rule set</dfn> of a <a>graph</a> is the set of all <a>rules</a> in the graph.
 						Rule sets can use the property <code>sh:includesRuleSet</code> to (transitively) include other rule sets.
 					</p>
 					<p class="syntax">
 						<span data-syntax-rule="RuleSet-nodeKind">All <a>SHACL instances</a> of <code>sh:RuleSet</code> have a <a>IRI</a>.</span>
-						<span data-syntax-rule="memberRule"><a>Rule sets</a> can have <a>values</a> for <code>sh:memberRule</code> and those values are <a>rules</a>.</span>
+						<span data-syntax-rule="hasRule"><a>Rule sets</a> can have <a>values</a> for <code>sh:hasRule</code> and those values are <a>rules</a>.</span>
 						<span data-syntax-rule="includesRuleSet-nodeKind">The <a>values</a> of <code>sh:includesRuleSet</code> at a <a>rule set</a> are <a>IRIs</a>.</span>
 					</p>
 					<aside class="example" title="Example of rules and rule sets" id="example-ruleset">
@@ -2156,11 +2156,11 @@ ex:Rule2
 ex:RuleSet1
 	a sh:RuleSet ;
 	sh:includesRuleSet ex:RuleSet2 ;
-	sh:memberRule ex:Rule1 .
+	sh:hasRule ex:Rule1 .
 
 ex:RuleSet2
 	a sh:RuleSet ;
-	sh:memberRule ex:Rule2 .
+	sh:hasRule ex:Rule2 .
 							</div>
 						</div>
 						<p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2133,40 +2133,39 @@ ex:ExampleRectangle ex:area 56 .
 					<p>
 						The input to a SHACL rules processor is a set of rules called a <dfn>rule set</dfn>.
 						A <a>rule set</a> is identified by an <a>IRI</a>.
-						The property <code>sh:ruleSet</code> can be used to declare that a <a>rule</a> is member of a <a>rule set</a>.
+						The property <code>sh:memberRule</code> can be used to declare that a <a>rule set</a> has a given <a>rule</a> as a member.
 						The <dfn>default rule set</dfn> of a <a>graph</a> is the set of all <a>rules</a> in the graph.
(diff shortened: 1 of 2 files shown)


shacl12-sparql/index.html
```diff
@@ -2128,6 +2128,63 @@ <h2>General Definition of SHACL Rules</h2>
...
+					<p>
+						The input to a SHACL rules processor is a set of rules called a <dfn>rule set</dfn>.
+						A <a>rule set</a> is identified by an <a>IRI</a>.
+						The property <code>sh:ruleSet</code> can be used to declare that a <a>rule</a> is member of a <a>rule set</a>.
+						The <dfn>default rule set</dfn> of a <a>graph</a> is the set of <a>rules</a> in the graph that do not have
+						any <a>value</a> for <code>sh:ruleSet</code>, or whose only value for <code>sh:ruleSet</code> is
+						<code>sh:DefaultRuleSet</code>.
+					</p>
+					<p>
+						Rule sets can use the property <code>sh:includedRuleSet</code> to include other rule sets.
```
```suggestion
						Rule sets can use the property <code>sh:includesRuleSet</code> to include other rule sets.
```
otherwise it can be read as "included rule set (of)" which is the inverse relationship.

shacl12-sparql/index.html
```diff
@@ -2128,6 +2128,63 @@ <h2>General Definition of SHACL Rules</h2>
...
 				</p>
+				<section id="ruleSet">
+					<h3>Rule Sets</h3>
+					<p>
+						The input to a SHACL rules processor is a set of rules called a <dfn>rule set</dfn>.
+						A <a>rule set</a> is identified by an <a>IRI</a>.
+						The property <code>sh:ruleSet</code> can be used to declare that a <a>rule</a> is member of a <a>rule set</a>.
+						The <dfn>default rule set</dfn> of a <a>graph</a> is the set of <a>rules</a> in the graph that do not have
+						any <a>value</a> for <code>sh:ruleSet</code>, or whose only value for <code>sh:ruleSet</code> is
+						<code>sh:DefaultRuleSet</code>.
```
mgberg: So it would not be allowable for the following to exist then?
```turtle
ex:SomeRule
    sh:ruleSet ex:RuleSet1 , sh:DefaultRuleSet .
```

shacl12-sparql/index.html
```diff
@@ -2635,7 +2693,7 @@ <h3>General Execution Instructions for SHACL Rules</h3>
 				</p>
 				<div class="algorithm">
 					<pre class="text">
-						For all <a>layers</a> in the rule set (in ascending order):
+						For all <a>layers</a> in the <a>rule set</a> (in ascending order):
```
mgberg, line 2680: Should the algorithm include the ruleset building step, e.g. finding the transitive closure of `sh:includesRuleSet` if other than the default is specified?

shacl12-sparql/index.html
```diff
@@ -2128,6 +2128,63 @@ <h2>General Definition of SHACL Rules</h2>
...
+						A <a>rule set</a> is identified by an <a>IRI</a>.
+						The property <code>sh:ruleSet</code> can be used to declare that a <a>rule</a> is member of a <a>rule set</a>.
+						The <dfn>default rule set</dfn> of a <a>graph</a> is the set of <a>rules</a> in the graph that do not have
+						any <a>value</a> for <code>sh:ruleSet</code>, or whose only value for <code>sh:ruleSet</code> is
+						<code>sh:DefaultRuleSet</code>.
+					</p>
+					<p>
+						Rule sets can use the property <code>sh:includesRuleSet</code> to include other rule sets.
+						<code>sh:DefaultRuleSet</code> is a reserved <a>IRI</a> that can be used as a <a>subject</a> of
+						<a>triples</a> with the <a>predicate</a> <code>sh:includesRuleSet</code> to include other rule sets into the <a>default rule set</a>.
```
mgberg: I suppose this could be a reason why 
```
ex:SomeRule
    sh:ruleSet ex:RuleSet1 , sh:DefaultRuleSet .
```
would not be allowable.

Might it be good to fold this sentence (at least the part regarding allowing `sh:DefaultRuleSet` be a subject of `sh:includesRuleSet`) into the definition of default rule set in the prior paragraph?

HolgerKnublauch:
I have just simplified the notion of the default rule set. My original idea was too complex and caused all kinds of non-intuitive cases. I had thought it would be good to allow the default rule set to be disjoint from the other rule sets, and then needed to introduce sh:DefaultRuleSet as a way to explicitly include others. But in retrospect I think it would be closer to the user expectation to simply count *all* rules in the graph as the default rule set.  So sh:DefaultRuleSet is now gone and I hope all use cases are still covered: if anyone wants to run only a subset, they need to introduce a named sh:RuleSet for it.

(This is similar to how SPARQL stores have a notion of default graph: in some cases the default graph is a dedicated graph, while in others it is the union of all graphs. Here it is now the union.)

Let me know if this isn't working for anyone.

shacl12-sparql/index.html
HolgerKnublauch: See my comment below, this is indeed an ugly case that would be too difficult to explain. Simplified by making the default rule set the union of all rules.

shacl12-sparql/index.html
HolgerKnublauch: I hope that's not needed, because the section on rule sets already does that (and I added the term "transitive") to clarify this. If you think that's not enough, please add a specific diff suggestion to Apply.

shacl12-sparql/index.html
HolgerKnublauch: Yeah good question, but see my comment below: sh:DefaultRuleSet is gone now.

liviorobaldo (approved):

PR #1167, opened 2026-08-18
Define evalRule as a function

afs:
Final sweep before renaming the document.

* [See this section here](https://raw.githack.com/w3c/data-shapes/srl-evalRule/shacl12-rules/index.html#eval-rule)

Define `evalRule` as a function.

There are also some small fixes in the evaluation algorithms and some CSS rationalization.

highlight.js is switched off for "evaluation" algorithm blocks for the moment.
With the latest changes, highlight.js was making some weird choices of highlight coloring.
This now allows for HTML inside `<pre>` so `<sub>` and `<var>` work

### DIFF
--- a/shacl12-rules/index.html
+++ b/shacl12-rules/index.html
@@ -12,10 +12,8 @@
         github: "w3c/data-shapes",
         //preProcess:     [  ],
         edDraftURI: "https://w3c.github.io/data-shapes/shacl12-rules/",
-        shortName: "shacl12-rules",
+        shortName: "sparql12-rules",
         copyrightStart: "2025",
-        // Can remove SHACL12-*?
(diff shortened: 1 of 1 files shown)


shacl12-rules/index.html
```diff
@@ -2049,24 +2022,37 @@ <h3>Evaluation of a Rule</h3>
...
+define evalRule(R, G, GD):
+    where 
+        R is a well-formed rule
+        G and GD are RDF graphs
+    let B be R.body 
+        where each blank node in a triple pattern in R.body
+        is replaced by a variable which is not used in the rule. 
+        The same variable is used for each occurrence of the
+        same blank node, and a different variable is used for 
+       different blank node.
```
```suggestion
       each different blank node.
```

recalcitrantsupplant (approved):
looks good - grammar comment only

PR #1169, opened 2026-08-19
#1159: Added definition of sh:TripleRule

### LINKED ISSUE #1159: SHACL Rules: Add sh:TripleRules
The sh:TripleRule rule type was defined in SHACL 1.1 AF but is currently absent from 1.2.

https://www.w3.org/TR/shacl-af/#TripleRule

They are quite trivial to define and the new SHACL 1.2 Rules stand-alone document makes it possible to define new rule types that do not depend on SPARQL.

They can act as a bridge between rules and node expressions - some computations are easier to express as node expressions than in SPARQL strings.

The main simplification I would make compared to the 1.1 draft is to make $this the default value if subject, predicate or object are missing, because writing [ shnex:var "focusNode" ] is too verbose and noisy.

Example to compute a childCount property:

```
ex:Person
    a sh:ShapeClass ;
    sh:rule [
        a sh:TripleRule ;
        sh:predicate ex:childCount ;
        sh:object [
            shnex:count [
                 shnex:pathValues ex:child 
            ]
        ]
    ] .
```

which is an alternative to this derived property

```
ex:Person
    a sh:ShapeClass ;
    sh:property [
        a sh:PropertyShape ;
        sh:path ex:childCount ;
        sh:values [
            shnex:count [
                 shnex:pathValues ex:child 
            ]
        ]
    ] .
```

The difference is that the sh:rule counts as an inference that will always lead to inferred triples, and such rules would iterate, giving them more power than sh:values rules that are computed just once and usually not inferred as triples.

We would be clear in the Conformance section that not all implementations must support this feature.

Does anyone see problems with adding this rule type, @scotthenninger @mgberg @liviorobaldo @simonstey @nicholascar ?

HolgerKnublauch:
Closes #1159

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1159/shacl12-sparql/index.html#TripleRule)

I am aware that more is needed here, including test cases, but that will be much easier once the rules section was moved into its own document, as discussed. Until then, I suggest we put this here so that the document is reasonably feature-complete before the end of the month and the clean up that follows.

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2018,7 +2018,7 @@ ex:Cougar
 			<p>
 				The <a>SHACL rules</a> feature defined in this section includes a general framework using the properties
 				such as <code>sh:rule</code> and <code>sh:condition</code>, plus an extension mechanism for specific <a>rule types</a>.
-				This document defines one such rule type: <a>SPARQL rules</a>.
+				This document defines two such rule types: <a>SPARQL rules</a> and <a>triple rules</a>.
 			</p>
 			
 			<section id="rules-examples" class="informative">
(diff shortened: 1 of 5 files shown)


shacl12-sparql/index.html
```diff
@@ -2780,6 +2780,159 @@ <h3>SPARQL Rules</h3>
...
+					</tr>
+				</table>
+				<div class="def def-text">
+					<div class="def-header">EXECUTION OF TRIPLE RULES</div>
+					<div class="def-text-body">
+						Let <code>S</code>, <code>P</code> and <code>O</code> be the sets of <a>nodes</a> produced by evaluating
+						the <a>node expressions</a> that are the values of <code>sh:subject</code>, <code>sh:predicate</code>
+						and <code>sh:object</code> respectively at the <a>triple rule</a>.
+						Where <code>sh:subject</code>, <code>sh:predicate</code>, or <code>sh:object</code> are absent, use
+						the list consisting of the current <a>focus node</a>.
```
mgberg: What would happen if one of these were missing on a _global_ triple rule? Or should this be explicitly conditional on not being a global rule?

HolgerKnublauch: Good point, I have clarified that this is the empty list for global rules. It will produce no triples.

shacl12-sparql/index.html
mgberg: I was scanning through the rest of the document and found that this needed to be updated:

```suggestion
					The <a>SHACL instances</a> of <code>sh:Rule</code>, including those of its <a>subclasses</a>,
					are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
					SHACL has a flexible, extensible design in which multiple types of rules can be supported,
					and this document defines two of them: <a>SPARQL rules</a> and <a>triple rules</a>.
```

mgberg: I don't know why the suggestion doesn't seem to render correctly at least for me, but:
- The first line is altered to make it clear that other subclasses of `sh:Rule` beyond the one defined here are also SHACL Rules
- The last line is altered to include the fact that triple rules are being added.

HolgerKnublauch: Yes thanks. I have updated this text in this spirit. We need to explicitly spell out sh:SPARQLRule and sh:TripleRule because we want those to count as SHACL rules even if the triple sh:SPARQLRule rdfs:subClassOf sh:Rule is not present in the rules graph.

mgberg (approved):
Looks good to me

Issue #1165, opened 2026-08-18
[UI] Decision: should the specification define a rendering model separate from presentation?


smessie:
Split from #1083.

## Question

Should the specification define an abstract representation of what is to be rendered, distinct from how it is rendered, and use that representation as the basis for conformance testing?

Such a representation would contain the resolved outcome of applying shapes to data: which fields are present, their order and grouping, resolved labels, the selected widget for each, cardinality, and current values. It would say nothing about styling, layout, or platform controls.

## Use cases

- #1160: this is what makes "the same fields, labels and order" checkable rather than a matter of comparing screenshots
- #1142: most UI frameworks expect objects or arrays rather than triples, so an intermediate representation is useful independently of conformance testing

## Why this may help

@HolgerKnublauch raised in #1083 that verifying compliance across browsers, screen sizes, input methods and accessibility needs is very hard. That is true of rendered output. It is much less true of the model behind it.

Separating the two lets the specification be precise about the model while saying nothing at all about appearance, and it gives us something a test suite can compare. It also means non-visual implementations are not excluded.

@danielbeeke noted in the meeting that this should not become a requirement to expose a particular internal structure. The intent is a shared structure for testing behaviour, not a mandated implementation architecture.

## Open questions

- Whether the representation is normative or a testing artefact
- Whether it is required output of an implementation, or only something an implementation must be able to produce for conformance testing
- Serialisation, if any

danielbeeke:
_Should the specification define an abstract representation of what is to be rendered, distinct from how it is rendered, and use that representation as the basis for conformance testing?_

**Yes, I agree**

Maybe we can call it a conformance model.
I think all implementations of the taskforce MUST implement it.
Further I think we can only supply the conformance testing files and not require implementations to support this.
They can claim conformance but not proof it. I think, this is what happens anyway when software is proprietary.

bergos:
No. This is optional, and given our limited time, we need to prioritize more important topics.

danielbeeke:
Side note:

While thinking and discussing the conformance testing with @smessie we realized that if we test for outcomes instead of explicit widget implementations we can have the following conformance tests:

- Run conformance test with spec scoring graph and check if the algorithm for matching and ordering is correct.
- Run conformance test with scoring graph of implementations and check that an editor is returned for, for example, xsd:boolean.

smessie:
> No. This is optional, and given our limited time, we need to prioritize more important topics.

No worries about the timing; I can open a PR for that.
[...]

Issue #1164, opened 2026-08-18
[UI] Decision: what does a conformance claim guarantee, and do claims name what they cover?


smessie:
Split from #1083. The specification's conformance section is currently a TODO, and much of the rest depends on what goes in it.

## Questions

1. What does "conforms to SHACL UI" guarantee to a shape author or an adopting organisation?
2. Do we adopt profiles or declared features, and if so, must a conformance claim name the ones it meets?
3. Which profiles or features, if any, are required for a claim to be made at all?

## Use cases

- #1160: Moving a deployment to a different implementation, including the procurement dimension
- Application profiles the organisation does not control (see use case issue to be created by @danielbeeke)
- #1142: whether a build-time generator can make a conformance claim, and of what

## Proposed resolution

There was broad agreement in the task force meeting of 18 August on a direction, and I would like to write it down here so people can say whether they agree.

**1. Map the specification to features.** The specification defines a set of named features. Each feature covers a coherent piece of behaviour and is something an implementation can meaningfully say it does or does not support. Roughly: label and language resolution, ordering, widget matching, and so on. The exact set is work to be done, and doing that work is the next step.

**2. An implementation lists the features it implements.** This is what makes implementations comparable, and it's the part I'd most like to see agreed. An adopter can see what an implementation covers, and a shape author knows what to target.

**3. The specification designates which features are required for conformance.** Not every feature is required. The specification says which set an implementation must support before it can claim conformance at all.

**4. Start with features, and decide about profiles afterwards.** Rather than designing a profile mechanism now, I would suggest we define the features first and see how many there are and how they group. It is possible that one required set is all we need, in which case we may not need the term "profile" at all. If the features turn out to fall into natural groupings, we can name those then, with the evidence in front of us.

The thing I would ask us to settle regardless of how much structure we end up with: **a conformance claim says what it covers.** No unqualified "conforms to SHACL UI" that leaves an adopter unable to tell what they are getting.

Anything published as a feature set or profile should be a deliverable in the specification, subject to the same process and review as the rest of it.

## What this leaves open

- The feature set itself: what the features are and where the boundaries fall
- Which features are required for a conformance claim
- Whether we need more than one required set, and whether "profile" is the right term for that
- How an implementation records and publishes its feature support

danielbeeke:
I agree with proposed solution 1, 3 and 4.
I think we can ask 2 but I do not see a way of enforcing. We could have a simple checklist that makes it easy for implementations to publish this information.

[...]

Issue #1162, opened 2026-08-18
[UI] Decision: what behaviour is defined when `sh:order` is absent?


smessie:
Split from #1083.

## Question

Where `sh:order` is present, is an implementation required to respect it? Where it is absent, what does the specification say about the resulting order?

## Use cases

- Application profiles that the deploying organisation does not control (see use case issue to be created by @danielbeeke): published profiles frequently carry no `sh:order` at all, so the absent case is the common case, not the edge case
- Moving a deployment to a different implementation (see use case issue #1160)
- Varying field order per deployment (see use case issue #1161)
- #1142: build-time form generation from closed shapes

## Options

Three were presented in the task force meeting of 11 August.

**Option 1.** `sh:order` MUST be respected, plus a defined fallback algorithm when it is absent. Shapes without `sh:order` then render in the same order across conforming implementations. Innovation such as ordering by usage frequency or pinned fields remains possible, with an empty context used when testing conformance.

**Option 2.** `sh:order` MUST be respected, fallback behaviour left open. Guarantees hold to the extent that `sh:order` values are given. A global setting such as `sh:orderFallback` could express intent, defaulting to positive infinity as some DASH implementations do today.

**Option 3.** `sh:order` optional or informative, to allow fully custom dynamic sorting.

## Arguments

For a defined fallback: profiles that carry no ordering render unpredictably without it, which is the case that matters most for the application profile use case above.

Against: usage-based or context-dependent ordering depends on runtime data that is not in the shapes graph, so it cannot be made interoperable by specification text in any case.

## Proposed resolution

I would propose something between options 1 and 2, close to what @robert-david suggested in #1083: `sh:order` MUST be respected where it is given, and the specification describes an algorithm implementations SHOULD follow where it is absent, rather than leaving the absent case undefined.

Concretely, I propose the text currently in the specification following #1104, with one change:

- **`shui:defaultOrder` in the global configuration**, where present, MUST be respected. Unchanged from the current text.
- **Members with no effective order.** The current text says these MUST be placed after all members that have one. I propose changing that MUST to a SHOULD, so an implementation can apply a smarter default ordering of its own where it has reason to, while the described behaviour remains the one an author can expect by default. This is the change that accommodates the concerns raised against a fully defined fallback.
- **Tie-breaking.** Keep the current text as it stands: renderers MUST apply a deterministic tie-breaking algorithm, SHOULD apply the one described, and MAY use an alternative provided it yields deterministic results within that implementation.

[...]

Issue #1160, opened 2026-08-18
[UI] Use case: moving an existing deployment to a different SHACL UI implementation


smessie:
## Context

An organisation has invested in a set of shapes, either authored in-house or adopted from a published application profile, and uses them with one SHACL UI implementation. It later wants to move to a different implementation, for example because of procurement rules, cost, licensing, support, platform, or because the current supplier's product no longer fits.

## The need

The organisation wants its shapes to keep working. Its investment is in the shapes, and it should not have to re-do that work to change supplier.

Concretely, on moving to another conforming implementation:

- the same properties should appear
- labels should resolve the same way, including language selection
- the ordering it relies on should be preserved
- values already stored should still be editable, including for properties that hold several values

Where the new implementation cannot support something, the organisation needs to be able to find that out before migrating, not after.

## Related to procurement

This is also how conformance claims get used in practice. Public sector tenders reference specifications formally. If a tender says "must conform to SHACL UI", the evaluator needs to be able to compare bids against it. If two implementations can both conform and produce materially different forms from the same shapes, the requirement does not help the buyer distinguish them.

## What this implies for SHACL UI

A conformance claim needs to convey what an implementation actually guarantees, and different implementations claiming conformance need to agree on enough that shapes move between them.

This does not require that they look alike. It requires that the same shapes graph resolves to the same fields, labels, order and value handling.

PR #1173, opened 2026-08-19
[UI] Define the SHACL UI feature set

smessie:
Addresses the **first open point** of #1164 — *the feature set itself: what the features are and where the boundaries fall* — discussed in the 18 August task-force meeting and proposed by @edmondchuc as the next step.

This is a **proposal for discussion**, not settled text.

## What it does

Replaces the `TODO` in the Conformance section of `shacl12-ui/index.html` with:

- **Conformance defined in terms of named features.** An implementation records which features it supports, and a claim always states what it covers — no unqualified "conforms to SHACL UI".
- **Interpretation vs presentation.** Features constrain interpretation (which fields, order, grouping, labels, widget, value handling) and are defined so conformance is checkable from machine-readable output. Presentation (styling, layout) is unconstrained.
  This checkability depends on **#1165** — whether the specification defines an abstract rendering model, distinct from presentation, as the basis for conformance testing. If that model is not defined, the machine-readable output these features are checked against is not specified.
- **Cross-cutting contracts:** no silent omission, fallback on unknown widget IRIs, extensibility beyond any required set, and presentation left unconstrained.
- **A feature catalogue:** a single index table (number · feature · what supporting it means · specification sections · **suggested conformance level**). The authoritative behaviour of each feature stays in the linked sections; the table does not restate the normative requirements, so a section and its feature entry are not kept in sync separately (single source of truth).

## Granularity principle

One rule decides the boundaries: **a behaviour graded SHOULD or MAY gets its own feature**, so it is independently reportable and can be marked not-required, rather than being bundled with a stronger requirement. This lets an implementation's list of supported features tell a user precisely what to expect.
Property paths are split by **path kind** and by **view/edit mode**, so a read-only viewer and an editor report path support separately.

## What is in the document vs. in this PR

- **In the specification:** the feature index table (number, feature, behavioural description, section mapping, suggested-level column), the cross-cutting contracts, and editor's-note / issue-box markers for provisional or not-yet-defined areas. The behavioural definition of each feature lives in the linked sections, not in the table.
- **In this PR body:** the proposal framing, the granularity principle above, the boundary rationale below, and the open questions — the material meant for task-force discussion rather than the published draft.

## Proposed features (26)

| Feature | Suggested level |
|---|---|
| Language preference via `sh:languageIn` | Required |
| Application and browser language preference | Recommended |
| Cross-language fallback | Optional |
| Label resolution | Required |
| Local name humanization | Recommended |
| Direct role annotation | Required |
| Qualified role annotation | Recommended |
| Explicit `sh:order` | Required |
| Default order configuration (`shui:defaultOrder`) | Required |
| Unordered-member fallback | Recommended |
| Deterministic tie-breaking | Required |
| Standard tie-breaking algorithm | Recommended |
| Grouping (`sh:group`) | Required |
| Nested groups | Recommended |
| Built-in widgets | Required |
| Widget scoring | Required |
| Default widget selection | Recommended |
| Declared widgets (`shui:editor`/`shui:viewer`) | Required |
| Widget switching | Optional |
| Value preservation | Required |
| Predicate and inverse paths — view | Required |
| Predicate and inverse paths — edit | Required |
| Alternative paths — view | Recommended |
| Alternative paths — edit | Recommended |
| Complex paths — view | Recommended |
| Complex paths — edit | Optional |

## Boundary rationale

- **Language** is split into `sh:languageIn` (MUST → Required), application/browser preference (SHOULD → Recommended) and cross-language fallback (MAY → Optional), so an implementation without browser context can report only what it does.
- **Label resolution** keeps the property and value-node chains together (both MUST) but separates **local name humanization** (SHOULD).
- **Property roles** separate **direct** (Required) from **qualified** precedence (SHOULD, incl. the RDF 1.2 triple-annotation form).
- **Ordering** follows the normative lines of #1162: explicit `sh:order` and `shui:defaultOrder` (Required), unordered-member fallback (Recommended — #1162 proposes MUST → SHOULD, dependency noted in the draft), deterministic tie-breaking (Required) split from the specific label-then-identifier algorithm (Recommended).
- **Grouping** separates flat membership (Required) from nested groups (Recommended; the draft already makes nesting optional).
- **Widgets** separate scoring (Required) from default selection (SHOULD → Recommended), declared widgets (Required), and switching (MAY → Optional).
- **Value preservation** is distinct from the no-silent-omission contract: the contract covers properties an implementation cannot handle at all; the feature covers not dropping values for properties it does handle.
- **Property paths** are split by kind and mode so viewing and editing are reported independently.

## Open questions

- **Granularity.** How finely to divide the feature set is open. I favour a more granular set: the finer the features, the more clearly and precisely an implementation can state what it supports, which is the whole point of reporting features. This PR therefore leans fine-grained — property paths, for example, are split into six features by kind and by view/edit mode. The open question is whether we find any of these too fine and want them merged. Resolving it also informs the profiles question (#1164 third open point), since we can only see how features cluster once the set exists.
- **Required set.** The suggested levels are only suggestions (#1164 second open point).
- **Widget switching** is reported as a capability but is hard to check from output; it may be better treated as an affordance than a feature.
- **RDF 1.2 support** is a `SHOULD` (RDF Abstract Model Compatibility). It is kept in that section rather than added as a feature; whether it should also be a reportable feature is open.
- **Recording and publishing feature support** (#1164 fourth open point) is out of scope here.

Related: #1083, #1162, #1165, #1146, #1160, #1142, #823.

---

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/feature-set/shacl12-ui/index.html)


danielbeeke:
I like this very much, thanks for making (a start of) the list of features.

I think it is a good start!

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
       </section>
       <section id="conformance">
-        <p>TODO: define conformance criteria for SHACL UI implementations.</p>
+        <p>
+          A SHACL UI implementation is a <a>SHACL Renderer</a> as defined in
+          <a href="#renderer">SHACL Renderer</a>. Conformance is defined in terms of a set of named
+          <dfn data-lt="feature|features">features</dfn>, each covering a coherent piece of
+          behaviour that an implementation can state it does or does not support. An implementation
+          records which features it supports, and a conformance claim always states which features it
+          covers; there is no unqualified claim of conformance to SHACL UI.
```
danielbeeke: I am not sure about the last part of this sentence: "there is no unqualified claim of conformance to SHACL UI". I feel it wants to dictate how someone may claim conformance. I think in practice this is complicated. Lets say a tender requires conformance to SHACL UI. Should a company then put a check in the checkbox and state all the features? Or can they just put a check, which implicitly tells that the implementation at least conforms to all the required features?

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
+          how values are handled. Conformance to a feature is defined so that it can be checked by
+          comparing an implementation's machine-readable output. Features do not constrain
+          <dfn>presentation</dfn>, i.e., styling, layout, reflow, platform controls, touch targets and
+          animation are unconstrained by this specification.
+        </p>
+
+        <div class="issue" data-number="1164"></div>
+
+        <p class="ednote">
+          The feature set below is under discussion. The <em>suggested conformance level</em> of each
```
danielbeeke, line 669: great, this way we can start with this and add or fine tune this later.

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
+          feature (required, recommended, or optional) is a suggestion; which features are required
+          for a conformance claim, and how an implementation records and publishes its feature
+          support, are not yet settled.
+        </p>
+
+        <section id="feature-set-contracts">
+          <h4>Cross-cutting Contracts</h4>
+
+          <p>The following contracts apply to every feature and to every implementation.</p>
```
danielbeeke: The following contracts apply to every feature ~~and to every implementation~~.

Would this be better? I mean isn't it implied that any conforming implementation must adhere to the spec?

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
+          <ol>
+            <li>
+              <strong>No silent omission.</strong> A <a>property shape</a> an implementation cannot
+              handle MUST be surfaced rather than silently omitted. A property that is not rendered,
```
danielbeeke: I find this sentence hard to read.

danielbeeke (changes requested):
Looks good, some request for changes

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
+
+        <section id="feature-set-catalogue">
+          <h4>SHACL UI Features</h4>
+
+          <p>
+            The following table indexes the features. For each feature it gives a short description of
+            what supporting it means, the sections that define the behaviour, and the
+            conformance level. The authoritative behaviour of each feature is defined in the linked
+            sections; the table does not restate the normative requirements.
+          </p>
```
edmondchuc, line 718: Should we add an identifier column for each feature, preferably using an IRI, so that they are referenceable across implementations and reports?

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
...
+                <td>Apply the ordering rules recursively to nested property groups.</td>
+                <td><a href="#grouping-and-ordering">Grouping, Ordering, and Layout Hints</a></td>
+                <td>Recommended</td>
+              </tr>
+              <tr>
+                <td>15</td>
+                <td id="feature-builtin-widgets">Built-in widgets</td>
+                <td>Provide the built-in editors and viewers defined by this specification so they participate in scoring and selection.</td>
+                <td><a href="#builtin-widgets">Built-in Widgets</a></td>
+                <td>Required</td>
```
edmondchuc, line 834: I'm wondering whether this should be recommended rather than required. We could perhaps split this so that a small set of widgets capable of handling IRIs, blank nodes, and literals are required and can be selected by the scoring system as fallbacks, while all other widgets are recommended.

The idea is that this small set of required widgets would still allow a SHACL Renderer to be functional without requiring implementations to support every widget defined by the specification.

shacl12-ui/index.html
```diff
@@ -621,7 +621,279 @@ <h3>Document Conventions</h3>
         </p>
       </section>
       <section id="conformance">
-        <p>TODO: define conformance criteria for SHACL UI implementations.</p>
+        <p>
+          A SHACL UI implementation is a <a>SHACL Renderer</a> as defined in
+          <a href="#renderer">SHACL Renderer</a>. Conformance is defined in terms of a set of named
```
bergos: We never defined that and it's not part of concept/feature discussion.

bergos (changes requested):

shacl12-ui/index.html
edmondchuc: Is this section somewhat related to https://github.com/w3c/data-shapes/issues/1178?

I'm assuming that when it says a property is not rendered, it means that the widget for that property/value (property UI component) is not rendererd?

shacl12-ui/index.html
smessie: Good point. I've dropped the "no unqualified claim" framing and instead given the bare claim a precise meaning: an implementation **conforms to SHACL UI when it supports every feature designated as required**, and it may additionally state which further features it supports. So in your tender example a company can just tick "conforms", which now unambiguously means "supports all required features" — and optionally list the extra recommended/optional features it also covers. Updated in a7d5b869.

shacl12-ui/index.html
smessie: Thanks — yes, that's the intent: start with the suggested levels and refine them as the task force settles which features are required.

shacl12-ui/index.html
smessie: Agreed, it was redundant. Removed — it now reads "The following contracts apply to every feature." Updated in a7d5b869.

shacl12-ui/index.html
smessie: Rephrased for readability: "If an implementation cannot handle a property shape, it MUST surface this rather than omitting the property silently. A silently omitted property is not rendered and not held in form state, so writing the node back drops its values with no error." Updated in a7d5b869.

shacl12-ui/index.html
```diff
@@ -621,7 +621,463 @@ <h3>Document Conventions</h3>
...
+              <tr>
+                <td>14</td>
+                <td id="feature-nested-groups"><a href="#feature-nested-groups">Nested groups</a></td>
+                <td>Apply the ordering rules recursively to nested property groups.</td>
+                <td><a href="#grouping-and-ordering">Grouping, Ordering, and Layout Hints</a></td>
+                <td>Recommended</td>
+              </tr>
+              <tr>
+                <td>15</td>
+                <td id="feature-autocomplete-editor"><a href="#feature-autocomplete-editor">Auto-complete editor</a></td>
```
danielbeeke: What if we would record outcomes instead of specific widgets?
Also we could group them together something like:

| Feature                                                                                        | What supporting it means                                            | Specification sections | Conformance level |
| ---------------------------------------------------------------------------------------------- | ------------------------------------------------------------------- | ---------------------- | :---------------: |
| Editor for IRIs                                                                                | Edit an IRI value.                                                  | [shui:IRIEditor](#IRIEditor)         |     Required      |
| Editor for Blank nodes                                                                         | Edit a value that is a blank node via its nested shape.             | [shui:BlankNodeEditor](#BlankNodeEditor)   |     Required      |
| Editors for xsd:boolean, xsd:date, xsd:dateTime, xsd:integer, xsd:double, xsd:float, xsd:decimal, xsd:string | It is possible to edit literals that have these data types.         | [shui:BooleanEditor](#BooleanEditor), [shui:DatePickerEditor](#DatePickerEditor), [shui:DateTimePickerEditor](#DateTimePickerEditor), [shui:NumberFieldEditor](#NumberFieldEditor), [shui:TextFieldEditor](#TextFieldEditor) and [shui:TextAreaEditor](#TextAreaEditor)                |     Required      |
| Editors for rdf:langString                                                                     | It is possible to edit literals with language and set the language. | [shui:TextAreaWithLangEditor](#TextAreaWithLangEditor) and [shui:TextFieldWithLangEditor](TextFieldWithLangEditor)                |     Required      |
| Editor for rdf:HTML                                                             | It is possible to edit literals that have these data types.    | [shui:RichTextEditor](#RichTextEditor)                |    Recommended    |

shacl12-ui/index.html
danielbeeke: @edmondchuc what are your thoughts on https://github.com/w3c/data-shapes/pull/1173#discussion_r3821111924? There is a somewhat smaller set of required widgets and only the outcome is defined and not the exact IRI of the widget that must be implemented. This gives implementers freedom for example, to create an editor that handles both blank nodes and IRIs so that you can give the possibility to change a blanknode to an IRI.

shacl12-ui/index.html
smessie: I've removed this in https://github.com/w3c/data-shapes/pull/1173/changes/5ec59778f4faac5533fc6c70c3aa0f59142743b4

shacl12-ui/index.html
smessie: Yes, good suggestion! Added referenceable identifiers, but as self-links on each feature name rather than a separate column, to keep the table narrow. Every feature name now links to its own row, so each feature has a stable fragment IRI (e.g.  …/shacl12-ui/#feature-language-in ) you can cite in reports. Updated in https://github.com/w3c/data-shapes/pull/1173/changes/35d1bdfaf337da168d2b10dde0fdadef900cea3e.

shacl12-ui/index.html
smessie: Thanks for the feedback, both! I've now applied @danielbeeke's suggestion from the linked comment to define the expected behavior with linked conformance levels, rather than defining the expected widget implementations to be supported.

shacl12-ui/index.html
smessie: Yes, I like this suggestion! I've applied it in https://github.com/w3c/data-shapes/pull/1173/changes/c4a022268f4ce615f1dfa55244c8fb69d998fe9f and have also taken into account the viewers.

bergos (changes requested):
I think the feature list would be useful as a matrix mapped against use cases, to help identify profiles.

It should not become part of the specification. It's unnecessarily complex, and there are already established ways to define varying conformance within a single specification, such as profiles, conformance levels, or conformance classes. I don't see the point of introducing something new.

nicholascar:
@bergos

> I think the feature list would be useful as a matrix mapped against use cases, to help identify profiles.

The Use Cases the UI Task Force has defined are informal - not registered in any way with the W3C - so they are helpful for this TF's work but they cannot be quoted in the document. I'm not sure if you were suggesting that or just a comparison of Features & Use Cases in this PR discussion.

I suggest that you, and other UI UCR authors, compare the list of Features to your Use Cases and see if you can identify any collisions, perhaps in the UCR Issues. Having the Features listed should make this easy.

Regarding Profiles:

If Features can be easily identified and their Conformance levels known, there will be no need for this specification to also articulate Profiles as well - that's another mechanism. The Required Features will make a Core, which you could think of as a Core Profile if you like but it need not be described beyond just the set of Required Features. 

Others could do that after this specification's publication, for other purposes.

> It should not become part of the specification.

I think it, or something like it, should. The major issue currently seems to be what things are mandatory and what are not. A listing of Features with conformance imperatives - Required, Recommended, Optional as above - would make this clear. The set of Required Features would make the Core of this specification.

> It's unnecessarily complex, and there are already established ways to define varying conformance within a single specification, such as profiles, conformance levels, or conformance classes. I don't see the point of introducing something new.

I don't think this is complex: it's just a listing of the parts of SHACL UI and the level of requirement for them to be implemented.

It's true there are _different_ ways of implementing conformance things within a specification - Conformance Classes, Profiles etc - but feature listing with imperatives is a reasonable way of doing this and it reveals the Core, as indicated above.

My only suggestion is that the features should be able to be ordered in their table based on their Conformance Level so that the Core - those with level Required - are able to be grouped easily. 

I recommend that this PR be merged and the discussion about what is and isn't mandatory to be implemented be piked up per-Feature, noting that elsewhere I've suggested that data preparation for an implementing engine could likely handle at least one of your Use Cases (#1142).

bergos:
@nicholascar 

> The Use Cases the UI Task Force has defined are informal - not registered in any way with the W3C - so they are helpful for this TF's work but they cannot be quoted in the document. I'm not sure if you were suggesting that or just a comparison of Features & Use Cases in this PR discussion.

I was proposing that we use the current list of features to compare against the use cases, and then shape the concepts based on that outcome. Since that's not part of this PR, it probably shouldn't happen here.

> 
> Regarding Profiles:
> 
> If Features can be easily identified and their Conformance levels known, there will be no need for this specification to also articulate Profiles as well - that's another mechanism. The Required Features will make a Core, which you could think of as a Core Profile if you like but it need not be described beyond just the set of Required Features.
> 
> Others could do that after this specification's publication, for other purposes.

I agree with that. I made the same point with different wording here: https://github.com/w3c/data-shapes/issues/1164#issuecomment-5352972195. I wanted to continue the discussion in the issue I created for that topic, as requested, but no feedback has come through yet.

*If we agree to work with concepts, we also agree on profiles. Otherwise, structuring the specification into concepts does not make sense. The set of concepts an implementation supports would already be a profile, at a very low level. I hope we can agree on profiles at a higher level of abstraction.*

> 
> > It should not become part of the specification.
> 
> I think it, or something like it, should. The major issue currently seems to be what things are mandatory and what are not. A listing of Features with conformance imperatives - Required, Recommended, Optional as above - would make this clear. The set of Required Features would make the Core of this specification.

The features, as currently structured, can't be traced back to their actual definitions. A single feature may correspond to just part of a sentence, or to an entire paragraph within a section. For example, look at the features listed for Grouping, Ordering, and Layout Hints, then compare them against that section itself.

> 
> > It's unnecessarily complex, and there are already established ways to define varying conformance within a single specification, such as profiles, conformance levels, or conformance classes. I don't see the point of introducing something new.
> 
> I don't think this is complex: it's just a listing of the parts of SHACL UI and the level of requirement for them to be implemented.
> 

What I actually requested — and this is why I use the term "concepts" to distinguish them from "features" — is a list where each entry maps to a clearly defined, discrete part of the specification.

caribouW3:
There seems to be a confusion between requirements derived from use cases, and features defined in the spec.
A feature can help satisfy several requirements, and is generally defined in a easily identifiable portion of the spec.

Documenting which feature satisfies which requirement is useful, and should preferably be kept out of the spec.

> The features, as currently structured, can't be traced back to their actual definitions. A single feature may correspond to just part of a sentence, or to an entire paragraph within a section. For example, look at the features listed for Grouping, Ordering, and Layout Hints, then compare them against that section itself.

Mapping Use cases -> requirements -> features would help clarifying what really needs to go into the spec prose.
Later on, when defining which features are mandatory or not,  we can define conformance classes, or any other mechanism that is needed to help a user know what an implementation can or can't do.

> What I actually requested — and this is why I use the term "concepts" to distinguish them from "features" — is a list where each entry maps to a clearly defined, discrete part of the specification.

+1 to map use cases, requirements, and then features.

shacl12-ui/index.html
edmondchuc: @danielbeeke good idea, I just checked the other comment, and it looks great! Thanks both.

PR #1185, opened 2026-08-23
Refine relation to SPARQL section

afs:
Use full name in title
Clarify focus is SPARQL pattern matching

* [See this section rendered online here](https://raw.githack.com/w3c/data-shapes/srl/sparql-pattern-matching/sparql12-rl/index.html#srl-sparql-relationship)


recalcitrantsupplant (approved):
Looks good to me

Issue #1163, opened 2026-08-18
[UI] Decision: is widget matching required for conformance, or may an implementation rely only on declared widgets?


smessie:
Split from #1083.

## Question

Must a conforming implementation implement the widget matching mechanism (Select, Score, Match), or may it conform while rendering only properties that carry an explicit widget declaration such as `shui:editor`?

A related sub-question: is an explicit widget declaration an input to matching, at its highest confidence level, or a bypass around it?

## Use cases

- Application profiles the deploying organisation does not control (see use case issue to be created by @danielbeeke): these carry no UI annotations, so an implementation that only honors declared widgets cannot render them at all without the organisation first annotating every property in a profile it does not own
- #1146: guaranteeing scoring-based fallback rendering for validation-only shapes
- #1142: build-time form generation from closed shapes

## Arguments

For requiring matching: without it, a conformance claim does not tell an adopter whether their existing shapes will render. The practical consequence is annotating each property, which the application profile use case rules out.

Against: implementations may have reason not to implement matching, and requiring it raises the barrier to a conforming implementation.

## On declarations as a bypass

If an explicit declaration bypasses matching rather than feeding into it, three things need answering that matching already handles:

- what happens when the declared widget IRI is one the implementation does not know
- what happens when the declared widget conflicts with the constraints, for example a numeric editor on a property with `sh:datatype xsd:string`
- how an author reasons about shapes that annotate some properties and not others

Treating a declaration as the highest confidence input to matching answers all three without new machinery.

## Possible resolutions

1. Matching is required for conformance
2. Matching is required, but only within a profile or feature that implementations declare
3. Matching is optional, and an implementation may conform while supporting only declared widgets

danielbeeke:
I agree with:
**Matching is required for conformance.**

### About a SHACL renderer that displays as simple HTML and submits to a back-end:

- The SHACL renderer is in that case the combination of the front-end and the back-end.
- The scoring system still has value for:
  - Changing schema (datatype changes / cardinality changes)
  - Invalid data (importing legacy data)
- The only reason I have seen for not implementing the scoring system is complexity and effort. I do not think this should steer the spec into a space where conformance is just implementing at least one concept.
- Important, the biggest part of the ecosystem uses SHACL for validation of data. We should honor this user-base and make it easy for them to render their shapes without the need of adding `shui:editor` statements. For this reason I think the scoring system is a base requirement.

smessie:
I agree with @danielbeeke.

[...]

PR #1195, opened 2026-08-24
[UI] Revert #994: "Update scoring graph descriptions to clarify use of WidgetMatcher"

bergos:
<!--
# Pull request instructions
(Note that text between the HTML comments will be recorded in this pull request's source for future editing, but will not render on Github.  Text is suggested with portions needing substitution enclosed in curly braces.  The curly braces also need to be removed when substituting.)

Please use "Closing keywords" for the associated Issue if this PR should close the Issue once merged.
https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

If this PR edits a document, please consider including a "Live render."  The non-comment template text suggests one mechanism.  If including the link line(s), please test-load the link(s) while drafting the PR.
-->
                                                                                                                                                                                                                                                                                                                   
This reverts all four commits from #994 (squash-merged as 560951b1):

* bd882824 — Update scoring graph descriptions to clarify WidgetMatcher usage
* 03234764 — Address review: scoring graph in matcher, best parameter for select
* 31ae34d2 — Replace the select function with scoring graph preparation
* 8c1cba6b — Remove the best parameter from the score function

## Why
The changes in #994 are substantive and not editorial: they replace the Select function with a new Scoring Graph Preparation step, remove the `Best` parameter from the Score function, and introduce the `shui:WidgetMatcher`/`shui:WidgetAcceptMatcher` terminology throughout the scoring algorithm section. Changes of this kind to the scoring system are the sort of design decision the Working Group would normally expect to discuss before merging. #994 was merged without such discussion, during a period when @bergos — who has a longstanding, active interest in this area of the spec — was known to be unavailable to review or weigh in. This revert restores the pre-#994 text so the scoring system design can be discussed on its merits, with the relevant participants present, before any of these changes are reintroduced.               

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/revert-994/shacl12-ui/index.html)


bergos:
@smessie , in the call you stated that #1163 covers the changes made in #994. That is not accurate. You introduced the section "Scoring Graph Preparation." The previous version only described the algorithm of the functions. After your changes, it now requires preprocessing the scoring graph beforehand. That requirement is not necessary, and it's unrelated to the outcome of #1163, which doesn't mention it at all.

Let's revert the changes and update the specification through the formal review process once we have an outcome on #1163.

@smessie , it would have been nice if you had mentioned this PR. After the incident with the other pull request that was merged fast-track, in the same time frame when I was absent, transparency is important here. The PR could have been referenced in issue #1163. Unlike the other PR, this one is hard to spot as containing a major change, since you didn't update the PR description after making it. The change was added late, in a commit on Friday, July 31st, and the PR was merged on August 5th. I only noticed it because of a follow-up PR.

@nicholascar @PapoutsoglouE

PR #1198, opened 2026-08-25
GH#1184: Improve SHACL Rules layers example

### LINKED ISSUE #1184: SHACL Rules layers-example
https://github.com/w3c/data-shapes/blob/gh-pages/shacl12-test-suite/tests/sparql/rules/layers-example.ttl

Rule 1 and Rule 2 can be in the same layer. As a test, that would then show "run to fixed-point" across two rules.

It is, arguably, more natural for manual authoring to have the query rules (Q3,Q4,Q5,Q6) in the same layer.  Alternatively, Q4 and Q6 can go in the same layer as Rule1 and Rule2 because they are positive query rules and don't depend on rules 1 and 2 having finished; however as rules about the outcome, manual authoring might well put them all separate from the calculation rules.

erdemonal:
Closes #1184

Updates the layers example so that `Rule1` and `Rule2` are in the same layer and `Q3` through `Q6` are grouped in the following layer. The expected results remain unchanged.

### DIFF
--- a/shacl12-test-suite/tests/sparql/rules/layers-example.ttl
+++ b/shacl12-test-suite/tests/sparql/rules/layers-example.ttl
@@ -49,14 +49,14 @@ ex:Rule1
 
 ex:Rule2
     a sh:SPARQLRule ;
-    sh:layer 1 ;
+    sh:layer 0 ;
     sh:construct """
         CONSTRUCT { ?a :connected ?b .} WHERE { ?a :connected/:connected ?b }
     """ .
 
 ex:RuleQ3
     a sh:SPARQLRule ;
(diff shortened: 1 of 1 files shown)


HolgerKnublauch (approved):

HolgerKnublauch:
Thanks, @erdemonal

afs:
@HolgerKnublauch You merged without the IPR check passing.

I have marked it as "non-substantive" this time; I can do that only because it enacts my suggestion.

If there is any way that it is not completely clear it is 100% non-substantive it must be raised to the WG or W3C staff.

HolgerKnublauch:
Thanks, I had no clue about this rule!

PR #1201, opened 2026-08-25
[UI] Proposal for shui:searchQuery

### LINKED ISSUE #846: [Core & UI] Referencing external dynamic IRIs in sh:in for federated SHACL validation and UI option lists.

danielbeeke:
# Feature Proposal: `shui:searchQuery` Extension

**This proposal supersedes #1112** 

## 1. Abstract & Context

In SHACL UI, a property's value nodes are typically enumerated via `sh:in`, optionally computed dynamically with a `sh:select` expression (including federated `SERVICE` queries), as allowed by Dynamic SHACL in SHACL 1.2.

Such baseline queries resolve a full candidate set and cannot be parameterized with a user's live search input, nor take advantage of vendor-specific full-text search (FTS) extensions.

This proposal introduces **`shui:searchQuery`**, an optional property directly on a **property shape**, that lets shape authors provide a dedicated SPARQL `SELECT` query for resolving candidate value nodes from live, user-provided search input (e.g. in an `AutoCompleteEditor`). It can be used standalone, or alongside `sh:in`/`sh:select` as a live-search override with a defined fallback path.

---

## 2. Vocabulary Definition

| Term              | Detail                                                                 |
| :---------------- | :---------------------------------------------------------------------- |
| **Property Name** | `shui:searchQuery`                                                      |
| **Domain**        | `sh:PropertyShape` — appears directly on the property shape, not nested inside `sh:in` |
| **Range**         | `xsd:string` (containing a SPARQL `SELECT` query string)                |
| **Cardinality**   | At most one value per property shape                                    |
| **Conformance**   | Optional / MAY be supported by SHACL UI renderers                       |

---

## 3. Specification Rules

1. **Rendering & Execution Behavior:**
   When an end user enters or changes search input for a property rendered with a widget that consumes candidate value nodes, a renderer that supports this extension MAY evaluate the query given by `shui:searchQuery` to retrieve the candidate values offered to the user.

2. **Reserved Execution Variables:**
   The query MUST expect the following variables to be pre-bound by the renderer before evaluation:
   * **`$searchTerm`**: an `xsd:string` literal containing the current search input entered by the user.
   * **`$uiLanguage`**: a language-tagged literal identifying the user's current UI language, as determined by Language Resolution.

3. **Query Structure & Constraints:**
   * **Output Mapping:** the query MUST project exactly one result variable, representing matching IRIs or literals.
   * **Paging:** the query MUST NOT include a `LIMIT` or `OFFSET` solution modifier — pagination and limits remain under the exclusive control of the renderer.

4. **Interaction with `sh:in`:**
   A property shape's permissible value nodes can still be computed via `sh:in`/`sh:select` (including `SERVICE`) per Dynamic SHACL. When a property shape combines `sh:in` with `shui:searchQuery`, a renderer that supports `shui:searchQuery` evaluates it in place of the `sh:in`-derived query while the user is searching. A renderer that does not support `shui:searchQuery`, but does support Dynamic SHACL, MAY fall back to the baseline `sh:in`/`sh:select` query instead. `shui:searchQuery` may also be used on its own, without any `sh:in`.

5. **Editor Selection:**
   The presence of `shui:searchQuery` on a property shape signals that a live search widget is appropriate, even without an explicit `shui:editor` statement — it now contributes `+20` to the `AutoCompleteEditor`'s selection score.

6. **Validation Scope:**
   `shui:searchQuery` is never evaluated as part of ordinary SHACL validation and MUST NOT be treated as a constraint by a validation engine. A new subsection defines an *optional* mechanism for implementations that additionally want to validate value nodes returned by `shui:searchQuery`:
   * Shape authors SHOULD write `shui:searchQuery` so its results are a subset of (or equal to) the value nodes permitted by the property shape's other constraints (`sh:class`, `sh:node`, `sh:in`, etc.).
   * Implementations MAY validate each returned value node in isolation against the surrounding property shape.
   * For `SERVICE`-backed `sh:in` expressions, this requires a validation engine supporting federated (Dynamic SHACL) validation, and the isolated check SHOULD use only the `sh:in` constraint — not the shape's other constraints — since a value node retrieved via `SERVICE` typically has no local triples and would otherwise trigger spurious violations.

---

## 4. Example Usage

**Standalone (no `sh:in` needed):**

```turtle
ex:TaskShape
    a sh:NodeShape ;
    sh:targetClass ex:Task ;
    sh:property [
        sh:name "Assignee" ;
        sh:path ex:assignee ;
        sh:class ex:Person ;
        shui:searchQuery """
            PREFIX ex: <http://example.com/>
            PREFIX text: <http://jena.apache.org/text#>
            SELECT ?value WHERE {
                ?value text:query ($searchTerm $uiLanguage) .
                ?value a ex:Person .
            }
        """ ;
    ] .
```

**Combined with `sh:in` for federated fallback:**

```turtle
ex:BookShape
    a sh:NodeShape ;
    sh:targetClass ex:Book ;
    sh:property [
        sh:name "Author" ;
        sh:path dct:creator ;
        sh:in [
            sh:select """
                PREFIX ex: <http://example.com/>
                SELECT ?value WHERE {
                    SERVICE <http://example.com/sparql> {
                        ?value a ex:Person .
                    }
                }
            """ ;
        ] ;
        shui:searchQuery """
            PREFIX ex: <http://example.com/>
            PREFIX text: <http://jena.apache.org/text#>
            SELECT ?value WHERE {
                SERVICE <http://example.com/sparql> {
                    ?value text:query ($searchTerm $uiLanguage) .
                    ?value a ex:Person .
                }
            }
        """ ;
    ] .
```

If the renderer doesn't support `shui:searchQuery` but does support Dynamic SHACL, it falls back to the baseline `sh:in`/`sh:select` query, returning all `ex:Person` instances.

## 5. Implementation Notes & Scope Boundary

- **Decoupled from Property Roles:** this spec does not mandate how property roles (e.g. `shui:LabelRole`, `shui:DepictionRole`) are resolved or injected into the search query. Renderers MAY combine filtering and label retrieval into one query or perform them separately.
- **Deterministic Execution:** isolating `shui:searchQuery` as a standalone SPARQL contract parameterized by `$searchTerm` and `$uiLanguage` keeps execution predictable without relying on string manipulation or automatic query restructuring by renderers.
- **Not a validation hook by default:** `shui:searchQuery` results are not checked against the shape unless an implementation opts in to the new optional validation mechanism described above.

Closes #846

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/shui-searchQuery/shacl12-ui/index.html)


robert-david:
I have some questions about this proposal. My understanding is that you want to query something different (different nodes) compared to what the 'standard' sh:select query does. In the example it is not clear how the two queries interact.

1. What is the reason this cannot be expressed by the sh:select?
2. Is the semantics of having two queries such that both result sets are merged?
3. The sh:in, as presented in the example above, is not a valid SelectExpression. I think we need to change this in the [shacl12-sparql](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-sparql/index.html) spec, right?

danielbeeke:
**What is the reason this cannot be expressed by the sh:select?** 
- sh:select is a query to get all valid value nodes. It is for validation. We need to a have a short list that only gives what the end user is searching for.
- We could stretch `s:select` its usage and pre-bind a variable $searchTerm. This would put a burden on shape writers to make a very precise sh:select query so that UIs work. Requiring them to put something like `if (bound($searchTerm) ? etc...)`. Seems complicated and error prone.

**Is the semantics of having two queries such that both result sets are merged?** No, when `shui:searchQuery` is available, it is only used to populate the list of options in an autocomplete. `shui:searchQuery` is never used in validation. Maybe we should add that the shape writer SHOULD make sure that the list of results in the `shui:searchQuery` is a sub set of the query in `sh:in`.

**The sh:in, as presented in the example above, is not a valid SelectExpression. I think we need to change this in the [shacl12-sparql](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-sparql/index.html) spec, right?** It is a valid SelectExpression, I carefully read the spec, I might have missed something. I do not read [the normative part](https://raw.githack.com/w3c/data-shapes/gh-pages/shacl12-sparql/index.html#dfn-select-expression) as blocking having more predicates with values. We are just extending the `SelectExpression`.

danielbeeke:
I made changes to the PR and the issue description after a call with Robert David,

Summary of what changed vs. the original proposal: 
- shui:searchQuery moved off the sh:in/sh:select blank node onto the property shape itself (so it no longer requires sh:in at all)
- Reserved variables switched from ?searchTerm/?uiLanguage to $searchTerm/$uiLanguage
- Its presence now feeds into AutoCompleteEditor editor-selection scoring
- A new optional "Validation of Search Query Results" section was added covering how (and when) implementations may validate returned values against the shape, including guidance for federated/SERVICE-backed cases.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
       </p>
     </section>
 
+    <section id="search-query">
+      <h2>Search Query</h2>
+
+      <p>
+        This section defines <code>shui:searchQuery</code>, an optional property of a <a>property shape</a>
+        that lets shape authors provide a dedicated SPARQL <code>SELECT</code> query for resolving candidate
```
robert-david:
```suggestion
        that lets shape authors provide a dedicated SPARQL <code>SELECT</code> query for retrieving candidate
```

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+              <a href="#lang-resolution">Language Resolution</a>.
+            </td>
+          </tr>
+        </tbody>
+      </table>
+
+      <p>
+        Using the <a href="../shacl12-sparql/#sparql-prefixes">prefix handling rules</a>, the SPARQL query derived
+        from the value of <code>shui:searchQuery</code> MUST project exactly one variable in its <code>SELECT</code>
+        clause, representing the matching IRIs or literals. The query MUST NOT include a <code>LIMIT</code> or
```
robert-david: I suggest to remove this constraint. We can leave that up to the implementer. A LIMIT of the query and a different LIMIT by the renderer could even be combined.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+            sh:select """
+                PREFIX ex: &lt;http://example.com/&gt;
+                SELECT ?value WHERE {
+                    SERVICE &lt;http://example.com/sparql&gt; {
+                        ?value a ex:Person .
+                    }
+                }
+            """ ;
+        ] ;
+        <ins>shui:searchQuery</ins> """
```
robert-david, line 2785: The example still shows the old approach, putting shui:searchQuery into the sh:in.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+        """ ;
+    ] .
+          </div>
+        </div>
+        <p>
+          As the user types into the <a href="#AutoCompleteEditor">AutoCompleteEditor</a> rendered for
+          <code>ex:BookShape</code>'s <code>Author</code> property, the renderer binds the typed text to
+          <code>$searchTerm</code> and the current UI language to <code>$uiLanguage</code>, then evaluates
+          <code>shui:searchQuery</code> to query the Apache Jena Text index via the <code>SERVICE</code> clause.
+          If the renderer does not support <code>shui:searchQuery</code> but does support Dynamic SHACL,
```
robert-david, line 2803: This is also outdated.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+          the baseline <code>sh:select</code> query is used instead, returning all instances of <code>ex:Person</code>.
+        </p>
+      </aside>
+
+      <h3 id="search-query-validation">Validation of Search Query Results</h3>
+
+      <p>
+        <code>shui:searchQuery</code> is never evaluated as part of ordinary SHACL validation: it plays no role in
```
robert-david: I don't think explaining this is necessary. It is somehow redundant and causes confusion. SHACL validation is defined elsewhere and it cannot anyways be changed by SHACL UI. Also the opposite holds.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+      <p>
+        Shape authors SHOULD write <code>shui:searchQuery</code> so that its result set, for any given
+        <code>$searchTerm</code>, is a subset of the value nodes permitted by the other constraints on the
+        <a>property shape</a> (such as <code>sh:class</code>, <code>sh:node</code>, or <code>sh:in</code>), or
+        exactly that same set. A <code>shui:searchQuery</code> that returns value nodes outside that set risks
+        offering end users candidate values that do not conform to the <a>property shape</a>.
+      </p>
+
+      <p>
+        Because a value node returned by <code>shui:searchQuery</code> is offered to the end user as though it
```
robert-david, line 2819: This is also outdated

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,213 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+        were drawn from <code>sh:in</code>, implementations MAY apply SHACL validation to the outcomes of
+        <code>shui:searchQuery</code>, where the outcomes are value nodes of the <a>property shape</a>.
+      </p>
+
+      <p>
+        Such validation runs the <a>property shape</a> in isolation once for each value node returned by the
```
robert-david, line 2825: I think this is redundant and blurs with validation. SHACL validation, also with SPARQL and SERVICE, is explained elsewhere. We should focus here on UI only.

robert-david (changes requested):
There is now a bit of inconsistency introduced with the changes. I think these leftovers should be cleaned up. Also, i mentioned some parts which i think are out of scope for UI. I hope that helps. Thanks

shacl12-ui/index.html
danielbeeke: Could it be that there was a caching issue or something? From my end it looks out of the `sh:in`.

shacl12-ui/index.html
danielbeeke: okay fine with me

shacl12-ui/index.html
danielbeeke: From my perspective this applies to the new situation. Would you rather not see mentions about how to fallback? What is it exactly that seems outdated?

shacl12-ui/index.html
danielbeeke: Okay, will apply

shacl12-ui/index.html
danielbeeke: This introduces the new concept of validating the result nodes of the shui:searchQuery.

shacl12-ui/index.html
danielbeeke: There is:
- SHACL validation on data (maybe in the backend or in the front end after changes)
- shui:searchQuery options fetching
- and then optional SHACL validation on the results of the shui:searchQuery, such that no invalid options can ever be presented. I do not think this is redundant. 

I will see if I can make it more clear.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,204 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+              A language-tagged literal identifying the user's current UI language preference, as determined by
+              <a href="#lang-resolution">Language Resolution</a>.
+            </td>
+          </tr>
+        </tbody>
+      </table>
+
+      <p>
+        Using the <a href="../shacl12-sparql/#sparql-prefixes">prefix handling rules</a>, the SPARQL query derived
+        from the value of <code>shui:searchQuery</code> MUST project exactly one variable in its <code>SELECT</code>
```
edmondchuc: I am wondering whether this needs to be restricted to exactly one projected variable. Would it be possible to instead require a specific binding name, e.g. `?value`, while allowing additional implementation-specific bindings?

This would give renderers a well-defined interoperable binding for the candidate value, while allowing application-specific autocomplete widgets to make use of additional information. For example, a query could also return a colour or logo for the item, a search result score from a database implementation such as Jena FTS, or a custom label that includes search result highlighting.

These implementation-specific use cases would not be possible if the SELECT clause is restricted to exactly one variable.

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,204 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+
+      <p>
+        When an end user enters or changes search input for a property rendered with a widget that consumes
+        candidate value nodes, a SHACL Renderer that supports this extension MAY evaluate the query given by
+        <code>shui:searchQuery</code> to retrieve the candidate value nodes offered to the user.
+      </p>
+
+      <p>
+        The query MUST expect the following variables to be pre-bound by the renderer before evaluation:
+      </p>
```
edmondchuc, line 2670: The current text doesn't explicitly say whether additional implementation-specific pre-bound variables are allowed. I have a use case where the renderer would supply the predicate used for a Jena FTS search.

Do you think we should make this extensibility point explicit? For example:

> Implementations MAY pre-bind additional implementation-specific variables.

edmondchuc (commented):
Does this PR supersede https://github.com/w3c/data-shapes/pull/1112?

shacl12-ui/index.html
danielbeeke: will do

shacl12-ui/index.html
danielbeeke: I loosened it up to require the first variable and allow any others to follow

danielbeeke:
> Does this PR supersede https://github.com/w3c/data-shapes/pull/1112?

Yes indeed. It is a clean up and a subset of it.

edmondchuc (approved):

shacl12-ui/index.html
```diff
@@ -2640,6 +2640,214 @@ <h4 id="displaying-validation-results">Displaying Validation Results</h4>
...
+            <th>Description</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td><code>$searchTerm</code></td>
+            <td>An <code>xsd:string</code> literal containing the current search input entered by the end user.</td>
+          </tr>
+          <tr>
+            <td><code>$uiLanguage</code></td>
```
robert-david, line 2685: I suggest to rename it to $language. It might be used for other purposes than the ui language, so better keep it open.

robert-david (approved):

shacl12-ui/index.html
danielbeeke: I think we should split languages. We could consider adding $contentLanguage? I am currently creating a comment about this language split.

danielbeeke: See https://github.com/w3c/data-shapes/issues/1208#issuecomment-5476625997. Does that make sense to you?

PR #1204, opened 2026-08-26
Update shacl-ui-validation.yml

caribouW3:
do not trigger on any push, restrict to the part you're working on


edmondchuc (approved):
Thank you! That makes much more sense.

PR #1206, opened 2026-08-26
[UI] Unsupported widgets in widget selection step

### LINKED ISSUE #1178: [UI] Filter/prune unsupported widgets in widget selection step
The current text says:

> The widget with the highest score SHOULD be selected as the default widget from the Scoring Algorithm results. Implementations MAY allow users to switch to any widget included in the results.
>
> Since multiple score entries may reference the same widget, a post-processing step SHOULD be performed to normalize the score results (for example, by aggregating or de-duplicating entries by widget) before presenting the widget choices to the user.

It should also define how widgets that are not supported by the SHACL Renderer are handled. In particular, the post-processing/selection step should filter out unsupported or unknown widgets before selecting the default widget or presenting widget choices to the user.

This would make the Widget Selection behaviour consistent with the existing requirement that an unknown declared widget falls back to widget scoring rather than causing rendering to fail.

danielbeeke:
Adds:
- implementation SHOULD only have implemented widget score shapes in scoring graph
- implementation MUST let the end user know that a widget implementation is not found, despite having score shapes.

Closes #1178

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/warning-about-scores-without-implementation/shacl12-ui/index.html)


shacl12-ui/index.html
```diff
@@ -1312,6 +1312,12 @@ <h3>Introduction</h3>
           conditions under which it applies, expressed as SHACL shapes referenced through
           `shui:shapesGraphShape` (validated against the shape node in the shapes graph)
           and `shui:dataGraphShape` (validated against the focus node in the data graph).
+        </p>
+        <p>
+          An implementation SHOULD make sure only widgets that are implemented have scoring
+          shapes in the scoring graph. When a widget is not implemented but its score shape
+          is present in the scoring graph and matches, the implementation MUST make known
+          to the user that the widget is not available.
```
```suggestion
          to the user that the widget is not available. Similarly, when no valid
          widget matches, the implementation MUST make this known to the user.
```

smessie (commented):
Looks good. One additional thing I would add though is that if no widget (score) matches, this must be made known to the user as well, such that no properties are silently omitted.

danielbeeke:
@smessie I added a message when no match has been found.

smessie (approved):
LGTM

PR #1209, opened 2026-08-26
Shui property roles

### LINKED ISSUE #619: Consider adopting DASH property roles in SHACL
The DASH vocabulary defines [property roles](https://datashapes.org/propertyroles.html) that are used to indicate the function or purpose of properties within a shape.

These roles are especially useful in UI use cases (labels or icons), but some may have broader usage outside the UI context.

The DASH specification currently defines the following property roles:
- `dash:LabelRole`: labels or names of the resource
- `dash:DescriptionRole`: summary texts describing the resource
- `dash:IconRole`: icons for the resource or its type
- `dash:DepictionRole`: images depicting the individual resource
- `dash:IDRole`: identifiers, often unique and machine-readable
- `dash:KeyInfoRole`: other relevant values that should be rendered in summaries

But there are also non-UI-specific roles, particular `dash:IDRole`. That property role, but maybe also others could be relevant for other specifications.

Are there existing or potential use cases where property roles could be used in other SHACL-related specifications?

Should we consider bringing the generic concept of property roles into SHACL namespace? UI or core?

danielbeeke:
Adds various DASH property roles. Only the ones clearly for UI and also adds shui:ClassificationRole for displaying skos:inScheme, rdf:type or schema:category etc.

Closes #619

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/shui-property-roles/shacl12-ui/index.html)


edmondchuc (approved):
Looks good! And I like the idea of the classification role. I will add this to my implementation soon and test it out.

shacl12-ui/index.html
```diff
@@ -3422,6 +3422,121 @@ <h4>shui:LabelRole</h4>
...
+    sh:nodeKind sh:IRI ;
+    sh:description "The type of the resource." ;
+    sh:name "type" ;
+    shui:propertyRole shui:ClassificationRole ;
+.
+          </div>
+        </div>
+        <p>
+          This example illustrates a property shape directly annotated with <code>shui:ClassificationRole</code>, indicating that its values
+          classify the resource, for instance by type, concept scheme, or category.
```
smessie, line 3473: The example illustrates a single classification role for rdf:type, so I propose the following change. Adding at least one additional example, such as for skos:inScheme, would be helpful.
```suggestion
          This example illustrates a property shape directly annotated with <code>shui:ClassificationRole</code>, indicating that its `rdf:type` values
          classify the resource.
```

shacl12-ui/index.html
```diff
@@ -3422,6 +3422,121 @@ <h4>shui:LabelRole</h4>
...
+        Properties marked with the property role <code>shui:IconRole</code> produce small images that may be displayed next to a resource's label,
+        for example in the corner of a card or to the left of a label in a tree view. Instances of the same class often share the same icon, which
+        may be computed using a <code>sh:values</code> rule or supplied via <code>sh:defaultValue</code>.
+      </p>
+
+      <p>
+        Common examples include <code>schema:logo</code> and <code>schema:thumbnail</code>. Values SHOULD have <code>sh:nodeKind sh:IRI</code> or
+        datatype <code>xsd:anyURI</code>, pointing at an image with a MIME type supported by the HTML <code>img</code> element. Icons SHOULD be
+        small, ideally vector graphics such as <code>.svg</code> files, since they are typically rendered at sizes between 16 and 48 pixels.
+      </p>
```
smessie, line 3515: "Icons SHOULD be small, ideally vector graphics" puts an untestable criterion under an RFC 2119 keyword — "small" has no criterion and "ideally" hedges inside the SHOULD. The 16–48 pixel range is carried over from DASH and assumes a particular class of desktop tree/card UI; SHUI renderers also target mobile, print and high-DPI output.

The substance is worth keeping, since the size/reuse contrast is what distinguishes this role from `shui:DepictionRole`. So rather than dropping it, this keeps the one testable SHOULD normative and demotes the rest to a note.


```suggestion
      <p>
        Common examples include <code>schema:logo</code> and <code>schema:thumbnail</code>. Values SHOULD have <code>sh:nodeKind sh:IRI</code> or
        datatype <code>xsd:anyURI</code>, pointing at an image with a MIME type supported by the HTML <code>img</code> element.
      </p>

      <p class="note">
        Icons are typically rendered at small sizes and are often reused across many resources. Scalable formats such as SVG are therefore
        usually a good choice. In contrast to <code>shui:DepictionRole</code>, icon values are not expected to be distinct per resource.
      </p>
```

shacl12-ui/index.html
```diff
@@ -3422,6 +3422,121 @@ <h4>shui:LabelRole</h4>
         Common examples of display label predicates include <code>rdfs:label</code>, <code>skos:prefLabel</code>, and <code>schema:name</code>.
       </p>
 
+      <h4>shui:DepictionRole</h4>
+
+      <p>
+        Properties marked with the property role <code>shui:DepictionRole</code> deliver images that represent the individual resource.
+        These images SHOULD be distinct for each resource and MAY have a medium to large resolution.
+        Values SHOULD have a well-known image MIME type.
+      </p>
```
smessie, line 3431: Same point as on `shui:IconRole` below: "MAY have a medium to large resolution" isn't testable. Suggesting the parallel treatment so the two roles stay symmetrical, with the icon/depiction contrast stated explicitly instead of left implicit in a resolution hint.

```suggestion
      <p>
        Properties marked with the property role <code>shui:DepictionRole</code> deliver images that represent the individual resource.
        These images SHOULD be distinct for each resource. Values SHOULD have a well-known image MIME type.
      </p>

      <p class="note">
        Depictions are typically rendered larger than icons and are not constrained to a square box. In contrast to
        <code>shui:IconRole</code>, a depiction is expected to be specific to the individual resource rather than shared across instances.
      </p>
```

smessie (commented):
The `sh:order` precedence model is inherited from `shui:LabelRole`, where selecting a single winner is the right behaviour, i.e., you want exactly one label.
For `shui:ClassificationRole` I'm not sure that carries over: a resource can legitimately belong to several groups at once, and a UI would plausibly want to render all of them as parallel badges rather than pick one.

That leaves two things underspecified:

1. **Across shapes**: if two property shapes both carry `shui:ClassificationRole` with different `sh:order` values, does the smaller `sh:order` win and suppress the other, or do both contribute with `sh:order` only determining display order?
2. **Within one shape**: if a single classification property has multiple values, should a renderer show all of them or just one?

Worth settling before implementations diverge. @edmondchuc, since you mentioned
you'd be implementing this: which reading did you assume?
Or do we consider this out of scope?

PR #1215, opened 2026-08-29
#1212: Updated use of terms "base graph", "inference graph" and "evaluation graph"

### LINKED ISSUE #1212: Review wording on SHACL Rules as an update language
SHACL Inf Rules is an update language - it changes the input graph. That ought to be noted.

SPARQL-RL is not an update language -- (graph, ruleset) -> new graph the original graph is not changed - the user may choose to update the original graph - it is their decision.

_Originally posted by @afs in https://github.com/w3c/data-shapes/pull/1186#discussion_r3881690945_

The original wording on the update behavior is from SHACL-AF and could likely be aligned with what SRL does.

HolgerKnublauch:
These SHOULD now have the same interpretation as in SRL (unless I made mistakes). The intention was always that, but the old wording from SHACL-AF wasn't clear.

Closes #1212

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1212/shacl12-inference-rules/index.html#rules-execution)

CC (with thanks for raising this): @afs

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -1044,7 +1044,7 @@ ex:RuleSet2
 			<p>
 				It is sometimes useful for inference rules to produce triples that are only visible during the execution
 				of other rules but do not end up in the final inferences.
-				A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the inference graph contains a
+				A <dfn>temporary triple</dfn> is an inferred <a>triple</a> for which the <a>inference graph</a> contains a
 				<a>reifier</a> with the <a>value</a> <code>sh:tempTriple true</code>.
 				<a>Temporary triples</a> and their <a>reifiers</a> are visible to executing rules.
 				Rules can produce such triples as illustrated in the following example:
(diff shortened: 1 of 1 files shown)


mgberg:
Not directly related to this PR, but there is a ReSpec error:
> Couldn't find "SHACL rule engine" in this document or other cited documents:

shacl12-inference-rules/index.html
```diff
@@ -1109,20 +1109,16 @@ <h2>Temporary Triples</h2>
...
 				The new <a>triples</a> that are produced by a rules engine are called the <dfn data-lt="inferring|infer">inferred</dfn> triples.
-			</p>
-			<p>
-				Note that, from a logical perspective, the <a>data graph</a> will be <em>modified</em> if <a>triples</a> get inferred.
-				This means that rules can trigger after other triples have been inferred.
-				However, in cases where the original data should not be modified, implementations may construct a logical <a>data graph</a>
-				that has the original data as one subgraph and a dedicated inferences graph as another subgraph, and where
-				the inferred triples get added to the inferences graph only.
+				The <dfn>inference graph</dfn> consists of these <a>inferred</a> triples.
+				The <dfn>evaluation graph</dfn> is the union of the <a>base graph</a> and the <a>inference graph</a>.
```
mgberg, line 1121: Given that the evaluation graph is now defined as the union of other graphs, and that the current evaluation algorithm states

> Delete the <a>derived triples</a> (except those that were also inferred by rules) and their <a>reifiers</a>

Might it be convenient to say that derived triples end up in their own graph that is also part of the union, which I think might:
- Simplify deletion of derived triples at the end (simply drop the graph)
- Simplify replacement of derived properties on recomputation per iteration (again by dropping the graph in the naive case), and
- Simplify figuring out whether derived triples were also inferred by rules (not necessary once the graph is dropped)?

afs: Just a note if copying text:

SPARQL-RL does not have "union graphs". It works on sets (immutable) and rather than "add to a set" it makes a new one: e.g. 

```
   GE = GE ∪︀ Y
```
Implementations notice that GE is replaced, so the old value is not used again, and an implementation mutating a datastructure set would work.

HolgerKnublauch: As a response to both, I have no strong opinion. Both forms of describing the unions (including Matt's proposal) should produce the same results. They are NOT meant as implementation advice (certainly not the GE = GE u Y approach which indicates that all triples are copied each time). So Matt, yes I agree that implementations could use that extra subgraph, assuming any rules actually use sh:expectedPredicate. But in the end this is a conceptual specification of what the outcome should be, and there are many ways to implement those. If you want to change that, I'd be happy to review a PR.

With the current wording, are there any factual errors in the definitions? Do these definitions prevent certain desirable implementations?

@afs how are you implementing this in Jena? Are you using some TDB transaction that can see the extra triples without modifying the underlying graph?

HolgerKnublauch:
> Not directly related to this PR, but there is a ReSpec error:
> 
> > Couldn't find "SHACL rule engine" in this document or other cited documents:

Yes this is fixed in https://github.com/w3c/data-shapes/pull/1214/changes - I am waiting with the merge until after the Monday meeting to give people reasonable time to review it.

shacl12-inference-rules/index.html
mgberg: I don't see any errors/roadblocks at the moment, but I don't know if @afs is seeing something I'm missing.

Regardless, I think a follow up PR to clarify something may be necessary based upon the conversation above and in #1223.

HolgerKnublauch: Ok thanks, I have re-requested your review to see if this can be merged for now. I believe it's at least a step in the right direction, although we can always fine tune later.

mgberg (approved):

PR #1216, opened 2026-08-29
Created a section on using sh:order and sh:layer; updated intro paragraph to Variations of SHACL Rule Implementations

### LINKED ISSUE #1183: Relationship between `sh:order` and `sh:layer`
https://www.w3.org/TR/shacl12-sparql/#rule-variations

> In particular, the algorithm is non-deterministic in the sense that unless the order of rules is specified explicitly, the results may differ across executions. This is, for example, the case when rules draw conclusions from the number of certain triples yet those triples may be produced by other rules. In this document, the responsibility of producing a predictable ordering and layering/grouping of rules is left to the rule author.

Editorial:
"the order of rules" is ambiguous - it should be about the general concept but it can read as if specific to `sh:order`. The example needs `sh:layer` in general, for counting.  It _might_ get away with `sh:order` for specific data although the new repeated execution per layer has a considerable impact on that.

More generally, when is it appropriate to use `sh:order` with SHACL 1.2 Inf Rules, given the repeated layer execution until fixed point? Some text about this to assist the user (and reveiwer) would be helpful.

In https://www.w3.org/TR/shacl12-sparql/#rule-order, the example comment says:
(Aside - `sh:order` is not necessary in the example)

```
     sh:order 1 ;   # Will be evaluated before 2
```
but that isn't the whole story because it will also be run after rule 2 --
```
	do
		Execute one iteration over all iterating rules in the layer
```
means there is always a last "execute iteration" that produces no triples from any of the layer rules.
If any rule in a layer produces at least one triple, then all rules in the layer run at least twice.

Also:

> and the [stata](https://www.w3.org/TR/shacl12-sparql/#rule-layers) (ignoring sh:layer).

s/stat/layer/

scotthenninger:
Created a section on using sh:order and sh:layer; updated intro paragraph to Variations of SHACL Rule Implementations

Closes #1183

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/order-layer-inference-rules/shacl12-inference-rules/index.html#orderVsLayer)

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -850,7 +850,9 @@ ex:Granddaughter
 					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code> 
 					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
 					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
-					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					subsequent passes.</p>
+					
+<p>Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
 					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
 					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
 					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -845,7 +845,7 @@ ex:Granddaughter
 					of each rule in the sequence given by <code>sh:order</code> - as a phase for general (iterating) rules.
 				</p>
 				<p>
-					Use <code></code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
+					Use <code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
 					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers 
 					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code> 
 					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -855,7 +855,10 @@ ex:Granddaughter
 <p>Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
 					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
 					computations whose partial results should not persist, as in <a href="#tempTriple-example">Example 10</a>, where a layer-0 rule constructs ex:offspring as temporary 
-					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
+					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. 
+</p>
+<p>
+					<code>sh:order</code> is 
 					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
 					where a layer-1 run-once rule executes only after the iterating rules in the preceding layer have finished. 
 					n short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -861,7 +861,7 @@ ex:Granddaughter
 					<code>sh:order</code> is 
 					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
 					where a layer-1 run-once rule executes only after the iterating rules in the preceding layer have finished. 
-					n short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
+					In short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
 					when you need phase separation around the per-layer fixpoint and cleanup semantics.
 				</p>
 			</section>
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -857,7 +857,7 @@ ex:Granddaughter
 					computations whose partial results should not persist, as in <a href="#tempTriple-example">Example 10</a>, where a layer-0 rule constructs ex:offspring as temporary 
 					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
 					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
-					where a run-once rule placed in <code>sh:layer 1</code> executes only after the iterating rules in the preceding layer have finished. 
+					where a layer-1 run-once rule executes only after the iterating rules in the preceding layer have finished. 
 					n short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
 					when you need phase separation around the per-layer fixpoint and cleanup semantics.
 				</p>
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -854,7 +854,7 @@ ex:Granddaughter
 					
 <p>Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
 					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
-					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
+					computations whose partial results should not persist, as in <a href="#tempTriple-example">Example 10</a>, where a layer-0 rule constructs ex:offspring as temporary 
 					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
 					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
 					where a run-once rule placed in <code>sh:layer 1</code> executes only after the iterating rules in the preceding layer have finished. 
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -1,327 +1,327 @@
 <!DOCTYPE html>
 <html>
-	<head>
+    <head>
     <meta charset="utf-8" />
     <meta name="color-scheme" content="light dark">
-		<title>SHACL 1.2 Inference Rules</title>
-		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer></script>
-		<script class="remove">
+        <title>SHACL 1.2 Inference Rules</title>
+        <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer></script>
+        <script class="remove">
 
-			function prepareSyntaxRules() {
-				document.querySelectorAll("[data-syntax-rule]").forEach(element => {
-					let ruleId = element.getAttribute("data-syntax-rule");
-					let tr = document.createElement("tr");
-					tr.classList.add("syntax-rule-tr");
-					let td1 = document.createElement("td");
-					td1.classList.add("syntax-rule-id");
-					let a = document.createElement("a");
-					a.classList.add("syntax-rule-id-a");
-					a.href = "#syntax-rule-" + ruleId;
-					a.textContent = ruleId;
-					td1.appendChild(a);
-					let td2 = document.createElement("td");
-					td2.innerHTML = element.innerHTML;
-					tr.appendChild(td1);
-					tr.appendChild(td2);
+            function prepareSyntaxRules() {
+                document.querySelectorAll("[data-syntax-rule]").forEach(element => {
+                    let ruleId = element.getAttribute("data-syntax-rule");
+                    let tr = document.createElement("tr");
+                    tr.classList.add("syntax-rule-tr");
+                    let td1 = document.createElement("td");
+                    td1.classList.add("syntax-rule-id");
+                    let a = document.createElement("a");
+                    a.classList.add("syntax-rule-id-a");
+                    a.href = "#syntax-rule-" + ruleId;
+                    a.textContent = ruleId;
+                    td1.appendChild(a);
+                    let td2 = document.createElement("td");
+                    td2.innerHTML = element.innerHTML;
+                    tr.appendChild(td1);
+                    tr.appendChild(td2);
 
-					// Replace <dfn> elements inside `tr` with <a> elements
-					tr.querySelectorAll("dfn").forEach(dfn => {
+                    // Replace <dfn> elements inside `tr` with <a> elements
+                    tr.querySelectorAll("dfn").forEach(dfn => {
     					let a = document.createElement("a");
     					a.textContent = dfn.textContent;
     					dfn.replaceWith(a);
-					});
+                    });
 
-					document.getElementById("syntax-rules-table").appendChild(tr);
-					element.setAttribute("id", "syntax-rule-" + ruleId);
-				});
-			}
+                    document.getElementById("syntax-rules-table").appendChild(tr);
+                    element.setAttribute("id", "syntax-rule-" + ruleId);
+                });
+            }
 
-			function prepareValidators() {
-				document.querySelectorAll("[data-validator]").forEach(element => {
-					let validatorId = element.getAttribute("data-validator") + "ConstraintComponent";
-					let tr = document.createElement("tr");
-					tr.classList.add("validator-tr");
-					let td = document.createElement("td");
-					tr.appendChild(td);
-					let a = document.createElement("a");
-					a.classList.add("validator-id-a");
-					a.href = `#validator-${validatorId}`;
-					a.textContent = `sh:${validatorId}`;
-					td.appendChild(a);
-					td.innerHTML += ": " + element.innerHTML;
-					document.getElementById("validators-table").appendChild(tr);
-					element.id = "validator-" + validatorId;
-				});
-			}
+            function prepareValidators() {
+                document.querySelectorAll("[data-validator]").forEach(element => {
+                    let validatorId = element.getAttribute("data-validator") + "ConstraintComponent";
+                    let tr = document.createElement("tr");
+                    tr.classList.add("validator-tr");
+                    let td = document.createElement("td");
+                    tr.appendChild(td);
+                    let a = document.createElement("a");
+                    a.classList.add("validator-id-a");
+                    a.href = `#validator-${validatorId}`;
+                    a.textContent = `sh:${validatorId}`;
+                    td.appendChild(a);
+                    td.innerHTML += ": " + element.innerHTML;
+                    document.getElementById("validators-table").appendChild(tr);
+                    element.id = "validator-" + validatorId;
+                });
+            }
 
-			document.addEventListener("DOMContentLoaded", () => {
-				// Replace code div blocks in shapes, data, and results graph with tabs
-				for (const graph of document.querySelectorAll(".shapes-graph, .data-graph, .inferences-graph")) {
-					const tabs = document.createElement("aside");
-					tabs.classList.add("ds-selector-tabs");
-					graph.firstElementChild.classList.add("selected");
+            document.addEventListener("DOMContentLoaded", () => {
+                // Replace code div blocks in shapes, data, and results graph with tabs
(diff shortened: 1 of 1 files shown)


liviorobaldo:
I can't see it. I clicked on "[See this document rendered online here](https://raw.githack.com/w3c/data-shapes/%7BBRANCH_NAME%7D/%7BFOLDER%7D/index.html)" and I get a 404:

image

mgberg:
Based on the diff I think there was an indent conversion issue

scotthenninger:
Ok any suggestions on how to fix that?

scotthenninger:
Try now. I made a change to the tabs that should work.

shacl12-inference-rules/index.html
```diff
@@ -1204,14 +1227,17 @@ <h2>Tracking the Rule that has produced a Triple (sh:sourceRule)</h2>
...
-				In this document, the responsibility of producing a predictable ordering and layering/grouping of
-				rules is left to the rule author.
+				The <a href="#rules-execution">general execution algorithm</a> described above and the SPARQL language 
+				itself offer implementations considerable flexibility. Consequently, the execution of rules may be 
+				non-deterministic. With respect to <code>sh:order</code> in particular, note that rules sharing the same 
+				<code>sh:order</code> value within a layer are executed in an arbitrary order, so their combined output 
+				may vary across executions. This is especially problematic when rules draw conclusions from the number 
+				of certain triples, yet those triples may themselves be produced by other rules. It is the responsibility 
+				of the rule developer to avoid such non-determinism by appropriately defining the execution order and 
+				layering of rules where necessary. By contrast, <a href="../sparql12-rl/">SPARQL-RL</a> rules do not
```
HolgerKnublauch, line 1242: We should not mention SPARQL-RL until the relationship between the two languages has been clarified. Right now they are disconnected. My latest proposal to address this is https://github.com/w3c/data-shapes/pull/1219

Also, the current text is way too positive about SRL and sounds like SHACL rules have a serious problem, while in practice those cases where the order even matters are not all that common and are easy to work around. If we talk about the benefits of SRL, we should also talk about the limitations (e.g. in expressiveness and its limitation to global rules that are detached from the ontology/shapes).

HolgerKnublauch (changes requested):

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
 			</section>
+			<section id="orderVsLayer">
+				<h3>Using sh:order and sh:layer</h3>
+				<p>
+					The choice between <code>sh:order</code> and <code>sh:layer</code> depends on whether a rule set requires simple sequencing within a 
+					single pass of the iteration loop or genuine separation between phases.It is useful to think of one such pass - a single execution 
+					of each rule in the sequence given by <code>sh:order</code> - as a phase for general (iterating) rules.
+				</p>
+				<p>
+					Use <code></code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples
```
```suggestion
					Use <code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+					single pass of the iteration loop or genuine separation between phases.It is useful to think of one such pass - a single execution 
+					of each rule in the sequence given by <code>sh:order</code> - as a phase for general (iterating) rules.
+				</p>
+				<p>
+					Use <code></code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
+					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers 
+					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code> 
+					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has
```
```suggestion
					subsequent passes.</p>
					
<p>Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+				</p>
+				<p>
+					Use <code></code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
+					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers 
+					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code> 
+					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
+					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary
```
```suggestion
					computations whose partial results should not persist, as in <a href="#tempTriple-example">Example 10</a>, where a layer-0 rule constructs ex:offspring as temporary 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers 
+					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code> 
+					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
+					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
+					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is
```
```suggestion
					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. 
</p>
<p>
					<code>sh:order</code> is 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+					(<code>sh:order 2</code>), is representative of this case. Note that <code>sh:order</code> sequences rules only within a pass. Because 
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
+					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
+					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
+					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
+					where a run-once rule placed in <code>sh:layer 1</code> executes only after the iterating rules in the preceding layer have finished.
```
```suggestion
					where a layer-1 run-once rule executes only after the iterating rules in the preceding layer have finished. 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
+					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
+					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
+					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
+					where a run-once rule placed in <code>sh:layer 1</code> executes only after the iterating rules in the preceding layer have finished. 
+					n short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code>
```
```suggestion
					In short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
```

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,29 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+					the layer's iteration loop repeats while new triples are produced, a later rule's output persists and is visible to earlier rules on 
+					subsequent passes. Use <code>sh:layer</code>, by contrast, when a later group of rules must run only after an earlier group has 
+					reached its fixpoint and after that group's intermediate results have been cleaned up. This is appropriate for multi-step 
+					computations whose partial results should not persist, as in Example 10, where a layer-0 rule constructs ex:offspring as temporary 
+					triples and a layer-1 rule consumes them to infer only the final <code>ex:youngestOffspring</code> triples. <code>sh:order</code> is 
+					also appropriate for post-processing that must follow recursive closure, as in <a href="#example-rules-before-after">Example 6</a>, 
+					where a run-once rule placed in <code>sh:layer 1</code> executes only after the iterating rules in the preceding layer have finished. 
+					n short, use <code>sh:order</code> when you need deterministic sequencing of rules within an iteration pass, and <code>sh:layer</code> 
+					when you need phase separation around the per-layer fixpoint and cleanup semantics.
```
liviorobaldo, line 865: @scotthenninger , what do you mean by "cleanup semantics"?

scotthenninger: This refers to cleaning up temporary triples, etc. I suppose it could enumerate all of the triples cleaned up after a layer finishes.

shacl12-inference-rules/index.html
mgberg: I definitely agree that it would be good to cast the non-determinism as more of a tradeoff that comes with the flexibility/extensibility of these rules vs SRL.

shacl12-inference-rules/index.html
liviorobaldo: @scotthenninger, I do not understand what you mean. Temporary triples are not "enumerated"; they are simply "marked" by placing their reifications in `tempTriple`.

liviorobaldo: By the way, I have just noticed that in the [General Execution Instructions for SHACL Rules](https://raw.githack.com/w3c/data-shapes/order-layer-inference-rules/shacl12-inference-rules/index.html#rules-execution), temporary triples are deleted _both_ at the end of the entire inference process _and_ at the end of each layer.

Deleting them at the end of the entire process is, of course, redundant if they are already deleted at the end of each layer (including the last one).

In any case, I think we should delete them only at the end of the entire process, because it could be unintuitive for developers to have to keep in mind that temporary triples are deleted when moving from one layer to the next.

Finally, I don't understand what "except those that were also inferred by rules" means.

@scotthenninger @HolgerKnublauch @mgberg

HolgerKnublauch: @liviorobaldo you may have misread those instructions. What they say is that the *derived* triples are deleted in each layer. The derived triples stem from sh:values and sh:defaultValue statements. They are distinct from explicitly created temporary triples (sh:tempTriple).

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,34 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+			<section id="orderVsLayer">
+				<h3>Using sh:order and sh:layer</h3>
+				<p>
+					The choice between <code>sh:order</code> and <code>sh:layer</code> depends on whether a rule set requires simple sequencing within a 
+					single pass of the iteration loop or genuine separation between phases.It is useful to think of one such pass - a single execution 
+					of each rule in the sequence given by <code>sh:order</code> - as a phase for general (iterating) rules.
+				</p>
+				<p>
+					Use <code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
+					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers
```
HolgerKnublauch: In general I would suggest it is better to avoid using the Example number, because that number may change when other examples are added. The same applies below for Example 10.

shacl12-inference-rules/index.html
```diff
@@ -837,6 +837,34 @@ <h3>Run-once Rules (sh:runOnce)</h3>
...
+				<h3>Using sh:order and sh:layer</h3>
+				<p>
+					The choice between <code>sh:order</code> and <code>sh:layer</code> depends on whether a rule set requires simple sequencing within a 
+					single pass of the iteration loop or genuine separation between phases.It is useful to think of one such pass - a single execution 
+					of each rule in the sequence given by <code>sh:order</code> - as a phase for general (iterating) rules.
+				</p>
+				<p>
+					Use <code>sh:order</code> when rules form a straightforward producer-consumer chain in which a rule's inferred triples 
+					must be visible to a later rule within the same pass. <a href="#example-rule-order">Example 5</a>, where one rule infers 
+					<code>ex:uncle</code> (<code>sh:order 1</code>) before a second rule uses those triples to infer <code>ex:cousin</code>
```
HolgerKnublauch: I noticed that the example is unconvincing and should be changed. (This could become a separate ticket if necessary). The issue is that the sh:order here is not even needed. Even when the rules are executed in the opposite order, the same results will be computed because they are in the same layer, and the engine will iterate until no further triples are inferred. So if the uncle rule is executed second, the cousin rule will be reached in iteration 2. The only difference is that fewer iterations are needed when the order is "correct".

A better example may be using sh:runOnce true?

shacl12-inference-rules/index.html
liviorobaldo: Ok, sorry for the confusion. However, now that I have a better understanding of what this text means, I still find it quite confusing 😅

First, I find the terminology inconsistent: "expected derived triples" VS "derived triples (except those that were also inferred by rules)". In particular, I am still confused about what "except those that were also inferred by rules" exactly means.

It would also help to have two explicit pointers to §3.8 and §7 after the box following "The execution of a rule set is defined as follows:", so that the reader can easily refresh their memory about what we mean by "expected derived triples" and "temporary triples".

Then I was wondering: why must "expected derived triples" be deleted at the end of each layer? Why can't they be deleted at the end of the whole inference process? Note that in §3.8, layers are not even mentioned. We can, of course, have two rules producing two different (and inconsistent) values, e.g.:

```
ex:SmallRectangle ex:area 20 .
ex:SmallRectangle ex:area 30 .
```

and a shape reporting this inconsistency.

By deleting the "expected derived triples" at the end of each layer, the inconsistency is NOT detected when the two rules can belong to DIFFERENT layers. The inconsistency is detected ONLY when the two rules belong to the SAME layer. Conversely, by deleting "expected derived triples" at the end of the whole inference process, the inconsistency will ALWAYS be detected. I think the latter is what should be done here, rather than the former. So, in my view, "expected derived triples" should be deleted at the end of the whole process, not at the end of each layer.

If this is correct, and there are no other good reasons to delete "expected derived triples" at the end of each layer, another question arises: why don't we simply model "expected derived triples" as "temporary triples"? 😅 Instead of having `sh:value` and `sh:defaultValue`, we could have standard SPARQL rules with different `sh:order` values.

For instance, in Example 7, instead of writing this:

```
ex:RectangleShape-area
a sh:PropertyShape ;
sh:path ex:area ;
sh:defaultValue 1 ;
sh:values [
sparql:multiply ( [ shnex:pathValues ex:width ] [ shnex:pathValues ex:height ] )
] .

ex:Rectangle-computeSmall
a sh:SPARQLRule ;
rdfs:comment "This rule expects that the values of ex:area have been derived." ;
sh:expectedPredicate ex:area ;
sh:construct """
CONSTRUCT {
$this ex:isSmall true .
}
WHERE {
$this ex:area ?area .
FILTER (?area < 100) .
}
""" .
```

we could write THREE normal rules, with different `sh:order` values, and have the values of `ex:area` produced as temporary triples:

```
ex:Rectangle-computingArea
a sh:SPARQLRule ;
sh:order 0 ;
rdfs:comment "THIS IS A RULE WITH sh:order 0, 
WHICH EXECUTES BEFORE THE NEXT RULE!!!" ;
sh:construct """
CONSTRUCT {
$this ex:area ?value {| sh:tempTriple true |} .
}
WHERE {
$this ex:width ?w .
$this ex:height ?h .
BIND(?w * ?h AS ?value)
}
""" .

ex:Rectangle-assigningDefaultAreaIfPreviousRuleDoesNotTrigger
a sh:SPARQLRule ;
sh:order 1 ;
rdfs:comment "THIS RULE HAS sh:order 1, SO IT 
EXECUTES SECOND: IF NO AREA WAS COMPUTED
(BECAUSE EITHER WIDTH OR HEIGHT IS UNDEFINED), 
THE DEFAULT VALUE 1 IS SET!" ;
sh:construct """
CONSTRUCT {
$this ex:area 1 {| sh:tempTriple true |} .
}
WHERE {
FILTER NOT EXISTS {
$this ex:area ?value .
}
}
""" .

ex:Rectangle-computeSmall
a sh:SPARQLRule ;
sh:order 2 ;
rdfs:comment "This rule expects that the values of 
ex:area have been derived. THAT IS WHY IT HAS 
sh:order 2: IT EXECUTES AFTER THE TWO PREVIOUS
ONES!!!" ;
sh:construct """
CONSTRUCT {
$this ex:isSmall true .
}
WHERE {
$this ex:area ?area .
FILTER (?area < 100) .
}
""" .
```


Advantages:

- We remove §3.8 and all mentions of `sh:value` and `sh:defaultValue` from this document and the other documents (e.g., SHACL Core).
- We simplify the algorithm in §7.

I personally find `sh:value` and `sh:defaultValue` quite complex to use, even now that I fully understood how they work. I think they were introduced *before* we started thinking about temporary rules, which is why perhaps we did not immediately see the above "translation".

We need temporary rules and triples anyway, and they are more general than "expected derived triples". So I think we can now collapse the latter into the former, thereby simplifying both the format and its description.

What do you think? Am I missing something in the reasoning above? Is there anything that we can still do with `sh:value` and `sh:defaultValue` that cannot be done with temporary triples/rules?

@HolgerKnublauch @scotthenninger @mgberg

PS. In any case, I still think that "when you need phase separation around the per-layer fixpoint and cleanup semantics" by @scotthenninger in the above text is not correct.

afs: > [Temporary triples](https://github.com/w3c/data-shapes/pull/1216#dfn-temporary-triple) and their reifiers are visible to executing rules.

This and other uses of reifier look like they are "annotations".

The three concepts are (with their links):

* reifier -- subject of an `rdf:refiies` triple -- https://www.w3.org/TR/rdf12-concepts/#dfn-reifier
* reifying triples -  a triple `reifier rdf:reiifies <<( )>>` -- https://www.w3.org/TR/rdf12-concepts/#dfn-reifying-triple
* annotation -- "the subset of triples including the [reifier](https://www.w3.org/TR/rdf12-concepts/#dfn-reifier) as subject" -- https://www.w3.org/TR/rdf12-concepts/#dfn-triple-annotation -- the top level of  `{| |}`

> Delete the derived triples (except those that were also inferred by rules)
> 		and their reifiers

The derived triples section doesn't mention reifiers / annotations.

If adding annotations, do they get deleted even if the derived triple is inferred?

scotthenninger: @liviorobaldo I'd like to focus on the editing of Sections 3.7 (Using `sh:order` and `sh:layer`) and 10 (Variations of SHACL Rule Implementations).
 
On the broader proposal: removing `sh:defaultValue` and `sh:values` from all documents is considerably further reaching than the scope of this PR. Beyond scope, I'd also recommend against it on the merits. I see those properties as having genuine utility for SHACL as a whole, not just here. And while I take the point that "expected derived triples" and temporary triples/rules overlap conceptually, I'd note that your proposed three-rule translation is more complex than what's currently in the spec - both in verbosity and in the intellectual effort required to use it.
 
That said, by all means feel free to propose the change formally in a separate issue. It's worth discussing on its own terms rather than folding it into this PR.
 
On §3.7 and the use of `sh:layer`: let's assume for now that the spec is correct and that derived triples may be present. Making the "intermediate results" explicit is a clear improvement, since as written the phrase is vague about what gets cleaned up. And it's in fact two distinct categories (derived/implicit triples and temporary triples), not just the implicit ones. Would this opening sentence be more correct:
 
> Use `sh:layer`, by contrast, when a later group of rules must run only after an earlier group has reached its fixpoint and after that group's intermediate results - namely its derived (implicit) triples from `sh:defaultValue/sh:values`, except any that were independently inferred by rules, together with any temporary triples marked `sh:tempTriple` true - have been removed.
 
@HolgerKnublauch is that correct? Are temp triples removed after a layer completes? I wasn’t able to find anything that explicitly stated this.
 
On the PS: The reason for using `sh:layer` is intent-level: one has a multi-step computation whose earlier partial results must be fully resolved and then discarded before a later group runs — so those partials don't leak into the final inferences or get re-consumed on later passes. I'd suggest replacing it with something closer to:
 
> In short, use `sh:order` for deterministic sequencing within an iteration pass, and `sh:layer` when a later group must wait for an earlier group's computation to complete and its intermediate results to be discarded — as with Example 10's temporary `ex:offspring` triples, which are needed to reach a fixpoint but must not persist.

HolgerKnublauch: Temp triples are deleted at the very end of the inference process, see the general execution algorithm.

Issue #1196, opened 2026-08-24
Add description of translation between SPARQL-RL syntax and SPARQL CONSTRUCT


HolgerKnublauch:
> [Section 1](https://github.com/w3c/data-shapes/issues/1113#issuecomment-5363062547) belongs in the SHACL Inf Rules doc.
> 
> SRL already provides the relationship to SPARQL pattern matching (as requested). That also serves for SRL implementers.
> 
> The rest is SHACL Rules related. The complete translation is strongly connected to the details of the SHACL Rule framework (`sh:order`,`sh:layer`) and the consequences of evaluating converted SRL rules sets in a SHACL Rules rule sets.
> SHACL Inf Rules is still changing, and has more freedom to, if the SHACL Rules related part is in the SHACL Rules doc.
> 
> It could go in a rules overview document - which might possibly be a section of the SHACL Overview (but others in the WG might reasonably see that as detracting from all the non-rules work (constraints, profile, ui) that has been done by this working group).
> 
> Holger wrote:
> If future versions of SPARQL-RL are maintained by another WG, the editors can evolve that document in different ways if they want to.
> 
> Adding a feature tied closely to the current SHACL Rules impledes the scope to evolve SPARQL-RL. It is material to maintain and keep alignment (W3C stresses compatibly). There would be tests to write, and we'd need implementation experience.
>  

 _Originally posted by @afs in [#1113](https://github.com/w3c/data-shapes/issues/1113#issuecomment-5386161959)_

I don't think anyone argues that having such a translation is useful. The question is where to put it. The SPARQL-RL document already has a matching section that compares SPARQL and SRL, and IMHO adding a slightly more verbose description of how SET relates to BIND+filter and NOT relates to FILTER NOT EXISTS sounds like a sensible thing to do there.

afs:
> I don't think anyone argues that having such a translation is useful. 

What are the use cases?

HolgerKnublauch:
I just looked at https://www.w3.org/TR/sparql12-rl/#srl-sparql-relationship and it seems that the section is 95% there already. I didn't know that the SET vs BIND+FILTER was already spelled out. That is good.

[...]

Issue #1067, opened 2026-07-13
Use skolemization in blank node generation in head templates


afs:
Skolemization has been suggested as a way to remove restrictions on rules that include a blank node in the rule head. It exists in at least one implementation from outside the WG.

The idea is that a rule which will produce the same blank node (i.e. the same RDF Term) every time the head template is processed if the template is given the same input. The input is variables used in the rule head from solution mapping together with an id of the rule within the rule set. These are call _frontier variables_.

Well-formedness means that all variables in the head will have values in the solution mapping.

A _skolem key_ for a solution mapping from a rule is a unique symbol for that solution mapping in the current execution from a particular rule.
Example: Skolem key: `(rule index, var1=value1, var2=value2, ...)` where the variables are in the same order each time (e.g. sorted by name).

For rule set evaluation, there is a blank node allocator function `allocBlankNode(skolemKey, head blank node)` that returns a unique blank node for an argument pair.

With this capabilities, the "[closed dependency](https://www.w3.org/TR/shacl12-rules/#dfn-closed-dependency)" condition for rules with a blank node in the head can be removed.

Example: This simple example gives one triple for every triple matched by `?s ?p "ABC"`. There is one frontier variable, `?p`.

`RULE { [ ] ?p :q } WHERE { ?s ?p "ABC" }`

Different values of `?s`, with the same `?p` do not generate new blank nodes.

osavkovic:
Hi Andy, I think its a good idea. However, there might be a still a small problem.

Right now, if you run a rule twice with the same input, it makes a different blank node each time. Run-once exists to stop that. Your `allocBlankNode(skolemKey, ...)` fixes this directly. The same input always gives the same node.

However (if I got it right) there still a problem if there is a cyclic dependency between predicates — that can still make a new blank node every round, forever, simply because each round's input is different from the round before.

Simple case, a rule that feeds itself, same idea as `p([],Y) :- p(Y,X)`:

`RULE { [] :p ?y } WHERE { ?y :p ?x }`

[...]

Issue #752, opened 2026-01-31
Rule tuples - workspace during rule evaluation


afs:
[Datalog](https://en.wikipedia.org/wiki/Datalog) is based around "atoms". An atom consists of a name (a predicate symbol) and zero or more terms, where each term is a constant or a variable. Atoms are also called literals. Atoms do not contains atoms (unlike prolog).

Viewed as datalog, SHACL Rules has one type of atom used as triple/triple pattern/triple template, depending on position.

Having other tuples forms can be useful during rule execution.

A rule might emit a tuple to represent intermediate matching results to be shared by several other rules. Without some working space during execution, either the same pattern will be repeated in the other rules, or some intermediate triples are needed and these will end up in the output inference graph.

Rule tuples are workspace elements and are disjoint from triples. They are tuples of RDF terms (no variables).

Syntax of tuple patterns, templates and tuples:

```
TUPLE(termOrVar , ...)
```
with a shorthand:
```
$(termOrVar , ...)
```

Often, the first argument will be a fixed name - think of it as an SQL table name or a programming language struct.

There is a tuple store which holds tuples for the lifetime of the evaluation.
The tuple store holds duplicate data tuples (unlike an RDF graph which is a set).

A rule tuple pattern can appear in the body of a rule, where it matches in a similar fashion to a triple pattern -- look in the tuple store for matches based on arity and any constant RDF terms, returning a sequence of solution mappings. (Implementation note: indexing by first argument will greatly speed up access for the "fixed name" style.)

A rule tuple template can appear in the head of a rule, where it acts to generate grounded data tuples (i.e. no variables) which are added to the tuple store.

A rule data tuple can appear in a `DATA` block, where it is added to the tuple store at the start of execution.

The tuple store is ephemeral. It is not defined as an output of the evaluation process.

afs:
-- Converted the issue description into more of a summary --

Full description:

[Datalog](https://en.wikipedia.org/wiki/Datalog) is based around "atoms". An atom consists of a name (a predicate symbol) and zero or more terms, where each term is a constant or a variable. Atoms are also called literals. Atoms do not contains atoms (unlike prolog).

Viewed as datalog, SHACL Rules has one type of atom used as triple/triple pattern/triple template, depending on position.

Having other tuples forms can be useful during rule execution.

[...]

PR #1219, opened 2026-08-29
Proposal for CONSTRUCT as alternative syntax for SPARQL-RL

HolgerKnublauch:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/sparql-rl-construct/sparql12-rl/index.html#desc-construct-syntax)

Motivation: https://github.com/w3c/data-shapes/issues/1196#issuecomment-5465630556


afs (commented):
I'm not clear what benefit this brings SPARQL-RL users or to SHACL Rules users. Please explain.

SPARQL-RL executes rule sets.  A rule set is a document.
A SPARQL Query is a document.

SHACL Rules executes document units or triple rules at each step.

Familiarity of syntax is a trap:
```
CONSTRUCT { ?s :p true } WHERE { ?s :q 123 } 
CONSTRUCT { ?s :p false } WHERE { ?s :q 456 }
```
is not a `sh:SPARQLRule`.

This is a limited subset of CONSTRUCT - is it a useful real-world subset? I don't think it is because the fragments will be small. The value of SHACL Rules is the compute (aggregates, for example). Larger CONSTRUCT will involve syntax outside the subset.

Translation from this limited subset of CONSTRUCT covered here to RULE can be done by text processing. So the user doesn't have to see it.  Go outside that subset and the guarantees of SPARQL-RL do not apply and it's not SPARQL-RL.

Translation from SPARQL-RL rule set to SPARQL CONSTRUCT is also a text processing task and the user doesn't have to see it. It then needs to deal with multiple RULEs.

(The AI generated material contains many mistakes - I have read it in an optimistic light)

HolgerKnublauch:
Hi Andy, thanks for your response. Here is my opinion.

We are in a W3C working group.
The mission of the W3C is to define standards.
Standards are there to improve interoperability.

For example, it is great that SPARQL is a standard because it allows tools to interoperate and users to learn only one query language for the W3C semantic web stack instead of RQL, RDQL (Jena), SeRQL (Sesame/RDF4j). SHACL-AF has reused SPARQL CONSTRUCT syntax because it means that most potential users are already familiar with it, and implementers like us can reuse existing infrastructure such as YASQE. It also means that users can try our their SPARQL queries before copy-and-paste into a SHACL rules graph.

What you are proposing (and what is currently implemented) is that there would be two standards released by the Data Shapes WG, and in order to interoperate, implementations would need to convert between them using "text processing". In other words, SHACL Inference Rules and SPARQL-RL would interoperate just like SHACL Inference Rules and N3 or RIF or SWRL or the native Jena rule format. This IMHO completely misses the point of having standards. If that is the outcome of our WG then we have failed to complete our mission.

What this PR is proposing is to allow SHACL and SRL to interoperate by (slightly) extending SRL to understand a syntax that is compatible with a subset of SPARQL CONSTRUCT. This is technically possible because in terms of expressiveness and features, SRL is essentially a subset of CONSTRUCT (see my various discussions about SPARQL-Full vs SPARQL-RL).

I regard SRL as one possible subset of SPARQL along many others, similar to how OWL-RL and OWL-QL relate to OWL-Full. You stated yourself that you expect SRL to evolve further, so who knows if future versions of SRL will cover an even larger subset.

> I'm not clear what benefit this brings SPARQL-RL users or to SHACL Rules users. Please explain.

## Benefits to SHACL users:
- Seamless transition between the subsets. If they stay within SRL then fine, but they don't necessarily have to.
- SRL rule sets can be executed with vanilla SHACL engines, alongside other SHACL rules, assuming the layers are set correctly.
- Likewise, if they stay within SRL, a complete SHACL rules document can be sent to an SRL engine for execution with a trivial conversion into an .srl document.
- Users can hook into algorithms such as Stratification for development and debugging of SHACL rules. For example, they can write arbitrary SHACL rules, the system can identify those that fit into the SRL subset and then propose sh:layer and sh:runOnce values automatically, or flag mistakes in the layering.
- Users just need to learn one syntax and stay within the CONSTRUCT syntax that they already know.

## Benefits to SRL:
- Easier learning curve for newcomers. Most will already know CONSTRUCT.
- The name SPARQL-RL would actually make sense. No longer just a SPARQL-inspired own syntax.
- Ability to execute with SHACL Inference Rules engine which understands features like sh:deactivated.
- SRL can meaningfully be sold as a deliverable of the data shapes WG and no longer violates the charter.
- Copy and paste between SPARQL editors and the RDF document, e.g. to quickly test individual rules against a SPARQL endpoint.
- If future versions of SRL cover more of SPARQL, there is nothing left to do as the sh:construct syntax can still be used.

> SPARQL-RL executes rule sets. A rule set is a document. A SPARQL Query is a document.
> 
> SHACL Rules executes document units or triple rules at each step.
> 
> Familiarity of syntax is a trap:
> 
> ```
> CONSTRUCT { ?s :p true } WHERE { ?s :q 123 } 
> CONSTRUCT { ?s :p false } WHERE { ?s :q 456 }
> ```
> 
> is not a `sh:SPARQLRule`.

Correct, this is not a sh:SPARQLRule. And I am not proposing that. What I have done over the last few months is to extend SHACL Inf Rules to allow SRL to be represented and executed naturally. I have added sh:layer, sh:runOnce and sh:RuleSet for these exact reasons.

So the real representation of the example above would be a sh:RuleSet that has two sh:SPARQLRules. With a little bit of fine tuning, we can tell the SHACL engine that it needs to execute rule sets together as a unit, matching the expectations of the SRL engine. Basically: if a SHACL document contains a mix of general SPARQL rules and SRL-compatible CONSTRUCT rules, place all SRL rules into the lowest layers and then the general (more expressive) rules in higher layers. This way, the design assumptions of SRL still hold.


> This is a limited subset of CONSTRUCT - is it a useful real-world subset? I don't think it is because the fragments will be small. The value of SHACL Rules is the compute (aggregates, for example). Larger CONSTRUCT will involve syntax outside the subset.

Be careful what you wish for :) If you say that SRL is not a useful subset of CONSTRUCT then its whole expressiveness would be of limited use. I don't think you mean that. And I don't believe the SRL-CONSTRUCT subset is too small. I just walked through my SHACL rules that represent OWL-RL and the vast majority of them fall into the simple subset, while the remaining ones rely on rdf:rest*/rdf:first which doesn't exist directly in SRL and which would need additional rules to "flatten" these into temporary triples that directly attach the list members to their list.

> Translation from this limited subset of CONSTRUCT covered here to RULE can be done by text processing. So the user doesn't have to see it. Go outside that subset and the guarantees of SPARQL-RL do not apply and it's not SPARQL-RL.
> 
> Translation from SPARQL-RL rule set to SPARQL CONSTRUCT is also a text processing task and the user doesn't have to see it. It then needs to deal with multiple RULEs.

This is exactly what we should NOT do, as it defeats the purpose of having W3C standards as explained above.

> (The AI generated material contains many mistakes - I have read it in an optimistic light)

Hopefully quite easy to fix.

Note that I am not saying that the current SHACL Inference Rules draft is perfect to embed SRL. I would love to work with you to make it as good as possible though. So if you tell us what you need to see changed to better handle SRL rule sets then we will try to accommodate that. In particular I would suggest we introduce a formal plugin mechanism where the SRL stratification algorithm can plug into the engine without violating SRL guarantees. For example, whenever sh:layer and sh:runOnce are not specified, the SRL plugin could compute them. Or we could add some sh:sparqlProfile to explicitly state whether are rule or ruleset is expected to remain in the SRL CONSTRUCT subset and not the full CONSTRUCT subset.

recalcitrantsupplant:
I agree on the goals here but I think they are interoperable without changing the syntax. To me the most interoperable form would be to share the grammar where the meaning is the same, and only differ where there is some difference. This would follow "least surprise".

> - Seamless transition between the subsets. If they stay within SRL then fine, but they don't necessarily have to.
> - SRL rule sets can be executed with vanilla SHACL engines, alongside other SHACL rules, assuming the layers are set correctly.

Implementations that implement one or more forms will need to parse the syntax in any case, at which point a translation is trivial.

> - Copy and paste between SPARQL editors and the RDF document, e.g. to quickly test individual rules against a SPARQL endpoint.

This is useful, it's also the case where context (SRL, SPARQL) is quickly lost (chat messages etc.) and so the  original author's intent is unclear. At least if the conversion is explicit, e.g. via an export action, there is an opportunity to warn of differences or add comments etc.

HolgerKnublauch:
> I agree on the goals here but I think they are interoperable without changing the syntax.

Hi David, could you clarify what you are proposing exactly? How would they be interoperable without changing the syntax - we would again require two parsers. The heart of the proposal here is that with CONSTRUCT syntax the SRL rules can be directly understood as SHACL rules, too.

> To me the most interoperable form would be to share the grammar where the meaning is the same, and only differ where there is some difference.

The CONSTRUCT proposal does not change the meaning. They are just normal CONSTRUCT queries, only limited to a subset of CONSTRUCT in the same way like a profile (OWL-RL vs OWL-Full).

> This would follow "least surprise".

Most people new to SRL, reading SPARQL-RL in the title, will also be surprised that this is not really the SPARQL that they expect. They will be surprised that some functions don't exist for example. They will have to understand the new SET keyword. There are also further well-formedness rules https://www.w3.org/TR/sparql12-rl/#wellformed that users will need to learn before using either syntax. So how is that different?

What is needed in both cases is the custom parser of your grammar. Only that will give users enough feedback about what is allowed and what is not. And that parser can be placed either on top of a standard SPARQL editor (for CONSTRUCT) or used as only parser (native SRL files). Finally, the rule engine will throw errors on malformed rules, in both cases of RULE or CONSTRUCT.

recalcitrantsupplant:
> could you clarify what you are proposing exactly

That the syntax is interoperable without introducing an alternative syntax, and that most of the benefits listed are also met already.

> How would they be interoperable without changing the syntax - we would again require two parsers.
...
> What is needed in both cases is the custom parser of your grammar. Only that will give users enough feedback about what is allowed and what is not

I’m struggling to reconcile these two points - what would a single parser, say SPARQL, do with rules intended to be executed as SRL? It would need to at a minimum manage translation to INSERT, or merge CONSTRUCT results with a data graph, add fixed point logic - I would think these are more involved than a syntax transformation.

The interoperability comes from the majority of the syntax overlapping with SPARQL. This reduces the amount of new concepts a user needs to learn. For implementations, it enables code reuse, and regardless of the syntax, the differences are where an implementation will need to necessarily introduce different handling. I don't see the remaining differences as a barrier to interoperability; as soon as an implementation is involved it is trivial to convert. 

> Most people new to SRL, reading SPARQL-RL in the title, will also be surprised that this is not really the SPARQL that they expect. They will be surprised that some functions don't exist for example. They will have to understand the new SET keyword. There are also further well-formedness rules https://www.w3.org/TR/sparql12-rl/#wellformed that users will need to learn before using either syntax. So how is that different?

Users will need to learn something when looking at the new language, whether it is a subset (which parts are supported), or it is a subset with extensions. I don't think that is avoidable. It is least surprising if:
- Where tokens are reused, they have the same meaning in both languages.
  - if tokens are reused but have different meanings, at worst a user does not realise, at best they would likely be more surprised to learn that it had a different meaning (compared to if it had the same meaning)
- Where new tokens are introduced, they have new meanings
- (and perhaps, where tokens are similar, they have similar meanings)
Only the current use of FILTER does not follow this, with the benefit of users not needing to include BOUND statements.

> The CONSTRUCT proposal does not change the meaning.

The recursion/stratification is inherent in SRL - that is the difference in meaning.

Query strings inevitably get passed around without context. I have been sent turtle with the prefixes stripped; dealt with web scraped sources which have no mime types or file extension etc. In this vein, I would not like to be sent what appears to be a SPARQL CONSTRUCT query to later find out that the author intended it to be an SRL rule.

> What is needed in both cases is the custom parser of your grammar. Only that will give users enough feedback about what is allowed and what is not.

I agree - given this need, I reach a different conclusion: as such parsers can easily then do the transformations, why not have a syntax that is self contained/cannot be confused, and allow this implementation to transform on demand.

`RULE <iri> CONSTRUCT` and `RULE CONSTRUCT` (which I think would be more consistent than plain `CONSTRUCT`), would address my concern around copying strings around without context, but not the others.

HolgerKnublauch:
> I’m struggling to reconcile these two points - what would a single parser, say SPARQL, do with rules intended to be executed as SRL? It would need to at a minimum manage translation to INSERT, or merge CONSTRUCT results with a data graph, add fixed point logic - I would think these are more involved than a syntax transformation.

I have the feeling we are talking about very different things. Therefore we cannot seem to agree on anything and just turn in endless circles.

I am trying to solve the integration on the individual rule level while you and Andy point out that all SRL rules need to be executed together. I want to enable users of a SHACL rule editing and execution tool to have a unified experience, and our RDF files to share the same triple syntax. I don't care much about the implementation details of how those rules are executed under the hood. Engines may do the loop with a vanilla SHACL engine using layers or process them natively with an SRL engine. I don't care and this is unrelated to the syntax. So when I talk about using SPARQL, I mean this on the micro-level of each rule, not as a way to execute all rules at once and iterate with fix point logic.

> The interoperability comes from the majority of the syntax overlapping with SPARQL. This reduces the amount of new concepts a user needs to learn. For implementations, it enables code reuse, and regardless of the syntax, the differences are where an implementation will need to necessarily introduce different handling. I don't see the remaining differences as a barrier to interoperability; as soon as an implementation is involved it is trivial to convert.

"majority of syntax overlapping with SPARQL" is not good enough. With the current RULE+SET vs CONSTRUCT+BIND proposals there is not enough overlap, and some manual conversion is still needed. This makes us fail on our W3C standardization task. It's just like with SRL vs N3 or SHACL rules vs SWRL or any combination thereof. Not good enough when a clearer solution is obvious by allowing SRL rules to be written as a subset of SPARQL CONSTRUCTs.

> Users will need to learn something when looking at the new language, whether it is a subset (which parts are supported), or it is a subset with extensions. I don't think that is avoidable. 

Yes I agree. Some learning is required.

>   Only the current use of FILTER does not follow this, with the benefit of users not needing to include BOUND statements.

Did you mean SET instead of FILTER above?

> 
> > The CONSTRUCT proposal does not change the meaning.
> 
> The recursion/stratification is inherent in SRL - that is the difference in meaning.

No I am not asking for any of that as explained above. All I want is integration of the syntax of individual rules.

> Query strings inevitably get passed around without context. I have been sent turtle with the prefixes stripped; dealt with web scraped sources which have no mime types or file extension etc. In this vein, I would not like to be sent what appears to be a SPARQL CONSTRUCT query to later find out that the author intended it to be an SRL rule.

All I want to enable here is that people can use regular SPARQL syntax to edit and test individual rules.

afs:
#### SPARQL-RL evolution

> If future versions of SRL cover more of SPARQL, there is nothing left to do as the sh:construct syntax can still be used.

> If you say that SRL is not a useful subset of CONSTRUCT then its whole expressiveness would be of limited use.

SPARQL-RL may involve in ways that a single `CONSTRUCT` can not express - or even any combination of `CONSTRUCT`.

A few examples:

* Syntax (same expressibility, ease of use) for the enriched path capability so it can be written as one rule.
  The `CONSTRUCT` form requires at least two `CONSTRUCT`.
* `INVERSE`/`TRANSITIVE`/`SYMMETRIC` #779
* Rule tuples #752
* Synonyms - `owl:sameAs`, `owl:equivalent*` without an explosion of inferred triples. See also backward chaining.
* Addressing the run-to-infinity problem for blank nodes (one external implementation ready does this) #1067
* Well-behaved aggregates for rules

Translation can handle the case of one `RULE` becoming two `CONSTRUCT`, or even 2+ `RULE` becoming one `CONSTRUCT` as an optimization by knowing that _for the given data_, there is a faster way maybe using `sh:order`.

Translation puts an indirection step so that there isn't a syntax-lock-step.

When translating, the validity for use in SHACL Rules can be checked. Validity is not limited to syntax.

#### Use of rule sets published by 3rd parties

If only `CONSTRUCT` is supported by SHALC Inf Rules, then many SPARQL-RL publications will effectively be out of reach to SHACL Inf Rules.

#### Translation to `sh:RuleSet`

To make the stratification calculation available, translation of _a SPARQL-RL rule set_ would provide a `sh:RuleSet`.

1. Conversion of rules, including future forms - where possible and legal - to SPARQL Query `CONSTRUCT`, errors if not possible.
2. Package as a single graph with a named SHACL Rule rule set, using `sh:layer`. Translated rules URIs are first-class.

The SHACL Inf Rules user can use `owl:imports` and pick by URI, or use it as a well-behaved unit as a global rule (see below).

#### To make SHACL Inf rules better

1. Be able to execute an `sh:RuleSet` as a rule step or some other way to allow nesting, so that 3rd party published `sh:RuleSets` of any kind can be incorporated in the local shapes graph (and without numbering clashes).
2. Define what an extension can - and can't - do. For example, `sh:layer` and `sh:runOnce` are reserved by the framework.
3. Tests.

HolgerKnublauch:
(I think Andy's message was written before the Rules TF meeting in which I participated too. We had discussed the same questions there, so some of my responses are just a record in case anyone is reading this).

> #### SPARQL-RL evolution
> > If future versions of SRL cover more of SPARQL, there is nothing left to do as the sh:construct syntax can still be used.
> 
> > If you say that SRL is not a useful subset of CONSTRUCT then its whole expressiveness would be of limited use.
> 
> SPARQL-RL may involve in ways that a single `CONSTRUCT` can not express - or even any combination of `CONSTRUCT`.
> 
> A few examples:
> 
> * Syntax (same expressibility, ease of use) for the enriched path capability so it can be written as one rule.
>   The `CONSTRUCT` form requires at least two `CONSTRUCT`.
> * `INVERSE`/`TRANSITIVE`/`SYMMETRIC` [Abbreviations : `TRANSITIVE`, `SYMMETRIC`, and `INVERSE` #779](https://github.com/w3c/data-shapes/issues/779)
> * Rule tuples [Rule tuples - workspace during rule evaluation #752](https://github.com/w3c/data-shapes/issues/752)
> * Synonyms - `owl:sameAs`, `owl:equivalent*` without an explosion of inferred triples. See also backward chaining.
> * Addressing the run-to-infinity problem for blank nodes (one external implementation ready does this) [Use skolemization in blank node generation in head templates #1067](https://github.com/w3c/data-shapes/issues/1067)
> * Well-behaved aggregates for rules

And vice versa: SPARQL CONSTRUCT may be extended with new functions or features.

But that is fine. If CONSTRUCT and RULE can stay in lock-step then this is fine (and for 1.2 they do), but if they diverge that is also fine. Adding CONSTRUCT now does not prevent the evolution of RULE syntax.

> 
> Translation can handle the case of one `RULE` becoming two `CONSTRUCT`, or even 2+ `RULE` becoming one `CONSTRUCT` as an optimization by knowing that _for the given data_, there is a faster way maybe using `sh:order`.
> 
> Translation puts an indirection step so that there isn't a syntax-lock-step.
> 
> When translating, the validity for use in SHACL Rules can be checked. Validity is not limited to syntax.

Right, I understand even for CONSTRUCT there are other checks needed, as well-formedness rules. An SRL engine will do these tests anyway, regardless of RULE or CONSTRUCT.

> 
> #### Use of rule sets published by 3rd parties
> If only `CONSTRUCT` is supported by SHALC Inf Rules, then many SPARQL-RL publications will effectively be out of reach to SHACL Inf Rules.

And vice versa. If you don't support CONSTRUCT then all SHACL Rules are out of reach unless converted. And there are plenty of examples where CONSTRUCT users are not even aware that their rules fall into the SRL subset.

> 
> #### Translation to `sh:RuleSet`
> To make the stratification calculation available, translation of _a SPARQL-RL rule set_ would provide a `sh:RuleSet`.
> 
> 1. Conversion of rules, including future forms - where possible and legal - to SPARQL Query `CONSTRUCT`, errors if not possible.
> 2. Package as a single graph with a named SHACL Rule rule set, using `sh:layer`. Translated rules URIs are first-class.
> 
> The SHACL Inf Rules user can use `owl:imports` and pick by URI, or use it as a well-behaved unit as a global rule (see below).

Not sure what you mean here TBH.

> #### To make SHACL Inf rules better
> 1. Be able to execute an `sh:RuleSet` as a rule step or some other way to allow nesting, so that 3rd party published `sh:RuleSets` of any kind can be incorporated in the local shapes graph (and without numbering clashes).
> 2. Define what an extension can - and can't - do. For example, `sh:layer` and `sh:runOnce` are reserved by the framework.
> 3. Tests.

Ok certainly doable. Specific suggestions welcome.

afs:
(from the meeting)

> And vice versa. If you don't support CONSTRUCT then all SHACL Rules are out of reach unless converted.

Partial translation from SHACL Inf rule to SPAQL-RL - which definitely need checking applied so code is already involved.

> Adding CONSTRUCT now does not prevent the evolution of RULE syntax.

I disagree it is so simple. It's an unnecessary burden.

1. The natural expectation is there is a full(ish) coverage by CONSTRUCT.
2. Copying a `CONSTRUCT` over isn't a SPARQLRule - that has e sh:order, sh:layer, (and node expr for sh:deactivated?) It is not only about syntax.
3. There are now ways to write the same thing which is confusing.

HolgerKnublauch:
None of these choices are black and white. Both viewpoints are valid. It's a trade-off.

> > Adding CONSTRUCT now does not prevent the evolution of RULE syntax.
> 
> I disagree it is so simple. It's an unnecessary burden.
> 
> 1. The natural expectation is there is a full(ish) coverage by CONSTRUCT.

IMHO the natural expectation of a language called SPARQL-RL is that
1. It will support SPARQL
2. It will support -RL similar to how OWL-RL relates to OWL Full.

You are also right that when CONSTRUCT is used then some people will expect to see all features, but the same applies when users see a FILTER in SRL native RULE syntax yet are not allowed to use bound, coalesce and other excluded functions. It needs to be framed in the documentation and the parser with the SRL grammar will report issues either way.

> 2. Copying a `CONSTRUCT` over isn't a SPARQLRule - that has e sh:order, sh:layer, (and node expr for sh:deactivated?) It is not only about syntax.

Using sh:deactivated should be trivial. If such rules use sh:order, sh:layer or sh:runOnce directly then they cannot be sent to SRL for execution, except to verify that the layers and runOnce triples are correct. But that can be defined in the mapping.

> 3. There are now ways to write the same thing which is confusing.

Again, depends on viewpoint. It's either confusing or liberating to people who already know CONSTRUCT.

All these are opinions and nobody here can claim to know the absolute truth as we are inventing stuff on the fly without proper user experience.

The usual W3C process is trying to find compromises. The process has been designed to resolve different opinions. These (longish) Git discussions are one way to resolve them. Voting is another. People can raise serious concerns by voting -1 or formal objections. To avoid this, often the compromise is to include multiple options into the standard and let the users decide.

So moving forward, it is IMHO quite conceivable to have a hybrid solution in which 
a) SRL also allows CONSTRUCT syntax
b) SHACL includes a plugin point for stratification to be cleanly injected
c) SRL also defines a rule type for complete .srl snippets
d) SRL also defines a rule type for individual RULE syntax rules

However, I believe the easiest would be to only do a) and b) while c) and d) remain options for future versions when SRL indeed needs new language features that are not representable as individual CONSTRUCT rules (such as TRANSITIVE). But that isn't really needed now in 1.2. Doing a) and b) means that SRL doesn't need to worry much about SHACL at all.

PR #1220, opened 2026-08-31
[UI] Add SHACL UI logical constraints section

### LINKED ISSUE #533: [UI] Support for Core Constraint Components
SHACL UI should add support for core constraint components given that they are widely used and well supported by existing validation processors.

Add support for core constraint components for SHACL user-interfaces.

Discussion:
- Are all core constraint components supported?
- Are all core constraint components required?
- What are some core constraint components that should be optional due to their UI complexity?
- How are they used in view versus edit mode?

Links:
- [SHACL 1.2 Core - Core Constraint Components](https://www.w3.org/TR/shacl12-core/#core-components)

edmondchuc:
Initial placeholder section.

At a minimum, this PR will include at least support for `sh:or` with `sh:datatype` and `sh:class`, as described in [Form Generation using SHACL and DASH 3.1 Multiple Datatypes or Classes](https://datashapes.org/forms.html#multi).

Closes #533 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/logical-constraints/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -1827,10 +1827,13 @@ ex:PersonShapeName
       </section>
     </section>
 
-    <section id="core-constraints">
-      <h2>Core Constraints</h2>
+    <section>
+      <h2>Logical Constraints</h2>
 
-      <p class="ednote">TODO: This section may not be needed if we decide to fold in the use of different constraints for common form scenarios under the Patterns section.</p>
+      <p class="ednote">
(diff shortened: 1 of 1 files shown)


smessie:
I don't think that this PR should already close the mentioned issue, as this just adds a placeholder for where the contents should follow. The issue should remain open to actually follow up on the process of adding the actual contents.

PR #1224, opened 2026-09-01
Removing statements that require certain features to be implemented

danielbeeke:
From my perspective and per my interpretations of @caribouW3 comments we should not focus on conformance in this phase. The spec should be a set of features. Conformance text inside the chapters (features) is about the feature only, if it is implemented. Conformance is something to discuss later.

I would be happy to merge this if:

- @caribouW3 can conform that my interpretation regarding phases is correct, that there MUST be a phase where the task force or the working group must settle on conformance levels.
- @caribouW3 can conform that my interpretation regarding writing the spec is correct, that the spec itself is a set of features, where features can contain MUST / MAY / SHOULD and that those words only say something about that feature when it is implemented.
- Everyone in the taskforce agrees to merge this.

Essentially this would make the current working draft a set of features, without explicit requirements which features need to be implemented. We will get to deciding that in one of the next phases of writing the spec.

It remains of course that we have conversation about the features and if they are required, because the features can create an implicit dependency graphs which basically already gives input to the conformance discussion.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/removing-statements-about-features-that-must-be-implemented/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -643,9 +643,6 @@
           term <code>"false"^^xsd:boolean</code>.
         </p>
       </section>
-      <section id="conformance">
-        <p>TODO: define conformance criteria for SHACL UI implementations.</p>
-      </section>
 
       <section id="compatibility">
         <h2>RDF Abstract Model Compatibility</h2>
@@ -3634,8 +3631,7 @@ ex:Name a sh:PropertyShape ;
       <h3>Built-in Property Roles</h3>
 
       <p>
-        The following sections define the instances of <code>shui:PropertyRole</code> in the SHUI namespace. These property roles MUST be supported
-        by SHACL Renderers. Additional property roles may be defined in other namespaces.
+        The following sections define the instances of <code>shui:PropertyRole</code> in the SHUI namespace. Additional property roles may be defined in other namespaces.
       </p>
 
       <h4>shui:LabelRole</h4>


caribouW3:
I confirm that what is required to implement for conformance to a given conformance class (or level or whatever might be defined) can be defined later (and preferably only in a dedicated conformance section).

shacl12-ui/index.html
```diff
@@ -643,9 +643,6 @@ <h3>Document Conventions</h3>
           term <code>"false"^^xsd:boolean</code>.
         </p>
       </section>
-      <section id="conformance">
-        <p>TODO: define conformance criteria for SHACL UI implementations.</p>
-      </section>
```
edmondchuc, line 648: Should we leave this section in as a reminder to revisit it at a later date? We could add an ednote explaining that the intention is to discuss and define conformance at a later stage. This would also provide useful context for readers outside the TF/WG.

edmondchuc (approved):
I agree with the general direction, and I think this captures where we landed in yesterday's TF call regarding conformance. Thanks @danielbeeke!

shacl12-ui/index.html
smessie: Yes, I think that makes sense. This way, people working on the current draft can see that we still intend to address the topic of conformance, rather than leaving them with incorrect assumptions about the definition of conformance.

Issue #1228, opened 2026-09-03
Define the requirements for connecting shape instances to rules.


afs:
As a framework, SHACL Inf rules does not say how a new rule type should provide shape rule processing.

Pre-binding is SPARQL specific.

For all rule types, the requirement seems to be to restrict rules to the locale of a shape instance based on the target nodes, with `sh:condition` having been applied, to produce a set of focus nodes, supplied by the framework.

Maybe call this "shape focus scope".

For SPARQLRules, it would be outcome of pre-binding #999.

For "triple rules", it is `[shnex:var "focusNode"]` or the default for `sh:subject`.

afs:
It seems this may be requirements as well:
https://github.com/w3c/data-shapes/pull/1230#discussion_r3929530283

> ... a normal SHACL rules engine where each sh:Rule indeed represents a rule can offer things like cleanly canceling execution between rules, measuring individual rule performance,

HolgerKnublauch:
I am not sure what we can do to specify this further because rule types may be very different from what sh:SPARQLRule and sh:TripleRule do. So processing the focus node is IMHO difficult to constrain.

The current wording is

[...]

Issue #1226, opened 2026-09-01
A (longish) review of the SHACL 1.2 Inference Rules spec


osavkovic:
Also of interest to the rest of SRL team: @afs, @robert-david 

I recently became aware of the new shacl12-inference-rules.

My comments are only about the semantics, where I think the document currently leaves more open than it should.

I have been working with rules-like languages for most of my academic life, so I feel obliged to point out some problems :)

Just to add, I did only a first pass. If a closer look would be useful, I am happy to do one.

Organization

- I first present major/minor problems
- At the end, I discuss three possible ways to try to resolve this

# 1. Major

"The general execution algorithm described above is intentionally kept generic and offers a lot of flexibility to specific implementations. In particular, the algorithm is non-deterministic in the sense that unless the order of rules is specified explicitly, the results may differ across executions. This is, for example, the case when rules draw conclusions from the number of certain triples yet those triples may be produced by other rules. In this document, the responsibility of producing a predictable ordering and layering/grouping of rules is left to the rule author."

In essence, we are proposing something that is non-deterministic and leaves predictable ordering to the rule author and the implementer, which in turn makes conformance unpredictable. I think we should allow freedom in *how* to compute, but freedom in *what* to compute is a different thing, and every comparable specification (SQL, the RDF entailment regimes, the standard Datalog fragments) is strict about that.

Below I illustrate more what are those choices where the user/implementer can have confusions.

## 1.1 The order rules run in can change the output

Two rules, and both are in the default layer `0`:

```turtle
ex:RuleP
    a sh:SPARQLRule ;
    sh:construct """
        CONSTRUCT { $this ex:p true }
        WHERE { $this a ex:Node . FILTER NOT EXISTS { $this ex:q true } }
        """ .

ex:RuleQ
    a sh:SPARQLRule ;
    sh:construct """
        CONSTRUCT { $this ex:q true }
        WHERE { $this a ex:Node . FILTER NOT EXISTS { $this ex:p true } }
        """ .
```

In words: *add `p` if `q` is missing*, and *add `q` if `p` is missing*.
Start with a node that has neither.

- Run `ex:RuleP` first: `p` is added; `ex:RuleQ` no longer fires. **Output: `p`.**
- Run `ex:RuleQ` first: `q` is added; `ex:RuleP` no longer fires. **Output: `q`.**

So for the same data, the same rule set may produce two different results, decided by nothing but which rule the engine happened to run first.

### Where the non-determinism actually comes from

[...]

Issue #1113, opened 2026-08-03
Downplay SHACL Rules extension point or move SHACL-SPARQL rules into separate document


HolgerKnublauch:
The SHACL SPARQL Rules document currently still talks about a generic extension point for other rule languages, but it's hand-waving a lot there. The question arose if this is a generic SHACL Rules framework then it should be moved out of the SHACL-SPARQL document into a separate SHACL Rules document. To simplify editing and keeping the expectations realistic, it may be best to make this document about SHACL-SPARQL Rules only, and mention that this is potentially an extension point just in a side remark. It will also improve the flow of the document I hope.

liviorobaldo:
I am not entirely sure how you intend to modify the document. To me, it seems that the current version already "mentions that this is potentially an extension point only as a side remark" and then clearly states that: "This document defines one such rule type: SPARQL rules.".

I agree that the document should focus exclusively on SHACL-SPARQL Rules and mention extensions to other rule languages only as a side remark. However, it seems to me that the current version already does so.

mgberg:
I'm curious how much work it would be to pull these rules out into a separate SHACL 1.2 Rules document once SRL is moved to its final location as mentioned in the meeting earlier. It seems there would be a number of benefits to doing so, including:

- The different capabilities would be more cleanly separated.
    - The SPARQL document would contain only validation-related features.
    - The rules engine, which is largely a separate/additional capability from the validation engine (except for `sh:condition`), would have its own document instead of being just one section of the current document.
- The rules document could lean slightly into the fact that there could be other types of rules. If this is done:
    - We can leverage the fact that this framework can be extensible with custom features because of the relatively simple set of constraints on how a rule has to operate, as opposed to SRL which presumably could not be extensible in a similar manner because it is Datalog-based and due to the guarantees it seeks to provide.
    - Triple rules could (but not necessarily) be added back in since they already exist in AF.
    - Some implementations already have custom rule types (like TopQuadrant's JavaScript rules).
    - It could conceivably be the basis for a bridge between these rules and SRL (regardless of whether or not this is actually pursued).
- The content of what would be the new document is largely already written here.
- Any shared content with the SPARQL document could be referenced and reused (e.g. pre-binding).
- It would be more clear that something that supports SHACL-SPARQL does not have to include support for rules (although admittedly that is currently stated in the document).

HolgerKnublauch:
@mgberg This is certainly possible and does have some advantages, provided this doesn't risk our deliveries time line.

[...]

PR #1230, opened 2026-09-03
Add section for SRL-based rules

### LINKED ISSUE #1229: Add SPARQL-RL as a rule type in the SHACL Inf Rules framework.
SPARQL (Query) needs no change to work with SHACL Inf Rules.  SHACL Inf Rules provides the linkage to SHACL shape processing with [pre-binding](https://www.w3.org/TR/shacl12-sparql/#pre-binding). Pre-binding passes shape instance URIs into a CONSTRUCT template for `$this, allowing the rule to localize matching to a particular shape instance.

Including SPARQL-RL in SHACL Rules can be done in the same style.
This would be another built-in rule type along side SPARQL and triple rules.

Implementation (like the other rule types) is optional.
https://w3c.github.io/data-shapes/shacl12-inference-rules/#Built-in-RuleTypes
>> Note that not all implementations are required to implement all of these types for conformance.

&nbsp;

SPARQL-RL has guarantees and performance techniques of datalog not offered by SHACL SPARQL Rules.
SHACL SPARQL Rules has unique properties not offered by SPARQL-RL.

On #954 we have this example:
```turtle
[
      rdf:type srl:Rule;
      srl:text """
          RULE { ?x :bothPositive true . }
          WHERE { ?x :p ?v1  FILTER ( ?v1 > 0 )  ?x :q ?v2  FILTER ( ?v2 > 0 )  }
      """
]
```

That is only one `RULE`. It does not carry the necessary information for an evaluation order necessary to given the properties of SPARQL-RL in conjunction with other rules.

To keep these properties, more than a single SPARQL-RL `RULE` is allowed in one execution invocation.
`RULE`s go together to form application tasks; each task has guarantees.

This would also reduces work needed by a user in deciding SHACL Rules layers (layers didn't exist until recently).

This example calculates the "connected" paths rooted at `$this` 

```turtle
[
      rdf:type sh:SRL;
      rdfs:comment "Possible routes, avoiding weak links (links less than or equal to 5)" ;
      sh:prefixes ...
      sh:srlProgram """
      	  ## Condition for a safe link.
          RULE { ?x :weightSafe ?z } WHERE { ?x :link ?z {| :weight ?w |} FILTER(?w > 5) }
	  
          ## Initialization step
          RULE { $this :connected ?y } WHERE { $this :weightSafe ?y }
	  
          ## Induction step
          RULE { $this :connected ?y } WHERE { $this :connected ?z . ?z :weightSafe ?y }
      """ ;
]
```

This uses the word "program" because SHACL Inf Rules added the same words "rule set". A SPARQL-RL rule set and a SHACL Inf Rules rule set are not the same thing so the word "program" is used above until we find another name. "program" is taken from datalog terminology.

&nbsp;

SHACL Inf Rules is then an interleaving of steps of `CONSTRUCT` operations and steps of well-defined inference.
Interleaving steps meets the requirement for dynamic recalculation of targets.

Linkage to SHACL rule framework is passing `$this` in a SHACL Rules step as it is for SPARQL Query

The PR calls that "pre-binding" (using variable substitution) only for for simplicity although pre-binding (#999) is defined in a way that is SPARQL-only. #1128 separates requirements from implementation of pre-binding. For `sh:SRL`, the "shape focus scope" is provided by variable substitution for `$this`.

afs:
This PR adds a rule type to the SHACL Inference Rules framework as another sub-section in section 11.

The rule type is based on SPARQL-RL.

SPARQL-RL has unique properties and guarantees not offered by SHACL SPARQL Rules.
SHACL SPARQL Rules has unique properties not offered by SPARQL-RL.

This PR uses "program" because SHACL Inf Rules added the same terminology "rule set" after SPARQL-RL.
Other names would be possible.

See also #1228 for discussion of creating a general concept of linking SHACL shapes to SHACL Rule step evaluations necessary for any SHACL Rule rule type.

This PR uses variable binding for `$this`.

Closes #1229

* [See the section rendered online here](https://raw.githack.com/w3c/data-shapes/inf-rules-srl/shacl12-inference-rules/index.html#SRLRules)

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -565,10 +565,10 @@ ex:RectangleRulesShape
 		<section id="syntax">
 			<h2>Syntax of SHACL Rules</h2>
 			<p>
-				The <a>SHACL instances</a> of <code>sh:Rule</code>, including its subclasses <code>sh:SPARQLRule</code> and <code>sh:TripleRule</code>,
+				The <a>SHACL instances</a> of <code>sh:Rule</code>, including its subclasses <code>sh:SPARQLRule</code>, <code>sh:TripleRule</code> and <code>sh:SRL</code>,
 				are called <dfn data-lt="rule|rules|SHACL rule">SHACL rules</dfn>.
 				SHACL has a flexible, extensible design in which multiple types of rules can be supported,
-				but this document only defines two of them: <a>SPARQL rules</a> and <a>triple rules</a>.
+				but this document only defines three of them: <a>SPARQL rules</a>, <a>triple rules</a> and <a>SRL rules</a>.
 			</p>
 			<p>
 				Each <dfn>rule type</dfn> is identified by an <a>IRI</a> that is used as
@@ -1237,10 +1237,10 @@ _:id sh:sourceRule ex:SymmetricPropertyRule .</pre>
 			<h2>Built-in Rule Types</h2>
 			<p>
 				The SHACL inference rules framework defines a general syntax of rules and their execution algorithm.
-				This document defines the two <a>rule types</a> from the following two subsections.
+				This document defines the three <a>rule types</a> from the following subsections.
 			</p>
 			<p>
-				Note that not all implementations are required to implement both of these types for conformance.
+				Note that not all implementations are required to implement all of these types for conformance.
 				A <a>rules engine</a> that encounters a rule for which it does not implement any
 				<a>rule type</a> reports a <a>failure</a>, as defined in <a href="#rules-execution"></a>.
 			</p>
@@ -1456,6 +1456,59 @@ ex:Dad
 						</div>
 					</div>
 				</aside>
+			</section>
+			<section id="SRLRules">
+				<h3>SPARQL-RL Rules</h3>
+				<p>
+					This section defines a <a>rule type</a> called <dfn data-lt="SRL rules">SPARQL-RL</dfn>, often just "SRL",
+					identified by the <a>IRI</a> <code>sh:SRL</code>.
+					<a>SRL rules</a> have the following properties:
+				</p>
+				<table class="term-table">
+					<tr>
+						<th>Property</th>
+						<th>Summary and Syntax Rules</th>
+					</tr>
+					<tr>
+						<td><code>sh:srlProgram</code></td>
+						<td>
+							The SPARQL-RL program: a <a data-cite="SPARQL12-RL#dfn-rule-set">SPARQL-RL rule set</a>.
+							<span data-syntax-rule="srlProgram-count"><a>SRL rules</a> must have exactly one
+							<a>value</a> for the property <code>sh:srlProgram</code>.</span>
+							<span data-syntax-rule="srlProgram-datatype">The values of <code>sh:srlProgram</code>
+							are <a>literals</a> with datatype <code>xsd:string</code>.</span>
+						</td>
+					</tr>
+					<tr>
+						<td><code>sh:prefixes</code></td>
+						<td>
+							The prefixes to use to turn the <code>sh:srlProgram</code> into a SPARQL-RL program.
+							<a>SRL rules</a> may use the property <code>sh:prefixes</code> to declare a dependency on prefixes based on the
+							mechanism defined in <a data-cite="shacl12-sparql#sparql-prefixes">Prefix Declarations for SPARQL Queries</a>.
+							This mechanism allows users to abbreviate URIs in the <code>sh:srlProgram</code> strings.
+						</td>
+					</tr>
+				</table>
+				<div class="def def-text">
+					<div class="def-header">EXECUTION OF SPARQL-RL RULES</div>
+					<div class="def-text-body">
+						Let <code>PGM</code> be the SPARQL-RL program derived from the values of the properties
+						<code>sh:srlProgram</code> and <code>sh:prefixes</code> of the <a>SRL rule</a> in the <a>shapes graph</a>.
+						<ul>
+							<li>
+								If the rule is a <a>shape rule</a>:
+								For each <a>focus node</a>, execute the program <code>PGM</code>,
+								<a>pre-binding</a> by variable substitution of the variable <code>$this</code>
+								with the <a>focus node</a>, and <a>infer</a> the <a>triples</a> it produces.
+							</li>
+							<li>
+								If the rule is a <a>global rule</a>:
+								Execute the program <code>PGM</code> without any <a>pre-binding</a>,
+								and <a>infer</a> the <a>triples</a> it produces.
+							</li>
+						</ul>
+					</div>
+				</div>
 			</section>
 		</section>
 
--- a/shacl12-vocabularies/shacl.ttl
+++ b/shacl12-vocabularies/shacl.ttl
@@ -2048,6 +2048,28 @@ sh:object
 	rdfs:domain sh:TripleRule ;
 	rdfs:isDefinedBy sh: .
 
+## sh:SRL is a sh:SPARQLExecutable so that it can carry sh:prefixes, whose
+## rdfs:domain is sh:SPARQLExecutable. sh:SPARQLExprExpression is declared the
+## same way for the same reason. A separate executable class (in the style of
+## sh:SPARQLConstructExecutable) is not needed here because sh:srlProgram has no
+## user other than sh:SRL.
+
+sh:SRL
+	a rdfs:Class ;
+	rdfs:label "SPARQL-RL rule"@en ;
+	rdfs:comment "The class of SHACL rules based on SPARQL-RL programs."@en ;
+	rdfs:subClassOf sh:Rule ;
+	rdfs:subClassOf sh:SPARQLExecutable ;
+	rdfs:isDefinedBy sh: .
+
+sh:srlProgram
+	a rdf:Property ;
+	rdfs:label "SRL program"@en ;
+	rdfs:comment "The SPARQL-RL program to execute."@en ;
+	rdfs:domain sh:SRL ;
+	rdfs:range xsd:string ;
+	rdfs:isDefinedBy sh: .
+
 
 # -----------------------------------------------------------------------------
 # SHACL ADVANCED FEATURES (will likely be marked deprecated with 1.2)


liviorobaldo:
I am really getting confused... I no longer understand which version of the SHACL 1.2 Inference Rules should be considered the latest one. The problem is that there are too many open branches. I had been considering this branch as the latest:

https://raw.githack.com/w3c/data-shapes/order-layer-inference-rules/shacl12-inference-rules/index.html

However, as I also mentioned [in this comment](https://github.com/w3c/data-shapes/issues/1226#issuecomment-5510144459), it seems that at least @osavkovic is considering another branch to be the latest one, since at least §10 differs. §10 of the `inf-rules-srl` branch that Andy just created contains text similar to that in the `order-layer-inference-rules` branch above, although there is one important difference: SPARQL-RL is no longer mentioned there.

Finally, it seems that the following link (which should point to the main branch) is inactive:

https://www.w3.org/TR/shacl12-inference-rules

I get a 404:

image

If I go to the SHACL Core link:

https://www.w3.org/TR/shacl12-core/

and click on "SHACL 1.2 Inference Rules" under §SHACL Specifications, I also get a 404.

All the other links in the working drafts (SHACL 1.2 Node Expressions, SHACL 1.2 Profiling, SHACL 1.2 SPARQL Extensions, and SHACL 1.2 UI) work fine, but "SHACL 1.2 Inference Rules" gives me a 404.

Am I missing something? Can someone help me understand which version we should consider the latest? Thanks!

@HolgerKnublauch @scotthenninger @mgberg

afs:
Try : https://w3c.github.io/data-shapes/shacl12-inference-rules/ for the current version with no PRs.

It is the gh-pages branch.

That branch is the primary one of all merged material and will be the source for publication to /TR/.

This PR is branched off that, in the state it was today.
> [afs](https://github.com/afs) [force-pushed](https://github.com/w3c/data-shapes/compare/4d6462abcf3085f7596ff3b789bc686093354e4d..7687a9d5560e11683ab296d601b2e5cd8720617d) the inf-rules-srl branch from [4d6462a](https://github.com/w3c/data-shapes/commit/4d6462abcf3085f7596ff3b789bc686093354e4d) to [7687a9d](https://github.com/w3c/data-shapes/commit/7687a9d5560e11683ab296d601b2e5cd8720617d) ...

is re-syncing it with the primary copy, for example, to pick up the unrelated work on other documents.

shacl12-inference-rules/index.html
```diff
@@ -1235,10 +1235,10 @@ <h2>Variations of SHACL Rule Implementations</h2>
 			<h2>Built-in Rule Types</h2>
 			<p>
 				The SHACL inference rules framework defines a general syntax of rules and their execution algorithm.
-				This document defines the two <a>rule types</a> from the following two subsections.
+				This document defines the three <a>rule types</a> from the following subsections.
 			</p>
 			<p>
-				Note that not all implementations are required to implement both of these types for conformance.
+				Note that not all implementations are required to implement all of these types for conformance.
```
mgberg, line 1241: Would we actually want to require that all SHACL Rule engines must always support SRL? That seems like it could be a big lift if it needs to implement the full SRL capability.

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+				</p>
+				<table class="term-table">
+					<tr>
+						<th>Property</th>
+						<th>Summary and Syntax Rules</th>
+					</tr>
+					<tr>
+						<td><code>sh:srlProgram</code></td>
+						<td>
+							The SPARQL-RL program: a <a data-cite="SPARQL12-RL#dfn-rule-set">SPARQL-RL rule set</a>.
```
mgberg, line 1473: Can the rule set provided by this predicate include more than one rule? Can it include imports?

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+							<span data-syntax-rule="srlProgram-count"><a>SRL rules</a> must have exactly one
+							<a>value</a> for the property <code>sh:srlProgram</code>.</span>
+							<span data-syntax-rule="srlProgram-datatype">The values of <code>sh:srlProgram</code>
+							are <a>literals</a> with datatype <code>xsd:string</code>.</span>
```
mgberg, line 1477: I see in the SRL document that it would be possible to import rule sets by URI- so by this mechanism it wouldn't be possible to _reuse_ existing rule sets, they'd have to be provided inline here. Is that the intent?

shacl12-inference-rules/index.html
afs: > Would we actually want to require that all SHACL Rule engines must always support SRL? 

No.

Implementation (like the other rule types) is optional.
https://w3c.github.io/data-shapes/shacl12-inference-rules/#Built-in-RuleTypes already says:
>> Note that not all implementations are required to implement all of these types for conformance.

shacl12-inference-rules/index.html
afs: Yes, more than one SRL rule.
There is an example on #1229.
Without multiple `RULE`s, there's no stratification so less useful.

There is well-formedness - which means no half-outputs from the rule head and no unbound variable in WHERE which on then lead to over-matching:

```sparql
CONSTRUCT { ?thing rdf:type :hasChosenValue }
WHERE { BIND (1/0 AS ?object ) ?thing ?p ?object }
```
All `?thing ?p ?object` matches every triple.

OPTIONAL and UNION do the same - it's a known issue when writing and maintaining large CONSTRUCT queries. If they leave a variable unbound that is later used for pattern matching, there can be a related explosion of solutions can be hidden by the fact the results are a graph - set of triples. The CONSTRUCT just runs very slowly.

afs: Good point - it needs to say that no imports are allowed (IMPORTS are optional in SPARQL-RL anyway) inside a text block.

It is important that `owl:imports` is used so it is inspect-able by the execution environment.

shacl12-inference-rules/index.html
afs: We could allow a URL and read the document - i.e. an indirection.

Another property presumably, not overloading sh:slPorgram - but either should work.

It would be useful for global rules to provide rule-style entailment regimes, for example.

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+							<li>
+								If the rule is a <a>shape rule</a>:
+								For each <a>focus node</a>, execute the program <code>PGM</code>,
+								<a>pre-binding</a> by variable substitution of the variable <code>$this</code>
+								with the <a>focus node</a>, and <a>infer</a> the <a>triples</a> it produces.
+							</li>
+							<li>
+								If the rule is a <a>global rule</a>:
+								Execute the program <code>PGM</code> without any <a>pre-binding</a>,
+								and <a>infer</a> the <a>triples</a> it produces.
```
scotthenninger, line 1505: This is the only place I see where shape rule and global rule are introduced. I'd want some further information about these, such as how they are invoked (more of an issue for global rules). For example are global rules executed before or after shape rules and does that even matter - can it be left to the implementer? Are inferences from the one run first visible to the latter or are they isolated? Issues like that need some explanation IMHO.

afs: > This is the only place I see where shape rule and global rule are introduced. 

yes - compare this new section (11.3) with the existing text [11.1 SHACL SPARQL Rules](https://w3c.github.io/data-shapes/shacl12-inference-rules/#SPARQLRule).

The general SHACL Rules framework and execution is unchanged. 

[Section 3](https://w3c.github.io/data-shapes/shacl12-inference-rules/#syntax) covers global and shape rules:

> [SHACL rules](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule) can be divided into two categories:
> * A shape rule is a [rule](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule) that is the [object](https://www.w3.org/TR/rdf12-concepts/#dfn-object) of a sh:rule triple. Shape rules are executed for all [target](https://www.w3.org/TR/shacl12-core/#dfn-target) nodes of the [shape](https://www.w3.org/TR/shacl12-core/#dfn-shape) which is the [subject](https://www.w3.org/TR/rdf12-concepts/#dfn-subject) of the sh:rule triple.
> * A global rule is a [rule](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule) that is not linked to a [shape](https://www.w3.org/TR/shacl12-core/#dfn-shape) by a sh:rule [predicate](https://www.w3.org/TR/rdf12-concepts/#dfn-predicate).

Execution is section 8 [General Execution Instructions for SHACL Rules](https://w3c.github.io/data-shapes/shacl12-inference-rules/#rules-execution). 

> An execution of a [rule](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule) is the process that produces [inferred](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-inferring) triples from the [rule](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule) based on the [execution instructions](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-execution-instructions) of its [rule types](https://w3c.github.io/data-shapes/shacl12-inference-rules/#dfn-rule-type).

The addition of this PR is execution instructions for this rule type. 

As a SHACL "execution of a rule",  SRL takes the prefixes+text and outputs a graph of inferred triples, just like the other rule types. (SHACL Rules already implicitly assumes that any "execution of a rule" is atomic)

> Are inferences from the one run first visible to the latter or are they isolated?

Within SPARQL-RL, execution of each SRL `RULE` is visible to other rules; SPARQL-RL calculates execution order as needed for predicable, sound output.

Within the SHACL Rules framework, the outcome (graph of inferred triples) is visible: in section 8:
> Within an iteration, the inferred triples of one rule [i.e. sh:rule] become immediately visible to the next rule.

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+				<div class="def def-text">
+					<div class="def-header">EXECUTION OF SPARQL-RL RULES</div>
+					<div class="def-text-body">
+						Let <code>PGM</code> be the SPARQL-RL program derived from the values of the properties
+						<code>sh:srlProgram</code> and <code>sh:prefixes</code> of the <a>SRL rule</a> in the <a>shapes graph</a>.
+						<ul>
+							<li>
+								If the rule is a <a>shape rule</a>:
+								For each <a>focus node</a>, execute the program <code>PGM</code>,
+								<a>pre-binding</a> by variable substitution of the variable <code>$this</code>
```
HolgerKnublauch, line 1499: The term pre-binding is only defined for SPARQL. But SRL doesn't use SPARQL but has its own SPARQL-like grammar. So SRL would need to specify how pre-binding works and what limitations it has (e.g. w.r.t. SET). Alternatively, let's merge https://github.com/w3c/data-shapes/pull/1219 and limit these SRL snippets to CONSTRUCT syntax.

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+								with the <a>focus node</a>, and <a>infer</a> the <a>triples</a> it produces.
+							</li>
+							<li>
+								If the rule is a <a>global rule</a>:
+								Execute the program <code>PGM</code> without any <a>pre-binding</a>,
+								and <a>infer</a> the <a>triples</a> it produces.
+							</li>
+						</ul>
+					</div>
+				</div>
```
HolgerKnublauch, line 1509: This requires an example, ideally one that illustrates why people should even use this syntax. I guess that has to do with scenarios where SRL is mixed with other rule types.

shacl12-inference-rules/index.html
```diff
@@ -1455,6 +1455,59 @@ <h3>Triple Rules</h3>
...
+						<td><code>sh:prefixes</code></td>
+						<td>
+							The prefixes to use to turn the <code>sh:srlProgram</code> into a SPARQL-RL program.
+							<a>SRL rules</a> may use the property <code>sh:prefixes</code> to declare a dependency on prefixes based on the
+							mechanism defined in <a data-cite="shacl12-sparql#sparql-prefixes">Prefix Declarations for SPARQL Queries</a>.
+							This mechanism allows users to abbreviate URIs in the <code>sh:srlProgram</code> strings.
+						</td>
+					</tr>
+				</table>
+				<div class="def def-text">
```
HolgerKnublauch, line 1490: What happens if SRL RULEs declare URIs for individual rules? These are not really URIs but are local anchors within an SRL snippet? I find this does not align with RDF.

shacl12-inference-rules/index.html
```diff
@@ -565,10 +565,10 @@ <h3>An Example Triple Rule</h3>
 		<section id="syntax">
 			<h2>Syntax of SHACL Rules</h2>
 			<p>
-				The <a>SHACL instances</a> of <code>sh:Rule</code>, including its subclasses <code>sh:SPARQLRule</code> and <code>sh:TripleRule</code>,
+				The <a>SHACL instances</a> of <code>sh:Rule</code>, including its subclasses <code>sh:SPARQLRule</code>, <code>sh:TripleRule</code> and <code>sh:SRL</code>,
```
HolgerKnublauch, line 568: This naturally needs to go into the SPARQL-RL document. SHACL Rules define a framework with an extension point. It is not the role of SHACL Rules to then enumerate all such extensions. We also don't include a rule type for RIF or N3 here. You stated multiple times in the past that you expect SPARQL-RL to evolve further once the SHACL WG has ended. Dropping the SRL content here means you could no longer make changes to the integration without waiting for SHACL 1.3.

shacl12-vocabularies/shacl.ttl
```diff
@@ -2048,6 +2048,28 @@ sh:object
...
+## same way for the same reason. A separate executable class (in the style of
+## sh:SPARQLConstructExecutable) is not needed here because sh:srlProgram has no
+## user other than sh:SRL.
+
+sh:SRL
+	a rdfs:Class ;
+	rdfs:label "SPARQL-RL rule"@en ;
+	rdfs:comment "The class of SHACL rules based on SPARQL-RL programs."@en ;
+	rdfs:subClassOf sh:Rule ;
+	rdfs:subClassOf sh:SPARQLExecutable ;
```
HolgerKnublauch, line 2062: I find this whole layering is really ugly. Basically you suggest to abuse the term sh:Rule to mean a sh:RuleSet, and insert a parallel universe where your SRL rules also define rules and rule sets but within a text string. This is very much the language-within-language issue that you have criticised when I argued in favor of srl:text instead of SRL-as-blank-node-trees as RDF syntax. Here you suddenly seem to be OK with that.

The injection of SRL rule sets strings as individual SHACL rules introduces countless problems.

- You had to introduce a new term "program" to work around the duplication of the term rule set. SHACL users would expect to use sh:RuleSet here.
- The term "rule" is used in different ways between SRL and SHACL rules and SRL-in-SHACL. SHACL users would expect to have something like srl:Rule that represents an individual SRL rule.
- Concepts like sh:sourceRule don't work here, because all triples produced by an SRL "program" will be assigned to the same sh:SRL instance.
- A SHACL engine that executes SRL "programs" will by default treat such rule sets as black boxes. It will pass control to the SRL engine and (a while later) get some triples back. In contrast, a normal SHACL rules engine where each sh:Rule indeed represents a rule can offer things like cleanly canceling execution between rules, measuring individual rule performance, display progress to users about individual rules, deactivating individual rules (sh:deactivated). Also in your design sh:expectedPredicate would need to be specified for the whole SRL program instead of the rules like in SHACL.
- The mismatch between prefix declarations, e.g. what happens if a sh:srlProgram declares its own prefixes that clash with sh:prefixes?
- Any non-toy SRL use cases will consist of three or more rules. The sh:srlProgram strings will become very large, which is rather unusual for an RDF-based language. Here people cannot even reference individual rules in such strings.

I still think that all of these issues can be avoided if SRL rules are simply represented as sh:SPARQLRule, as my PR https://github.com/w3c/data-shapes/pull/1219 makes possible. This has the extra benefit that a normal SHACL SPARQL engine can execute SRL rules, assuming the stratification has been applied to compute the sh:layer/runOnce triples. (That particular integration point needs a bit of work but would fit well into SHACL). Then we would have a true integration and the best of both worlds, not some artificial attempt to fit a square peg into a round hole as done here.

HolgerKnublauch (changes requested):

shacl12-inference-rules/index.html
liviorobaldo: I agree with @HolgerKnublauch here. The work on SPARQL-RL is not yet complete; for instance, aggregates should be introduced in SPARQL-RL. I think that, rather than introducing _two_ rule types in this section, we should replace the proposed subsection with one explaining _how_ additional rule types can be defined. Subsequent WGs could then define new rule types for SPARQL-RL, RIF, N3, OWL, SHACL-JavaScript, etc.

afs: See #1228.

shacl12-vocabularies/shacl.ttl
afs: >Also in your design sh:expectedPredicate would need to be specified for the whole SRL program instead of the rules like in SHACL.

Yes.

As I understand the section on `sh:expectedPredicate` and the execution text:

```
	Compute the expected derived triples for all rules in the layer
```
that happens in the general execution by adding them to the evaluation graph. 
Is that how they get to a SHACL SPARQL Rules?

HolgerKnublauch: Yes.

afs: > Concepts like sh:sourceRule don't work here, because all triples produced by an SRL "program" will be assigned to the same sh:SRL instance.

1. That's a good thing - they are the output of the "program" (e.g. two ways to get the same triple; the "program" is the executed unit).

2. The `RULE` can add triples as well for finer detail.

afs: `srl`: if you want, but to users the difference between namespace prefixes from the same working group is confusing as we know from `rdf:` vs `rdfs:`

afs: > A SHACL engine that executes SRL "programs" will by default treat such rule sets as black boxes. It will pass control to the SRL engine and (a while later) get some triples back

The SRL security section covers this.

If that is important to have this in all extensions, then state it as an expectation. A single CONSTRUCT can do a lot of compute but the effect is masked by the "set of triples" of the result.

afs: > I still think that all of these issues can be avoided if SRL rules are simply represented as sh:SPARQLRule, 

SPARQL-RL rule sets can be run backwards (the shape rule for weighted paths is an example) for efficiency.

Reducing integration to one-rule SRL does not add much that SHACL IR can already do - `CONSTRUCT` can do it.
But `CONSTRUCT` in SHACL Inf Rules is not sound, safe and there is no completeness defined -- not reproducible.

It would means general publications of SRL can not be reused in SHACL IR for global rules.

While I would support a translation of *SRL rule sets* to *SHACL Rules rule sets* that it is different usage from incorporating well-defined steps of SPARQL-RL intermixed with general computation.

afs:
> This naturally needs to go into the SPARQL-RL document.

> Dropping the SRL content here means you could no longer make changes to the integration without waiting for SHACL 1.3.

The content of this PR only relates to the the machinery of SHACL Inf Rules, not the features of SPARQL-RL.

What the new section is sensitive to is current and future changes in SHACL Inf Rules framework and in SHACL generally such targets. The contract is "text in, graph out" with the rule framework processing.

SHACL Inf Rules are deals with SPARQL 1.0, SPARQL 1.1 and SPARQL 1.2 and needs no changes for SPARQL 1.3, 1.4...

Separating it makes it inconvenient for the user. The user needs to understand SHACL Inf Rules.
See @scotthenninger's https://github.com/w3c/data-shapes/pull/1230#discussion_r3928519768

https://github.com/w3c/data-shapes/issues/1113#issuecomment-5188359672 lays out the relationships.

> You stated multiple times in the past that you expect SPARQL-RL to evolve further once the SHACL WG has ended.

I would like it to - and do things CONSTRUCT can't. But SHACL Inf Rules isn't dependent on that, in the same way it isn't dependent on SPARQL 1.1 vs SPARQL 1.2.

shacl12-inference-rules/index.html
afs: URI on RULE when used in SHACL Inf Rule (SHACL IR) isn't useful in this context.
So we can advise not using them while noting that reusing published SPARQL-RL rule sets may have them. 

The unit of execution for SPARQL-RL is rule set.  that's the URI that matters in the context of SHACL Inf Rules.
See also @mgberg's https://github.com/w3c/data-shapes/pull/1230#discussion_r3926515606

shacl12-vocabularies/shacl.ttl
afs: > I find this whole layering is really ugly.

What do you suggest for the modelling? 

We're stuck with SHACL Inf Rule rule set, which came after SPARQL-RL rule set whereas `sh:strata` got renamed as `sh:layer`. SPARQL-RL now carefully qualifies any use of layer with "Stratification".

These two "rule set" concepts are different. SRL rulesets have guarantees; SHACL Inf Rule rule sets do not.

---

This
```
## sh:SRL is a sh:SPARQLExecutable so that it can carry sh:prefixes, whose
## rdfs:domain is sh:SPARQLExecutable. sh:SPARQLExprExpression is declared the
## same way for the same reason. A separate executable class (in the style of
## sh:SPARQLConstructExecutable) is not needed here because sh:srlProgram has no
## user other than sh:SRL.
```
is an issue that Claude mentioned. I left it as a comment to get some help.

HolgerKnublauch: > These two "rule set" concepts are different. SRL rulesets have guarantees; SHACL Inf Rule rule sets do not.

This is what I would like to better understand. Yes, SRL rule sets have guarantees. But a SHACL rule set that uses only the same features (just with CONSTRUCT instead of RULE) can be interpreted to have the exact same guarantees. It's just another syntax.

Your previous RDF syntax for SRL (srl: namespace) also had classes for rule sets and rules, so what is missing to allow sh:RuleSet and sh:Rule to play the same role? Would some subclass help, e.g. srl:RuleSet rdfs:subClassOf sh:RuleSet? I don't think this is necessary, because a SHACL engine can detect whether all rules in a rule set are in SRL expressiveness, and then it can elect to delegate the execution to a native SRL engine if that has advantages. These things are implementation details IMHO.

HolgerKnublauch: Here is another thought. SHACL Rules could define an extension point that allows SRL to plug in and be required to be involved. Something like

```
ex:MySRLRuleSet
    a sh:RuleSet ;
    sh:ruleProcessor srl:SRL ;
```

where compliant engines are required to pre-process the rule set, e.g. to make sure that the stratification has been applied. Engines that don't know how to do that must report a Failure.

This would be similar to sh:entailment.

afs: What's the difference to defining a rule type?

HolgerKnublauch: Here are the various alternatives that I see floating around. I am using a toy example with three rules, two in SRL fragment, and one using full SPARQL features.

### 1: SHACL Inference Rules syntax, runOnce and layer materialized

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:area ?area }
            WHERE { ?r ex:width ?width . ?r ex:height ?height . BIND (?width * ?height AS ?area) }
        """ ;
        sh:runOnce true ;
    ] ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:large true }
            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
        """ ;
    ] ;
    sh:hasRule [    # This is outside of SRL, relying on unbound vars
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:color ?color }
            WHERE { OPTIONAL { ?r ex:large true } BIND (IF(?r, 'blue', 'green') AS ?color) }
        """ ;
        sh:layer 1 ;   # wait for other two rules to finish first
    ] .
```

### 2: SHACL syntax without runOnce/layer but requiring SRL processor

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:ruleProcessor srl:SRL ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:area ?area }
            WHERE { ?r ex:width ?width . ?r ex:height ?height . BIND (?width * ?height AS ?area) }
        """ ;
    ] ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:large true }
            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
        """ ;
    ] ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:color ?color }
            WHERE { OPTIONAL { ?r ex:large true } BIND (IF(?r, 'blue', 'green') AS ?color) }
        """ ;
    ] .
```

### 3. SHACL rule type for SRL Rule sets (PR #1230 but updated namespace)

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:hasRule [
        a srl:SRL ;
        srl:program """
            RULE { ?r ex:area ?area }
            WHERE { ?r ex:width ?width . ?r ex:height ?height . SET (?area := ?width * ?height) }

            RULE { ?r ex:large true }
            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
        """ ;
    ] .

ex:MyOtherRuleSet
    a sh:RuleSet ;
    sh:includesRuleSet ex:MyRuleSet ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:color ?color }
            WHERE { OPTIONAL { ?r ex:large true } BIND (IF(?r, 'blue', 'green') AS ?color) }
        """ ;
    ] .
```

### 4. SHACL rule type for individual SRL Rules (new potential proposal, not by me)

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:hasRule [
        a srl:Rule ;
        srl:text """
            RULE { ?r ex:area ?area }
            WHERE { ?r ex:width ?width . ?r ex:height ?height . SET (?area := ?width * ?height) }
        """ ;
    ] ;
    sh:hasRule [
        a srl:Rule ;
        srl:text """
            RULE { ?r ex:large true }
            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
        """ ;
    ] ;
    sh:hasRule [
        a sh:SPARQLRule ;
        sh:construct """
            CONSTRUCT { ?r ex:color ?color }
            WHERE { OPTIONAL { ?r ex:large true } BIND (IF(?r, 'blue', 'green') AS ?color) }
        """ ;
    ] .
```

It is difficult to work on these examples while we only have toy examples. The more interesting real-world scenarios would better illustrate the complexity that the maintenance and editing of such rule sets will require. 

But one point is that options 1, 2 and 4 keep individual rules as sh:Rule instances, linked to a RuleSet via sh:hasRule, while 3 wraps all SRL rules into a single large string literal and then needs a different syntax to represent the other SHACL rule.

With option 1, the vanilla SHACL engine has been given enough info to execute the rules with exactly the same output as an SRL engine would.

With option 2, a compliant engine must either use an SRL engine for the whole rule set, or use the SRL-compliant algorithm to compute the stratification before continuing with the SHACL engine. If would report a failure if it doesn't have SRL support.

Option 3 is basically the proposal from this PR here. It is hopefully obvious that this becomes rather inconsistent once people want to mix SRL and SPARQL-Full rules.

With option 4, an engine needs to be compliant with SRL because it otherwise doesn't know how to parse srl:Rule/srl:text. This could be combined with some other flags like sh:ruleProcessor srl:SRL or use srl:RuleSet as rdf:type. But I am not really proposing this, although it's better than 3. This would be the next-best option if you guys totally cannot live with CONSTRUCT syntax anywhere near SRL.

Option 2 looks reasonable to me. It clearly communicates to the engine that the stratification has not been computed yet, and that SRL's contracts need to be enforced. Yet, it uses normal SHACL syntax, which means that once the sh:layer/runOnce triples are added, the exact same output will be produced with a vanilla SHACL engine. So applying SRL on option 2 produces option 1.

mgberg: > With option 2, a compliant engine must either use an SRL engine for the whole rule set, or use the SRL-compliant algorithm to compute the stratification before continuing with the SHACL engine. If would report a failure if it doesn't have SRL support.
> 
> Option 3 is basically the proposal from this PR here. It is hopefully obvious that this becomes rather inconsistent once people want to mix SRL and SPARQL-Full rules.

One potential downside to option 2 vs option 3 is that there would be multiple ways to determine whether or not an engine has the ability to execute a rule. In option 3, it just simply looks at the rule type- if it doesn't have a type that it supports (e.g. `srl:SRL`, or some custom rule type), it reports a failure. With option 2, it would have to be a combination of rule type and processor.

HolgerKnublauch: @mgberg not exactly. With option 2 *only* a custom ruleProcessor is needed (that computes the stratification). The rules themselves are just normal SHACL rules (sh:SPARQLRule) and can be executed as such, individually. This makes it consistent with the SPARQL Full rules handled by the same engine.

And, perhaps more importantly, the stratification can be computed statically (once) and then even be asserted in the form of sh:layer/runOnce triples in the rules graph. So the workload to produce stratification is only needed once, on the computer (IDE) where the rules are developed, not in the deployed runtime system prior to each execution.

On failure detection: If an engine cannot handle slr:SRL rule type, it will report a failure. The same will happen if it doesn't know how to handle sh:ruleProcessor "SRL".

shacl12-inference-rules/index.html
HolgerKnublauch: Although this PR adds support for shape rules to SRL, this would not be the same as shape rule support in general SHACL because it means that *all* rules in the SRL block would need to be applicable to the same shape. That is unlikely in practice except for toy examples. The whole point of sh:rule is to distribute rules across the shapes/classes where they apply best, in order to break up long flat lists of rules that are otherwise hard to maintain and understand.

shacl12-inference-rules/index.html
afs: It uses variable substitution, as used by several SHACL systems for pre-binding.

Since it seem pre-binding has gone in a different direction, "prebinding" has been replaced by variable substitution.

shacl12-inference-rules/index.html
```diff
@@ -1264,10 +1264,10 @@ <h2>Custom Rule Processors (sh:ruleProcessor)</h2>
 			<h2>Built-in Rule Types</h2>
 			<p>
 				The SHACL inference rules framework defines a general syntax of rules and their execution algorithm.
-				This document defines the two <a>rule types</a> from the following two subsections.
+				This document defines the three <a>rule types</a> from the following subsections.
 			</p>
 			<p>
-				Note that not all implementations are required to implement both of these types for conformance.
+				Note that not all implementations are required to implement all of these types for conformance.
```
recalcitrantsupplant:
```suggestion
				Note that implementations are not required to implement all of these types for conformance.
```

recalcitrantsupplant (commented):
grammar change

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1484,59 @@ <h3>Triple Rules</h3>
...
+					<a>SRL rules</a> have the following properties:
+				</p>
+				<table class="term-table">
+					<tr>
+						<th>Property</th>
+						<th>Summary and Syntax Rules</th>
+					</tr>
+					<tr>
+						<td><code>sh:srlProgram</code></td>
+						<td>
```
liviorobaldo, line 1501: I now understand from [PR 1263](https://github.com/w3c/data-shapes/pull/1263) that `sh:srlProgram` allows us to _recursively_ associate a rule set with a rule set:

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:hasRule [
        a srl:SRL ;
        sh:srlProgram """
            RULE { ?r ex:area ?area }
            WHERE { ?r ex:width ?width . ?r ex:height ?height . SET (?area := ?width * ?height) }

            RULE { ?r ex:large true }
            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
        """ ;
    ] .
```

The string that is the object of `sh:srlProgram` can contain multiple rules, but note that `sh:hasRule` is singular.

I do not like this. Not only do I find it syntactically convoluted, but I also think it will amplify the problems that I was trying to "mitigate" in [PR 1269](https://github.com/w3c/data-shapes/pull/1269). With the current approach, we could associate a `sh:RuleSet` with multiple (sub)sets of rules, e.g.:

```
ex:MyRuleSet
    a sh:RuleSet ;
    sh:hasRule [
        a srl:SRL ;
        sh:srlProgram """
            RULE{ … }WHERE{ … }
            RULE{ … }WHERE{ … }
            Etc.
        """ ;
    ];
    sh:hasRule [
        a srl:SRL ;
        sh:srlProgram """
            RULE{ … }WHERE{ … }
            RULE{ … }WHERE{ … }
            Etc.
        """ ;
    ] .
```

How should the order of the rules be computed in this case, i.e., which `sh:order` and `sh:layer` should be determined? We should therefore advise against this as well, just as we advise against having rules of different types in the same rule set.

I think we should instead replace `sh:srlProgram` with an RDF property associating an instance of `srl:SRL` with a string representing a _single_ "RULE-WHERE". In this way, the `sh:RuleSet` will contain a _single_ subset of SPARQL-RL rules (possibly, and preferably, a _proper_ subset if it does not also contain rules of other types).

PR #1239, opened 2026-09-08
First draft of report extension for infos on matched nodes

### LINKED ISSUE #1221: Validation report reporting on matched target nodes
It seems to be fairly important in practice to report on the 'positive' matches of target nodes to shapes.
Previous issues addressing this are
- https://github.com/w3c/data-shapes/issues/325
- https://github.com/w3c/data-shapes/issues/518
In these issues, the difficulties for introducing such a feature in the validation report are described.

As recent discussions brought up this topic again, the difficulties to address are mainly
- different target declarations, e.g. `sh:targetClass`, `sh:targetNode`, ...
- a target node an be selected by multiple shapes
- a potentially high number of reported target nodes
- performance issues

A simple practical proposal would be to report on the total number of matching target nodes and the total number of violating target nodes. A `sh:ValidationReport` can optionally report on these numbers using the properties `sh:conformingNodes` and `sh:violatingNodes`. The numbers refer to the total numbers of actually validated nodes. This would help integrators of SHACL processors to verify if validation actually worked correctly, i.e. if it 'hits' the right target nodes.

In addition, we can still have the optional `sh:hasTargetNodes`, which lists the actual focus nodes (of all shapes).
If there are target declarations other than `sh:targetNode`, the list contains the respective grounded target nodes.

@HolgerKnublauch @ajnelson-nist @afs @bergos

robert-david:
This pull request provides a first draft to extend the validation report with infos on targeted nodes.
The extensions are designed to not break backwards compatibility with SHACL 1.0.
It provides the following extensions:
- Node Assignment Count (sh:nodeAssignmentCount) - total number of pairs of targeted nodes and shapes
- Assignment (sh:assignment) - a property to link a concrete assignment (instance of sh:ShapeAssignment)
- Validation Details (sh:ShapeAssignment) - the actual assignment reporting on focus node, shape and optionally on conformance

Closes #1221 

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1221-report-extension/shacl12-core/index.html)

### DIFF
--- a/shacl/shacl.ttl
+++ b/shacl/shacl.ttl
@@ -171,12 +171,28 @@ sh:conforms
 	rdfs:range xsd:boolean ;
 	rdfs:isDefinedBy sh: .
 
+sh:nodeAssignmentCount
+	a rdf:Property ;
+	rdfs:label "Node assignment count"@en ;
+	rdfs:comment "The total number of shape assignments of shapes to focus nodes."@en ;
+	rdfs:domain sh:ValidationReport ;
+	rdfs:range xsd:integer ;
+	rdfs:isDefinedBy sh: .
+
 sh:result
 	a rdf:Property ;
 	rdfs:label "result"@en ;
 	rdfs:comment "The validation results contained in a validation report."@en ;
 	rdfs:domain sh:ValidationReport ;
 	rdfs:range sh:ValidationResult ;
+	rdfs:isDefinedBy sh: .
+
+sh:assignment
+	a rdf:Property ;
+	rdfs:label "assignment"@en ;
+	rdfs:comment "The shape assignments that the validation was processed on."@en ;
+	rdfs:domain sh:ValidationReport ;
+	rdfs:range sh:ShapeAssignment ;
 	rdfs:isDefinedBy sh: .
 
 sh:shapesGraphWellFormed
@@ -198,6 +214,13 @@ sh:ValidationResult
 	a rdfs:Class ;
 	rdfs:label "Validation result"@en ;
 	rdfs:comment "The class of validation results."@en ;
+	rdfs:subClassOf sh:AbstractResult ;
+	rdfs:isDefinedBy sh: .
+
+sh:ShapeAssignment
+	a rdfs:Class ;
+	rdfs:label "Shape assignment"@en ;
+	rdfs:comment "The class of assignments of shapes to nodes."@en ;
 	rdfs:subClassOf sh:AbstractResult ;
 	rdfs:isDefinedBy sh: .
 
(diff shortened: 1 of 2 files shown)


shacl12-core/index.html
```diff
@@ -2619,6 +2619,26 @@ <h5>Processor Configuration (sh:ProcessorConfiguration)</h5>
 							maximum recursion depth, or other runtime parameters.
 						</p>
 					</section>
+					<section id="nodeAssignmentCount"><!-- ISSUE-1221 -->
+						<h5>Node Assignment Count (sh:nodeAssignmentCount)</h5>
+						<p>
+							A valdiation report MAY contain one value for the property <code>sh:nodeAssignmentCount</code>.
+							The value MUST be equal to the total number of shape assignments of a <a>focus node</a> to a <a>shape</a>.
+							It MUST include focus nodes of target declarations other than <a href="#targetNode"><code>sh:targetNode</code></a>, which are determined by grounding the target nodes based on the specific target declaration.
```
ajnelson-nist, line 2627: "other than" feels wrong here.

If I write a shape targeting a node but typo the node IRI, I want the report to hint to me 0 matches.

But, when a shapes graph has many targeting shapes, that 0-count will be obscured information.

This leans me to thinking `sh:nodeAssignmentCount` would go into some kind of reporting node about a shape.  Now, node shape, or property shape, or both?

My day to day shapes usage includes graphs that lean on blank nodes for inlining property shapes, many of which appear redundant and impossible to distinguish if separated from their parent node.  So, maybe `sh:nodeAssignmentCount` would only reported on shapes using some kind of targeting predicate.

And this leads me to thinking there's a verbosity-scale for this optional feature.  Maybe I want `sh:nodeAssignmentCount` for all shapes that had an assignment; and in a next-level verbosity, I want `sh:nodeAssignmentCount` for all shapes that could have an assignment (i.e., having a targeting predicate).

Does this help design considerations?

HolgerKnublauch:
Would you mind adding the definitions of the new terms in the shacl.ttl to make it even more specific? I am for example curious how the assignment concept relates to sh:Result/AbstractResult

shacl/shacl.ttl
```diff
@@ -201,6 +217,13 @@ sh:ValidationResult
 	rdfs:subClassOf sh:AbstractResult ;
 	rdfs:isDefinedBy sh: .
 
+sh:ShapeAssignment
+	a rdfs:Class ;
+	rdfs:label "Shape assignment"@en ;
+	rdfs:comment "The class of assignments of shapes to nodes."@en ;
+	rdfs:subClassOf sh:AbstractResult ;
```
ajnelson-nist, line 224: What is the relationship between this class and its sibling `sh:ValidationResult`?  Is a `ValidationResult` implicitly a `ShapeAssignment`?

ajnelson-nist: Or, never (i.e., classes are disjoint, to be encoded in SHACL-SHACL)?

robert-david: Rather never, i guess.

PR #1240, opened 2026-09-09
#1228: Add executeRule function and formal rule-types section

### LINKED ISSUE #1228: Define the requirements for connecting shape instances to rules.

HolgerKnublauch:
This introduces the executeRule API function and moves the rule type discussion into its own sub-section.

Closes #1228

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1228/shacl12-inference-rules/index.html#rule-types)

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -1149,9 +1149,7 @@ ex:SetYoungestOffspringsRule
 				For a given <a>rule set</a> in the <a>shapes graph</a>
 				an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
 				specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-				Within an iteration, the inferred triples of one rule become immediately visible to the next rule, but
-				rules with the same order can not see each other's inferred triples until the iteration has completed.
-				This means that rules with the same order may be executed in parallel.
+				Within an iteration, the inferred triples of one rule become immediately visible to the next rule.
 			</p>
 			<p>
 				The <dfn>execution of a rule set</dfn> is defined as follows:
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -399,7 +399,7 @@
 					This specification describes conformance criteria for:
 				</p>
 				<ul>
-					<li><strong>SHACL inference rule processors</strong> as processors that support the evaluation of <a>rules</a></li>
+					<li><strong><a>SHACL rules engines</a></strong> as processors that support the evaluation of <a>rules</a></li>
 				</ul>
 				<p>
 					Also see the discussion of well-formedness in the <a data-cite="shacl12-core#conformance">Conformance section of SHACL Core</a>.
@@ -961,7 +961,7 @@ ex:SmallRectangle ex:isSmall true .
 		<section id="ruleSet">
 			<h2>Rule Sets</h2>
 			<p>
-				The input to a SHACL rules processor is a set of rules called a <dfn>rule set</dfn>.
+				The input to a SHACL rules engine is a set of rules called a <dfn>rule set</dfn>.
 				A <a>rule set</a> is identified by an <a>IRI</a>.
 				The property <code>sh:hasRule</code> can be used to declare that a <a>rule set</a> has a given <a>rule</a> as a member.
 				The <dfn>default rule set</dfn> of a <a>graph</a> is the set of all <a>rules</a> in the graph.
@@ -1314,7 +1314,7 @@ _:id sh:sourceRule ex:SymmetricPropertyRule .</pre>
 				</div>
 				<p class="note">
 					For <a>SPARQL rules</a> that are <a>shape rules</a> (i.e., linked to a shape with <code>sh:rule</code>),
-					a <a>SHACL rules engine</a> also counts as a SHACL-SPARQL processor
+					a <a>SHACL rules engine</a> also counts as a SHACL-SPARQL processor (as defined by <a data-cite="shacl12-sparql"></a>)
 					and the syntax limitations required by <a>pre-binding</a> do apply to shape rules.
 					Since <a>global</a> SPARQL rules do not use <a>pre-binding</a>, the syntax limitations
 					required by <a>pre-binding</a> do not apply to them.
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -570,21 +570,26 @@ ex:RectangleRulesShape
 				SHACL has a flexible, extensible design in which multiple types of rules can be supported,
 				but this document only defines two of them: <a>SPARQL rules</a> and <a>triple rules</a>.
 			</p>
-			<p>
-				Each <dfn>rule type</dfn> is identified by an <a>IRI</a> that is used as
-				their <code>rdf:type</code>.
-				Each rule type also defines <dfn>execution instructions</dfn> that can be implemented by rule engines.
-			</p>
 			<p class="syntax">
 				<span data-syntax-rule="rule-type">Each <a>SHACL rule</a> has at least one <code>rdf:type</code>
 				which is a <a>IRI</a>.</span>
 			</p>
-			<p>
-				Rules can have multiple types, e.g., to provide instructions that work either in SPARQL or JavaScript,
-				depending on the capabilities of the engine.
-				The creator of such rules needs to make sure that such rules have consistent semantics.
-				Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
-			</p>
+			<section id="rule-types">
+				<h3>Rule Types</h3>
+				<p>
+					A <dfn>rule type</dfn> is identified by an <a>IRI</a>.
+					Each <a>rule type</a> defines <dfn>execution instructions</dfn> that can be implemented by rule engines through the function
+					<code>execute(rule, focusNodes) -&gt; inferredTriples</code> where
+					<code>rule</code> is the rule to execute, <code>focusNodes</code> is a set of <a>nodes</a> provided by a rule engine,
+					and <code>inferredTriples</code> is the graph of (output) <a>triples</a> that are not in the <a>base graph</a>.
+				</p>
+				<p>
+					<a>Rule</a> <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
+					Rules can have multiple types, e.g., to provide instructions that work either in SPARQL or JavaScript,
+					depending on the capabilities of the engine.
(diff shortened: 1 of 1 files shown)


afs:
> a [SHACL rules engine](https://raw.githack.com/w3c/data-shapes/issue-1228/shacl12-inference-rules/index.html#dfn-rules-engine) also counts as a SHACL-SPARQL processor.

There are several pieces of terminology that are related but what the differences are between a "SHACL rules engine", "rule engines", a "SHACL inference rule processor" and a "SHACL-SPARQL processor"?

shacl12-inference-rules/index.html
```diff
@@ -570,21 +570,26 @@ <h2>Syntax of SHACL Rules</h2>
...
-				depending on the capabilities of the engine.
-				The creator of such rules needs to make sure that such rules have consistent semantics.
-				Rule <code>R</code> has <a>rule type</a> <code>T</code> if <code>R</code> is a <a>SHACL instance</a> of <code>T</code>.
-			</p>
+			<section id="rule-types">
+				<h3>Rule Types</h3>
+				<p>
+					A <dfn>rule type</dfn> is identified by an <a>IRI</a>.
+					Each <a>rule type</a> defines <dfn>execution instructions</dfn> that can be implemented by rule engines through the function
+					<code>execute(rule, focusNodes) -&gt; inferredTriples</code> where
```
afs, line 582:
```suggestion
					<code>execute(rule, focusNodes, baseGraph, evaluation graph) -&gt; inferredTriples</code> where
```
and evaluation graph is the input to the layer.

shacl12-inference-rules/index.html
```diff
@@ -570,21 +570,26 @@ <h2>Syntax of SHACL Rules</h2>
...
+					<code>rule</code> is the rule to execute, <code>focusNodes</code> is a set of <a>nodes</a> provided by a rule engine,
+					and <code>inferredTriples</code> is the graph of (output) <a>triples</a> that are not in the <a>base graph</a>.
```
afs, line 584: "Base graph" or "evaluation graph" - the latter is new-this-iteration.

Isn't it the rules engine that updates the inference graph? The rule step code does not need the inference graph except as part of the evaluation graph. It can then return the result of CONSTRUCT directly.

shacl12-inference-rules/index.html
```diff
@@ -1122,18 +1127,31 @@ <h2>General Execution Instructions for SHACL Rules</h2>
...
+				After execution of a rule, the inferred triples from the <code>execute</code> function are added to the <a>inference graph</a>.
 			</p>
 			<p>
 				For a given <a>rule set</a> in the <a>shapes graph</a>
 				an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
 				specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-				Within an iteration, the inferred triples of one rule become immediately visible to the next rule.
+				Within an iteration, the inferred triples of one rule become immediately visible to the next rule, but
+				rules with the same order can not see each other's inferred triples until the iteration has completed.
+				This means that rules with the same order may be executed in parallel.
```
afs:
```suggestion
				This means that rules with the same order may be executed concurrently.
```

HolgerKnublauch:
> There are several pieces of terminology that are related but what the differences are between a "SHACL rules engine", "rule engines", a "SHACL inference rule processor" and a "SHACL-SPARQL processor"?

Good point, I have attempted to clarify this in the current PR.

osavkovic:
Hi Holger,

Just to say, I believe there is a slight difference in the "solutions".

I think you are saying (please correct me if I am wrong):

- "The same order cannot see each other's inferred triples," so apply each rule in isolation until the fixpoint, and only at the end take the union of all inferred triples and move to the next layer.
- *My proposal was:* All the rules in the same order and layer are applied on the same graph and the result is unioned. We repeat this until we reach a fixpoint, and only then move to the next layer (and order).

The first one was @livio's algorithm:

```
AlgL
for i = 0; i < max(sh:order); i++:
    collect the set S of all rules having sh:order = i
    execute all rules in S in parallel, i.e., isolated from each other
    merge all graphs, each inferred by one of the rules, into a single graph
```

This was my alternative:

```
GenAlgorithm
G' := G
for each layer i, in ascending order:
    G' := G' ∪ ⋃ { r(G') | r run-once in R^i }        // all together
    repeat
        G_prev := G'
        for each order value o occurring in R^i, ascending:
            G' := G' ∪ ⋃ { r(G') | r in R^i_o }       // all together
    until G' = G_prev
delete the temporary triples and their reifiers
```

Example.

Take a 4-node path as base data: E(a,b), E(b,c), E(c,d). Same-order rules:

```
p(X,Y) :- E(X,Y).
p(X,Y) :- q(X,Z), E(Z,Y).
q(X,Y) :- E(X,Y).
q(X,Y) :- p(X,Z), E(Z,Y).
```

AlgL doesn't produce anything beyond the base edges (no transitive closure). GenAlgorithm produces the full transitive closure on paths.

Expressivity-wise, I don't think we lose anything, since every positive rule set (even stratified) can be rewritten, by introducing auxiliary predicates, into an "equivalent" rule set. 

For instance:

```
order 1
r(X, Y) :- E(X, Y).
r(X, Y) :- p(X, Y).
r(X, Y) :- q(X, Y).

order 2
r(X, Y) :- r(X, Z), E(Z, Y).

order 3
p(X, Y) :- r(X, Y).
q(X, Y) :- r(X, Y).
```

In this case, the translation is rather simple. However, a general translation might be more involved.

afs:
> Rules may specify their relative execution order within the same layer as defined in this section.

Case 1:

`RA order 1 depends on other rules in the layer` less, same, or higher order.
`RB order 2 depends on RA and other rules in the layer`

RB does `count()` (or `optional`, `not exist` etc - any closed operation).

In round 1, RB/round=1 runs on the eval graph after RA/round=1.
In round 2, RA/round=2 runs again and can see the outcome of all rules it depends on in the layer. This includes RB/round=1. 

`sh:order` is not saying RA completes before RB.

If RA has counting, it outputs potential different counts.

This needs more time to investigate -- maybe the intent of `sh:order` is OK with that. #1216 would help to frame the intent of `sh:order`. 

This case does not depend on this PR - it does relate the goal of non-determinism within the framework, regardless of whether the rule types are deterministic or not.

IMO the current text allows concurrent execution albeit with non-determinism from the framework.

Case 2:

If we now add

`RC order 1 - no overlap with RA and RB` example: `CONSTRUCT {} WHERE { }`

and now in the iteration round=1 loop, RB runs on the state before RA runs, different to case 1. Different counts - adding an unrelated rule has caused a change.

(It outputs a count from before RA it didn't before, and then probably the same counts as case 1, shifted up one round.)


Summary: 
It's complicated. 

Maybe the examples are not intended for `sh:order`; maybe the examples are wrong (there were done in limited time).

afs:
@HolgerKnublauch --

> close #1226.

If you leave that in the description, this PR will close 1226.
(this comment will not - GH only looks in descriptions)

shacl12-inference-rules/index.html
HolgerKnublauch: If we change this here then the whole algorithm in the execution instructions also needs to change. I am aware that conceptually the evaluation graph can be a fresh graph each time, yet I would consider that an implementation choice and personally find the current wording slightly better.

So who is willing to rewrite all this here? Or do we once again end up with another stale branch?

shacl12-inference-rules/index.html
HolgerKnublauch: On the question of whether the rule returns all triples or just the new triples, I also have no strong opinion but as written now, the expectation is that it will only return new triples (that's how "inferred" is defined). So again this would be an overarching change across multiple sections and terms.

My preference remains to do micro-steps into the right direction or otherwise just leave it be unless someone can create a full rewrite that still makes sense.

HolgerKnublauch:
> Hi Holger,
> 
> Just to say, I believe there is a slight difference in the "solutions".
> 
> I think you are saying (please correct me if I am wrong):
> 
> * "The same order cannot see each other's inferred triples," so apply each rule in isolation until the fixpoint, and only at the end take the union of all inferred triples and move to the next layer.
> * _My proposal was:_ All the rules in the same order and layer are applied on the same graph and the result is unioned. We repeat this until we reach a fixpoint, and only then move to the next layer (and order).
> 
> The first one was @livio's algorithm:
> 
> ```
> AlgL
> for i = 0; i < max(sh:order); i++:
>     collect the set S of all rules having sh:order = i
>     execute all rules in S in parallel, i.e., isolated from each other
>     merge all graphs, each inferred by one of the rules, into a single graph
> ```
> 
> This was my alternative:
> 
> ```
> GenAlgorithm
> G' := G
> for each layer i, in ascending order:
>     G' := G' ∪ ⋃ { r(G') | r run-once in R^i }        // all together
>     repeat
>         G_prev := G'
>         for each order value o occurring in R^i, ascending:
>             G' := G' ∪ ⋃ { r(G') | r in R^i_o }       // all together
>     until G' = G_prev
> delete the temporary triples and their reifiers
> ```
> 
> Example.
> 
> Take a 4-node path as base data: E(a,b), E(b,c), E(c,d). Same-order rules:
> 
> ```
> p(X,Y) :- E(X,Y).
> p(X,Y) :- q(X,Z), E(Z,Y).
> q(X,Y) :- E(X,Y).
> q(X,Y) :- p(X,Z), E(Z,Y).
> ```
> 
> AlgL doesn't produce anything beyond the base edges (no transitive closure). GenAlgorithm produces the full transitive closure on paths.
> 
> Expressivity-wise, I don't think we lose anything, since every positive rule set (even stratified) can be rewritten, by introducing auxiliary predicates, into an "equivalent" rule set.
> 
> For instance:
> 
> ```
> order 1
> r(X, Y) :- E(X, Y).
> r(X, Y) :- p(X, Y).
> r(X, Y) :- q(X, Y).
> 
> order 2
> r(X, Y) :- r(X, Z), E(Z, Y).
> 
> order 3
> p(X, Y) :- r(X, Y).
> q(X, Y) :- r(X, Y).
> ```
> 
> In this case, the translation is rather simple. However, a general translation might be more involved.

Ok I guess this topic is better handled on #1226 now that this PR no longer talks about sh:order. You may want to repost your observations there @osavkovic

mgberg (approved):
This looks good to me, but I'm not sure if @afs's outstanding comments should be addressed in a new issue.

HolgerKnublauch:
I'll merge it in. Andy's comments can be picked up in another ticket but I believe the semantics will be the same either way. It's just another way of defining the algorithm.

PR #1241, opened 2026-09-09
[UI] Improvements to seach query

danielbeeke:
I cleaned up the shui:searchQuery a bit more. It had sometimes to much text.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/improvements-to-seachQuery/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -2675,7 +2675,7 @@ ex:config
       </p>
 
       <p>
-        The query MUST expect the following variables to be pre-bound by the renderer before evaluation:
+        The SHACL renderer MUST pre-bind the following variables:
       </p>
 
       <table class="term-table">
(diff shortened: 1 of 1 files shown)


shacl12-ui/index.html
edmondchuc, line 2734: > beyond the first described above

Looks like we've moved the description of additional bindings. Should this be updated too?

danielbeeke: I think it was not moved:

"Implementations MAY pre-bind additional implementation-specific variables."

Maybe the sentence is unclear and can be cleaned up

edmondchuc: Ah yes, you're right! I saw the diff and thought we removed the text around additional variables.

> The query MAY project
        additional variables beyond the first, and a SHACL Renderer MAY add bindings for such additional variables
        to the candidate value nodes it offers to the end user, for example to carry a label or other data
        associated with each candidate value node.

edmondchuc (approved):

PR #1242, opened 2026-09-09
#1227: Replaced "reifier" with "triple annotations" in a few places

### LINKED ISSUE #1227: Reifiers and annotations
> [Temporary triples](https://github.com/w3c/data-shapes/pull/1216#dfn-temporary-triple) and their reifiers are visible to executing rules.

This and other uses of reifier look like they are "annotations".

The three concepts are (with their links):

* reifier -- subject of an `rdf:refiies` triple -- https://www.w3.org/TR/rdf12-concepts/#dfn-reifier
* reifying triples -  a triple `reifier rdf:reiifies <<( )>>` -- https://www.w3.org/TR/rdf12-concepts/#dfn-reifying-triple
* annotation -- "the subset of triples including the [reifier](https://www.w3.org/TR/rdf12-concepts/#dfn-reifier) as subject" -- https://www.w3.org/TR/rdf12-concepts/#dfn-triple-annotation -- the top level of  `{| |}`

(( Partial extract of https://github.com/w3c/data-shapes/pull/1216#discussion_r3895123080 ))

HolgerKnublauch:
Closes #1227

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1227/shacl12-inference-rules/index.html)

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -211,6 +211,7 @@
 						<dfn data-cite="rdf12-concepts#dfn-rdf-node" data-lt="node|nodes">node</dfn> of an RDF graph,
 						<dfn data-cite="rdf12-concepts#dfn-datatype" data-lt="datatype|datatypes">datatype</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-reifier">reifier</dfn>,
+						<dfn data-cite="rdf12-concepts#dfn-triple-annotation">triple annotation</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-rdf-term" data-lt="term|terms">RDF term</dfn>, and
 						<dfn data-cite="rdf12-concepts#dfn-subject" data-lt="subject|subjects">subject</dfn>,
 						<dfn data-cite="rdf12-concepts#dfn-predicate" data-lt="predicate|predicates">predicate</dfn>, and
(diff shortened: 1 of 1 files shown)


afs:
>> The property sh:sourceRule can be used in a [reifier](https://www.w3.org/TR/rdf12-concepts/#dfn-reifier) of a triple in the

As in `:s :p :o {| sh:sourceRule  :rule57 |}`?
-- that's in an annotation.

HolgerKnublauch:
> > > The property sh:sourceRule can be used in a [reifier](https://www.w3.org/TR/rdf12-concepts/#dfn-reifier) of a triple in the
> 
> As in `:s :p :o {| sh:sourceRule :rule57 |}`? -- that's in an annotation.

I think in this case the term reifier is better. That term is also still used elsewhere in the document, with the same meaning: the subject of a rdf:reifies triple. The sentence "The property sh:sourceRule can be used in a reifier" means that it is permitted to have a value for sh:sourceRule at a subject of rdf:reifies (for that triple). The term annotation is better when we talk about the complete set of triples that have the reifier as subject, but not for the case when the sh:sourceRule triple is not even created yet.

Anyway, both terms could be used, although if we use annotation here then we should eliminate the other two usages of reifier too.

mgberg:
Section 7 says
> A temporary triple is an inferred triple for which the inference graph contains a reifier with the value sh:tempTriple true.

So then the alternative phrasing would be something like
> A temporary triple is an inferred triple for which the inference graph contains a triple annotation with the value sh:tempTriple true.

and/or

> The property sh:sourceRule can be used in a triple annotation for a triple in the inference graph to link the triple with the rule.

I don't have an opinion one way or the other at the moment, I haven't talked about RDF 1.2 reification enough to know which makes more sense/is easier to understand.

PR #1243, opened 2026-09-10
Updated Custom Rule Processors section introducing sh:ruleProcessor property

HolgerKnublauch:
This clarifies that custom rule processors (such as a SPARQL-RL stratification algorithm) can only compute sh:layer, order, runOnce when no values are explicitly given. This means that authors can leave those values out when they know that SRL will handle them, and "fix" other rules by giving them explicit layers.

The property sh:ruleProcessor can be used to enforce expectations, e.g. that stratification MUST be applied beforehand. See examples of this in my comment https://github.com/w3c/data-shapes/pull/1230#discussion_r3964203449

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor)

CC: @afs @robert-david @recalcitrantsupplant

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -664,7 +664,7 @@ ex:SymmetricPropertyRule
 				<h3>Grouping of Rules into Layers (sh:layer)</h3>
 				<p>
 					Rules may be grouped into <dfn data-lt="layer">layers</dfn>, identified by a numeric value.
-					During execution, a SHACL rules engine will iterate over all rules in the same layer
+					During execution, a SHACL rules engine will iterate over all rules with the same numeric layer value
 					before moving to the next layer.
 					Layers with a smaller numeric value will be executed before those with a larger number.
 				</p>
@@ -672,7 +672,7 @@ ex:SymmetricPropertyRule
 					<span data-syntax-rule="rule-layer-maxCount">Each <a>rule</a> may have at most one <a>value</a> for the
 					property <code>sh:layer</code>.</span>
 					<span data-syntax-rule="rule-layer-datatype">The values of <code>sh:layer</code> at <a>rules</a>
-					are <a>literals</a> with datatype <code>xsd:integer</code>.</span>
+					are <a>literals</a> with datatype <code>xsd:integer</code> or <code>xsd:decimal</code>.</span>
 				</p>
 				<p>
 					If unspecified and unless it has been computed by a <a href="#ruleProcessor">custom rule processor</a>, the default <a>layer</a> of a rule is <code>0</code>.
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -675,7 +675,7 @@ ex:SymmetricPropertyRule
 					are <a>literals</a> with datatype <code>xsd:integer</code>.</span>
 				</p>
 				<p>
-					If unspecified, then the default <a>layer</a> of a rule is <code>0</code>.
+					If unspecified and unless it has been computed by a <a href="#ruleProcessor">custom rule processor</a>, the default <a>layer</a> of a rule is <code>0</code>.
 				</p>
 				<p>
 					An example of <code>sh:layer</code> to control the execution order of rules is provided in <a href="#example-rules-before-after"></a>.
@@ -693,7 +693,7 @@ ex:SymmetricPropertyRule
 					are <a>literals</a> with datatype <code>xsd:decimal</code> or <code>xsd:integer</code>.</span>
 				</p>
 				<p>
-					If unspecified, then the default <a>execution order</a> is <code>0</code>.
+					If unspecified and unless it has been computed by a <a href="#ruleProcessor">custom rule processor</a>, the default <a>execution order</a> is <code>0</code>.
 					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
 				</p>
 				<aside class="example" title="Rule order example">
@@ -756,6 +756,7 @@ ex:RuleOrderExampleShape
 					A <dfn>run-once rule</dfn> is a <a>rule</a> for which at most one <a>iteration</a> is performed (per <a>shape</a>, if it is a <a>shape rule</a>), i.e. it is executed at most once per <a>focus node</a>.
 					All other rules are called <dfn>iterating rules</dfn>.
 					A <a>rule</a> that has <code>true</code> as its <a>value</a> for <code>sh:runOnce</code> is a <a>run-once rule</a>.
+					<a href="#ruleProcessor">Custom rule processors</a> may compute additional run-once rules when <code>sh:runOnce</code> is absent.
 				</p>
 				<aside class="example" title="Example of run-once rules and layers" id="example-rules-before-after">
 					<div class="shapes-graph">
@@ -1198,8 +1199,8 @@ _:id sh:sourceRule ex:SymmetricPropertyRule .</pre>
 			</p>
 		</section>
 
-		<section id="rule-variations">
-			<h2>Variations of SHACL Rule Implementations</h2>
+		<section id="ruleProcessor">
+			<h2>Custom Rule Processors (sh:ruleProcessor)</h2>
 			<p>
 				The <a href="#rules-execution">general execution algorithm</a> described above is intentionally
 				kept generic and offers a lot of flexibility to specific implementations.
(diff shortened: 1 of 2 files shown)


liviorobaldo:
I report here the comments I made on the other PR:

I don't understand the claim in §3.5:

"If unspecified and unless it has been computed by a [custom rule processor](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor), the default [execution order](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#dfn-execution-order) is 0."

Perhaps I missed some communications from our meetings, but I don't recall us deciding that rules computed by a custom rule processor have no order. It is also not clear to me how this is supposed to work.

As for the context of §10, please have a look at https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205, at the bottom of Issue https://github.com/w3c/data-shapes/issues/1226. Perhaps we should leave the text more open, without referring to specific details such as "(non-)determinism" and the RAND function.

mgberg:
> "If unspecified and unless it has been computed by a [custom rule processor](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor), the default [execution order](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#dfn-execution-order) is 0."

Before this PR, the default order for rules that did not have an order specified was 0. I believe that rules in the same layer and order have no defined execution order, which means they should be parallelizable.

HolgerKnublauch:
> I don't understand the claim in §3.5:
> 
> "If unspecified and unless it has been computed by a [custom rule processor](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor), the default [execution order](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#dfn-execution-order) is 0."
> 
> Perhaps I missed some communications from our meetings, but I don't recall us deciding that rules computed by a custom rule processor have no order. It is also not clear to me how this is supposed to work.

The steps are:
1. If the sh:Rule has a sh:order explicitly stated, then use that sh:order
2. If the sh:Rule declares a sh:ruleProcessor or is in a RuleSet that does so, then ask the ruleProcessor to compute the order of the rule
3. If it's still undefined, use 0

Do you see problems with those steps, and how can I make this clearer in the text? From my reading I find it clear but I was of course the author.


> As for the context of §10, please have a look at [#1226 (comment)](https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205), at the bottom of Issue #1226. Perhaps we should leave the text more open, without referring to specific details such as "(non-)determinism" and the RAND function.

I think here it is uncontroversial that RAND is non-deterministic. And there are plenty of other features in SPARQL-Full that are like that. What matters more is that the framework itself is deterministic, and with the changes to same-order we have IMHO achieved that.

HolgerKnublauch:
> > "If unspecified and unless it has been computed by a [custom rule processor](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor), the default [execution order](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#dfn-execution-order) is 0."
> 
> Before this PR, the default order for rules that did not have an order specified was 0. I believe that rules in the same layer and order have no defined execution order, which means they should be parallelizable.

Correct. (Was this a question, a comment, or a change request :) ?)

mgberg:
> Correct. (Was this a question, a comment, or a change request :) ?)

Just a comment

liviorobaldo:
> The steps are:
> If the sh:Rule has a sh:order explicitly stated, then use that sh:order
> If the sh:Rule declares a sh:ruleProcessor or is in a RuleSet that does so, then ask the ruleProcessor to compute the order of the rule
> If it's still undefined, use 0
> Do you see problems with those steps, and how can I make this clearer in the text? From my reading I find it clear but I was of course the author.

OK, now I see what you mean. And yes, the text does actually say that; it just needs to be read carefully, and possibly more than once.

However, my biggest source of confusion is always when I try to imagine how things work when (1) we mix rules of different types in the same rule set, or (2) multiple rule sets are defined in the same graph. Indeed, I got confused by this sentence because I was imagining a situation in which (1) and/or (2) occurred.

I think we should address (1) and (2) explicitly. At the moment, the document is silent on these cases, so it leaves open the possibility that both (1) and (2) can occur, possibly together.

Regarding (1), as I said at the meeting, I think the solution is _not_ to allow rules of different types in the same rule set. Therefore, a rule set can contain ONLY _all_ SPARQL rules, _all_ Triple Rules, or _all_ SPARQL-RL rules. The `sh:ruleProcessor` property would then apply only to a rule set, rather than to an individual rule.

Regarding (2), some more substantial amendments would be required, and these are definitely outside the scope of this PR. However, let me briefly summarize the issue here, since I have already mentioned it.

§4 says: "The input to a SHACL rules processor is a set of rules called a rule set." Example 8, however, shows a graph containing _TWO_ rule sets. So, how exactly does this work? Is there a method like TopBraid's `RuleUtil.executeRules` that takes the IRI of a rule set as input? Or does the processor take a whole RDF file as input, in which case is the intended input "the set of all global rules defined in the RDF file"?

Note that in Example 8, `ex:Rule1` and `ex:Rule2` are actually global rules. So, to me, there are _three_ sets of rules here: `ex:RuleSet1`, `ex:RuleSet2`, and the set of global rules. According to what I am reading here, each of these could potentially be provided as "the input to a SHACL rules processor". How exactly is this supposed to work?

liviorobaldo:
> As for the context of §10, please have a look at https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205, at the bottom of Issue https://github.com/w3c/data-shapes/issues/1226. Perhaps we should leave the text more open, without referring to specific details such as "(non-)determinism" and the RAND function.
> I think here it is uncontroversial that RAND is non-deterministic. And there are plenty of other features in SPARQL-Full that are like that. What matters more is that the framework itself is deterministic, and with the changes to same-order we have IMHO achieved that.

Yes, as both @afs and I said in the comments I linked above, we all agree that "the framework itself is deterministic, and with the changes to same-order we have IMHO achieved that".

However, to spell this out we would then need to explain in the document the distinction between "the framework", which is deterministic, and the additional functions from SPARQL that are still allowed in SHACL 1.2 and may be non-deterministic. I think this explanation could become quite convoluted, potentially confusing readers. They may wonder why we cared to explain this distinction and then try to understand why it is important.

I don't think it is important. We just have to say "rules with the same `sh:order` are executed concurrently". Then, a reader who investigates SHACL 1.2 in depth (let's call them "a SHACL 1.2 nerd") may eventually realise: "Ah! If rules with the same `sh:order` are executed concurrently, then their output is always the same, i.e., the framework is deterministic... wait, that does not mean that the full SHACL 1.2 language is deterministic, because `RAND` and other similar functions are still allowed. So I can still mess up the output; I need to pay attention to those functions."

Why do we need to spell all this out in the spec? Let the nerds figure it out for themselves 😄

That's why I was saying (and I understood @afs to agree):

As for "The goal is only about the framework", I agree, but I wouldn't mention that in the spec. Nor would I mention determinism/non-determinism. Sometimes it is better to avoid specifying too many details and leave the text more open, to avoid what we call in Italian "un vespaio di polemiche" (a hornet's nest of criticism)... or, depending on the country in which the SHACL 1.2 nerd lives, to "avoid letting the cat out of the bag".

So I would remove every occurrence of "determinism" and "non-determinism" from the spec. I would simply leave "rules with the same `sh:order` are executed concurrently".

HolgerKnublauch:
Yes makes sense, @liviorobaldo I have updated this branch to simplify it. Please check again.

On whether rule sets may mix rule types, I am less concerned. For example I expect that many people will not even use rule sets, but instead just use the default rule set (i.e. no sh:hasRule links). In those cases, they will naturally mix SPARQLRule and TripleRule, for example, and I don't see why they shouldn't be allowed to. If a ruleProcessor isn't able to handle a mixed rule set, it always has the choice to report a failure. This can be left to the implementations that use this extension point.

liviorobaldo:
@HolgerKnublauch, the text is now indeed more generic, so that's fine with me. I'm going to approve this PR so that we can move forward, but I think we still need to address the following points:

> On whether rule sets may mix rule types, I am less concerned. For example I expect that many people will not even use rule sets, but instead just use the default rule set (i.e. no sh:hasRule links). In those cases, they will naturally mix SPARQLRule and TripleRule, for example, and I don't see why they shouldn't be allowed to. If a ruleProcessor isn't able to handle a mixed rule set, it always has the choice to report a failure. This can be left to the implementations that use this extension point.

Mixing SPARQLRule and TripleRule is not a problem. The problem could instead arise when mixing SPARQLRule/TripleRule with SPARQL-RL rules. If you are fine with allowing different rule types to be mixed, that's also fine with me. I also think that users will almost always use a SINGLE type, but I'm quite sure that many users will nevertheless ask themselves (perhaps just out of curiosity): "What happens if I mix rules of different types?" This is different from the issue of concurrently executing same-order rules, which is something only true nerds will probably worry about, so we can avoid going into those details. I think the question of mixing different rule types is much more likely to arise.

Therefore, perhaps this is the case where we should use SHOULD language 😅. We could say something like: "Users _should_ always use a single type of rule. If they mix different types, different rule processors may assign each rule an `sh:order` based on the custom rule processor's behaviour and the general criteria explained in §3.5. As a result, the behaviour of the rules could become quite difficult to control. We therefore advise against mixing different rule types."

Similar considerations apply to (2) above. Currently, the spec allows "(2) multiple rule sets are defined in the same graph". You write: "I expect that many people will not even use rule sets, but instead just use the default rule set"... OK, but what happens if they do? Again, we can advise against it, but it is still unclear to me how I can ask a rule processor to use one rule set rather than another, as I mentioned above:

> §4 says: "The input to a SHACL rules processor is a set of rules called a rule set." Example 8, however, shows a graph containing TWO rule sets. So, how exactly does this work? Is there a method like TopBraid's `RuleUtil.executeRules` that takes the IRI of a rule set as input? Or does the processor take a whole RDF file as input, in which case is the intended input "the set of all global rules defined in the RDF file"?

> Note that in Example 8, `ex:Rule1` and `ex:Rule2` are actually global rules. So, to me, there are THREE sets of rules here: `ex:RuleSet1`, `ex:RuleSet2`, and the set of global rules. According to what I am reading here, each of these could potentially be provided as "the input to a SHACL rules processor". How exactly is this supposed to work?

liviorobaldo (approved):

HolgerKnublauch:
Ok I have added a 'should' statement, albeit without going into too many details why. I would argue this depends on the evolution of the relation between SHACL rules and SRL, and what new rule types will be introduced at all. We can fine tune that after the final decisions on that have been made.

liviorobaldo:
> Ok I have added a 'should' statement, albeit without going into too many details why.

I would actually add more details 😅 And we still need to figure out what to do about the other problem, namely (2) multiple rule sets being defined in the same graph. However, this is definitely outside the scope of this PR. Therefore, I took some notes and will possibly create an issue, and later a PR, after today's meeting.

I have already approved this PR. As far as I'm concerned, you can merge it.

PR #1244, opened 2026-09-10
#1226: Added test case and changed contract for rules with same order

HolgerKnublauch:
As a starting point for a reworking of same-order logic, I have added a test case that is using Oggy's example. Both triples will always be inferred.

Could someone else extend this PR to update the text or are there already parallel branches for that? In the latter case, I suggest we merge this here into them.

### DIFF
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -695,6 +695,7 @@ ex:SymmetricPropertyRule
 				<p>
 					If unspecified, then the default <a>execution order</a> is <code>0</code>.
 					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
+					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed.
 				</p>
 				<aside class="example" title="Rule order example">
 					<div class="shapes-graph">
@@ -1133,7 +1134,10 @@ ex:SetYoungestOffspringsRule
 				For a given <a>rule set</a> in the <a>shapes graph</a>
 				an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
 				specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-				Within an iteration, the inferred triples of one rule become immediately visible to the next rule.
+				Within an iteration, the inferred triples of one rule (or group of same-order rules) become immediately
+				visible to the following rule (or group of same-order rules).
+				When rules have the same order, they must be executed concurrently and must not see each other's inferences before they
+				have all completed.
 			</p>
 			<p>
 				The <dfn>execution of a rule set</dfn> is defined as follows:
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -694,8 +694,8 @@ ex:SymmetricPropertyRule
 				</p>
 				<p>
 					If unspecified, then the default <a>execution order</a> is <code>0</code>.
-					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
-					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed.
+					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values. The inferred triples of one rule (or group of same-order rules) become immediately visible to the subsequent rule (or group of same-order rules) in the order.
+					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed. The sets of triples individually inferred by these rules are then _unioned_ to produce the inferred graph.
 				</p>
 				<aside class="example" title="Rule order example">
 					<div class="shapes-graph">
@@ -1134,10 +1134,6 @@ ex:SetYoungestOffspringsRule
 				For a given <a>rule set</a> in the <a>shapes graph</a>
 				an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
 				specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-				Within an iteration, the inferred triples of one rule (or group of same-order rules) become immediately
-				visible to the following rule (or group of same-order rules).
-				When rules have the same order, they must be executed concurrently and must not see each other's inferences before they
-				have all completed.
 			</p>
 			<p>
 				The <dfn>execution of a rule set</dfn> is defined as follows:
--- a/shacl12-inference-rules/index.html
+++ b/shacl12-inference-rules/index.html
@@ -694,8 +694,10 @@ ex:SymmetricPropertyRule
 				</p>
 				<p>
 					If unspecified, then the default <a>execution order</a> is <code>0</code>.
-					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values. The inferred triples of one rule (or group of same-order rules) become immediately visible to the subsequent rule (or group of same-order rules) in the order.
-					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed. The sets of triples individually inferred by these rules are then _unioned_ to produce the inferred graph.
+					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
+					The inferred triples of one rule (or group of same-order rules) become immediately visible to the subsequent rule (or group of same-order rules) in the order.
+					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed.
+					The sets of triples individually inferred by these rules are merged into the inference graph.
 				</p>
 				<aside class="example" title="Rule order example">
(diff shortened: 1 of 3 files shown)


shacl12-test-suite/tests/inference-rules/same-order.ttl
```diff
@@ -0,0 +1,52 @@
...
+<same-order>
+  rdf:type sht:Infer ;
+  rdfs:label "Test of two rules that have the same order (and layer), making sure they don't see each other's triples" ;
+  rdfs:seeAlso <https://github.com/w3c/data-shapes/issues/1226#issue-5313787732> ;
+  mf:action [
+      sht:dataGraph <> ;
+      sht:shapesGraph <> ;
+    ] ;
+  mf:result (
+    ( ex:SomeNode ex:p true )
```
afs, line 49: A good place to use triple terms! 

```suggestion
    <<( ex:SomeNode ex:p true )>>
```

HolgerKnublauch: See RDF 1.2 Primer: "Triple terms may only appear in the object position, and should be used with the special reifies predicate of reifying triples."

Anyway, I don't think this matters much, either syntax is fine. We already have other test cases that use these nested lists (and my implementation uses those), so I see no strong reason to change that here.

liviorobaldo:
> Could someone else extend this PR to update the text, or are there already parallel branches for that? If the latter, I suggest we merge this into them here.

I could... but I'm afraid I'm not expert enough on GitHub to do all this (extending PRs, merging branches, etc.). I barely learned how to create individual branches and PRs during this WG (forgive me, I'm an academic 😅).

I can create a PR to implement [what I suggested here](https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205): modifying the algorithm in [§8](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#rules-execution) and fixing the text in [§10](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor) to avoid mentioning non-determinism. I can also re-read the spec from the top and fix related claims here and there.

However, which branch should I branch off from? I see you just created another PR for §10 ([PR 1243](https://github.com/w3c/data-shapes/pull/1243)). Perhaps I should propose the §10 amendments through that PR.

Then, I still see the issue I mentioned [in this comment](https://github.com/w3c/data-shapes/pull/1230#issuecomment-5527192495): the SHACL 1.2 Inference Rules page is not reachable from SHACL 1.2 Core or the other sections. This makes me even more confused about which parent branch I should use for the PR...

shacl12-test-suite/tests/inference-rules/same-order.ttl
afs: That is as opposed to the subject or predicate position. The context is about triples.

The primer also says:
See also: "Literals may only appear in the object position of a triple."

But if you want a purist solution:

```
<< ex:SomeNode ex:p true >>
```

HolgerKnublauch: If you feel strongly about this, I suggest you create a fresh issue and PR to change this everywhere. This isn't specific to the current PR.

HolgerKnublauch:
> > Could someone else extend this PR to update the text, or are there already parallel branches for that? If the latter, I suggest we merge this into them here.
> 
> I could... but I'm afraid I'm not expert enough on GitHub to do all this (extending PRs, merging branches, etc.). I barely learned how to create individual branches and PRs during this WG (forgive me, I'm an academic 😅).
> 
> I can create a PR to implement [what I suggested here](https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205): modifying the algorithm in [§8](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#rules-execution) and fixing the text in [§10](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor) to avoid mentioning non-determinism. I can also re-read the spec from the top and fix related claims here and there.
> 
> However, which branch should I branch off from? I see you just created another PR for §10 ([PR 1243](https://github.com/w3c/data-shapes/pull/1243)). Perhaps I should propose the §10 amendments through that PR.
> 
> Then, I still see the issue I mentioned [in this comment](https://github.com/w3c/data-shapes/pull/1230#issuecomment-5527192495): the SHACL 1.2 Inference Rules page is not reachable from SHACL 1.2 Core or the other sections. This makes me even more confused about which parent branch I should use for the PR...

I agree it's not good to have too many open branches.

The hidden joy of working in a distributed team with members on different schedules and, often, opinions, is that things can get stuck badly. Unfortunately we need to remain mindful of the deadlines of the formal process.

As we generally agree on the direction of what needs to change with same-order, I suggest we at least move ahead with the minimum required change. @liviorobaldo suggests to change the wording "that rules with the same sh:order must be executed concurrently". I have just updated this branch here to reflect this. If that is an OK step in the right direction at least, feel free to approve the branch or suggest fine tuning: https://github.com/w3c/data-shapes/pull/1244

On the branch of the sh:ruleProcessor, you can simplify the process by approving the branch, if you agree it's the right direction. I would then merge it in soon to declutter the backlog: https://github.com/w3c/data-shapes/pull/1243

shacl12-inference-rules/index.html
```diff
@@ -695,6 +695,7 @@ <h3>Ordering of Rules (sh:order)</h3>
 				<p>
 					If unspecified, then the default <a>execution order</a> is <code>0</code>.
 					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
+					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed.
```
```suggestion
					Rules with the same order are executed concurrently and must not see each other's inferences before they have all completed. The sets of triples individually inferred by these rules are then _unioned_ to produce the inferred graph.
```

shacl12-inference-rules/index.html
```diff
@@ -695,6 +695,7 @@ <h3>Ordering of Rules (sh:order)</h3>
 				<p>
 					If unspecified, then the default <a>execution order</a> is <code>0</code>.
 					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values.
```
```suggestion
					When the <a>rules</a> are executed, within the same <a>layer</a>, <a>rules</a> with larger order values will be executed after those with smaller values. The inferred triples of one rule (or group of same-order rules) become immediately visible to the subsequent rule (or group of same-order rules) in the order.
```

shacl12-inference-rules/index.html
```diff
@@ -1133,7 +1134,10 @@ <h2>General Execution Instructions for SHACL Rules</h2>
 				For a given <a>rule set</a> in the <a>shapes graph</a>
 				an <dfn data-lt="iterate">iteration</dfn> is a single <a>execution</a> of each individual rule in the order as
 				specified by <a href="#rule-order"></a>, skipping the rules that are <a href="#deactivated-rules">deactivated</a>.
-				Within an iteration, the inferred triples of one rule become immediately visible to the next rule.
+				Within an iteration, the inferred triples of one rule (or group of same-order rules) become immediately
+				visible to the following rule (or group of same-order rules).
+				When rules have the same order, they must be executed concurrently and must not see each other's inferences before they
+				have all completed.
```
```suggestion
```

I think this can be removed from here. There is already a backward pointer, "...as specified by <a href="#rule-order"></a>". Readers can follow it to see how `Execute one iteration over all iterating rules in the layer` works.

The sentence "Within an iteration, the inferred triples of one rule (or group of same-order rules) become immediately visible to the following rule (or group of same-order rules)." is useful. I have slightly modified it and moved it above. This content is therefore redundant and can be removed, IMHO.

liviorobaldo:
OK, let's incorporate my suggested amendments into this PR (if you and the others agree). As you can see, I have made some suggested modifications to the text.

>  On the branch of the sh:ruleProcessor, you can simplify the process by approving the branch, if you agree it's the right direction. I would then merge it in soon to declutter the backlog: https://github.com/w3c/data-shapes/pull/1243

Let's first close and merge this PR, and then we can move on to that one. Indeed, I mostly disagree with what I read there.

In particular, I don't understand the claim in §3.5:

"If unspecified and unless it has been computed by a [custom rule processor](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#ruleProcessor), the default [execution order](https://raw.githack.com/w3c/data-shapes/issue-1202-ruleProcessor/shacl12-inference-rules/index.html#dfn-execution-order) is 0."

Perhaps I missed some communications from our meetings, but I don't recall us deciding that rules computed by a custom rule processor have no order. It is also not clear to me how this is supposed to work.

As for the context of §10, please have a look at [yesterday's discussion between me and @afs](https://github.com/w3c/data-shapes/issues/1226#issuecomment-5614827205), at the bottom of Issue #1226. Perhaps we should leave the text more open, without referring to specific details such as "(non-)determinism" and the `RAND` function.

HolgerKnublauch:
Ok, all applied, please review again if it's consistent now. Let's continue the other topic on its individual ticket.

liviorobaldo (approved):

liviorobaldo:
Approved. And I move the comments to the other PR.

HolgerKnublauch:
Merged in but, as always, further fine tuning PRs are welcome.

PR #1246, opened 2026-09-10
Sec 3 update

### LINKED ISSUE #875: SHACL Profile proposal: disallowing recursive constraints

nicholascar:
Minor updates to Profiling Section 3: Profiling of SHACL.

(the PR looks big because it builds on PR #1111 but when that's merged, it will appear small!)

See this rendered at:

* [Section 3](https://raw.githack.com/w3c/data-shapes/sec-3-update/shacl12-profiling/index.html#profiling-of-shacl)
* [Appendix A](https://raw.githack.com/w3c/data-shapes/sec-3-update/shacl12-profiling/index.html#profiles)

@osavkovic I have included you as a reviewer here due to the definition of the [Non-Recursive Profile](https://raw.githack.com/w3c/data-shapes/sec-3-update/shacl12-profiling/index.html#non-recursion-profile), so please could you take a look at that.

Still to do: Update A.3.2 - A.3.4 the other non-Specification profiles, after which I will take this out of Draft.


osavkovic:
Just a quick observation. 

Do you want a non-recursive profile or this is meant to be called postive shacl profile?

btw, i am not sure you will be able to define non-recursve profile by only excluding SHACL elements.

nicholascar:
> btw, i am not sure you will be able to define non-recursve profile by only excluding SHACL elements.

Yes, these are the questions I thought would be hard to answer! It does seem that people will want a computationally easy profile of SHACL so I'm really asking: what's the best way we can achieve this? We could just say "don't do recursion" and people can say they conform to that, but that's not really helping people achieve it. So what can we do, in terms of limiting SHACL inputs to computation, not the computation engines, to achieve this? I though I read that excluding these elements, and perhaps a few other things, would get pretty close to no recursion.

> Do you want a non-recursive profile or this is meant to be called postive shacl profile?

What do you think is the most computationally-limited profile we _could_ define and how would we do that? I'm happy to give such a limited profile a more accurate name, like positive profile, if you think that makes sense.

scotthenninger:
See https://raw.githack.com/w3c/data-shapes/ISSUE

There is a shack-shacl shape meant to prevent some types of recursion. Comments and corrections welcomed.

YoucTagh:
@scotthenninger I think this is [the file](https://github.com/w3c/data-shapes/blob/sec-3-update/shacl12-profiling/examples/profile-nr.ttl) you wanted to point to. If not please correct me.

scotthenninger:
It’s the section titled “ Profile for Disallowing Recursive Constraints”. The pR is at https://github.com/w3c/data-shapes/pull/978

osavkovic:
@nicholascar A bit longer asnwer...

### The non-recursive case can't be defined down by syntax alone

SHACL isn't built in a way that lets you spot "non-recursive" shapes just by looking at which keywords are used. The only real way to define it is to build the graph of which shape depends on which, and say that graph can't have any loops. No other condition/restriction is needed. 

If you try to get a rough, syntax-only stand-in for that instead, you'd have to forbid `sh:node`, `sh:property`, `sh:not`, `sh:and`, `sh:or`, `sh:xone`, and `sh:qualifiedValueShape` . This way we loosing too much of expressivity. 

### My proposal for "well-behaved" recursion: define a positive SHACL core

Staying inside the positive profile only means leaving out four things: `sh:not`, `sh:qualifiedMaxCount`, `sh:qualifiedValueShapesDisjoint`, and `sh:xone`. Everything else, including recursion itself, stays available.  Mathematically it is well defined, and it can be easily implemented on top of the existing algorithms and it runs fast (polynomial time).

Defining a profile like this to get good computational behavior isn't unusual in the semantic web world. Eg., OWL 2's profiles are a good example of the same idea.

### No restrictions is still possible for recursion

This also possible and still be defined precisely, and it's actually simple to implement, just we know that one can't run in polynomial time in the worst case (unless P=NP).

I would be also happy write up an algorithm for the unrestricted case and test cases (we already have a few from our earlier papers). 

### Possible fragments/profiles

- **No-recursion fragment** — not something you can define as a syntax profile, only by checking the shape dependency graph for cycles.
- **Positive, well-behaved profile** — recursion is allowed, and it's guaranteed to always work, in polynomial time.
- **No restrictions** — full expressivity, recursion fully allowed, but computationally expensive.

I fine whatever you guys decide to pick as long as we have some support for the recursion

ajnelson-nist:
Does PR #978 block this PR?

ajnelson-nist:
@nicholascar - I've filed [shacl-resources PR 12](https://github.com/w3c/shacl-resources/pull/12) to set the home for the "production" profile graph.  I started it in part from the example profile graph that's part of this PR.  Will both be retained?

Issue #1248, opened 2026-09-10
[UI] Use shui:editor and shui:viewer properties in shui:WidgetScore


bergos:
A `shui:WidgetScore` identifies its widget with `shui:widget`, the same property for editors and viewers. The mode a score applies to is therefore not part of the score, it is only implied by the class of the referenced widget. As a result, the scoring system returns widgets of all kinds, not just the one asked for.

The editor and the viewer of a kind are separate score nodes that repeat the same matcher shapes, for example `shui:isIRI` in both `iri-editor.ttl` and `iri-viewer.ttl`.

Pinning a widget on a property shape already uses `shui:editor` and `shui:viewer`, so the scoring graph names the same thing with a different property.

## Proposal

A `shui:WidgetScore` identifies its widget with `shui:editor` or `shui:viewer`, at least one and at most one of each. The score function reads the widget through the widget predicate it receives as an input.

This makes the mode a plain lookup by predicate, lets one node cover both modes where the matcher and the score coincide, and uses the same properties as pinning:

```turtle
# before
shui:iriEditorScore20
    a shui:WidgetScore ;
    shui:widget shui:IRIEditor ;
    shui:score 20 ;
    shui:dataGraphShape shui:isIRI ;
.

shui:iriViewerScore20
    a shui:WidgetScore ;
    shui:widget shui:IRIViewer ;
    shui:score 20 ;
    shui:dataGraphShape shui:isIRI ;
.

# after
shui:iriScore20
    a shui:WidgetScore ;
    shui:editor shui:IRIEditor ;
    shui:viewer shui:IRIViewer ;
    shui:score 20 ;
    shui:dataGraphShape shui:isIRI ;
.
```

## Open to discuss

Should `shui:WidgetAcceptMatcher` follow the same pattern? It still uses `shui:widget`, which keeps one pair of properties out of the scoring graph, but also keeps a matcher that is often mode independent free of a mode distinction.

Or should the inverse property be used, attaching the accept matcher to the widget?

```turtle
shui:BooleanViewer shui:acceptMatcher shui:isBooleanMatcher .
```

That fits the idea that the accept matcher is part of the widget implementation, makes the lookup direction explicit, and removes the need for a matcher to enumerate the widgets it serves.

danielbeeke:
I think we should ditch shui:widget completely. This looks like a great cleanup to me.

PR #1249, opened 2026-09-10
[UI] Simpler version of the scoring system

### LINKED ISSUE #1225: What are the requirements for the scoring system algorithm?
This topic came up in our taskforce, its part of a broader conversation.

Here is a list of requirement that were taking into account when discussing #994

1. It MUST be possible to hard wire widgets with `shui:editor` and `shui:viewer`
2. It MUST be possible to create a custom widget and not touch any scoring graph triples. Using this custom widget can then only be done via `shui:editor` and `shui:viewer`. By adding a scoring shape, it can be picked automatically and the `shui:editor` and `shui:viewer` statement are not needed then.
3. Implementation of a custom widget MUST not require prior RDF knowledge outside of reading the value of the term and updating the term in case its an shui:Editor
4. Scoring system algorithmic functions MUST have simple singular return types. 
5. The scoring system MUST use generic programming language features available to most languages
6. The algorithm most have a reasonably optimal performance
7. Hard coded widgets should have a fallback mechanism in case the accept matcher does not accept them.
8. Closed world assumption must be possible. It can be achieved via the removal of all accept matchers and hard wire all widgets in the shapes.
8. Not finding a widget is reported to the interface.

bergos:
This PR contains a simpler version of the scoring system, based on the requirements described in #1225. It also provides a fix for #1248.

Closes #1225

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/ui/scoring/shacl12-ui/index.html)

### DIFF
--- a/shacl12-ui/index.html
+++ b/shacl12-ui/index.html
@@ -60,9 +60,9 @@
       }
 
       document.addEventListener("DOMContentLoaded", () => {
-        // Replace code div blocks in shapes, data, scoring, and results graph with tabs
+        // Replace code div blocks in shapes, data, and results graph with tabs
         for (const graph of document.querySelectorAll(
-          ".shapes-graph, .data-graph, .scoring-graph, .results-graph"
+          ".shapes-graph, .data-graph, .results-graph"
         )) {
           const tabs = document.createElement("aside");
           tabs.classList.add("ds-selector-tabs");
@@ -603,15 +603,6 @@
           </div>
         </div>
 
-        <div class="scoring-graph">
-          <div class="turtle">
-# This box represents an input scoring graph
-          </div>
-          <div class="jsonld">
-            <pre class="text">// This box represents an input scoring graph</pre>
-          </div>
-        </div>
-
         <div class="results-graph">
           <div class="turtle">
 # This box represents an output results graph
@@ -954,7 +945,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
         </aside>
 
         <p class="note">
-          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is
+          Note that `schema:honorificPrefix` has no `sh:minCount`, so it is 
           optional. The `rdfs:label` node expression should therefore handle
           missing values. For brevity, this example assumes that
           `schema:honorificPrefix` is always present.
@@ -1154,7 +1145,7 @@ ex:PersonShapeNameGroup a sh:PropertyGroup ;
 
         <p>
           We have only briefly touched on cardinality constraints. Other useful
-          but more complex topics include logical constraints and using
+          but more complex topics include logical constraints and using 
           <a href="../shacl12-node-expr/#dfn-dynamic-shacl">Dynamic SHACL</a>
           to populate the values for the enum and autocomplete editors.
           You can find these in the <a href="#patterns">Patterns</a> section.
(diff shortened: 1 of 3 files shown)


shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
-        </p>
-        <p>
-          A widget can also be attached to a shape directly, using `shui:editor` or `shui:viewer`. Such a widget
-          takes part in scoring like any other: the scoring graph is expected to contain a `shui:WidgetScore`
-          whose `shui:shapesGraphShape` matches shapes that declare the widget in this way.
-          <a href="#scoring-algorithm-scoring-graph-preparation">Scoring graph preparation</a> supplies such a
-          `shui:WidgetScore` for declared widgets that have none, so that a widget attached to a shape is never
-          silently ignored.
-        </p>
-      </section>
```
danielbeeke, line 1350: I think this text is still helpful

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
 
-      <section id="scoring-band-conventions">
-        <h3>Score Conventions</h3>
-        <p>
-          A `shui:score` is any `xsd:integer` or `xsd:decimal`. The <a href="#scoring-algorithm-score-function">score
-          function</a> simply orders widgets by descending score. The absolute value of a score has no meaning of its own.
```
danielbeeke, line 1356: tie break algo is still helpful to define.

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
-        </table>
         <p class="note">
-          The built-in widgets occasionally use an intermediate score, such as `5`, to break a tie between two widgets that
-          would otherwise share a band. Such values are fine-tuning between the bands above and are not themselves bands.
-        </p>
-        <p class="note">
-          To declare that a widget is <em>not</em> applicable under some condition, give the widget a
-          `shui:WidgetAcceptMatcher` whose shapes match the applicable case, rather than a `shui:WidgetScore` with a low or
-          negative score.
+          An implementation can use an alternative algorithm to the one described in this section, as long as it results in the same <a href="#scoring-algorithm-widget-selection">widget selection</a>.
```
danielbeeke, line 1302: Seems unrealistic to me. If we are to create a suite of checks these two examples will fail for more complex combinations of input:

shape and focus node where shape wants a date and the focus node is xsd:string.

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
 
-      <section id="scoring-algorithm-scoring-graph-preparation">
-        <h3>Scoring Graph Preparation</h3>
-        <p>
-          A widget can be attached to a shape directly, with `shui:editor` or `shui:viewer`, without the
-          <strong>scoring graph</strong> saying anything about that widget. Scoring graph preparation extends the
-          <strong>scoring graph</strong> with a `shui:WidgetScore` for each such widget, so that it participates in
-          the <a href="#scoring-algorithm-score-function">score function</a> like any other widget.
-        </p>
+      <section id="select-function">
```
danielbeeke, line 1307: What is the difference between the select and score function? Would it make sense to merge them and let an optional argument determine the mode (score / select)?

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
+
         <p>Output:</p>
         <ul>
-          <li>A <strong>prepared scoring graph</strong> — the <strong>scoring graph</strong> extended with a
-            `shui:WidgetScore` for each explicitly declared widget that has none.
+          <li>A list of records, ordered by descending <strong>Score</strong>, each with the following members:
+            <ul>
+              <li><strong>Widget IRI</strong> — The IRI of the widget.</li>
+              <li><strong>Widget Score IRI</strong> — Optional. The IRI of the <code>shui:WidgetScore</code> instance.</li>
+              <li><strong>Score</strong> — Optional. The <code>shui:score</code> of the <code>shui:WidgetScore</code> instance.</li>
```
danielbeeke, line 1333: So the output type of the function is:

```TypeScript
export type ScoreResults = { widgetIri: IRI, widgetScoreIri?: IRI, score?: number }[]
```

It solves the problem of having different outputs somewhat.

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
+          <li>If the <strong>Shape node</strong> has a <strong>Widget IRI</strong> defined via <strong>Widget predicate:</strong>
             <ol>
-              <li>Collect the set of IRI <a>values</a> <code>W</code> of the property <code>P</code> of the <a>shapes</a> in the
-                <strong>shapes graph</strong>. These are the widgets that are explicitly declared on a shape. A value of
-                <code>P</code> at a node that is not a shape does not declare a widget and is ignored. A value of
-                <code>P</code> that is not an IRI is also ignored: a widget is identified by an IRI, and a non-IRI value
-                such as a blank node cannot be matched reliably with `sh:hasValue` across graphs.
+              <li>Find the <code>shui:WidgetAcceptMatcher</code> instance <code>AM</code> with a matching <code>shui:widget</code> value.</li>
+              <li>If there is no <strong>Accept Matcher</strong> defined for the widget, return a list holding a single record whose
+                <strong>Widget IRI</strong> is the <strong>Widget IRI</strong>.</li>
```
danielbeeke, line 1344: A bit unclear language.

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
-                      </li>
-                    </ul>
-                  </li>
-                </ol>
+              <li>If the return value is <code>true</code>, return a list holding a single record whose
+                <strong>Widget IRI</strong> is the <strong>Widget IRI</strong>.</li>
+              <li>If the return value is <code>false</code>, continue with the next step. When a widget is not accepted it falls through to scoring.</li>
+            </ol>
+          </li>
+          <li>Call the <strong>score</strong> function with the following inputs and store the return value in an array <code>matches</code>:
```
danielbeeke: Would the word list instead of array be more consistent here?

shacl12-ui/index.html
```diff
@@ -1315,230 +1297,111 @@ <h3>Introduction</h3>
...
-        </p>
+      <section id="select-function">
+        <h3>Select Function</h3>
+
         <p>
-          A scoring graph MUST be prepared by this function before it is passed to the
-          <a href="#scoring-algorithm-score-function">score function</a>. Without preparation, a widget that is only
-          attached to a shape with `shui:editor` or `shui:viewer` is never returned.
+          The <strong>Select</strong> function determines which widgets can be used for a given combination of a focus node and a property shape,
+          ordered by preference.
```
danielbeeke, line 1312: Also "ordered by preference" is unclear to me. It should be ordered by score I think.

Further more:

I think we have the following use cases of the scoring system:
- retrieve list of all widgets applicable for the property (UIs widget widget switcher)
- Retrieve the first valid widget (UIs widget widget switcher)

I think the last option needs a lot less computation. The algorithms should make this possible. The current functions are entangled in such a way that the score function call, within the select function call, always receives all results, while for optimal performance it is not needed to call validate for each widget, just for the first one that is accepted.

shacl12-ui/index.html
```diff
@@ -1621,7 +1485,7 @@ <h3>Matcher Function</h3>
               <li><strong>Focus node</strong> as <strong>focus node</strong>.</li>
               <li><strong>Data graph</strong> as <strong>target graph</strong>.</li>
               <li>The value of the `shui:dataGraphShape` property of <strong>matcher node</strong> as <strong>shape node</strong>.</li>
-              <li><strong>Scoring graph</strong> as <strong>shapes graph</strong>.</li>
+              <li><strong>Shapes graph</strong> as <strong>shapes graph</strong>.</li>
```
danielbeeke: This change seems unnecessary to me and very abstract. It is the scoring graph that is needed here yes?

shacl12-ui/index.html
```diff
@@ -1632,40 +1496,36 @@ <h3>Matcher Function</h3>
       <section id="scoring-algorithm-score-function">
         <h3>Score Function</h3>
           <p>
-            The score function determines which widgets should be used for a given combination of a focus node and a
-            shape node. It is the entry point to the scoring system: it returns the widgets whose
-            <code>shui:WidgetScore</code> matches and that are accepted according to their
-            <code>shui:WidgetAcceptMatcher</code> instances, ordered from best to worst match.
+            The score function used to find an ordered list of matching widgets.
```
danielbeeke, line 1499: I think it is helpful to specify which functions are public and which ones are private to the scoring system.

shacl12-ui/index.html
```diff
@@ -1632,40 +1496,36 @@ <h3>Matcher Function</h3>
...
+          <li><strong>Scoring graph</strong> — The RDF graph containing the Widget Score definitions.</li>
+          <li><strong>Widget predicate</strong> — The predicate used to specify the widget, typically <code>shui:editor</code> or <code>shui:viewer</code>.</li>
         </ul>
         <p>Output:</p>
         <ul>
-          <li>An ordered sequence of <strong>widget results</strong>, sorted by descending <code>shui:score</code> so that
-            the best match comes first. The sequence is empty if no widget matches and is accepted.
+          <li>A list of records, ordered by descending <strong>Score</strong>, each with the following members:
+            <ul>
+              <li><strong>Widget IRI</strong> — The IRI of the widget.</li>
```
danielbeeke, line 1516: Here the type is stronger than the select function. This is part of what I previously disliked. Fine to keep it, if it solves the requirements.

danielbeeke (commented):
I tried to go through this PR, it was difficult for me. It does not seem to solve the requirements posted in https://github.com/w3c/data-shapes/issues/1225.

I am now trying to create my own implementation that fixes these problems. 
I hope to be able to bring this to the meeting tomorrow.

shacl12-ui/index.html
bergos: That's an implementation detail and doesn't belong in the spec.

shacl12-ui/index.html
bergos: That's part of the score function. Do you want it more explicit, or what makes you think it would help?

shacl12-ui/index.html
bergos: That's easy to implement and, for very large datasets, probably quite useful. Especially if you only want the first match. Create a map: use the most common constraints (data type and node kind) as the key, concatenating their values in a fixed, consistent order. For each key, store the widget, its score, and, if the key doesn't cover all constraints, a reduced widget score that skips the constraints already covered by the key. Then, if you find a match in the map and its score is higher than any score among widgets not in the lookup table, you can use the lookup result directly for an optimized search.

shacl12-ui/index.html
bergos: The score function only handles scoring based on the scoring graph. The select function is what the renderer calls. It combines the results of the score function, the accept matcher, and pinned widgets. Keeping them separate simplifies the logic.

shacl12-ui/index.html
bergos: Which part? Would attaching the accept matcher to the widget IRI solve the problem?

shacl12-ui/index.html
bergos:
```suggestion
          <li>Call the <strong>score</strong> function with the following inputs and store the return value in an list <code>matches</code>:
```

shacl12-ui/index.html
bergos: Maybe *ordered* doesn't fully cover it, but with *preference* I meant to capture that the score in the widget score isn't the only source. It could also just be the first item. The *Widget Selection* section covers that, but it's near the end, so it's easy to miss. I've considered making *Scoring System* a subsection of *Widget Selection*, and adding a second subsection just for the pinned-widgets definition. That would also make it easier to say 'SHACL UI consists of: Widget Selection, ...'. Would that solve the problem? I'd add that in a separate PR.

shacl12-ui/index.html
bergos:
```suggestion
              <li><strong>Scoring graph</strong> as <strong>shapes graph</strong>.</li>
```

shacl12-ui/index.html
danielbeeke: The point I am trying to make it is, that it is would not be helpful to tell people that can use *completely* different algorithms. How you have written it, it seems like it is very doable to have an algorithm that for each input has a correct output. You are not exactly writing *completely* but it can be read that way.

Further more, when community widgets take off, it can be the case there are multiple widgets that claim to be the canonical widget for a certain datatype. 

Maybe there is a community widget AwesomeDatePicker, it should have a higher data graph driven score such that it will be the default, so implementations must use the same core mechanism and not only copy the outputs for a limited amount of score shapes.

shacl12-ui/index.html
danielbeeke: This can indeed be resolved, I thought it was about something else.

shacl12-ui/index.html
bergos: I don't think the current text covers private/public, that's an implementation detail, and the *Widget Selection* section should mention what the entry point of the algorithm is.

danielbeeke:
I implemented a simplified version of the scoring system that has a lot of commonality.

https://github.com/ShapeThing/shapething/blob/main/packages/shacl-everything/src/scoring/score.ts

The pre processing of the shapes graph is gone.
I also have organized the functions a little different such that the algorithm is optimal in performance.

`score()` is only used to get a list
`select()` just returns an IRI

These two functions do not call each other. 
Ill try to make a branch that is inline with my code.

PR #1250, opened 2026-09-10
GH#1234: Allow `sh:name` and `sh:description` on node shapes

### LINKED ISSUE #1234: Make the use of 'sh:name' and 'sh:description' consistent between node and property shapes
*Description updated after feedback from Holger and Miel and Daniël on 2026-09-07.*

### Observation

For historic reasons, some of which are documented in #209 and its duplicate #590, the SHACL Core standard currently contains the following asymmetry:

* For property shapes, a human-readable label (`sh:name`) and description (`sh:description`) can be given for the corresponding property *in the context of the surrounding shape*.
* For node shapes, this support is not available, even though a similar need exists to give a human-readable label and description for the corresponding class *in the context of the surrounding shape*.

Version 1.2 of the SHACL Core standard contains the following text, which solidifies this asymmetry ([link](https://www.w3.org/TR/shacl12-core/#name)):

> Note that sh:name and sh:description should NOT be used for node shapes. For those, the properties rdfs:label and rdfs:comment are well-established and already used for class definitions.

Notice that this is not a good idea, because `rdfs:label` and `rdfs:comment` are already used to give a human-readable label and description to the node shape itself, rather than to its corresponding class. Notice that neither `rdfs:label` nor `rdfs:comment` are intended to name/describe the class *in the context of the surrounding shape*. In the current situation, the use of `rdfs:label` and `rdfs:comment` on a node shape is therefore ambiguous:

1. They may express the human-readable name and description of the node shape (e.g. "Employee shape"). This meaning is defined in the RDF Schema standard, where the meaning of `rdfs:label` is "A human-readable name for the subject." Since the subject is the node shape, `rdfs:label` is the name of the node shape, and is not necessarily identical to the name of its corresponding class.
2. They may express the human-readable name and description of the corresponding class (e.g. "Employee"). This meaning is stated in the SHACL Core standard, see the quote above.

Notice that this ambiguity occurs whenever the node shape and the class are distinct IRIs. This is a common use case (i.e. not a corner case), and includes all use cases in which an application profile is built on top of an existing ontology (e.g. DCAT-AP-NL and DCAT-AP on top of DCAT, NL-SBB on top of SKOS). In such cases, it is uncommon and unwanted to re-use the class IRI as the node shape IRI, because the class IRI is maintained by a different organization. (E.g. DCAT is maintained by W3C, DCAT-AP is maintained by SEMIC, DCAT-AP-NL is maintained by Geonovum.) Notice that this ambiguity does not occur when the node shape and class IRI are the same, e.g. when [implicit class targets](https://www.w3.org/TR/shacl12-core/#implicit-targetClass) are used. In these use cases, conflating the node shape and the class is on purpose, and naming/describing the node shape is the same thing as naming/describing the class.

### Suggestion

* Let's broaden the domain of `sh:name` and `sh:description` from `sh:PropertyShape` to `sh:Shape`.
* This removes the current asymmetry between property shapes and node shapes.
* This removes the current ambiguity of `rdfs:label` and `rdfs:comment`, when node shape and class are distinct IRIs.
* This does not affect current use cases in which the node shape and the class have the same IRI (implicit class targets), which can still be named/described with `rdfs:label` and `rdfs:comment`.
* This does not have any other negative consequences AFAICT.

PS: I sometimes see that `sh:name` and `sh:description` are applied to a node shape, even though the current standard considers this incorrect (e.g. in [NL-SBB](https://github.com/pldn/nederlands-profiel-voor-stelselcatalogi/blob/main/profiles/skos-ap-nl.ttl#L97)). (Notice that incorrect use of a standard does not imply that the standard should be changed. This example therefore does not constitute an argument for or against this proposal, and is only included as background information.)

danielbeeke:
- Broadens the domain of sh:name and sh:description to sh:Shape
- Clears up confusion between rdfs:label and sh:name
- Broadens prose about sh:name and sh:description to include node shapes
- Added basic example containing sh:name and rdfs:label
- Added example of sh:name in case of different targeting mechanism than sh:targetClass

Closes #1234

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/allow-sh-name-description-on-node-shapes/shacl12-core/index.html#name)

### DIFF
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5960,7 +5960,7 @@ ex:PersonShape
 				</aside>
 
 				<p class="note">
-					<code>sh:name</code> and other label properties (<code>rdfs:label</code>, <code>skosprefLabel</code>, etc.) are complementary.
+					<code>sh:name</code> and other label properties (<code>rdfs:label</code>, <code>skos:prefLabel</code>, etc.) are complementary.
 					Other label properties name the shape resource itself.
 					<code>sh:name</code> is not intended to name the shape resource, but rather
 					the shape's defining characteristic, as described above.
@@ -5987,10 +5987,6 @@ ex:SocialNetworkParticipantShape
 				</aside>
 
 				<p>
-					If present, tools should prefer such locally specified names
-					over globally specified labels at the <code>rdf:Property</code> or the <code>rdfs:Class</code> itself.
-					For example, if a form displays a node that is in the target of a given property shape
-					with an <code>sh:name</code>, then the tool should use the provided name.
 					The <a>values</a> of <code>sh:name</code> must be <a>literals</a> with <a>datatype</a>
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, or <code>rdf:langString</code>.
 				</p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5982,7 +5982,7 @@ ex:TargetSubjectsOfExampleShape
 				</p>
 				<p>
 					Similarly, shapes may have values for <code id="description">sh:description</code>
-					to provide descriptions of the property or node in the given context.
+					to provide descriptions of the property or target in the given context.
 					The <a>values</a> of <code>sh:description</code> must be <a>literals</a> with <a>datatype</a>
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, <code>rdf:langString</code>, or <code>rdf:HTML</code>.
 				</p>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5936,11 +5936,14 @@ ex:Concept2
 				<h3>sh:name and sh:description</h3>
 				<p>
 					Shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels:
+					to provide a human-readable name for the shape's defining characteristic,
+					i.e. the property path in the case of a property shape,
+					or the kind of resource targeted by a node shape:
 				</p>
 				<ul>
-					<li>For property shapes, it is the label of the property path in the target where it appears.</li>
-					<li>For node shapes, it is the label of the shape's target nodes.</li>
+					<li>For property shapes, <code>sh:name</code> names the property path.</li>
+					<li>For node shapes, <code>sh:name</code> names what the shape's targets have in common,
+						i.e. the kind of resource described by the shape's constraints.</li>
 				</ul>
 
 
@@ -5958,10 +5961,12 @@ ex:PersonShape
 
 				<p class="note">
 					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
-					<code>rdfs:label</code> specifies a label for the shape. <code>sh:name</code> does not.
+					<code>rdfs:label</code> names the shape resource itself.
+					<code>sh:name</code> is not intended to name the shape resource, but rather
+					the shape's defining characteristic, as described above.
 				</p>
 
-				<aside class="example" title="An example distinguishing labels for the shape from labels for its targets">
+				<aside class="example" title="An example distinguishing rdfs:label for the shape from sh:name for what it characterizes">
 					<div class="shapes-graph">
 						<div class="turtle">
 ex:SocialNetworkParticipantShape
@@ -5970,11 +5975,11 @@ ex:SocialNetworkParticipantShape
 	sh:targetSubjectsOf ex:knows ;
 	sh:targetObjectsOf ex:knows ;
 
-	# The following describe the shape itself
+	# The following name and describe the shape resource itself
 	rdfs:label "Social Network Participant Shape" ;
 	rdfs:comment "The constraints to apply to all participants in a social network." ;
 
-	# The following describe the shape's targets
+	# The following name and describe the kind of resource the shape's constraints characterize
 	sh:name "Social Network Participant" ;
 	sh:description "A participant in a social network." .
 						</div>
@@ -5982,7 +5987,7 @@ ex:SocialNetworkParticipantShape
 				</aside>
 
 				<p>
-					If present, tools should prefer such locally specified labels
+					If present, tools should prefer such locally specified names
 					over globally specified labels at the <code>rdf:Property</code> or the <code>rdfs:Class</code> itself.
 					For example, if a form displays a node that is in the target of a given property shape
 					with an <code>sh:name</code>, then the tool should use the provided name.
@@ -5991,11 +5996,11 @@ ex:SocialNetworkParticipantShape
 				</p>
 				<p>
 					Similarly, shapes may have values for <code id="description">sh:description</code>
-					to provide descriptions:
+					to provide a human-readable description of the same defining characteristic:
 				</p>
 				<ul>
-					<li>For property shapes, it is the description of the property path in the given context.</li>
-					<li>For node shapes, it is the description of the shape's target nodes.</li>
+					<li>For property shapes, a description of the property path in the given context.</li>
+					<li>For node shapes, a description of the kind of resource described by the shape's constraints.</li>
 				</ul>
 				<p>
 					The <a>values</a> of <code>sh:description</code> must be <a>literals</a> with <a>datatype</a>
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5935,29 +5935,60 @@ ex:Concept2
 			<section id="name">
 				<h3>sh:name and sh:description</h3>
 				<p>
-					Property shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels for the property in the target where it appears.
+					Shapes may have one or more <a>values</a> for <code>sh:name</code>
+					to provide human-readable labels.
+					In case of property shapes it is the label of the property in the target where it appears.
+					In case of node shapes, it is the label of the shape's target.
+				</p>
+
+
+				<aside class="example" title="A simple example of sh:name">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:PersonShape
+	a sh:NodeShape ;
+	sh:name "Person" ;
+	rdfs:label "Person shape v3.2" ;
+	sh:targetClass ex:Person .
+						</div>
+					</div>
+				</aside>
+
+				<aside class="example" title="An example of a subjects-of target with sh:name">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:TargetSubjectsOfExampleShape
+	a sh:NodeShape ;
+	sh:name "Subjects that know someone" ;
+	rdfs:label "Subjects that know someone shape v3.2" ;
+	sh:targetSubjectsOf ex:knows .
+						</div>
+					</div>
+				</aside>
+
+				<p class="note">
+					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
+					<code>rdfs:label</code> specifies a label for the shape. <code>sh:name</code> does not.
+				</p>
+
+
+				<p>
 					If present, tools should prefer such locally specified labels
-					over globally specified labels at the <code>rdf:Property</code> itself.
+					over globally specified labels at the <code>rdf:Property</code> or the <code>rdfs:Class</code> itself.
 					For example, if a form displays a node that is in the target of a given property shape
 					with an <code>sh:name</code>, then the tool should use the provided name.
 					The <a>values</a> of <code>sh:name</code> must be <a>literals</a> with <a>datatype</a>
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, or <code>rdf:langString</code>.
 				</p>
 				<p>
-					Similarly, property shapes may have values for <code id="description">sh:description</code>
-					to provide descriptions of the property in the given context.
+					Similarly, shapes may have values for <code id="description">sh:description</code>
+					to provide descriptions of the property or node in the given context.
 					The <a>values</a> of <code>sh:description</code> must be <a>literals</a> with <a>datatype</a>
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, <code>rdf:langString</code>, or <code>rdf:HTML</code>.
 				</p>
 				<p>
 					Both <code>sh:name</code> and <code>sh:description</code> may have
 					multiple <a>values</a>, but should only have one <a>value</a> per language tag.
-				</p>
-				<p>
-					Note that <code>sh:name</code> and <code>sh:description</code> should NOT be used for <a>node shapes</a>.
-					For those, the properties <code>rdfs:label</code> and <code>rdfs:comment</code> are well-established
-					and already used for class definitions.
 				</p>
 			</section>
 			<section id="intent">
--- a/shacl12-core/index.html
+++ b/shacl12-core/index.html
@@ -5936,10 +5936,12 @@ ex:Concept2
 				<h3>sh:name and sh:description</h3>
 				<p>
 					Shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels.
-					In case of property shapes it is the label of the property in the target where it appears.
-					In case of node shapes, it is the label of the shape's target.
+					to provide human-readable labels:
 				</p>
+				<ul>
+					<li>For property shapes, it is the label of the property path in the target where it appears.</li>
+					<li>For node shapes, it is the label of the shape's target nodes.</li>
+				</ul>
 
 
 				<aside class="example" title="A simple example of sh:name">
@@ -5954,23 +5956,30 @@ ex:PersonShape
 					</div>
 				</aside>
 
-				<aside class="example" title="An example of a subjects-of target with sh:name">
-					<div class="shapes-graph">
-						<div class="turtle">
-ex:TargetSubjectsOfExampleShape
-	a sh:NodeShape ;
-	sh:name "Subjects that know someone" ;
-	rdfs:label "Subjects that know someone shape v3.2" ;
-	sh:targetSubjectsOf ex:knows .
-						</div>
-					</div>
-				</aside>
-
 				<p class="note">
 					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
 					<code>rdfs:label</code> specifies a label for the shape. <code>sh:name</code> does not.
 				</p>
 
+				<aside class="example" title="An example distinguishing labels for the shape from labels for its targets">
+					<div class="shapes-graph">
+						<div class="turtle">
+ex:SocialNetworkParticipantShape
+	a sh:NodeShape ;
+	sh:targetClass ex:Person ;
+	sh:targetSubjectsOf ex:knows ;
+	sh:targetObjectsOf ex:knows ;
+
+	# The following describe the shape itself
+	rdfs:label "Social Network Participant Shape" ;
+	rdfs:comment "The constraints to apply to all participants in a social network." ;
+
+	# The following describe the shape's targets
+	sh:name "Social Network Participant" ;
+	sh:description "A participant in a social network." .
+						</div>
+					</div>
+				</aside>
 
 				<p>
 					If present, tools should prefer such locally specified labels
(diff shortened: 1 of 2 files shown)


shacl12-core/index.html
```diff
@@ -5935,30 +5935,61 @@ <h2>Non-Validating Shape Characteristics</h2>
 			<section id="name">
 				<h3>sh:name and sh:description</h3>
 				<p>
-					Property shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels for the property in the target where it appears.
+					Shapes may have one or more <a>values</a> for <code>sh:name</code>
+					to provide human-readable labels.
+					In case of property shapes it is the label of the property in the target where it appears.
```
robert-david: Shouldn't this be the label of the property path? This is how i understood the SHACL 1.0 definition.

danielbeeke: yes it is the label for the property in the target where it appears.
it is a complex way of saying a label for the property. This part did not change much.
It is now also allowed to put it on node shapes.

shacl12-core/index.html
```diff
@@ -5935,30 +5935,61 @@ <h2>Non-Validating Shape Characteristics</h2>
...
+				<p class="note">
+					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
+					<code>rdfs:label</code> specifies a label for the shape. <code>sh:name</code> does not.
+				</p>
+
+
+				<p>
 					If present, tools should prefer such locally specified labels
-					over globally specified labels at the <code>rdf:Property</code> itself.
+					over globally specified labels at the <code>rdf:Property</code> or the <code>rdfs:Class</code> itself.
```
mgberg, line 5991: I think it would probably be best to not suggest that `sh:name` overrides the label of the class, but rather is a name that could be applied to the set of targets of the shape.

As mentioned in more detail in #1234, If the intent is to actually render resources by class, then it likely makes the most sense to continue to use the label of the class. But if the intent is instead to render resources targeted by a node shape (which in some cases may just happen to be a single sh:targetClass), then it makes sense to look at the `sh:name` of the node shape.

Along those lines, it may be useful to use an example similar to the following to make it clear that `sh:name` is _not_ a label override for a class:

```turtle
ex:SocialNetworkParticipantShape
    a sh:NodeShape ;
    sh:targetClass :Person ;
    sh:targetSubjectsOf foaf:knows ;
    sh:targetObjectsOf foaf:knows ;

    # The following are about the shape itself
    rdfs:label "Social Network Participant Shape" ;
    rdfs:comment "The constraints to apply to all participants in a social network." ;

    # The following are about the set of targets of this shape
    sh:name "Social Network Participant" ;
    sh:description "A participant in a social network." ;
.
```

That's at least my opinion- I don't know if anyone else from the UI group (@edmondchuc, @bergos, @smessie) has an opinion on this.

danielbeeke: yes generally I agree, however it is also the cases that the problem space is relatively small.

I guess its a hierarchy of best matches, to less good matches but still mostly fitting and sometimes real misses.
The cases where rdfs:label can not be used as a fallback for a missing sh:name are small.
It boils down to targets that do not resemble a class anymore. 

I guess it would be good to discuss the other ways of targeting apart from sh:targetClass and see if the label matching algorithm is still good, or that we should narrow it to only shapes with targetClass or implicit target classes.

shacl12-vocabularies/shacl.ttl
```diff
@@ -1863,8 +1863,8 @@ sh:defaultValue
 sh:description
 	a rdf:Property ;
 	rdfs:label "description"@en ;
-	rdfs:comment "Human-readable descriptions for the property in the context of the surrounding shape."@en ;
-	rdfs:domain sh:PropertyShape ;
+	rdfs:comment "Human-readable descriptions for the property or the target in the context of the surrounding shape."@en ;
```
HolgerKnublauch:
```suggestion
	rdfs:comment "Human-readable descriptions for the property path (for property shapes) or the target nodes (for node shapes). Not to be used as description of the shape itself."@en ;
```

shacl12-vocabularies/shacl.ttl
```diff
@@ -1893,8 +1893,8 @@ sh:intent
 sh:name
 	a rdf:Property ;
 	rdfs:label "name"@en ;
-	rdfs:comment "Human-readable labels for the property in the context of the surrounding shape."@en ;
-	rdfs:domain sh:PropertyShape ;
+	rdfs:comment "Human-readable labels for the property or the target in the context of the surrounding shape."@en ;
```
HolgerKnublauch:
```suggestion
	rdfs:comment "Human-readable labels for the property path (for property shapes) or the target nodes (for node shapes). Not to be used as labels for the shape itself."@en ;
```

shacl12-core/index.html
```diff
@@ -5935,30 +5935,61 @@ <h2>Non-Validating Shape Characteristics</h2>
 			<section id="name">
 				<h3>sh:name and sh:description</h3>
 				<p>
-					Property shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels for the property in the target where it appears.
+					Shapes may have one or more <a>values</a> for <code>sh:name</code>
+					to provide human-readable labels.
+					In case of property shapes it is the label of the property in the target where it appears.
+					In case of node shapes, it is the label of the shape's target.
```
HolgerKnublauch:
```suggestion
					In case of node shapes, it is the label of the shape's target nodes.
```

shacl12-core/index.html
HolgerKnublauch: Best split this into a <li> list with two items for property shapes and node shapes. And yes, better say property path instead of just property.

shacl12-core/index.html
HolgerKnublauch: I think the example provided by @mgberg is very helpful here and should be included.

On overriding, yes I think it should be consistent: As sh:name overrides the rdfs:label of an rdf:Property, so should sh:name override the rdfs:label of a Class. It only really affects the forms, i.e. mainly SHACL UI.

shacl12-core/index.html
```diff
@@ -5935,30 +5935,61 @@ <h2>Non-Validating Shape Characteristics</h2>
...
 					For example, if a form displays a node that is in the target of a given property shape
 					with an <code>sh:name</code>, then the tool should use the provided name.
 					The <a>values</a> of <code>sh:name</code> must be <a>literals</a> with <a>datatype</a>
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, or <code>rdf:langString</code>.
 				</p>
 				<p>
-					Similarly, property shapes may have values for <code id="description">sh:description</code>
-					to provide descriptions of the property in the given context.
+					Similarly, shapes may have values for <code id="description">sh:description</code>
+					to provide descriptions of the property or target in the given context.
```
HolgerKnublauch: Again this should become a <li> list where it's clear that property path is for property shapes and target nodes for node shapes.

shacl12-core/index.html
danielbeeke: applied

shacl12-core/index.html
```diff
@@ -5935,30 +5935,76 @@ <h2>Non-Validating Shape Characteristics</h2>
 			<section id="name">
 				<h3>sh:name and sh:description</h3>
 				<p>
-					Property shapes may have one or more <a>values</a> for <code>sh:name</code>
-					to provide human-readable labels for the property in the target where it appears.
+					Shapes may have one or more <a>values</a> for <code>sh:name</code>
+					to provide human-readable labels:
+				</p>
+				<ul>
+					<li>For property shapes, it is the label of the property path in the target where it appears.</li>
```
robert-david: what does '.. where it appears.' mean? Is is 'from the focus nodes'?

shacl12-core/index.html
```diff
@@ -5935,30 +5935,76 @@ <h2>Non-Validating Shape Characteristics</h2>
...
+	sh:name "Person" ;
+	rdfs:label "Person shape v3.2" ;
+	sh:targetClass ex:Person .
+						</div>
+					</div>
+				</aside>
+
+				<p class="note">
+					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
+					<code>rdfs:label</code> specifies a label for the shape. <code>sh:name</code> does not.
```
robert-david: '... does not.' is an insufficient definition. Maybe '... is not indented to describe the shape, but ...'

shacl12-core/index.html
```diff
@@ -5935,30 +5935,76 @@ <h2>Non-Validating Shape Characteristics</h2>
...
 					<code>xsd:string</code>, <code>rdf:dirLangString</code>, or <code>rdf:langString</code>.
 				</p>
 				<p>
-					Similarly, property shapes may have values for <code id="description">sh:description</code>
-					to provide descriptions of the property in the given context.
+					Similarly, shapes may have values for <code id="description">sh:description</code>
+					to provide descriptions:
+				</p>
+				<ul>
+					<li>For property shapes, it is the description of the property path in the given context.</li>
```
robert-david: It is highly confusing to have a property that changes semantics depending on where it is used. Why does sh:name for a property shape not describe the targets, but something else?
I strongly recommend to streamline this in some way. If it does not define a clear purpose, users will just use it as a label.

shacl12-vocabularies/shacl.ttl
```diff
@@ -1863,8 +1863,8 @@ sh:defaultValue
 sh:description
 	a rdf:Property ;
 	rdfs:label "description"@en ;
-	rdfs:comment "Human-readable descriptions for the property in the context of the surrounding shape."@en ;
-	rdfs:domain sh:PropertyShape ;
+	rdfs:comment "Human-readable descriptions for the property path (for property shapes) or the target nodes (for node shapes). Not to be used as description of the shape itself."@en ;
```
robert-david: Same here for unclear purpose. Generally i can put anything in a description.

shacl12-vocabularies/shacl.ttl
```diff
@@ -1893,8 +1893,8 @@ sh:intent
 sh:name
 	a rdf:Property ;
 	rdfs:label "name"@en ;
-	rdfs:comment "Human-readable labels for the property in the context of the surrounding shape."@en ;
-	rdfs:domain sh:PropertyShape ;
+	rdfs:comment "Human-readable labels for the property path (for property shapes) or the target nodes (for node shapes). Not to be used as labels for the shape itself."@en ;
```
robert-david: We can even say that sh:name is a name for a shape. For property shapes, the path is the main defining element, so this would fit well. For node shapes, it describes (or summarises) the constraints, which is somehow describing what the target nodes need to have in common. 
Anyways, sh:name should have some common ground on node shapes and property shapes.

robert-david (changes requested):
I find the different semantics for node shapes and property shapes highly confusing. I expect users to just use it as a label if the purpose is not simple and clearly defined.

danielbeeke:
@robert-david I tried to process your feedback, does this make it clearer for you? Are things you would like to tweak further?

shacl12-core/index.html
```diff
@@ -5935,30 +5935,81 @@ <h2>Non-Validating Shape Characteristics</h2>
...
+						</div>
+					</div>
+				</aside>
+
+				<p class="note">
+					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
+					<code>rdfs:label</code> names the shape resource itself.
+					<code>sh:name</code> is not intended to name the shape resource, but rather
+					the shape's defining characteristic, as described above.
+				</p>
```
mgberg, line 5967: Might it be valuable to add an additional sentence here explicitly saying that `sh:description` follows a similar pattern?

shacl12-core/index.html
```diff
@@ -5935,30 +5935,81 @@ <h2>Non-Validating Shape Characteristics</h2>
...
+	sh:name "Person" ;
+	rdfs:label "Person shape v3.2" ;
+	sh:targetClass ex:Person .
+						</div>
+					</div>
+				</aside>
+
+				<p class="note">
+					<code>sh:name</code> and <code>rdfs:label</code> are complementary.
+					<code>rdfs:label</code> names the shape resource itself.
```
mgberg: Do we want to say _only_ `rdfs:label` here? Or would we want to say something along the lines of
```suggestion
					<code>sh:name</code> and other label properties (<code>rdfs:label</code>, <code>skosprefLabel</code>, etc.) are complementary.
					Other label properties name the shape resource itself.
```
given that there are users that use `skos:prefLabel` or `dc:title` or others beside `rdfs:label`, and that there is [support for specifying preferred label predicates in SHACL UI](https://w3c.github.io/data-shapes/shacl12-ui/#global-configuration)?

shacl12-core/index.html
mgberg: I'm still not sure we should say "sh:name overrides the rdfs:label of a Class". There is in general a many-to-many relationship between node shapes and classes (and beyond that, since shapes don't have to target classes).

Suppose we decide to also allow `:Organization`s to be social network participants:
```
ex:SocialNetworkParticipantShape
    a sh:NodeShape ;
    sh:targetClass :Person ;
    sh:targetClass :Organization ;
    sh:targetSubjectsOf :follows ;
    sh:targetObjectsOf :follows ;

    # The following are about the shape itself
    rdfs:label "Social Network Participant Shape" ;
    rdfs:comment "The constraints to apply to all participants in a social network." ;

    # The following are about the set of targets of this shape
    sh:name "Social Network Participant" ;
    sh:description "A participant in a social network." ;
.
```

If we say that "sh:name overrides the rdfs:label of a Class", then the labels for both `:Person` and `:Organization` classes would be "Social Network Participant", and I don't think that would be desired if you were looking at resources on a class-by-class basis.

Furthermore, suppose we then add the following shapes to the above:
```
ex:PersonShape
    a sh:NodeShape ;
    sh:targetClass :Person ;

    # The following are about the shape itself
    rdfs:label "Person Shape" ;
    rdfs:comment "The constraints to apply to all people." ;

    # The following are about the set of targets of this shape
    sh:name "Person" ;
    sh:description "A person." ;
.

ex:OrganizationShape
    a sh:NodeShape ;
    sh:targetClass :Organization ;

    # The following are about the shape itself
    rdfs:label "Organization Shape" ;
    rdfs:comment "The constraints to apply to all organizations." ;

    # The following are about the set of targets of this shape
    sh:name "Organization" ;
    sh:description "An organization." ;
.
```
Now, each class has two `sh:name`s to pick from. How would you know which `sh:name` would you use?

I believe both issues are sidestepped if `sh:name` on a node shape labels only the set of targets of that shape and is not treated as a class-level override. Anything targeted by `ex:SocialNetworkParticipantShape` is a "Social Network Participant" regardless of its class, and `:Person` keeps its own `rdfs:label`.

This is also somewhat consistent with how `sh:name` already works on property shapes, in the sense that it doesn't globally replace an `rdf:Property`'s name but only relabels it in a given context. That allows, for example, the property `dcterms:identifier` to be displayed with different user-friendly labels/descriptions per class. The node shape case is the same idea: the label applies in the context of a target set, not in the context of class instances.

danielbeeke: @mgberg thanks for the in depth comment. I think I might understand your perspective. Maybe it works best to check by scenarios:

### Showing a list of buttons to create new things, eg. "Create a new person"

If I understand you correct the following is problematic:

```SPARQL
select ?class ?label where {
	{
	# Class based targets
		?s a sh:NodeShape ;
			sh:targetClass ?class ;
			sh:name ?label ;
	} union {
	# Implicit class targets
		?class a sh:NodeShape ;
			a rdfs:Class ;
			sh:name ?label ;
	}
}
```
_(lets not deal with multilingual labels in these examples)_

This might give incorrect labels because:

- We do not know how many shapes directly target ex:Person
- When there are multiple shapes, there is no clear way to determine the canonical shape
- The shapes might target ex:Person but for example also ex:Product via another sh:targetClass. Multiple targeting ways may exist in 1 node shape.

Observations
- the more targeting ways inside a shape, a higher change it does not exactly represent a class
- the more constraints added, a higher chance a nodeShape does not exactly represent a class

If we could mark a shape as the canonical shape for a class, we would have less of this problem.
Maybe something like sh:CanonicalNodeShape?

One could make the argument that when there is only one shape in the shapes graph that targets a certain class via sh:targetClass and that is the only targeting in the shape, that it is maybe closer to representing the class fully.

Possible solution for this particular context:

```SPARQL
select ?class ?label where {
	{
		?s a sh:NodeShape ;
			sh:targetClass ?class ;
		?class rdfs:label ?label ;
	} union {
		?class a sh:NodeShape ;
			a rdfs:Class ;
		?class rdfs:label ?label ;
	}
}
```

Observation: sh:name can be used as long as start from the shape, but not when we start from classes and go through sh:targetClass to the shapes. Then there is more chance sh:name is incorrect.

### Show a list of target nodes of a shape

- The shape can be followed to get target nodes
- sh:name can be used to caption the list

No problems here as sh:name labels the target nodes.

### Showing "Edit Person John Doe"

- This is fine, a SHACL renderer has the interface:
	- data graph
	- shapes graph
	- focus node (ex:JohnDoe)
	- nodeShape (ex:PersonShape)
- As we have one nodeShape, we can just get its sh:name

### Summary:

It is incorrect to assume that a node shape with sh:targetClass {class} will have an sh:name that labels {class}, it can be a very narrow subset of {class} that it labels.

When starting from the shape sh:name always labels the target nodes correct.

### Observation and slight sidetrack:
sh:name also labels a property path, it might also be incorrect to assume it overrides the property label from the ontology, it labels the property contextually:

Example:
```turtle
sh:property [
	sh:name "Ingredient" ;
	sh:path [ 
		sh:alternativePath ( 
			schema:value schema:unitCode schema:name 
		) 
	] ;
] ;
```

In this particular example the property was about Ingredients:

schema:value 2
schema:unitCode  ex:Tablespoon
schema:name Mustard

And sh:alternativePath was used to mean more like sh:concatenatedPaths

In that case it can create the label: "2 Tablespoon Mustard".

Not sure if we should go that route. But basically a property path ends with properties but it is not said that the rdfs:label's of the property path leaves always represents whatever is captured.

[Example](https://storybook-shacl-renderer.shapething.com/?path=/story/showcases-recipes-and-chefs--recipes-and-chefs-view)

### Proposal:
- Remove any text that hints at override about sh:name and sh:description in SHACL core
- Write a guidance in SHACL UI, where we explain how to show a list of classes that can be created with the shapes from the shapes graph.

shacl12-core/index.html
danielbeeke: processed

shacl12-core/index.html
YoucTagh: Update the diagram to reflect changes.

danielbeeke: processed

YoucTagh: One alternative would be to move them to `sh:Shape` and then delete them from `sh:NodeShape` and `sh:PropertyShape`.

danielbeeke: thanks, applied

YoucTagh (approved):

HolgerKnublauch (approved):

shacl12-core/index.html
danielbeeke: @HolgerKnublauch what do you think? Can we merge this PR?

HolgerKnublauch: Yes but it's officially blocked. @robert-david request changes still valid?

robert-david (approved):

shacl12-core/index.html
danielbeeke: Okay it seems now to be fully accepted, do you want to merge it? Or am I allowed to do that?

HolgerKnublauch:
You can go aheadSent from my iPhoneOn 17 Sep 2026, at 18:36, Daniel Beeke ***@***.***> wrote:﻿
@danielbeeke commented on this pull request.



In shacl12-core/index.html:
>  					If present, tools should prefer such locally specified labels
-					over globally specified labels at the <code>rdf:Property</code> itself.
+					over globally specified labels at the <code>rdf:Property</code> or the <code>rdfs:Class</code> itself.

Okay it seems now to be fully accepted, do you want to merge it? Or am I allowed to do that?

PR #1251, opened 2026-09-11
GH#999: First attempt at `VALUES`-based pre-binding definition

### LINKED ISSUE #999: Update pre-binding to align with SPARQL 1.2

HolgerKnublauch:
Here is a (likely naive) attempt at defining pre-binding using VALUES. I am not a SPARQL expert on the grammar level nor have I implemented anything like this, so please report obvious flaws and don't hit me :)

Closes #999

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-999/shacl12-sparql/index.html#pre-binding)

### DIFF
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2047,7 +2047,7 @@ ex:Cougar
 						<code>V = "VALUES (?this ?value ?other) { (ex:FocusNode 42 UNDEF) }"</code>
 					</p>
 					<p>
-						Note that for the purpose of this definition we assume that the processor is capabable preserving the identity of <a>blank nodes</a>,
+						Note that for the purpose of this definition we assume that the processor is capable of preserving the identity of <a>blank nodes</a>,
 						even though the current <a data-cite="sparql12-query/#grammarBNodes">SPARQL 1.2 grammar</a> does not support this.
 					</p>
 					<p>
--- a/shacl12-sparql/index.html
+++ b/shacl12-sparql/index.html
@@ -2007,16 +2007,11 @@ ex:Cougar
 			<p>
 				Some features of SHACL-SPARQL rely on the concept of <a>pre-binding of variables</a> as defined in this section.
 			</p>
-			<p class="issue atrisk" data-number="999">
-				This feature is "at risk" and may be changed (not removed)
-				to align with currently ongoing work in the RDF and SPARQL 1.2 work in this area.
-				Originally discussed as <a href="https://github.com/w3c/data-shapes/issues/647">Issue 647</a>.
-			</p>
 			<div class="syntax" data-syntax-rule="pre-binding-limitations">
 				<p>
 					The definition of pre-binding used by SHACL requires the following restrictions on SPARQL queries.
-					SHACL-SPARQL processors MUST report a <a>failure</a> when it is operating on a <a>shapes graph</a>
-					that contains SHACL-SPARQL queries (via <code>sh:ask</code>, <code>sh:construct</code> and <code>sh:select</code>)
+					A SHACL-SPARQL processor MUST report a <a>failure</a> when it is operating on a <a>shapes graph</a>
+					that contains SPARQL queries (via properties such as <code>sh:ask</code>, <code>sh:construct</code> and <code>sh:select</code>)
 					that are executed with pre-bound variables and violate any of these "MUST" restrictions.
 					Note that the term <em>potentially pre-bound variables</em> includes the variables <code>this</code>,
 					<code>value</code> (for ASK queries),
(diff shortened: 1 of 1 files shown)


shacl12-sparql/index.html
```diff
@@ -2037,37 +2032,71 @@ <h2>Pre-binding of Variables in SPARQL Queries</h2>
...
+						<code>V = "VALUES (?this ?value ?other) { (ex:FocusNode 42 UNDEF) }"</code>
 					</p>
 					<p>
-						Define the <em>Values Insertion</em> function <code>Replace(X, μ)</code> to
-						replace each occurrence <code>Y</code> of a
-						<a data-cite="sparql12-query/#sparqlTranslateBasicGraphPatterns">Basic Graph Pattern</a>,
-						<a data-cite="sparql12-query/#sparqlTranslatePathExpressions">Property Path Expression</a>,
-						<a data-cite="sparql12-query/#sparqlTranslateGraphPatterns"><code>Graph(Var, pattern)</code></a>
-						in <code>X</code> with <code>join(Y, Table(μ))</code>.
+						Note that for the purpose of this definition we assume that the processor is capabable preserving the identity of <a>blank nodes</a>,
```
mgberg, line 2050:
```suggestion
						Note that for the purpose of this definition we assume that the processor is capable of preserving the identity of <a>blank nodes</a>,
```

afs:
As I said at Monday's meeting, I won't have time next week. I've removed myself from the reviewers.

One question:

This is different to the previous version, and also it is different to how several implementations currently do pre-binding.
Does it matter/will it change SHACL 1.0?

shacl12-sparql/index.html
```diff
@@ -2037,37 +2032,71 @@ <h2>Pre-binding of Variables in SPARQL Queries</h2>
...
-				<div class="def-header">DEFINITION: <dfn>Values Insertion</dfn></div>
+				<div class="def-header">DEFINITION: <dfn data-lt="pre-binding|pre-bind|pre-bound|pre-bound variables|pre-binds">Pre-binding of variables</dfn></div>
 				<div class="def-text-body">
 					<p>
-						For solution mapping <code>μ</code>, define <code>Table(μ)</code> to be the multiset formed from <code>μ</code>.
+						Let <code>P</code> be a map from variables to <a>nodes</a> <code>P = { ?var1: node1, ?var2: node2, ... }</code> (the <em>pre-bound variables</em>)
+						where any of the nodes may be unbound,
+						and <code>V</code> be a <a data-cite="sparql12-query#inline-data">VALUES clause</a> consisting of the same mappings from <code>P</code>
+						representing a <a data-cite="sparql12-query/#defn_sparqlSolutionSequence">sequence</a> with exactly one solution where each variable 
+						appears exactly once, using <code>UNDEF</code> for unbound variables.
```
mgberg, line 2042: I read through the SPARQL 1.2 spec and it says in [section 18.6.2](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval):

> We define eval(D(G), AQE, μctx) as the evaluation of an algebraic query expression AQE with respect to a dataset D having active graph G in correlation with solution mapping μctx.
>
> The active graph is initially the default graph of D and μctx is initially the empty solution mapping μ0.
>
>Note
> The case in which μctx may be different from μ0 is when evaluating an expression of the form EXISTS { pattern } or NOT EXISTS { pattern }, as defined in 17.4.1.4 NOT EXISTS and EXISTS.

Assuming I'm interpreting this correctly, I'm wondering if it would make more sense to define the pre-binding in terms of μctx not initially the empty mapping, but rather replacing it with P like is done for EXISTS.

Or perhaps it would be necessary to replace the rule for ContextSolution with `eval( D(G), ContextSolution, μctx ) = Join( { μctx }, { P } )` to ensure that the prebound variables are always present and don't get filtered out?

If either of these work, then it might not be necessary to say the bit about preserving the identity of blank nodes despite the VALUES grammar.

I definitely am not an expert on the algebra, let alone the algebra changes for 1.2, so someone should definitely check me on this.

HolgerKnublauch:
> As I said at Monday's meeting, I won't have time next week. I've removed myself from the reviewers.

I believe it will be difficult to make progress on this topic without your input. There are very few people in the world that have actually implemented any such algorithm.

> One question:
> 
> This is different to the previous version, and also it is different to how several implementations currently do pre-binding. Does it matter/will it change SHACL 1.0?

For this someone would need to demonstrate an example where the outcomes are different.

My question is actually why we are doing this in the first place. The original motivation of the issue was to update pre-binding to changes in SPARQL 1.2. But what changes in SPARQL have specifically happened that made the old definition unsuitable?

Asking because an alternative, low-risk solution may be just to leave the old definition in place.

shacl12-sparql/index.html
```diff
@@ -2007,16 +2007,11 @@ <h2>Pre-binding of Variables in SPARQL Queries</h2>
...
 			<div class="syntax" data-syntax-rule="pre-binding-limitations">
 				<p>
 					The definition of pre-binding used by SHACL requires the following restrictions on SPARQL queries.
-					SHACL-SPARQL processors MUST report a <a>failure</a> when it is operating on a <a>shapes graph</a>
-					that contains SHACL-SPARQL queries (via <code>sh:ask</code>, <code>sh:construct</code> and <code>sh:select</code>)
+					A SHACL-SPARQL processor MUST report a <a>failure</a> when it is operating on a <a>shapes graph</a>
+					that contains SPARQL queries (via properties such as <code>sh:ask</code>, <code>sh:construct</code> and <code>sh:select</code>)
 					that are executed with pre-bound variables and violate any of these "MUST" restrictions.
 					Note that the term <em>potentially pre-bound variables</em> includes the variables <code>this</code>,
 					<code>value</code> (for ASK queries),
```
```suggestion
					Note that the term <em>potentially pre-bound variables</em> includes the variables <code>this</code>
					and <code>value</code> (for ASK queries),
```

shacl12-sparql/index.html
```diff
@@ -2037,37 +2032,71 @@ <h2>Pre-binding of Variables in SPARQL Queries</h2>
...
-				<div class="def-header">DEFINITION: <dfn data-lt="pre-binding|pre-bind|pre-bound|pre-bound variables|pre-binds">Pre-binding of variables</dfn></div>
-				<div class="def-text-body">
 					<p>
-						The evaluation of the <a data-cite="sparql12-query/#idp2427544">SPARQL Query</a>
-						<code>Q = (E, DS, QF)</code> with <em>pre-bound</em> variables <code>μ</code>
-						is defined as the evaluation of SPARQL query <code>Q' = (Replace(E, μ), DS, QF)</code>.
+						For a SPARQL query <code>Q</code> and the pre-bound variables <code>P</code>,
+						the results of the query must be identical to a variation <code>Q'</code>
+						in which <code>V</code> is inserted into the beginning of each <code>{ ... }</code> block of the WHERE clause
+						where that is syntactically valid.
```
MPvHarmelen, line 2057: > into the **beginning** of each <code>{ ... }</code> block

I would define this to be at the **end**: otherwise something like this becomes problematic when pre-binding `?problematic`:

```sparql
select * {
  bind("something" as ?problematic)
}
```

becomes

```sparql
select * {
  values ?problematic { "something" }
  bind("something" as ?problematic)
}
```

which contains a syntax error because of

> The variable introduced by the BIND clause must not have been used in the group graph pattern up to the point of use in BIND.


Inserting at the end is fine, because `values` has no such restriction:

```sparql
select * {
  bind("something" as ?problematic)
  values ?problematic { "something" }
}
```

MPvHarmelen: > each <code>{ ... }</code> block of the WHERE clause

I have a hunch this could be interpreted ambiguously, what about (includes suggested edit for https://github.com/w3c/data-shapes/pull/1251/changes#r4013179900):

```suggestion
						in which <code>V</code> is inserted at the end of each <code>{ ... }</code> block that occurs anywhere inside the WHERE clause
						where that is syntactically valid, including at the end of the <code>{ ... }</code> of the WHERE clause itself.
```

shacl12-sparql/index.html
MPvHarmelen: My understanding of the discussion in https://github.com/w3c/data-shapes/issues/999 is that we want a definition for pre-binding that can be implemented client-side:

> A client-side parameterization design would make it SPARQL engine neutral, including work over the SPARQL Query protocol and work for SPARQL 1.1.

from: https://github.com/w3c/data-shapes/issues/999#issuecomment-5118327252

Additionally, a definition in terms of SPARQL syntax makes the definition a lot more accessible to users who never implemented a SPARQL engine: everyone that wants to make use of the pre-binding feature will know SPARQL syntax and will be able to understand a definition in terms of it. Only very few of those are familiar with the SPARQL algebra.

A definition in terms of SPARQL syntax also makes the implementation verifiable by hand. This is important especially when a user encounters unexpected behaviour: they can then verify their own understanding of the definition by applying the rewrite rules themselves.

HolgerKnublauch: How would this work with pre-binding of blank nodes? This remains a show stopper. And these use cases are real, e.g. to validate a sh:PropertyShape, rdf:List or owl:Restriction that is a blank node.

shacl12-sparql/index.html
HolgerKnublauch: The current draft still contains the limitation that overlapping VALUES and BIND with the same pre-bound variables are illegal anyway, so the cases you mention cannot really happen AFAIK.

I personally find it more intuitive to have the VALUES at the beginning of the block, because it clarifies to the average reader that the values are set before everything else is executed.

HolgerKnublauch: Applied in spirit, but without changing beginning to end.

shacl12-sparql/index.html
MPvHarmelen: Yes, the definition must definitely cover blank nodes too. I understood from https://github.com/w3c/data-shapes/issues/999#issuecomment-5172671627 that you had the following solution in mind:

> For blank nodes, the spec could say that implementations need to find their own native solutions that produce the same results, because SPARQL doesn't have a consistent syntax for bnodes. The spec only needs to describe what the outcome should be, it doesn't have to define implementation details.

shacl12-sparql/index.html
MPvHarmelen: Ah I suddenly realise there is an issue with inserting at the end:

Imagine pre-binding the variable `?preBound` in the following query: `select * { bind(bound(?preBound) as ?bound) }`.

Then the following is problematic:

```sparql
select * {
  bind(bound(?preBound) as ?bound)
  values ?preBound { "something" }
}
```

but inserting at the start works as intended:

```sparql
select * {
  values ?preBound { "something" }
  bind(bound(?preBound) as ?bound)
}
```

So indeed, let's do insertion at the start!

Issue #781, opened 2026-02-11
Evaluation of parameterized rules 


afs:
Using `$this` in #765 is an example of rule parameterization.

Is there a general mechanism here?

In SPARQL Query, [`VALUES`](https://www.w3.org/TR/sparql12-query/#inline-data) is used to let one or more variables to range over some given settings.

nicholascar:
Using SPARQL, we tend to write out parameterized queries from application code and just use `VALUES` and `BIND`. 

In the interests of an initial implementation, I think a just `VALUES` equivalent would be fine. that would let us easily slot variables into rules from application code. 

How would the RDF syntax work?

afs:
> Using SPARQL, we tend to write out parameterized queries from application code and just use VALUES and BIND.

First question is whether this is parameterization of a single rule or the rule set.

> How would the RDF syntax work?

If it's one rule, it would be possible to have it as the first element of the body.
If its the rule set, attach to the rule set node.

(Sketch)
```
        srl:values ( [ 
            [ srl:varName "x" ;  srl:value <uri> ]
            [ srl:varName "y" ;  srl:value 123 ]
        ]  ;
```
----

This does not need to be a specific feature.
This could use Rule Tuples #752.

* Evaluation takes an optional collection of tuples to seed the tuple store.
* SRL syntax adds a "tuples" block c.f. `DATA`.

Both an advantage and disadvantage is that it distinguishes between accessing parameterization variables and triple pattern variables.
Pro: stops accidentally running without parameters settings which may be very expensive (c.f. `WHERE { ?s ?p ?o . ?x ?y ?z }`).
Con: the rule is only for parameterized usage, not general usage as well. c.f. `WHERE { ?x rdfs:subClassOf $T `}

Another design possibility is that using `$` for a variable is understood (mandated) as a parameter and it's not just a convention.

afs:
Marking this as "Postponed". If there is more information, we can reconsider that.

afs:
Global setting for paramaterization (p13n) introduce a risk that parameters intended for one rule in the rule set accidentallly affecting other rules.

Just setting `?this` everywhere might affect a rule that used that variable name naturally.

`?this` might be `SET`-assigned, which breaks well-formedness.

These problems are compounded by rule set imports.

The idea sketched out here makes parameter usage "opt-in". A rule is has access to parameters if it requests them. Alternatives include rules fetching the parameters by matching.

### Pulling parameters - Explicit syntax 

```
[...]

PR #1257, opened 2026-09-15
#1256: SHACL SPARQL rule templates

### LINKED ISSUE #1256: Custom SHACL Rule Types
Similar to SPARQL-based constraint components and custom node expressions, the SHACL Inference Rules spec should support Custom Rule Types. They are basically SPARQL Rules that can be parameterized, allowing the same SPARQL query to be reused in multiple places, lifting the abstraction level of the syntax.

We have used such parameterized rules for many years, e.g. in SPIN Templates. Among others, we have used them for a higher-level mapping language with repeatable patterns. Produces nicely readable triples instead of geeky SPARQL queries.

The framework already has the extension point for rule types so this new feature will hook into that.

The PR will work out all the details.

HolgerKnublauch:
Closes #1256

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/issue-1256/shacl12-inference-rules/index.html#SPARQLRuleTemplate)


shacl12-inference-rules/index.html
```diff
@@ -421,6 +421,7 @@ <h2>Getting Started with SHACL Inference Rules</h2>
 				The <a>SHACL rules</a> feature defined in this document includes a general framework using the properties
 				such as <code>sh:rule</code> and <code>sh:condition</code>, plus an extension mechanism for specific <a>rule types</a>.
 				This document defines two such rule types: <a>SHACL SPARQL rules</a> and <a>triple rules</a>.
+				This document also defines a declarative mechanism for <a href="#SPARQLRuleType">custom rule types</a>.
```
mgberg: If we call `sh:TripleRule` and `sh:SPARQLRule` rule types, and rule types are an extension mechanism, then any other type of rule would be a "custom rule type". Perhaps there is a better term for this new capability than "custom rule types" for this that would be less confusing/wouldn't conflict. Maybe `sh:SPARQLRuleTemplate` instead?

liviorobaldo: I agree with @mgberg. I think these custom rules are different from the rule types described in "§3.1 Rule Types”.

Precisely to avoid confusion with the rule types in §3.1, perhaps we should change “§12. Custom Rule Types” to “§12. Parameterizable Rules”.

Even better... why don't we incorporate them into “§11.1 SHACL SPARQL Rules”? I understood that it makes sense to define these parameters only for SHACL SPARQL Rules. Neither Triple Rules nor SPARQL-RL Rules have parameters, at least not in their current form.

If we call them “custom rules” (or “parameterizable rules”, or whatever we decide to call them), we would be implying that all types of rules (Triple Rules, SPARQL-RL Rules, and any new rule types introduced in the future) must support this parametrization.

That said, I really like parameterizable rules... I have always liked customizations ☺️

mgberg: > Even better... why don't we incorporate them into “§11.1 SHACL SPARQL Rules”?

Good idea, unless there's an intent to allow other rule types to also be parameterizable (e.g. a `JavaScriptRule` and corresponding `JavaScriptRuleTemplate` or similar), in which case it might be best to still have a Parameterizable Rules section but just incorporate the specifics of the SPARQL-based ones into 11.1.

scotthenninger: I agree with Parameterizable Rules, but I’d suggest it should be a section 11.3. I think 11.1 needs an example or a pointer to examples to align it with the other subsections.

HolgerKnublauch: I have meanwhile replaced "custom" with "parameterizable rule types" which also avoids the conflict with "custom rule processors". So this is a step in the right direction.

I had also thought about whether this mechanism could be generalized to all rule types. This would unfortunately imply that we expect all future rule types to be parameterizable and I cannot speak for something like SPARQL-RL or JavaScript rules here. TripleRules already have an argument mechanism so they wouldn't be the problem. For SPARQL-RL there was even a ticket https://github.com/w3c/data-shapes/issues/781 but that was closed 3 weeks ago. I don't know why exactly. For JavaScript rules I wouldn't expect problems to pass in arguments, esp if they already have support for this for the "this" focus node.

However, if we broaden it to TripleRules, the metaclasses become more complex. We would need to introduce something like

```
sh:ParameterizableRuleType
	a rdfs:Class ;
	rdfs:label "Parameterizable rule type"@en ;
	rdfs:comment "The (meta) class for parameterizable rule types. Instances of this are classes whose own instances are the actual rule instances."@en ;
	rdfs:subClassOf rdfs:Class ;
	rdfs:subClassOf sh:Parameterizable ;
	rdfs:isDefinedBy sh: .

sh:SPARQLRuleType
	a rdfs:Class ;
	rdfs:label "SPARQL rule type"@en ;
	rdfs:comment "The class of SHACL rule types based on parameterized SPARQL CONSTRUCT queries."@en ;
	rdfs:subClassOf sh:ParameterizableRuleType ;
	rdfs:subClassOf sh:SPARQLConstructExecutable ;
	rdfs:isDefinedBy sh: .

sh:TripleRuleType
	a rdfs:Class ;
	rdfs:label "Triple rule type"@en ;
	rdfs:comment "The class of SHACL rule types based on parameterized subject/predicate/object triple templates."@en ;
	rdfs:subClassOf sh:ParameterizableRuleType ;
	rdfs:subClassOf sh:TripleExecutable ;
	rdfs:isDefinedBy sh: .

sh:TripleExecutable
	a rdfs:Class ;
	rdfs:label "Triple executable"@en ;
	rdfs:comment "The class of executables that produce triples based on a subject/predicate/object template."@en ;
	rdfs:subClassOf rdfs:Resource ;
	rdfs:isDefinedBy sh: .

sh:subject
	a rdf:Property ;
	rdfs:label "subject"@en ;
	rdfs:comment "An expression producing the nodes that shall be inferred as subjects. Defaults to the focus node only."@en ;
	rdfs:domain sh:TripleExecutable ;
	rdfs:isDefinedBy sh: .

sh:predicate
	a rdf:Property ;
	rdfs:label "predicate"@en ;
	rdfs:comment "An expression producing the nodes that shall be inferred as predicates. Defaults to the focus node only."@en ;
	rdfs:domain sh:TripleExecutable ;
	rdfs:isDefinedBy sh: .

sh:object
	a rdf:Property ;
	rdfs:label "object"@en ;
	rdfs:comment "An expression producing the nodes that shall be inferred as objects. Defaults to the focus node only."@en ;
	rdfs:domain sh:TripleExecutable ;
	rdfs:isDefinedBy sh: .
```

Note we would need an abstraction of each rule type: sh:SPARQLConstructExecutable already exists, but we would need to add sh:TripleExecutable too, and same for all other future rule types, to allow mix-in types and a proper metaclass hierarchy.

This is certainly doable in principle, yet we could also leave this for future work (1.3). TripleRules themselves haven't received much traction in the past and I am worried we are preemptively creating more work for very few users here, and complicate the metamodel along the way. Given that we would also risk making it harder for languages like SPARQL-RL that don't support parameterization yet to adopt this framework, I would caution against overgeneralizing this too early.

The good thing is that the current approach CAN be retrofitted without breaking anything if in the future we decide to introduce the TripleRuleType class. The metaclass sh:ParameterizableRuleType is never instantiated by end users.

If we keep it limited to sh:SPARQLRuleType, I suggest we make sure the terminology remains generic enough and clarify that SPARQLRuleType is the only parameterizable rule type in the current version?

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1487,179 @@ <h3>Triple Rules</h3>
 			</section>
 		</section>
 
+		<section id="SPARQLRuleType">
+			<h2>Parameterizable Rules</h2>
+			<p>
+				The class <code>sh:SPARQLRuleType</code> can be used to declare new <a>rule types</a> called <dfn>parameterizable rule types</dfn>.
```
```suggestion
				The class <code>sh:ParameterizableRuleType</code> can be used to declare new <a>rule types</a> called <dfn>parameterizable rule types</dfn>.
```

Consistently with the new title. We should align the terminology all over the section indeed.

HolgerKnublauch: I kept that intentionally, so that in the future we can generalize other rule types.

liviorobaldo: I'm not following. Above, you wrote:

``` 
sh:SPARQLRuleType
        ...
    rdfs:subClassOf sh:ParameterizableRuleType ;
```

which means that every `sh:SPARQLRuleType` is also a `sh:ParameterizableRuleType`.

Maybe I should have made it clearer that my answer to this question:

> If we keep it limited to sh:SPARQLRuleType, I suggest we make sure the terminology remains generic enough and clarify that SPARQLRuleType is the only parameterizable rule type in the current version?

is "yes". But to achieve this (i.e., to make sure the terminology remains generic enough), in this section we should talk about `sh:ParameterizableRuleType`, not `sh:SPARQLRuleType`. `sh:SPARQLRuleType` is only one of the (possibly many) subclasses of `sh:ParameterizableRuleType`.

HolgerKnublauch: The current PR does not include the triple

```
sh:SPARQLRuleType rdfs:subClassOf sh:ParameterizableRuleType ;
```

That was just hypothetical, in case we want to generalize this approach. If we go down that route we could indeed then also declare the parallel solution for TripleRules. I have no strong opinion, but right now it's limited to SPARQLRuleType directly, but the superclass could be added in the future.

liviorobaldo: I think readers (who are unaware of the generalization we have in mind) could easily get confused. I got confused, and I'm one of the editors 😅... although I don't know how representative I am of the `HomoSapiensSapiens` class 🤣

I think we should call them `sh:ParameterizableRuleType` and explicitly state that, for now, only `sh:SPARQLRuleType` and `sh:TripleRuleType` support parameters, by adding the `rdfs:subClassOf` statements above.

mgberg: > I think readers (who are unaware of the generalization we have in mind) could easily get confused. I got confused, and I'm one of the editors 😅... although I don't know how representative I am of the `HomoSapiensSapiens` class 🤣
> 
> I think we should call them `sh:ParameterizableRuleType` and explicitly state that, for now, only `sh:SPARQLRuleType` and `sh:TripleRuleType` support parameters, by adding the `rdfs:subClassOf` statements above.

I think that if we intend to only support parameterized SPARQL-based rules and we note that we could generalize in future work like stated [here](https://github.com/w3c/data-shapes/pull/1257#discussion_r4021521043), then I think I'm okay with the current state.

If there's some confusion that still needs to be resolved, then we should deal with that. If my above assumptions are true, then would renaming the section to "Parameterizable SPARQL Rules" resolve the ambiguity? That would allow defining a general "Parameterizable Rules" section in future work.

HolgerKnublauch: So I just tried to generalize the mechanism to also support sh:TripleRuleType. This is technically not a drama but when I tried to convert either of the two examples that I gave for SPARQL rule types, none of them could be easily expressed in Triple Rules. The problem is that node expressions don't easily support the path (shnex:pathValues) to be a variable, and this would require SPARQL yet again. While this may be a general limitation of node expressions (and thus triple rules) I don't think this is urgent enough to boil the ocean here.

To me this confirmed that limiting the approach to SPARQL rules for now is the right call. I did however just commit a change to introduce sh:ParameterizableRuleType as a superclass so that others can more easily add their own parameterizable rule types in the future. I also made the rdfs:subClassOf sh:Rule triple mandatory because this way anyone can already add such a mechanism and the contract stays consistent: all rules are instances of sh:Rule and must have a type with an IRI that identifies the rule type.

I did rename "parameterizable rule types" to  "parameterizable SPARQL rule types" to be clear that the specified mechanism in the 1.2 document is only about SPARQL rule types.

I hope this addresses the concerns for now. Any other problems or can this be approved?

shacl12-vocabularies/shacl.ttl
```diff
@@ -2053,6 +2053,22 @@ sh:object
...
+	rdfs:subClassOf sh:Parameterizable ;
+	rdfs:isDefinedBy sh: .
+
+sh:SPARQLRuleType
+	a rdfs:Class ;
+	rdfs:label "SPARQL rule type"@en ;
+	rdfs:comment "The class of SHACL rule types based on parameterized SPARQL CONSTRUCT queries."@en ;
+	rdfs:subClassOf sh:ParameterizableRuleType ;
+	rdfs:subClassOf sh:SPARQLConstructExecutable ;
+	rdfs:isDefinedBy sh: .
```
liviorobaldo, line 2063: I see that:

```
sh:SPARQLRule
	a rdfs:Class ;
	rdfs:label "SPARQL rule"@en ;
	rdfs:comment "The class of SHACL rules based on SPARQL CONSTRUCT queries."@en ;
	rdfs:subClassOf sh:Rule ;
	rdfs:subClassOf sh:SPARQLConstructExecutable ;
	rdfs:isDefinedBy sh: .
```

Shouldn't we therefore add:

```
sh:SPARQLRuleType
	rdfs:subClassOf sh:Rule ;
```

However, I actually think we should declare `sh:SPARQLRuleType` directly as a subclass of `sh:SPARQLRule`.

Even better... I don't understand really the name `sh:SPARQLRuleType`. To me, it sounds like "the type of `sh:SPARQLRule`", which is not the case. Here, it is a different class, intended to be used _in place of_ `sh:SPARQLRule`. I would therefore replace `sh:SPARQLRuleType` with `sh:SPARQLParametrizableRule`, and this is a subclass of both `sh:SPARQLRule` and `sh:ParameterizableRuleType`, so it would be defined as follows:

```
sh:SPARQLParametrizableRule
	a rdfs:Class ;
	rdfs:label "SPARQL parametrizable rule"@en ;
	rdfs:comment "The class of parametrizable SHACL rules based on SPARQL CONSTRUCT queries."@en ;
	rdfs:subClassOf sh:ParameterizableRuleType ;
	rdfs:subClassOf sh:SPARQLRule ;
	rdfs:isDefinedBy sh: .
```

We could then replace `sh:SPARQLRuleType` with `sh:SPARQLParametrizableRule` everywhere in the document.

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1486,178 @@ <h3>Triple Rules</h3>
 			</section>
 		</section>
 
+		<section id="SPARQLRuleType">
+			<h2>Parameterizable SPARQL Rule Types</h2>
+			<p>
+				The class <code>sh:SPARQLRuleType</code> can be used to declare new <a>rule types</a> called <dfn>parameterizable SPARQL rule types</dfn>.
```
```suggestion
			<p>
				The class <code>sh:SPARQLRuleType</code> can be used to declare new <a>rule types</a> called <dfn>parameterizable SPARQL rule types</dfn>.
			</p>
<p>
  <code>sh:SPARQLRuleType</code> is a subclass of <code>sh:ParameterizableRuleType</code>
  (see <a href="https://github.com/w3c/data-shapes/pull/1257/changes#diff-e7ad12ee6ccf9c41e835d5d12955c739ac6ead85384719f9a3cd0127f7a1c5fd"><code>shacl12-vocabularies/shacl.ttl</code></a>),
  the abstract type that includes all rules with parameters. Currently, only instances of
  <code>sh:SPARQLRuleType</code> are instances of <code>sh:ParameterizableRuleType</code>, but
  additional rule types may be declared as subclasses of <code>sh:ParameterizableRuleType</code>
  in the future.
</p>
			<p>
```

shacl12-inference-rules/index.html
liviorobaldo: > To me this confirmed that limiting the approach to SPARQL rules for now is the right call. I did however just commit a change to introduce sh:ParameterizableRuleType as a superclass so that others can more easily add their own parameterizable rule types in the future.

I totally agree. However, I think we should clarify everything you mention in your comment. I also think that `sh:SPARQLRuleType` should be a subclass of `sh:SPARQLRule` and should therefore be renamed to `sh:SPARQLParametrizableRule`. See my comments next to the relevant HTML and TTL code.

shacl12-vocabularies/shacl.ttl
HolgerKnublauch: Ok naming is really an issue here. I am also not happy with the current terminology. The problem is that we are crossing three meta-levels, and the rule type term is overloaded.

To distinguish all this, what about using the term "rule template" instead, as @mgberg suggested earlier, with "SPARQL rule template" as the only built-in. We would say that rule templates have parameters, and then it should be clear.

mgberg: > To distinguish all this, what about using the term "rule template" instead, as @mgberg suggested earlier, with "SPARQL rule template" as the only built-in. We would say that rule templates have parameters, and then it should be clear.

This makes sense to me. I wouldn't mind if `sh:ParameterizableRuleType` was left in and was changed to something like `sh:RuleTemplate` to be forward-looking, but if you wanted to omit that for now as well then I'm fine with it.

liviorobaldo: Ok, makes sense

HolgerKnublauch: Renaming applied. I hope this is clearer. To simplify it further, I also removed the sh:ParameterizableRuleType for now. It isn't necessary to create new template types - people can copy and paste what sh:SPARQLRuleTemplate does.

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1486,183 @@ <h3>Triple Rules</h3>
...
+					Using a <a>pre-binding</a> map for each declared <code>sh:parameter</code> of <code>T</code> at <code>R</code>
+					similar to <a data-cite="shacl12-sparql#constraint-components-parameters">SPARQL-based constraint components</a>,
+					evaluate <code>Q</code> like a corresponding <a href="#SPARQLRule">SPARQL Rule</a> but using the extra pre-bound variables.
+				</div>
+			</div>
+			<p>
+				In the example below, the CONSTRUCT query is executed with the SPARQL variable <code>$property</code> pre-bound to the
+				value of <code>ex:property</code>.
+			</p>
+			<aside class="example" title="An example SPARQL rule type to declare symmetric properties" id="SPARQLRuleTemplate-symmetric-example">
```
mgberg:
```suggestion
			<aside class="example" title="An example SPARQL rule template to declare symmetric properties" id="SPARQLRuleTemplate-symmetric-example">
```

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1486,183 @@ <h3>Triple Rules</h3>
...
+				<p>
+					In this example, a SHACL rules engine will infer missing triples in the inverse direction for every value
+					of the given properties `ex:marriedTo` and `ex:overlaps`, using a SPARQL rule type.
```
mgberg:
```suggestion
					of the given properties `ex:marriedTo` and `ex:overlaps`, using a SPARQL rule template.
```

mgberg (approved):
I found a couple lingering usages of "SPARQL rule type" that got missed, but other than that looks good to me

HolgerKnublauch:
Thanks, Matt. I will merge this in tomorrow unless I hear other concerns.

shacl12-vocabularies/shacl.ttl
liviorobaldo, line 2026: The new version is improved, but I still see some points of confusion... apologies, maybe it's just me 😞

I see that `sh:SPARQLRule` and `sh:SPARQLRuleTemplate` are completely unrelated, yet their names suggest that they are related.

Maybe I misunderstood what @mgberg was suggesting. I see that we now have:

```
ex:SymmetricPropertyRule
    a sh:SPARQLRuleTemplate ;
    rdfs:subClassOf sh:Rule ;
```

But I understood that we wanted to declare all `sh:SPARQLRules` to have this template, i.e., to declare:

```
sh:SPARQLRule
    a sh:SPARQLRuleTemplate ;
```

(Note that this makes the connection between `sh:SPARQLRule` and `sh:SPARQLRuleTemplate`) Then we would have:

```
ex:SymmetricPropertyRule
    rdfs:subClassOf sh:SPARQLRule ;
```

Yes, this solution means that ALL instances of `sh:SPARQLRule` can have parameters. This is what I understood @mgberg to be suggesting.

If, instead, we want to allow _both_ SPARQL rules with parameters _and_ SPARQL rules without parameters, I would suggest making the following declarations in the vocabulary:

- `sh:SPARQLRule` remains as it is now.
- `sh:SPARQLRuleTemplate` is renamed to `sh:SPARQLParameterizable` and defined as follows (basically, as it is now, apart from the renaming `sh:SPARQLRuleTemplate` → `sh:SPARQLParameterizable`):

```
sh:SPARQLParameterizable
    a rdfs:Class ;
    rdfs:label "..."@en ;
    rdfs:comment "..."@en ;
    rdfs:subClassOf rdfs:Class ;
    rdfs:subClassOf sh:Parameterizable ;
    rdfs:subClassOf sh:SPARQLConstructExecutable ;
    rdfs:isDefinedBy sh: .
```

Then, if we want `ex:SymmetricPropertyRule` to be a simple SPARQL rule _without_ parameters, we write:

```
ex:SymmetricPropertyRule
    rdfs:subClassOf sh:SPARQLRule ;
```

If, instead, we want a SPARQL rule _with_ parameters, we write:

```
ex:SymmetricPropertyRule
    a sh:SPARQLParameterizable ;
    rdfs:subClassOf sh:SPARQLRule ;
```

`sh:SPARQLParameterizable` would therefore be a "label" indicating that a rule can accept parameters. In the future, we could symmetrically apply the same schema to other types of rules, for example:

```
sh:TripleParameterizable
    rdfs:subClassOf sh:Parameterizable ;
    # etc. Here we define how parameters work for Triple rules.
```

And then:

```
ex:SymmetricPropertyRule
    rdfs:subClassOf sh:TripleRule ;
```

rather than:

```
ex:SymmetricPropertyRule
    a sh:TripleParameterizable ;
    rdfs:subClassOf sh:TripleRule ;
```

The same could be done for the other types of rules.

Anyway, perhaps the confusion about the current formalization is only MY problem 😅 Feel free to ignore the above if so.

On the other hand, I just noticed the following in the vocabulary:

```
sh:SPARQLRuleTemplate
    a rdfs:Class ;
    rdfs:subClassOf rdfs:Class ;
```

So `sh:SPARQLRuleTemplate` is both an `rdfs:Class` and a subclass of `rdfs:Class`. Contrary to the point above, which could simply be my misunderstanding, I am rather sure that this is an error 😅

HolgerKnublauch: That last paragraph is exactly the point. sh:SPARQLRuleTemplate is a metaclass (i.e. both subClassOf and rdf:type rdfs:Class). This is intentional, because its instances are classes. And those classes become the rdf:type of the actual rule instances. I believe this makes sense and is correct.

As we have three metalayers here, the direct mapping to sh:SPARQLRule does not work as the sh:SPARQLRules are in reality at the same level as the template instances, e.g. they can be a value of sh:rule.

liviorobaldo: > That last paragraph is exactly the point. sh:SPARQLRuleTemplate is a metaclass (i.e. both subClassOf and rdf:type rdfs:Class). This is intentional, because its instances are classes.

Mhh... I don't think this is correct. Or perhaps it should apply to any subset of `rdfs:Class`: if X is a subset of the set of all sets, then X is also a set, and therefore X is also an instance of `rdfs:Class`.

I went to read the definitions in RDFS, and here:

https://www.w3.org/TR/rdf-schema/#ch_class

I read "`rdfs:Class` is an instance of `rdfs:Class`." Isn't this a property inherited by all subsets of `rdfs:Class`? My brain is starting to melt 😅

> As we have three metalayers here, the direct mapping to sh:SPARQLRule does not work as the sh:SPARQLRules are in reality at the same level as the template instances, e.g. they can be a value of sh:rule.

OK, _maybe_ I understand. `sh:SPARQLRule` is currently defined as:

```
sh:SPARQLRule
	a rdfs:Class ;
	rdfs:subClassOf sh:Rule ;
	rdfs:subClassOf sh:SPARQLConstructExecutable ;
```

while `sh:SPARQLRuleTemplate` as:

```
sh:SPARQLRuleTemplate
	a rdfs:Class ;
	rdfs:subClassOf rdfs:Class ;
	rdfs:subClassOf sh:Parameterizable ;
	rdfs:subClassOf sh:SPARQLConstructExecutable ;
``` 

So, since `rdfs:subClassOf sh:SPARQLConstructExecutable` occurs in both, declaring:

```
ex:SymmetricPropertyRule
	a sh:SPARQLRuleTemplate ;
	rdfs:subClassOf sh:Rule ;
```

is indeed different from:

```
ex:SymmetricPropertyRule
	rdfs:subClassOf sh:SPARQLRule ;
```

Are you therefore saying that `ex:SymmetricPropertyRule` is a _set_ of rules (rather than a _single_ rule): namely, the set of all rules obtained by varying the parameters in all possible ways? If so, I think this needs to be written explicit.

In addition, I still prefer the second:

```
ex:SymmetricPropertyRule
	rdfs:subClassOf sh:SPARQLRule ;
```

because it seems to me that `ex:SymmetricPropertyRule` is a _set_ of _SPARQL_ rules here. To explicitly encode that it has parameters, we could type it as `sh:Parameterizable`:

```
ex:SymmetricPropertyRule
    a sh:Parameterizable ;
    rdfs:subClassOf sh:SPARQLRule ;
```

because, in the `sh:SPARQLRuleTemplate` definition you wrote above, this is the only part we are interested in.

Now... to in principle allow parameters also on other types of rules, what I was proposing is to create a type for SPARQL parameters, a type for Triple rules parameters, etc., in this way:

```
sh:SPARQLParameterizable rdfs:subClassOf sh:Parameterizable.
sh:TripleParameterizable rdfs:subClassOf sh:Parameterizable.
#Etc.
```

by introducing these sub-types, this:

```
ex:SymmetricPropertyRule
    a sh:Parameterizable ;
    rdfs:subClassOf sh:SPARQLRule ;
```

should be specialized as:

```
ex:SymmetricPropertyRule
    a sh:SPARQLParameterizable ;
    rdfs:subClassOf sh:SPARQLRule ;
```

HolgerKnublauch: > Are you therefore saying that `ex:SymmetricPropertyRule` is a _set_ of rules (rather than a _single_ rule): namely, the set of all rules obtained by varying the parameters in all possible ways? If so, I think this needs to be written explicit.

Yes, ex:SymmetricPropertyRule is a set of rule instances. If you have specific suggestions, please feel free to make an edit request to the PR.

> In addition, I still prefer the second:
> 
> ```
> ex:SymmetricPropertyRule
> 	rdfs:subClassOf sh:SPARQLRule ;
> ```

This would not be correct. Keep in mind that the instances of sh:SPARQLRule must declare a sh:construct. But the sh:construct is already defined at the ex:SymmetricPropertyRule, and not at its instances.

Metaclasses are complicating...

> Now... to in principle allow parameters also on other types of rules, what I was proposing is to create a type for SPARQL parameters, a type for Triple rules parameters, etc., in this way...

I wouldn't worry about that for now. The topic is already complex enough :)

mgberg: > Yes, this solution means that ALL instances of sh:SPARQLRule can have parameters. This is what I understood @mgberg to be suggesting.

I wasn't suggesting that every rule could be parameterizable. I was suggesting that perhaps it would be useful to have other types of rules have parameterizable versions- so just as you can have `sh:SPARQLRule` and `sh:SPARQLRuleTemplate`, other rule types may wish to expose parameterizable versions.

I don't think it makes sense to make every rule parameterizable. Consider this example from the document:
```turtle
ex:RectangleRulesShape
	a sh:NodeShape ;
	sh:targetClass ex:Rectangle ;
	sh:rule [
		a sh:SPARQLRule ;
		sh:construct """
			CONSTRUCT {
				$this ex:area ?area .
			}
			WHERE {
				$this ex:width ?width .
				$this ex:height ?height .
				BIND (?width * ?height AS ?area) .
			}
			""" ;
		# Rule only applies to Rectangles that conform to ex:RectangleShape
		sh:condition ex:RectangleShape ;
	] .
```
That rule is associated with `ex:RectangleRulesShape`. I can't think of a situation where parameterizing a rule like that would make sense.

I think the current state is okay and there's already other examples following this same pattern, such as SPARQL-based target types:

```turtle
sh:Target             a rdfs:Class ; rdfs:subClassOf rdfs:Resource .
sh:SPARQLTarget       a rdfs:Class ; rdfs:subClassOf sh:Target ;
                      rdfs:subClassOf sh:SPARQLAskExecutable, sh:SPARQLSelectExecutable .

sh:TargetType         a rdfs:Class ; rdfs:subClassOf rdfs:Class ;   # <- metaclass
                      rdfs:subClassOf sh:Parameterizable .
sh:SPARQLTargetType   a rdfs:Class ; rdfs:subClassOf sh:TargetType ;
                      rdfs:subClassOf sh:SPARQLAskExecutable, sh:SPARQLSelectExecutable .
```
> Are you therefore saying that `ex:SymmetricPropertyRule` is a _set_ of rules (rather than a _single_ rule): namely, the set of all rules obtained by varying the parameters in all possible ways? If so, I think this needs to be written explicit.

Yes, in the same way that `sh:SPARQLRule` represents a set of rules. `ex:MarriedToRule` and `ex:OverlapsRule` are the instances of `ex:SymmetricPropertyRule` (and since `ex:SymmetricPropertyRule` is subclass of `sh:Rule` they are also instances of `sh:Rule`). [The green note box under the first example](https://raw.githack.com/w3c/data-shapes/issue-1256/shacl12-inference-rules/index.html#SPARQLRuleTemplate-symmetric-example) already states this.

> In addition, I still prefer the second: `ex:SymmetricPropertyRule rdfs:subClassOf sh:SPARQLRule ;` because it seems to me that `ex:SymmetricPropertyRule` is a _set_ of _SPARQL_ rules here.

Note that `rdfs:subClassOf sh:SPARQLRule` would make each instance a `sh:SPARQLRule`, and every `sh:SPARQLRule` must have exactly one value for `sh:construct`. `ex:MarriedToRule` has none, as the `sh:construct` lives on the template instead.

Note that `sh:SPARQLTargetType` above is deliberately *not* `rdfs:subClassOf sh:SPARQLTarget` for the same metalevel reason; it re-declares the executable mix-ins instead.

mgberg: @HolgerKnublauch you beat me to it 😆

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1486,183 @@ <h3>Triple Rules</h3>
 			</section>
 		</section>
 
+		<section id="SPARQLRuleTemplate">
+			<h2>SPARQL Rule Templates</h2>
+			<p>
+				The class <code>sh:SPARQLRuleTemplate</code> can be used to declare new <a>rule types</a> called <dfn>SPARQL rule templates</dfn>.
```
liviorobaldo, line 1492:
```suggestion
The class <code>sh:SPARQLRuleTemplate</code> can be used to declare new <a>rule types</a> called <dfn>SPARQL rule templates</dfn>. Being rule <i>types</i>, their instances are <i>sets</i> of rules, rather than individual rules (see example below).
```

shacl12-inference-rules/index.html
```diff
@@ -1484,6 +1486,183 @@ <h3>Triple Rules</h3>
...
+				</div>
+			</div>
+			<p>
+				In the example below, the CONSTRUCT query is executed with the SPARQL variable <code>$property</code> pre-bound to the
+				value of <code>ex:property</code>.
+			</p>
+			<aside class="example" title="An example SPARQL rule template to declare symmetric properties" id="SPARQLRuleTemplate-symmetric-example">
+				<p>
+					In this example, a SHACL rules engine will infer missing triples in the inverse direction for every value
+					of the given properties `ex:marriedTo` and `ex:overlaps`, using a SPARQL rule template.
```
```suggestion
In this example, a SHACL rules engine will infer missing triples in the inverse direction for every value of the given properties `ex:marriedTo` and `ex:overlaps`, using a SPARQL rule template. Note that `ex:SymmetricPropertyRule` is a new <i>rule type</i>, i.e., it denotes a <i>set</i> of rules; `ex:MarriedToRule` and `ex:OverlapsRule` are two instances of this set.
```

shacl12-vocabularies/shacl.ttl
liviorobaldo: Ok, THANKS A LOT TO BOTH OF YOU for your explanations (and for the necessary patience in writing them 😅). I think I understand it now, or at least most of it. The key was understanding that `ex:SymmetricPropertyRule` denotes a _set_ of rules. Perhaps other future readers of this document who are less familiar with metaclasses could have the same difficulty, so I have added two sentences to (hopefully) clarify the matter a bit further.

I'm still a bit unhappy with the name `sh:SPARQLRuleTemplate`. I now understand the use of the word "template", but I think the name should also convey the fact that the template enables parameters. Therefore, I think `sh:SPARQLParametrizableRuleTemplate` (subclass of `sh:Parameterizable`) would be better. However, we really don't need to discuss the choice of name much further; it's much more important to move forward, as the deadlines are approaching.

Thanks again! ❤️

HolgerKnublauch: There is nothing wrong with sh:SPARQLParametrizableRuleTemplate except that it is a bit too long and easy to get wrong (e.g. sh:SPARQLParametrizableRuleTemplate vs sh:SPARQLParametrizedRuleTemplate).

So I'll leave it simple for now and will merge the branch in. Thanks for the discussion, as it's always good to capture potential confusion points before our readers do.

Issue #627, opened 2025-10-28
Request for motivation of unsupported-sparql-004


MPvHarmelen:
The test [unsupported-sparql-004](https://github.com/w3c/data-shapes/blob/gh-pages/data-shapes-test-suite/tests/sparql/pre-binding/unsupported-sparql-004.ttl) is expected to fail, but having thoroughly read and implemented the [SHACL spec](https://www.w3.org/TR/shacl), I don't understand why this test should fail. Why is it expected to fail?

HolgerKnublauch:
Hi Martin, see https://www.w3.org/TR/shacl/#pre-binding all inner queries (sub-SELECTs) must return pre-bound variables - here it needs to project $this

MPvHarmelen:
Ah thank you, that clarifies a misinterpretation that I had of the spec.

MPvHarmelen:
@HolgerKnublauch, [sparql/pre-binding/pre-binding-006](https://github.com/w3c/data-shapes/blob/gh-pages/data-shapes-test-suite/tests/sparql/pre-binding/pre-binding-006.ttl) similarly confuses me. Would you mind clarifying why this test also expects a failure?

HolgerKnublauch:
I don't quite remember TBH. We should have documented that better :)

BTW for SHACL 1.2 the restriction on sub-selects has been removed, see https://github.com/w3c/data-shapes/issues/159#issuecomment-2958082918

Maybe the test 06 was also a result of some particular behavior of the SPARQL engine that I used back then, and with the removal of the condition above the test can likely also be deleted. Let me reopen the ticket to remove these tests.

afs:
It could be for the same reason - `SELECT *` does not necessarily include `$this` because `$this` is not in a graph pattern. If `*` is calculated at parse time, `$this` isn't there; if it is replaced by a no-op (optimization), it would be.

MPvHarmelen:
> SELECT * does not necessarily include $this because $this is not in a graph pattern

Thank you, that indeed clarifies the test!

> BTW for SHACL 1.2 the restriction on sub-selects has been removed, see https://github.com/w3c/data-shapes/issues/159#issuecomment-2958082918

That's great! Can we drop the other restrictions on queries used for SPARQL-based things in SHACL 1.2 too?

The restrictions greatly increase the learning burden of users and thereby decrease the usability.

HolgerKnublauch:
I don't think we can drop the other restrictions as they were there for a reason.

[...]

PR #1261, opened 2026-09-15
Remove some pre-binding restrictions

MPvHarmelen:
When defining pre-binding in terms of `values` insertion (at the end of `{}`-blocks), these two restrictions are no longer necessary: AFAIK all insertions will lead to legal queries. Even if insertion would lead to an illegal query, the "except when this leads to a syntax error" fallback covers those.


HolgerKnublauch:
We cannot merge this into main without the VALUES commit first. So I suggest you direct this PR against the other branch that I have started?

HolgerKnublauch:
Ah and also, it will cause problems because you are outside of the WG. So I would need to apply this change myself. Will do tomorrow.

HolgerKnublauch:
Thinking more about this, I am unconvinced this is the right step to make.

The change to define pre-binding through VALUES is just a change to the formal definition. It does NOT represent implementation advice. For example, blank nodes would not work like this anyway. Given that implementations may not rely on VALUES as a syntax (but continue to change the internal runtime data structure such as algebra), I don't think the enforcement of these syntax limitations can be taken for granted and instead we should continue to spell them out.

MPvHarmelen:
> The change to define pre-binding through VALUES is just a change to the formal definition. It does NOT represent implementation advice. For example, blank nodes would not work like this anyway. Given that implementations may not rely on VALUES as a syntax (but continue to change the internal runtime data structure such as algebra), I don't think the enforcement of these syntax limitations can be taken for granted and instead we should continue to spell them out.





Some more context on my reasoning:


1. engines are free to do whatever in their implementations, as long as all observable behaviour is spec-compliant
2. undefined situations are bad in a spec, because they lead to differing observable behaviour across implementations
2. these constructs were excluded from pre-binding because they would lead to undefined situations in the 1.1 definition of pre-binding
3. in the context of end-of-{} values insertion, these constructs no longer lead to undefined situations; in other words, their observable behaviour is pinned down exactly by the spec






If I understand correctly, you seem to say:


1. the definition is in terms of a possible implementation through SPARQL syntax manipulation ("The change to define pre-binding through VALUES is just a change to the formal definition.")
2. the definition is not implementation advice ("It does NOT represent implementation advice.")
3. implementations may, internally, deviate arbitrarily from the implementation that is used for the definition ("implementations may not rely on VALUES as a syntax (but continue to change the internal runtime data structure such as algebra)")
4. 3 implies that implementations could run in to implementation-specific issues ("Given (...), I don't think the enforcement of these syntax limitations can be taken for granted")
5. 4 implies that the spec should keep excluding these constructs ("we should continue to spell them out.")


I don't follow step 5. As far as I understand, the goal of a spec is not to exclude the possibility of implementation-specific issues. I understand the goal to be that the spec is unambiguous (among many other things of course, like implementable, usable, understandable). In the 1.1 definition, these constructs had to be excluded to avoid ambiguity in the spec. In the context of end-of-{} values insertion, these constructs no longer lead to ambiguity, so the restrictions can be lifted.

MPvHarmelen:
@HolgerKnublauch in light of https://github.com/w3c/data-shapes/pull/1251#discussion_r4034947424, I modified the pull request to only remove the restriction on `values` clauses. Does that sufficiently address your concerns of https://github.com/w3c/data-shapes/pull/1261#issuecomment-5691164953?

HolgerKnublauch:
> @HolgerKnublauch in light of [#1251 (comment)](https://github.com/w3c/data-shapes/pull/1251#discussion_r4034947424), I modified the pull request to only remove the restriction on `values` clauses. Does that sufficiently address your concerns of [#1261 (comment)](https://github.com/w3c/data-shapes/pull/1261#issuecomment-5691164953)?

I don't understand how we can remove that restriction. People still cannot use VALUES ?prebound ... or what has changed that they could?

MPvHarmelen:
> People still cannot use VALUES ?prebound

Why not? 

Historically, the `values` clause was banned because it didn't interact well with the substitution-definition of pre-binding (see https://github.com/w3c/data-shapes/issues/627#issuecomment-5168813891). With the values insertion definition, this is no longer an issue.

HolgerKnublauch:
> > People still cannot use VALUES ?prebound
> 
> Why not?
> 
> Historically, the `values` clause was banned because it didn't interact well with the substitution-definition of pre-binding (see [#627 (comment)](https://github.com/w3c/data-shapes/issues/627#issuecomment-5168813891)). With the values insertion definition, this is no longer an issue.

I just tried

```
SELECT *
WHERE {
	VALUES ?a { 42 }
	VALUES ?a { 43 }
}
```

and it produces no results. So I don't believe that using VALUES with a different value than the pre-bound value is even working.

What is the use case for allowing this? Can you produce an example where a query re-defines the value of $this?

PR #1262, opened 2026-09-15
GH#1255: Add Datalog reference

### LINKED ISSUE #1255: Define or provide context for 'datalog'
The SPARQL-RL document starts with this in the abstract:

> This document defines SPARQL-RL, a datalog-style rules language for RDF.
>
> ...

No other mention is made of _datalog_ anywhere else in the document. 

If SPARQL-RL is to be defined as a "datalogy-style" something, then surely we need a bit more of a description of what datalog is or, at least, a reference to a datalog something.

erdemonal:
Closes #1255

Links the Datalog-style description of SPARQL-RL to a general reference on Datalog.

 [See this document rendered online here](https://raw.githack.com/erdemonal/data-shapes/issue-1255-datalog-reference/sparql12-rl/index.html)


afs:
Maybe a link to the wikipedia page is better.
https://en.wikipedia.org/wiki/Datalog 

It is hard to give a definitive reference. The wikipedia says it was emerging around 1977 (no reference) but suggested credit for the name isn't reported until much later (1995).

Some papers (1985) use the name expecting the reader to know what it is.
https://web.cecs.pdx.edu/harry/earley/datalog.pdf

Prolog is 1972.

RIF core isn't just datalog, it has extensions (as that links discusses).

erdemonal:
Thanks, that makes sense. I've updated the link to the Wikipedia page.

afs (approved):

Issue #1074, opened 2026-07-17
FOR ?v IN <shape>


afs:
_Originally posted by @simonstey in [#781](https://github.com/w3c/data-shapes/issues/781#issuecomment-4983967656)_

# The `FOR ?v IN <shape>` clause

`FOR ?v IN <shape>` re-introduces targeting on top of that model. It ties a single rule to a SHACL **shape** so the rule fires **only for the shape's target focus nodes that conform to the shape**, with a user-named "focus variable" pre-bound to each such node.

(It is an evolution of SHACL-AF's `sh:rule` + `$this`, refitted to current SRL semantics (solution mappings, stratification, run-once/general layers).

## Syntax

The clause is a **rule-level prefix**, sitting between the rule keyword (and its optional IRI) and the head template. 
⚠️  It is **not** a rule-body element  the spec-pure body grammar (triple patterns, `FILTER`, `NOT`, `SET`) is untouched.

Both concrete rule forms accept it:

```sparql
# RULE ... WHERE form
RULE ex:r FOR ?this IN ex:AdultShape
  { ?this ex:status ex:adult }
WHERE
  { ?this ex:age ?a . FILTER(?a >= 18) }

# IF ... THEN form
IF   { ?this ex:age ?a }
THEN ex:r FOR ?this IN ex:AdultShape { ?this ex:status ex:adult }
```

- `?this` is the **focus variable** -> author-chosen, any name. It is bound to each conforming focus node before the body runs, so it may appear in the body and the head.
- `ex:AdultShape` is a SHACL shape resolved against a separately supplied SHACL shapes graph.

##  Formal semantics

According to https://www.w3.org/TR/shacl12-rules/#eval-rule: 
- a **solution mapping** μ is a partial function from variables to RDF terms, and `μ0` is the empty solution mapping. 
- A rule body `B` evaluates over a **solution sequence** via `evalRuleElements(B, SEQ0, G)`, 
- starting from the initial sequence `SEQ0 = { μ0 }`; head templates are then instantiated once per `μ` in the resulting sequence `SEQ`. 
- **S** is the shapes graph, **G** the current data/inference graph.

### Abstract syntax

Extend a rule with an optional **targeting pair** `(v, s)` where `v` is a variable (the *focus variable*) and `s` is an IRI (the *shape*). A rule carrying a targeting pair is a **targeted rule**; one without is an ordinary rule (unchanged). 

```
[...]

PR #1263, opened 2026-09-16
First draft of a document to explain the vote options

HolgerKnublauch:
* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/rules-vote-prep/rules-vote.md)


rules-vote.md
```diff
@@ -0,0 +1,225 @@
...
+
+For SRL the cost is that the grammar requires a few additional lines so that either CONSTRUCT+BIND or RULE+SET
+can be used. They would both map to the same runtime objects, so implementation overhead is likely small.
+
+For SRL the benefit is that users that already know CONSTRUCT can get started immediately and even use
+existing SPARQL-based tooling to develop and test the rules.
+
+Another benefit for the SRL spec is that it doesn't need to worry about SHACL - the link becomes indirect through syntactical overlap.
+This would mean that the SRL Rule Type from Proposal 2 would not really be required.
+
```
recalcitrantsupplant, line 143: (commenting here but can move elsewhere if the intention was otherwise?)

the concerns I have with proposal #1 raised elsewhere summarised here:
1. for the use case where authors write rules with the intent of them being executed in a specific way, it is not clear without external context what the author's intent was. When text is copied / sent over chat/email etc. context can be lost
2. users may miss differences in semantics where the same tokens are used with different meanings. See https://dl.acm.org/doi/10.1145/3487051#sec-2-3-3

In addition I think implementations can provide translations fairly easily e.g. https://kurrawong.github.io/codemirror-lang-rdf/?sample=sparql-conversion (NB this is not an engine just a parser)

HolgerKnublauch: Yes there are arguments both ways for all options. It would be best to collect these arguments in the document to give them more visibility, so can you turn the change above into a diff change request that I can merge in?

recalcitrantsupplant, line 143:
```suggestion

### Concerns re. proposal `#1` - David Habgood

1. for the use case where authors write rules with the intent of them being executed in a specific way, it is not clear without external context what the author's intent was. When text is copied / sent over chat/email etc. context can be lost
2. users may miss differences in semantics where the same tokens are used with different meanings. See https://dl.acm.org/doi/10.1145/3487051#sec-2-3-3

In addition I think implementations can provide translations fairly easily e.g. https://kurrawong.github.io/codemirror-lang-rdf/?sample=sparql-conversion (NB this is not an engine just a parser)
```

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+While I acknowledge the desire of people in the SRL Task Force to produce a minimalistic and conceptually sound
+rule language, I believe there is a simple compromise that the SRL group could accept so that both languages
+can at least syntactically interoperate and so that we don't end up with two competing standards
+that unnecessarily confuse and divide the user community.
+
+## Proposal 0: No changes, keep them separate
+
+This could formally work if SRL is published next year and the WG charter is changed for 2027, to allow non-SHACL specs.
+But even formal objections to such a charter change are possible, as well as other objections to the document.
```
simonstey: If we want to have this proposal as voteable option it should also state the technical position it represents (two documents, two scopes, no mapping) and its costs for users. 

(Right now it reads as an option nobody is expected to pick; which is fine but it should then be also labelled as such imo)

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+has been rejected.  The only thing I have achieved was the deletion of the SRL native RDF syntax and this change was
+perhaps only made because I indicated that we would raise a formal objection otherwise.
+
+Furthermore, I believe that the process of how we reached the current draft was unfair:
+The Task Force made its key decisions (esp the introduction of the non-standard SPARQL keywords SET and NOT)
+before it was obvious that SHACL support was off the menu.
+As it is very difficult to change an existing draft (anyone can block commits and I am not even an editor of SRL),
+and all my requests to reconsider these choices were rejected, there was no way for me to cast -1 votes or block commits myself.
+So the current situation is the result of an asymmetric process in which I have no real power unless I cast formal objections
+to block the whole document.
```
simonstey: minor point but (imo) it would have been "cleaner" if we would split the process points from the technical ones. => a first-person account of the process inside the document about possible options invites the discussion to be about the narrative instead of the options. 

I would suggest moving it to a separate, linked position statement and keeping the Background section to facts the WG agrees on (or at least the objective/neutral stuff).

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Proposal 1: Generalize SRL Syntax to also support CONSTRUCT (aka SPARQL-Full vs SPARQL-RL)
+
+This was suggested by Holger across these tickets:
+- [SPARQL-Full vs SPARQL-RL](https://github.com/w3c/data-shapes/issues/1073)
+- [Add description of translation between SPARQL-RL syntax and SPARQL CONSTRUCT](https://github.com/w3c/data-shapes/issues/1196)
+
+The core observation is that the expressiveness of SRL is a subset of SPARQL CONSTRUCTs.
+SRL has introduced syntactic sugar (SET for BIND+FILTER bound and NOT for FILTER NOT EXISTS)
+but there is nothing in SRL that could not also be expressed in SHACL Rules via CONSTRUCT.
```
simonstey: this is not entirely true though:

1. [WHERE DATA / NOT DATA ("Matching Ground Data"): ](https://w3c.github.io/data-shapes/sparql12-rl/#ground-data) matches the base graph only, excluding inferred triples. shacl12-inference-rules states without exception that ["During execution of a rule, the evaluation graph is used as the active query graph",](https://www.w3.org/TR/shacl12-inference-rules/#rules-execution) and defines no base-graph-only mode. Neither NOT DATA (pattern match certain values on base graph only) nor WHERE DATA (entire rule body is pattern matched on base graph only) has no CONSTRUCT encoding today.

2. minor difference -> NOW() is fixed for the whole rule set evaluation in SPARQL-RL, but since in SHACL each rule execution is a separate query and SPARQL only fixes NOW() per query any NOW() used in CONSTRUCT queries will lead to different values compared to their SRL counterpart.

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+What SRL is adding is the ability to automatically compute the layers of rules (stratification).
+This is a nice feature that would be beneficial also for users of SHACL rules that don't even know
+that their rules fall into the SRL subset.
+
+If we recognize that SRL is a subset of CONSTRUCT then it *should* be possible to express any SRL rule set
+with SHACL Rules syntax.
+This would allow SRL rules to be executed by a SHACL engine, assuming that the layers have been set
+(either by hand or automatically through stratification).
+A key benefit is that users can step outside of the SRL expressiveness for individual rules,
+i.e. rule types can be mixed.
```
simonstey, line 77: SPARQL-RL's stratification condition is a precondition: ["if a rule set does not meet this condition, then this specification does not define an outcome for the evaluation of such a rule set."](https://w3c.github.io/data-shapes/sparql12-rl/#stratification-condition) 

SHACL Inference Rules has no equivalent and layering is the author's responsibility. 

=> an SRL rule set rewritten as CONSTRUCTs with incorrect `sh:layer` values produces wrong results rather than a failure, and a `NOT` inside an iterating layer becomes dependent on the iteration count. If Proposal 1 is adopted, SHACL Inference Rules should require a rules engine to report a failure when a layer's rules are not correctly stratified, i.e. when a negated pattern (`FILTER NOT EXISTS` / `MINUS`), or a run-once rule occurs inside a dependency cycle within a single layer

Otherwise the two documents might agree on syntax but produce different answers (which is arguably worse than having no bridge)

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
+        """ ;
+    ] .
+```
+
+which means the stratification can be applied once and is no longer needed before each execution.
+(In this case, only the sh:runOnce was computed, but there will certainly be more complex examples).
+
+For SRL the cost is that the grammar requires a few additional lines so that either CONSTRUCT+BIND or RULE+SET
+can be used. They would both map to the same runtime objects, so implementation overhead is likely small.
```
simonstey, line 130: yep, adding those would definitely be cheap but doing so wouldn't address the earlier raised charter concern as this would also not produce "a SHACL vocabulary" per se though.

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+            WHERE { ?r ex:width ?width . ?r ex:height ?height . BIND (?width * ?height AS ?area) }
+        """ ;
+    ] ;
+    sh:hasRule [
+        a sh:SPARQLRule ;
+        sh:construct """
+            CONSTRUCT { ?r ex:large true }
+            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
+        """ ;
+    ] .
```
simonstey, line 101: isn't this basically Proposal 2 but with the program string unpacked?

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+which means the stratification can be applied once and is no longer needed before each execution.
+(In this case, only the sh:runOnce was computed, but there will certainly be more complex examples).
+
+For SRL the cost is that the grammar requires a few additional lines so that either CONSTRUCT+BIND or RULE+SET
+can be used. They would both map to the same runtime objects, so implementation overhead is likely small.
+
+For SRL the benefit is that users that already know CONSTRUCT can get started immediately and even use
+existing SPARQL-based tooling to develop and test the rules.
```
simonstey, line 139: well.. not really though -> If SPARQL-RL's `CONSTRUCT` form accepts only the SRL fragment, then existing SPARQL tooling parses the same text and evaluates it under different semantics...

(basically what @recalcitrantsupplant mentioned below)

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+
+            RULE { ?r ex:large true }
+            WHERE { ?r ex:area ?area . FILTER (?area > 100) }
+        """ ;
+    ] .
+```
+
+Concerns have been recorded in the PR: <https://github.com/w3c/data-shapes/pull/1230>
+
+## Proposal 3: Support both
```
simonstey, line 210: what's the difference to proposal 0?

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+
+This acknowledges that none of us can fully predict the future so we could elect to give the choice to the users.
+
+If there is a clear winner in the coming years, future versions could deprecate one or the other syntax.
```
simonstey, line 216: None of the four options gives SRL a way to target SHACL shapes or classes

(which is the original charter issue raised in the beginning..  and what my #1074 (FOR ?v IN <shape>) tried to address) 

If charter compliance is the actual reason for the discussions here an option that cures it should be votable..

rules-vote.md
```diff
@@ -0,0 +1,231 @@
...
+so that optimizated algorithms such as incremental inferencing become possible.
+This wasn't the problem.
+But I did not anticipate that the language would completely sever any links to SHACL and instead
+become just another language for flat rule lists.
+
+So while I was observing the TF work from the outside (I don't have time to work on all documents myself)
+I was assuming that sooner or later such a link between rules and shapes was added.
+But the opposite happened, and gradually any remaining references to SHACL (even the namespace) were deleted.
+Anything that made SHACL Rules different from other rule languages was removed.
+SRL was not even properly able to match instances of a class, which is fundamental to linking rules to (SHACL) ontologies.
```
simonstey: this is not true anymore -> with recursion in the current draft, the `rdfs:subClassOf` closure is two rules ([cf. the :exposedTo example in the Recursion section](https://w3c.github.io/data-shapes/sparql12-rl/#example-rule-recursion)). 

what's still open though is shape-based targeting (which I tried to introduce with #1074 (`FOR ?v IN <shape>`)

simonstey (changes requested):
Thanks for the write up! I can understand your frustration but having been involved with both sides SRL + SHACL IR my ideal solution is one that defines the connection between the two "languages" and not two competing/differing standards for the same thing with minor differences (getting flashbacks to SHACL 1.0's final stretch and SHACL <-> ShEx "discussions").

My hope is that we can achieve this by:


> 1. SPARQL-RL keeps its grammar as it is.
> 2. SHACL Inference Rules gains a normative appendix mapping SPARQL-RL rule sets to  `sh:RuleSet` / `sh:SPARQLRule` / `sh:layer`, [referenced from the `sh:ruleProcessor` section (which already carries an editor's note anticipating exactly this).](https://www.w3.org/TR/shacl12-inference-rules/#ruleProcessor)
> 3. SHACL Inference Rules either gains base-graph-only matching or states that the `DATA` fragment is out of mapping scope.
> 4. The appendix explicitly lists the fragment that does not map ( `WHERE DATA` / `NOT DATA`, `NOW()` scope)
> 5. SHACL Inference Rules requires a rules engine to report a failure when a layer is not correctly stratified (a negated pattern or a run-once rule inside a dependency cycle within one layer) 
> 6. we put round-trip tests in the shared test suite -> so the mapping is part of the standard/rec.

Adopting a proposal along the lines of the poitns above would provide interoperability without either document having to adopt the other's surface syntax.

rules-vote.md
mgberg: > SPARQL-RL's stratification condition is a precondition: ["if a rule set does not meet this condition, then this specification does not define an outcome for the evaluation of such a rule set."](https://w3c.github.io/data-shapes/sparql12-rl/#stratification-condition)
> 
> SHACL Inference Rules has no equivalent and layering is the author's responsibility.

Is the Custom Rule Processors capability added in https://github.com/w3c/data-shapes/pull/1207 sufficient for this, or are there some details I'm not aware of that would prevent that from working?

rules-vote.md
HolgerKnublauch: Ok I am moving this block to a clearly marked section at the very end of the document.

rules-vote.md
HolgerKnublauch: I agree that 1. is a missing feature in SHACL and I suggest that discussion gets moved to https://github.com/w3c/data-shapes/issues/1271 (I don't expect this to be an obstacle to more complete coverage).

With the NOW() topic I had raised my concerns before, that this requires low-level control over SPARQL engines and makes it impossible to execute SRL rule sets using the generic SPARQL protocols and APIs. This is IMHO unfortunate and should be changed in SRL.

I will update the document to reflect those details so that your comment isn't lost.

rules-vote.md
HolgerKnublauch: Yes you are right, and I felt uneasy about this and was hoping that others who support this would make such additions. Thanks for the feedback.

I also had private feedback from @liviorobaldo who pointed out that the Proposal 0 is incomplete. So I am pushing a change that moves Proposal 0 as a Proposal 4 with two variations 4a and 4b. This has hopefully addressed your concerns and will make them more serious contenders. As usual, anyone is welcome to edit the document further, or make change requests (which I will always try to accommodate even if they don't reflect my personal preference).

On this occasion I have also added another option 2b which was also suggested by @mgberg and myself earlier (in https://github.com/w3c/data-shapes/pull/1230#discussion_r3964203449).

rules-vote.md
HolgerKnublauch: I think it does address the SHACL concern because I would feel more comfortable about adding a linking statement (e.g. as an appendix) to the SHACL 1.2 Inference Rules documents. Even without that, it would become much harder for anyone to defend a formal objection because it clearly was discussed by this WG and intended to be used this way.

Anyway, I have updated the document to add a paragraph that this doesn't even require a grammar change but could be handled by a pre-processor, as long as this pre-processor is a mandatory feature of SRL engines.

rules-vote.md
HolgerKnublauch: No it is very different because right now the program string does not use CONSTRUCT syntax. This is a key difference. The above CAN be executed by a vanilla SHACL engine while both Proposals 2 that add a custom rule type also require that a SHACL engine implements these rule types. This is a much bigger ask than only implementing the stratification algorithm as a plugin.

I have, however, meanwhile added option 2b to refine this distinction.

rules-vote.md
HolgerKnublauch: I hope this is clarified now.

rules-vote.md
HolgerKnublauch: I don't understand that point. There is no semantic difference here. CONSTRUCT means exactly the same in both cases, no keyword is interpreted differently. Could either of you clarify, e.g. with an example?

Of course, being a subset, a CONSTRUCT processor that is limited to SRL will report a failure when it encounters SPARQL Full elements.

rules-vote.md
HolgerKnublauch: We had discussed the rdfs:subClassOf* topic several times, and the hack with two rules has the problem that it causes an explosion of triples and those triples can interfere with rules that did not intend to see them (e.g. because they are only about the direct rdfs:subClassOf relationships).

On the shape-based targeting, I don't see any open PRs or Issues, so I don't have the impression that SRL folks are still aiming at that.

rules-vote.md
HolgerKnublauch: You would need to reopen #1074 if you want to bring this back into the game.

HolgerKnublauch:
> Thanks for the write up! I can understand your frustration but having been involved with both sides SRL + SHACL IR my ideal solution is one that defines the connection between the two "languages" and not two competing/differing standards for the same thing with minor differences (getting flashbacks to SHACL 1.0's final stretch and SHACL <-> ShEx "discussions").

Nobody here wants to repeat the SHACL vs ShEx challenges. This was an unfortunate process. But right now I am afraid we are heading towards a similarly disappointing outcome unless we all try to work on compromises. As I have written, the process how we reached the current state was IMHO already unfair.

> My hope is that we can achieve this by:
> 
> > 1. SPARQL-RL keeps its grammar as it is.
> > 2. SHACL Inference Rules gains a normative appendix mapping SPARQL-RL rule sets to  `sh:RuleSet` / `sh:SPARQLRule` / `sh:layer`, [referenced from the `sh:ruleProcessor` section (which already carries an editor's note anticipating exactly this).](https://www.w3.org/TR/shacl12-inference-rules/#ruleProcessor)

Yes some middle-ground as you outline can work, but I would at the minimum expect that the mapping from SRL to SHACL needs to live in the SPARQL-RL document. From a SHACL perspective, SRL is just one possible subset among others. It has defined the plugin mechanisms that other languages can hook into. Also SRL will evolve further so any future feature of SRL shouldn't require changes to the SHACL documents. Even the release cycle may be different, with SRL published only in 2027. Could you live with that?

> > 3. SHACL Inference Rules either gains base-graph-only matching or states that the `DATA` fragment is out of mapping scope.

Point 3 will be addressed by #1271

> > 4. The appendix explicitly lists the fragment that does not map ( `WHERE DATA` / `NOT DATA`, `NOW()` scope)

This would currently only be the semantics of NOW() which feels rather uncritical. SHACL engines could print a warning and move on if they encounter two NOW rules.

> > 5. SHACL Inference Rules requires a rules engine to report a failure when a layer is not correctly stratified (a negated pattern or a run-once rule inside a dependency cycle within one layer)

The sh:ruleProcessor mechanism can do this, and the details of such processing can be defined in an appendix of the SPARQL-RL document.

> > 6. we put round-trip tests in the shared test suite -> so the mapping is part of the standard/rec.
> 
> Adopting a proposal along the lines of the poitns above would provide interoperability without either document having to adopt the other's surface syntax.

Overall, is this another votable Proposal that you want to add to the document? Feel free to add it!

rules-vote.md
HolgerKnublauch: Yes, this is the intent of the recently added sh:ruleProcessor mechanism.

rules-vote.md
```diff
@@ -0,0 +1,306 @@
...
+    ] .
+```
+
+In this design, the existing rule set infrastructure of SHACL RDF syntax is used
+whereas 2a relied on nesting complete rule sets into (large) rule literals.
+It would allow users to use the syntactic sugar of SRL.
+And it would be a mechanism to enforce certain contracts as an engine that goes not fully recognize
+the srl:Rule type would need to throw a failure.
+
+A downside is that this cannot directly be executed with a vanilla SHACL engine.
```
mgberg: Given that a vanilla SHACL engine wouldn't have the stratification algorithm to evaluate SRL rules in the correct order anyway, wouldn't it be the case that a vanilla SHACL engine couldn't "fully/correctly" execute an SRL ruleset regardless of the syntax?

afs:
SPARQL-RL can integrate using the framework of the input document SHACL-AF rules.

The idea of using SHACL-AF as-is and including a new rule type, where the rule step is one SPARQL-RL Rule set is not new.

The discussion on PR #452 includes a link to ["Attaching Rules To Shapes"](https://gist.github.com/afs/2547eac679b1411acc078959ac2b3d31#attaching-rules-to-shapes)  (Aug 2025).

A recent discussion #765/#766 (Feb 2026) also covers this approach; #766 where there was a request was to leave SHACL-AF Rules alone, and #765 (Feb 2026) where there is the sketch:

```
[] rdf:type sh:NodeShape ;
    sh:rule 
      [ a srl:SHACLRule ;
        srl:ruleSet "...";    ## SRL syntax
        sh:prefixes ... ;
      ];
```

The charter, [section 1](https://www.w3.org/2024/12/data-shapes.html#background), cover this:

> Motivation and Background
>
> The primary motivations for forming this working group are to:
>
> * provide SHACL handling of for RDF-star and SPARQL-star
> * make minor updates to the SHACL core and extended specifications from 5+ years experience of their use
>
> The secondary motivation is to:
>
> * develop new specifications that assist with extended SHACL use
>
> For this secondary motivation, the areas of UI generation, reasoning rule definition and profiling descriptions are expected to be addressed. Others may be too, pending resourcing.

afs:
#1113, specifically https://github.com/w3c/data-shapes/issues/1113#issuecomment-5188359672.
@nicholascar:
> Since I believe we can have an SRL → SHACL-SPARQL translation (but not the reverse) and SHACL SPARQL has an RDF syntax, we will need to remove 5. Concrete Syntax forms for the Shape Rules Language from SRL or else we will have two competing RDF syntaxes. I hope this will be a net win for SRL: less stuff to
maintain!

We can translate a valid SRL rule set to a SHACL Inf Rules rule set, with two translation functions: 

1. Conversion of each `RULE`, and the `DATA` block to the `CONSTRUCT` subset (after checking well-formedness)
2.  Encoding the stratification as `sh:layer` (and not using `sh:order`)

(1) is static, (2) needs the stratification algorithm.

then produce a shapes graph. 

This allows SHACL Inf Rules to access the stratification calculation.

This address the proposed text #1269 which steers the user away from proposal 2. Proposal 2 allows for choice - incremental evolution of SHACL SPARQL Rules and SRL based rules, picking the rule step most suited to the local needs.

As languages develop, translations can evolve. It might be 2 SHACL Inf Rules to one SPARQL-RL rule, it might be 2 SPARQL-RL rules to one SHACL Inference Rule framework steps.

There is no advantage to have a fixed 1-1 text syntax linkage. Acknowledging the existence of programmatic translation does not freeze a fixed syntax linkage based on the current state of documents. It allows both SPARQL-RL and SHACL Inference Rules to evolve; SHACL Inference Rules is still evolving.

We are now running out of time, but a NOTE may yet be possible for a translation based on the current state.

rules-vote.md
```diff
@@ -0,0 +1,306 @@
...
+### Proposal 2b: Add Rule Type for individual SRL rules
+
+This is not covered by a PR yet but could make sense:
+
+```
+ex:MyRuleSet
+    a sh:RuleSet ;
+    sh:hasRule [
+        a srl:Rule ;
+        sh:construct """
```
mgberg: I believe the other examples where this came from used `srl:text` instead of `sh:construct`, which would make sense given that values would be an SRL string instead of a CONSTRUCT query in this option.

HolgerKnublauch: Good catch, this was my copy-and-paste mistake.

rules-vote.md
HolgerKnublauch: Text updated, now hopefully clarified.

HolgerKnublauch:
> SPARQL-RL can integrate using the framework of the input document SHACL-AF rules.
> 
> The idea of using SHACL-AF as-is and including a new rule type, where the rule step is one SPARQL-RL Rule set is not new.
> 
> The discussion on PR #452 includes a link to ["Attaching Rules To Shapes"](https://gist.github.com/afs/2547eac679b1411acc078959ac2b3d31#attaching-rules-to-shapes) (Aug 2025).
> 
> A recent discussion #765/#766 (Feb 2026) also covers this approach; #766 where there was a request was to leave SHACL-AF Rules alone

People can read the full discussion and make up their own mind about this "request to leave SHACL-AF Rules alone". I stated explicitly that I am OK to leave SHACL-AF alone assuming that sh:SPARQLRule will become part of SHACL Rules (which then evolved into SRL): https://github.com/w3c/data-shapes/issues/766#issuecomment-3863108674

Basically everybody else on the discussion on that ticket (including well-known external members of the semantic web community) supported keeping CONSTRUCT and sh:SPARQLRule in the language. To address those requirements we had to create the separate SHACL Rules document because SRL didn't address the original requirements.

So claiming that "there was a request was to leave SHACL-AF Rules alone" is misleading.

> The charter, [section 1](https://www.w3.org/2024/12/data-shapes.html#background), cover this:
> 
> > Motivation and Background
> > The primary motivations for forming this working group are to:
> > 
> > * provide SHACL handling of for RDF-star and SPARQL-star
> > * make minor updates to the SHACL core and extended specifications from 5+ years experience of their use
> > 
> > The secondary motivation is to:
> > 
> > * develop new specifications that assist with extended SHACL use
> > 
> > For this secondary motivation, the areas of UI generation, reasoning rule definition and profiling descriptions are expected to be addressed. Others may be too, pending resourcing.

Yes, and the SHACL 1.2 Inference Rules document addresses that part of the charter. This makes the SPARQL-RL document, in principle, redundant. It currently still violates the charter because it's not based on SHACL.

HolgerKnublauch:
> #1113, specifically [#1113 (comment)](https://github.com/w3c/data-shapes/issues/1113#issuecomment-5188359672). @nicholascar:
> 
> > Since I believe we can have an SRL → SHACL-SPARQL translation (but not the reverse) and SHACL SPARQL has an RDF syntax, we will need to remove 5. Concrete Syntax forms for the Shape Rules Language from SRL or else we will have two competing RDF syntaxes. I hope this will be a net win for SRL: less stuff to
> > maintain!
> 
> We can translate a valid SRL rule set to a SHACL Inf Rules rule set, with two translation functions:
> 
> 1. Conversion of each `RULE`, and the `DATA` block to the `CONSTRUCT` subset (after checking well-formedness)
> 2. Encoding the stratification as `sh:layer` (and not using `sh:order`)
> 
> (1) is static, (2) needs the stratification algorithm.
> 
> then produce a shapes graph.
> 
> This allows SHACL Inf Rules to access the stratification calculation.
> 
> This address the proposed text #1269 which steers the user away from proposal 2. Proposal 2 allows for choice - incremental evolution of SHACL SPARQL Rules and SRL based rules, picking the rule step most suited to the local needs.
> 
> As languages develop, translations can evolve. It might be 2 SHACL Inf Rules to one SPARQL-RL rule, it might be 2 SPARQL-RL rules to one SHACL Inference Rule framework steps.
> 
> There is no advantage to have a fixed 1-1 text syntax linkage. Acknowledging the existence of programmatic translation does not freeze a fixed syntax linkage based on the current state of documents. It allows both SPARQL-RL and SHACL Inference Rules to evolve; SHACL Inference Rules is still evolving.
> 
> We are now running out of time, but a NOTE may yet be possible for a translation based on the current state.

My hope with starting this vote document was to collect all arguments in one condensed space. We already had spent months with endless discussion threads. I suggest anyone enumerates their arguments in the vote document (you can edit directly, I am not safeguarding this document).

The reason for the vote is exactly what you write: we are running out of time. Whatever the outcome of the vote is, I don't see any proposal that would take more than a week to implement.

rules-vote.md
HolgerKnublauch: Also note that even with SPARQL-RL native grammar, a SRL engine needs to check the https://www.w3.org/TR/sparql12-rl/#wellformed conditions after parsing. So people can write a malformed SRL rule that passes the Grammar test but fails at execution. Similar situation when such an engine encounters a SPARQL CONSTRUCT rule that contains unsupported elements.

afs:
What has been lost in the discussions is there is a downside to a rigidly interconnected set of specifications.

To implement the SHACL Inference Rules framework requires SHACL Core, which in turn brings in SHACL node expressions, and it requires SHACL SPARQL-Related features.

We should be able to use best-of-breed implementations. Pick one implementation of the SHACL Inference Rules framework, and separately pick a rules engine. Different rules engines will focus on different aspects: scale, performance, truth maintenance/incremental update, explainability, ... . 

Expecting each rule engine to have custom code for each SHACL Inf Rules implementation, or each SHACL Inf Rules implementation to have custom code for each engine will impede SHACL for rules.

afs:
> This makes the SPARQL-RL document, in principle, redundant. 

Compare like-for-like.

The SHACL solution is "SHACL-AF rules framework + SPARQL-RL".

HolgerKnublauch:
> What has been lost in the discussions is there is a downside to a rigidly interconnected set of specifications.
> 
> To implement the SHACL Inference Rules framework requires SHACL Core, which in turn brings in SHACL node expressions, and it requires SHACL SPARQL-Related features.
> 
> We should be able to use best-of-breed implementations. Pick one implementation of the SHACL Inference Rules framework, and separately pick a rules engine. Different rules engines will focus on different aspects: scale, performance, truth maintenance/incremental update, explainability, ... .

Yes absolutely. One reason why I am so keen on seeing our languages interoperate is that SHACL rules users can benefit from optimized incremental rule engines. These are more likely to emerge with SRL as starting point than full SPARQL.

> Expecting each rule engine to have custom code for each SHACL Inf Rules implementation, or each SHACL Inf Rules implementation to have custom code for each engine will impede SHACL for rules.

Here is a fresh attempt at finding a compromise that keeps the standards aligned without depending on each other or over-complicating things: https://github.com/w3c/data-shapes/pull/1278

PR #1264, opened 2026-09-17
Add property-path tests with literal value nodes (oneOrMore, zeroOrMore, sequence)

styk-tv:
`shacl12-test-suite/tests/core/path/` has no test in which a literal is the value reached by a property path. Every existing path test uses IRI values throughout, so an implementation can pass all of them while dropping every literal a path reaches.

SHACL property paths are defined over SPARQL property paths, whose results routinely include literals — a literal is a valid path endpoint, it simply cannot be a path origin, since a literal cannot be an RDF subject.

### How this surfaced

I hit this while building property-path coverage tests for a SHACL implementation: literal values reached through recursive and sequence paths were not being validated at all, and a literal partway along a multi-hop path also suppressed violations on a sibling branch that was still reachable.

The cause was in [rudof](https://github.com/rudof-project/rudof), and the maintainers fixed it the same day it was reported — [#818](https://github.com/rudof-project/rudof/issues/818) / [#819](https://github.com/rudof-project/rudof/pull/819), merged.

The part that seems relevant here is what happened to the tests. rudof runs `path-oneOrMore-001` and `path-zeroOrMore-001` from this suite against **both** of its validation engines, and both passed throughout — there was nothing in the fixtures for them to catch, since the values are all IRIs. That is the gap these three cases are meant to close, and it applies to any implementation rather than to that one.

### What this adds

Three tests, named per the directory's existing convention (`path-<pathType>-<NNN>`, numbered per path type):

| File | Case |
|---|---|
| `path-oneOrMore-002.ttl` | literal as the value of `sh:oneOrMorePath` |
| `path-zeroOrMore-002.ttl` | literal as the value of `sh:zeroOrMorePath` |
| `path-sequence-003.ttl` | literal at an intermediate step of a sequence, with a sibling branch that still yields a value |

Each uses `sh:nodeKind sh:IRI`, so the literal value node is the violation. Each is self-contained — data, shapes, manifest entry and expected report in one file — in the style of the existing tests. `manifest.ttl` gains three `mf:include` lines beside their same-type neighbours.

**Why these three and not the other path types.** They are the only Core path constructs where a reached literal can be lost, and this follows from the SPARQL definitions rather than from sampling. `sh:oneOrMorePath` and `sh:zeroOrMorePath` are the two that recurse *from* the terms they reach, via `ALP`, so they are the two that can attempt to continue from a literal. A sequence chains through an intermediate term, so it can attempt the same at any step but the last. By contrast `sh:alternativePath` is a `Union` of branches each evaluated from the focus node, and `sh:zeroOrOnePath` likewise evaluates only from the focus node — neither ever continues from a reached term, so a literal there is an ordinary single-step result. And a bare `sh:inversePath` reaches subjects, which cannot be literals at all. Adding literal cases for those would not test anything the existing fixtures leave open.

### Notes for review

- `path-zeroOrMore-002` expects **one** result, not two. The zero-length path contributes the focus node itself, which is an IRI and satisfies the constraint; only the literal violates it.
- `path-sequence-003` places the literal on a *sibling* branch of `ex:child`, alongside an IRI-valued branch that reaches `ex:name "Alice"`. That is what makes it a regression test rather than a third endpoint case: an implementation that aborts the path when it meets the literal also loses `"Alice"`, and reports nothing at all.
- The expected results are **derived from the normative text**, not just observed: SHACL §1.5 defers path semantics to SPARQL; §3.7 defines value nodes as a set of *nodes* with no term-type restriction; and SPARQL 1.1 §18.4's `ALP` does `add x to V` *before* `X = eval(x,path)`, where `eval` from a literal is the empty set rather than an error. The order of those two lines is what these fixtures pin.
- The expected reports were then verified on **two independent implementations**: pySHACL 0.30.1 (rdflib 7.1.4) and a second engine carrying the rudof fix. Both produce exactly the report each file declares. Before that fix, the second engine reported `sh:conforms true` with no results for all three.
- I also ran the **whole of `tests/core/path/` through pySHACL**, resolving `sht:dataGraph` / `sht:shapesGraph` per test: **14/14 agree with their declared result**, so these additions do not disturb the existing entries and the manifest still lists exactly the files on disk.
- `mf:status` is `sht:approved` to match the neighbouring entries. Happy to change these to `sht:proposed` if that is preferred for new submissions.


HolgerKnublauch:
Thanks for your contribution! I have confirmed the tests also work in our (Java) implementation.

As there are IPR constraints in the W3C repo, we cannot directly merge your PR into ours. I have instead created a fresh clone under my name: https://github.com/w3c/data-shapes/pull/1265

HolgerKnublauch:
Once the other PR is merged, I suggest you close this one here?

styk-tv:
Thank you @HolgerKnublauch  :superhero_man:

Issue #1266, opened 2026-09-19
Two unrunnable fixtures in shacl12-test-suite/tests/core/node/: in-003.ttl (undeclared shsh: prefix) and in-002.ttl (undefined ex:TestShape)


sparkling:
Hello — while running the SHACL 1.2 test suite against an independent implementation we hit two fixtures in `shacl12-test-suite/tests/core/node/` that we could not execute. Both appear to be small transcription slips rather than anything to do with the spec, so I wanted to flag them in case they are useful. Apologies if either is already known.

Observed at commit `eedda09f93c39be1d2e978f3f942631494ae25a0`, and re-checked on `gh-pages` today — both files are byte-identical on the current default branch. Neither has been modified since 1c2916e0 (PR #451, "#429: Have SHACL-SHACL warn about unsatisfiable usage of `sh:in` and `sh:xone`").

---

### 1. `in-003.ttl` — undeclared `shsh:` prefix, plus one term that does not resolve

The file uses the `shsh:` prefix at lines 16, 18, 19, 22, 24 and 50, but the `@prefix` block (lines 1–8) never declares it. An undeclared prefix is a Turtle syntax error, so the file cannot be parsed and the test cannot be loaded at all.

The sibling fixture `xone-003.ttl`, added in the same commit, does declare it on line 7:

```turtle
@prefix shsh: <http://www.w3.org/ns/shacl-shacl#> .
```

Adding that one line to `in-003.ttl` makes it parse (33 triples).

There is a second, smaller point in the same file. Line 19 reads:

```turtle
  sh:property shsh:inSubjectsShapeXonePropertyShape ;
```

That term is not defined in the fixture, and does not appear in `http://www.w3.org/ns/shacl-shacl#`. Comparing against `xone-003.ttl` it looks like a leftover from the copy: the corresponding line there is `sh:property shsh:xoneSubjectsShapeXonePropertyShape`, which does resolve. The property shape actually defined in `in-003.ttl` (line 22) — and the one its expected result cites via `sh:sourceShape` (line 50) — is `shsh:inSubjectsShapeInPropertyShape`, matching the namespace document. So line 19 reads as though it should be:

```turtle
  sh:property shsh:inSubjectsShapeInPropertyShape ;
```

[...]

PR #1267, opened 2026-09-19
Fix in-003 test fixture

erdemonal:
Addresses part of #1266.

Fixes the missing `shsh:` prefix declaration and the incorrect `shsh:inSubjectsShapeXonePropertyShape` reference in `in-003.ttl`.

`in-002.ttl` is left unchanged.


HolgerKnublauch:
Thanks! I have reimplemented this in a separate PR to avoid the IPR problems. Closing this one here.

PR #1269, opened 2026-09-20
Add advisory on mixing rule types in rule sets

liviorobaldo:
As previously discussed, I have just added a short paragraph advising against mixing different rule types in a rule set due to potential unpredictability.

* [See this document rendered online here](https://raw.githack.com/w3c/data-shapes/nomixtypes/shacl12-inference-rules/index.html)


shacl12-inference-rules/index.html
```diff
@@ -1004,6 +1004,12 @@ <h2>Rule Sets</h2>
 					(via <code>ex:RuleSet2</code>), while <code>ex:RuleSet2</code> contains only <code>ex:Rule2</code>.
 				</p>
 			</aside>
+			<p>
+			    Although rule sets may contain rules of different types, users should include only rules of the same type within a rule set.
```
scotthenninger:
```suggestion
			    Although rule sets may contain rules of different <a>rule types</a>, users should include only rules of the same type within a rule set.
```

scotthenninger (commented):
Link to the definition of a rule type.

shacl12-inference-rules/index.html
liviorobaldo: > Link to the definition of a rule type.

I agree... but does "\<a\>rule types\<\/a\>" work? Shouldn't it be "\<a href="\#rule-types"\>rule types\<\/a\>"?

scotthenninger: I looked and there isn’t an anchor defied for the section and this is how the reference is defined elsewhere. I assume there is a script or post-processing to resolve the link.

HolgerKnublauch (approved):
I see no problems with making such a statement as it is not a MUST.

afs:
> I see no problems with making such a statement as it is not a MUST.

nor is it a SHOULD (there is one SHOULD in the whole doc in section 9).

liviorobaldo:
Sorry @afs, I don't quite understand:

> nor is it a SHOULD (there is one SHOULD in the whole doc in section 9).

It is a "should":

> "Although rule sets may contain rules of different <a>rule types</a>, users should include only rules of the same type within a rule set."

I looked at Section 9. Do you mean that I should write "SHOULD" in uppercase, like this?

> "Although rule sets may contain rules of different <a>rule types</a>, users SHOULD include only rules of the same type within a rule set."

afs:
> Sorry @afs, I don't quite understand:
> 
> > nor is it a SHOULD (there is one SHOULD in the whole doc in section 9).
> 
> It is a "should":
> 
> > "Although rule sets may contain rules of different rule types, users should include only rules of the same type within a rule set."
> 
> I looked at Section 9. Do you mean that I should write "SHOULD" in uppercase, like this?
> 
> > "Although rule sets may contain rules of different rule types, users SHOULD include only rules of the same type within a rule set."

You need to decide.

"SHOULD" and "should" are different - see the conformance section.

"should" is informal. "SHOULD" has a specific weight/meaning; as Holger notes, it is weaker than "MUST".

liviorobaldo:
Ok, I changed it into "SHOULD" capitalized.

